PR #351, opened 2014-09-14
Added 'OnPCStatCalcEvent' trigger

MrKeiKun:
credits: @QQfoolsorellina, @AnnieRuru
rAthena/rathena@27a0f3f


shennetsind:
Insanely resource intensive, you're calling it even before parsing equips (which can, and will result in the event being triggered about a dozen times for a fully geared player on each action (whereas once would be expected), everytime that status calc is necessary).
Should be a plugin.

PR #361, opened 2014-09-19
Fix nullpo function needed for plugin

AnnieRuru:
http://hercules.ws/board/topic/7188-market/?p=43751
this fix doesn't include the hooking to assert_report function ...
I don't think people ever need it anyway

Ind or Haru please double check then merge,
because its just annoying that we have to add assert_report every time dealing with nullpo

### DIFF
--- a/src/common/nullpo.c
+++ b/src/common/nullpo.c
@@ -12,6 +12,8 @@
 
 #include "../common/showmsg.h"
 
+struct nullpo_interface nullpo_s;
+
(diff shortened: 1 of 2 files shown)


shennetsind:
As this pull request has been without feedback for a while I've taken in my hands to implement this.
https://github.com/HerculesWS/Hercules/commit/57789fb70b899eda61206a59283e1a849e544397

PR #364, opened 2014-09-23
Add color constants for announcement and mes

AnnieRuru:
just declare a bunch in the constant file, then viola !
new bunch of color code can be use without refering to external sites for color effects
http://hercules.ws/board/topic/7024-add-color-constants-for-announce/

@MishimaHaruna might love this ?


shennetsind:
I'm on maid mode. Closing this due to lack of feedback from the user

PR #374, opened 2014-10-18
Deprecation of legacy script commands

MishimaHaruna:
Marking some legacy script commands as deprecated, for future removal.

**_NOTE: This is a work in progress, please DO NOT merge.**_

Goals of the project:
- Discourage the use in new scripts of certain legacy script commands that would promote bad scripting habits.
- Press users to update their custom scripts to remove said commands.
- Provide user documentation on how to replace those commands from custom scripts.
- Prepare for the complete removal of those script commands.

Non goals:
- Prevent any existing scripts from working.
- Remove any script commands.

Commands slated for removal and suggested alternatives (list will be integrated with other commands)
- [x] `menu` -> `select`, `prompt`
- [ ] `set` -> direct assignment (NOTE: This can't be done at the moment, as there are some edge cases where `set` is still necessary, unfortunately. We may provide a non-deprecated synonym for those edge cases.
- [x] `setr` (internal use only)
- [x] `checkquest` -> `questprogress` (NOTE: This requires a re-commit of 4ac673941714032ada6d26fb60936ec510bbe496)
- [ ] `goto` -> proper use of `if`, `while`, `for`, `switch`. A synonym for edge cases (and for internal use) might be provided.
- [x] `jump_zero` -> proper use of `if`, `while`, `for`, `switch`. A prefixed synonym internal use is provided.
- [x] `setdragon`, `setmadogear`, `setriding` -> `setmount`
- [x] `checkriding`, `checkdragon`, `checkmadogear` -> `checkmount`
- [x] `petheal` -> `petskillsupport`
- [ ] `cardscnt` -> `getequipcardcnt`
- [ ] `basicskillcheck` -> `getbattleflag("basic_skill_check")`
- [x] `save` -> `savepoint`
- [x] `cmdothernpc` -> `donpcevent`
- [x] `enablearena`, `disablearena` -> `enablewaitingroomevent`, `disablewaitingroomevent`
- [ ] `setmapflagnosave` -> `setmapflag` (requires source changes to `setmapflag`)
- [x] `petskillattack` -> `petskillattack2` (and rename back to `petskillattack`, eventually). `petskillattack` is the same as `petskillattack2` with the `div` argument set to 0.
- [x] `isday` -> `!isnight()`
- [ ] `checkequipedcard` -> Never used. Loop through `getinventorylist`'s results if necessary.
- [ ] `getusersname` -> Never used. It could be replaced by a new command that returns an array of users (IDs, names), since we now have support for very large arrays.
- [ ] `recovery` -> Never used. It could be replaced by a new command that returns an array of users (IDs, names), and a loop on said array.
- [ ] `globalmes` -> Never used. It could be replaced by a new command that returns an array of users (IDs, names), and a loop on said array.
-    `awake` -> use of NPC timers instead. [keeping it for the time being, see discussion below]
- [x] `checkre` -> We should export a constant instead.
- [ ] `consumeitem` -> `itemeffect` (misleading name, and two synonyms for the same command is one too many)
- [ ] `classchange`

`petheal` can't be replaced by `petskillsupport` - see comment below -- nevermind, it can be replaced after all, it's custom.

---

This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/374)

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -1457,164 +1457,110 @@ Gotos are considered to be harmful and should be avoided whenever possible.
 
 *menu("<option_text>", <target_label>{, "<option_text>", <target_label>, ...})
 
-This command will create a selectable menu for the invoking character.
-Only one menu can be on screen at the same time.
-
-Depending on what the player picks from the menu, the script execution
-will continue from the corresponding label. It's string-label pairs, not
-label-string.
+    @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+    @ /!\ This command is deprecated @
+    @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
 
 This command is deprecated and it should not be used in new scripts, as it
 is likely to be removed at a later time. Please consider using select() or
 prompt() instead.
 
-Options can be grouped together, separated by the character ':'.
+For the complete documentation, please see select and prompt.
 
-	menu("A:B", L_Wrong, "C", L_Right);
-
-It also sets a special temporary character variable @menu, which contains
-the number of option the player picked. Numbering of options starts at 1.
-This number is consistent with empty options and grouped options.
-
-	menu("A::B", L_Wrong, "", L_Impossible, "C", L_Right);
-	L_Wrong:
-	// If they click "A" or "B" they will end up here
-	// @menu == 1 if "A"
-	// @menu == 2 will never happen because the option is empty
-	// @menu == 3 if "B"
-	L_Impossible:
-	// Empty options are not displayed and therefore can't be selected
-	// this label will never be reached from the menu command
-	L_Right:
-	// If they click "C" they will end up here
-	// @menu == 5
+Unlike select, this command will jump to the label corrseponding to the user's
+selection. Note: It's string-label pairs, not label-string.
 
 If a label is '-', the script execution will continue right after the menu
-command if that option is selected, this can be used to save you time, and
-optimize big scripts.
-
-	menu("A::B:", -, "C", L_Right);
-	// If they click "A" or "B" they will end up here
-	// @menu == 1 if "A"
-	// @menu == 3 if "B"
-	L_Right:
-	// If they click "C" they will end up here
-	// @menu == 5
-
-Both these examples will perform the exact same task.
-
-If you give an empty string as a menu item, the item will not display.
-This can effectively be used to script dynamic menus by using empty string
-for entries that should be unavailable at that time.
-
-You can do it by using arrays, but watch carefully - this trick isn't high
-wizardry, but minor magic at least. You can't expect to easily duplicate
-it until you understand how it works.
-
-Create a temporary array of strings to contain your menu items, and
-populate it with the strings that should go into the menu at this
-execution, making sure not to leave any gaps. Normally, you do it with a
-loop and an extra counter, like this:
-
-	setarray(.@possiblemenuitems$[0], <list of potential menu items>);
-	.@j = 0; // That's the menu lines counter.
-
-	// We loop through the list of possible menu items.
-	// .@i is our loop counter.
-	for (.@i = 0; .@i < getarraysize(.@possiblemenuitems$); ++.@i) {
-		// That 'condition' is whatever condition that determines whether
-		// a menu item number .@i actually goes into the menu or not.
-
-		if (<condition>) {
-			// We record the option into the list of options actually
-			// available.
-			.@menulist$[.@j] = .@possiblemenuitems$[.@i];
-
-			// We just copied the string, we do need it's number for later
-			// though, so we record it as well.
-			.@menureference[.@j] = .@i;
-
-			// Since we've just added a menu item into the list, we
-			// increment the menu lines counter.
-			++.@j;
-		}
-
-		// We go on to the next possible menu item.
-	}
-
-This will create you an array .@menulist$ which contains the text of all
-items that should actually go into the menu based on your condition, and
-an array .@menureference, which contains their numbers in the list of
-possible menu items. Remember, arrays start with 0. There's less of them
-than the possible menu items you've defined, but the menu() command can
-handle the empty lines - only if they are last in the list, and if it's
-made this way, they are. Now comes a dirty trick:
-
-	// X is whatever the most menu items you expect to handle.
-	menu(.@menulist$[0], -, .@menulist$[1], -, ..., .@menulist$[<X>], -);
-
-This calls up a menu of all your items. Since you didn't copy some of the
-possible menu items into the list, it's end is empty and so no menu items
-will show up past the end. But this menu() call doesn't jump anywhere, it
-just continues execution right after the menu() command. (And it's a good
-thing it doesn't, cause you can only explicitly define labels to jump to,
-and how do you know which ones to define if you don't know beforehand
-which options will end up where in your menu?)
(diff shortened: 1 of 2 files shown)


MishimaHaruna:
Merged commits:

* b1b1b1a Deprecated command isday().
* 5e4e060 Merged petskillattack and petskillattack2.

MishimaHaruna:
ca877ec Deprecated command checkre().
ad9a149 Replaced checkre() calls with constants
513059c Removed deprecated command isday
c292b39 Added support for RENEWAL constants in scripts

AnnieRuru:
recovery and globalmes can remove now, nobody actually using them

What the... why remove the *awake script command ?
if you remove them, all my battleground script will bugged

the reason I still prefer to use sleep+awake rather than OnTimerxxx:
is because I can dynamically set a random variable to sleep them
```
sleep rand(10000,30000); // pause the script between 10~30 seconds
...
if ( <something happens> .. eg: player log out )
awake strnpcinfo(0);
```
however there is no way I can do the rand() with the OnTimerxxxx:
because the timer has be fixed

Ancyker:
I also use awake. While timers can be used in a random fashion, it's overly complicated and limits the use. Removing awake will encourage complex and messy scripts.

Ancyker:
Upon further thought, the work around for removing awake and not having messy timer code is adding another NPC. That's unacceptable. 

Increasing memory usage because "people use it wrong" is like saying "we're removing tables because people are using it when they should use a div and css". While people abuse tables in html (namely using tables for the page layout), it doesn't mean tables don't have a legitimate use. They are super useful when you want, you know, a table. Awake is useful for when you want randomness, because it's easier and cleaner than a regular timer, especially if you want to also use a regular timer.

MishimaHaruna:
Good points about `awake`, thank you both. It won't be removed unless a valid alternative will be provided. The reason why I wanted that gone is that I'm not too fond of scripts pausing in the middle as it happens with the `sleep`/`awake` pattern (which is often misused to create a `while(1)` run loop inside NPCs, waiting for events).

@AnnieRuru in the case you describe, I see that you mention 'player logout' as an example of 'something happens'. If we disregard the part about time interval duration randomness, which is perfectly valid and I can't suggest an alternative for right now, wouldn't it make more sense to use a player-attached timer and the `OnTimerQuit:` label in similar scenarios? That way you wouldn't have a script running permanently while waiting, but just a timer.

AnnieRuru:
`That way you wouldn't have a script running permanently while waiting, but just a timer.`
WAHHH .... now you say that, this is DEFINITELY true

@MishimaHaruna please check this sample battleground script
https://github.com/rathena/rathena/blob/master/npc/custom/battleground/bg_pvp.txt
this is a sample battleground script that has been inside rathena emulator since Fed 2014
this is how I used the sleep command, so the `.eventlasting` variable can be dynamically set

yes it is possible to use OnTimerxxxx: to replace the sleep command
if I want to `if disregard the part about time interval duration randomness`

the question I want to ask is
issit changing the `sleep;` into `initnpctimer; end;` and `awake strnpcinfo(0);` into `goto L_endmatch;`
will save server memory ?

MishimaHaruna:
While I don't have actual figures to back this, a timer should have a smaller memory footprint than a paused script.
As far as I'm aware, when you start an NPC timer, you're merely storing the NPC and RID, and starting a Hercules timer. When the timer ticks, it'll run the appropriate label in the given NPC, after attaching the RID, where necessary.
Using sleep/sleep2 instead, makes it keep the script state (including all local variables/arrays), and at the same time start a Hercules timer, to know when to re-run the script with the state loaded.

Unfortunately, there is no way to specify a dynamic duration for NPC timers (other than hacking something around setnpctimer), unless you can use addtimer (which only works with a player attached).

AnnieRuru:
`Using sleep/sleep2 instead, makes it keep the script state (including all local variables/arrays), and at the same time start a Hercules timer, to know when to re-run the script with the state loaded.`
I think this is the reason why most members prefer to stick to *sleep rather than npctimers
yes, the .@var type variable destroyed the moment it hit `end;` command
and many of us love to use this type of variable

because when we need to differentiate each player/party/guild
we usually use .@var type variable, then pause the script
if we use .var type variable, have to use setd/getd to simulate 2 dimension array to store the information
but with .@var type, this variable is unique to each script instance run, which makes it very convenient

thanks for the clarification, I really appreciate it =)

MishimaHaruna:
Yes, of course my last comment was based on the assumption that you don't need to save any local state during the wait, which is the case in the example script you link (after the line `sleep .eventlasting * 1000;` it only needs to know the global state, which you store in npc variables).

Having to resort to getd/setd is absolutely not desirable either.

MishimaHaruna:
When you mention

> so even if the one who attached the script logout
npc will still able to pronounce that player's party name

I assume you don't mean that the player is currently attached, do you? (what you describe should only happen if you detachrid, or the script would get terminated if I recall correctly)

AnnieRuru:
yes, of course the game/event should continue when the one who register for the party logout

in a party-based event script, there should always no player attaching the script
should save the party ID, then run the loop with getpartymember

if the script is constantly attaching to the player, it means if the one who register for the party logout,
the event will get stuck, because the timer stop when the attaching player logout
other party members will wonder why the monster isn't spawning ... etc

however, npc timer can also do this by saving more extra variables,
just like Ancyker said, it made the script more messy with npc timers compare to sleep

MishimaHaruna:
All the cases you described are suitable to be handled with a timer attached to the npc (events or games involving more than one player at a time). In any case, even if you use the sleep/awake/loop hack, you still need to use some global (likely npc) variables to store the event state (you don't want other people interacting with the npc and starting the script concurrently, now do you?). I'm confident that in those cases by using timers you can write a better structured and more efficient script.

AnnieRuru:
ok, let's do a live example then 
I do have a very complex script that uses a very complex timer system in this battleground script
http://herc.ws/board/topic/4514-wow-domination-battleground-olrox-is-not-invited/?p=29304

ok @MishimaHaruna , you don't have to click on the topic link,
because that battleground script is very very complex
so I highlight one part of this system

there are 5 flags to capture to win this game
when the one side of the battleground team capture the flag
if Alliance captured 1 flag, earn 1 point every 12 seconds
if Alliance captured 2 flags, earn 1 point every 9 seconds
if Alliance captured 3 flags, earn 1 point every 6 seconds
if Alliance captured 4 flags, earn 1 point every 3 seconds
if Alliance captured 5 flags, earn 3 points every 1 second
the side that earn 100 points first wins

and this is easily done with while(1) sleep x;
http://upaste.me/79a421835bfa299ad

now, how to do this with npc timers ?
this is going to be very ugly
http://upaste.me/295821836161a6278
look, if the team only captured 1 flag, it has to go to OnTimer12000:
however, the script will invoke OnTimer1000: labels and so on

do you have a better way of doing this using npc timers ?

MishimaHaruna:
Okay, I'll have a try at it. If I'm not forgetting anything, here you go: http://upaste.me/4e84be

AnnieRuru:
`	// Ugly hack (but still not nearly as ugly as sleep()), until timers can have dynamic labels.
	setnpctimer(60000 - .tick[.capturedamount]);`
HAHAHAHA !!! the moment I see this part I laugh so hard !
but yeah, now you convinced me already.
I have nothing else to say, but you are right.
*hands down*

AnnieRuru:
please add `classchange`, its useless now
we already have `changebase` and `changelook`

PR #397, opened 2014-11-17
Add item attribute what prevent item removing on use.

4144:
New item attribute: nodelonuse


kyeme:
I hope you fix this bug xD http://hercules.ws/board/index.php?app=tracker&showissue=5966&st=0#entry12939

PR #431, opened 2014-12-31
Cleaned up spirit charms implementation

MishimaHaruna:
Please do not merge (it's still a work in progress). I'm opening this just for discussion.

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -1251,8 +1251,8 @@ int64 battle_calc_defense(int attack_type, struct block_list *src, struct block_
 					def1 -= def1 * i / 100;
 					def2 -= def2 * i / 100;
 				}
-				if ( sd->spiritcharm[SPIRITS_TYPE_CHARM_LAND] > 0 ) // hidden from status window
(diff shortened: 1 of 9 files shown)


rud0lp20:
:+1:

MishimaHaruna:
I believe this is ready to merge, if anyone can review it.

I decided not to merge spirit balls and spirit charms implementations because of the different maximum amounts (and uncertainty as to whether it's officially possible to create a condition for a character to have both spirit balls and spirit charms at the same time).

rud0lp20:
not yet merge to spiritball?
' (and uncertainty as to whether it's officially possible to create a condition for a character to have both spirit balls and spirit charms at the same time).' 
-yes it is possible...

MishimaHaruna:
Then I think it's better to keep the implementation separated... if they were mutually exclusive, it'd make more sense to merge them and re-use the timers array. (or if you have a suggestion on an efficient way to merge them otherwise, I'd be happy to see it :3)

rud0lp20:
ok then..let's merge this...

PR #434, opened 2015-01-03
INF2_NOLP update. BL direction fixes. Combo minimum time increased.

csnv:


src/map/unit.c
```diff
@@ -1599,6 +1599,8 @@ int unit_skilluse_id2(struct block_list *src, int target_id, uint16 skill_id, ui
 	ud->skill_lv      = skill_lv;
 
 	if( casttime > 0 ) {
+		if (src->id != target->id) // self-targeted skills shouldn't show different direction
```
rud0lp20: ow... gonna check this on official...

csnv: These changes are meant to fix some visual glitches rather than changing behavior. I haven't see any direction change when a char casts a self targeted skill.

PR #435, opened 2015-01-03
Extend skills for using custom skills from plugins.

4144:


rud0lp20:
I don't think this is necessary..pre and post hooking is enough for me to create custom skills into a plugin..just saying...

4144:
Some functions near 1000 lines, for add own skill exist two ways:
1. copy function like this to plugin. this mean bunch of unmaintained code in plugin.
2. Reimplement part of this function and allow execution in old function too. this double executed code and slowdown code execution twice.

Both of this i think is bad. One for maintainability and second for performance.

With special hooks solved both issues.

MishimaHaruna:
I agree that this helps making plugins more maintainable, by providing hook points in the middle of large/complex functions.
I'm merging it, thank you very much :3

PR #437, opened 2015-01-08
Allow load custom jobs/classes by plugins from skill_tree.conf

4144:


src/map/pc.c
```diff
@@ -10242,6 +10254,15 @@ void pc_read_skill_tree(void) {
 		clif->skillinfoblock(sd);
 	mapit->free(iter);
 }
+
+int pc_convert_custom_class_name_to_id(const char* name)
+{
+	int val = -1;
+	if (script->get_constant(name, &val))
```
MishimaHaruna: This seems a good idea. I wonder if we should use this same method for official jobs, rather than hardcoding that ugly list in pc.c.

I'll take some time to think about this. In case, I'll make a pull request against your fork, to update this.

MishimaHaruna:
@shennetsind could you chime in?

What do you think of using `script->get_constant` (or any other similar method) to get rid of the `jnames` struct array in `pc_read_skill_tree`?

Performance isn't an issue here, because the skill tree is read once and never again. The main issue here is code maintainability / getting rid of those hardcoded strings.

dastgir:
@shennetsind 
could you please review it, its been pending from 2 months..

4144:
This pull request dont need anymore. Same features can be implimented from master branch already.

PR #438, opened 2015-01-09
Convert pcdb_checkid into function in pc interface

4144:


MishimaHaruna:
Makes sense for plugins. I'm merging this, thank you very much.

PR #443, opened 2015-01-17
Extract exp table validation into separate function.

4144:


src/map/pc.c
```diff
@@ -10354,16 +10354,7 @@ int pc_readdb(void) {
 		}
 	}
 	fclose(fp);
-	for (i = 0; i < JOB_MAX; i++) {
-		if (!pc->db_checkid(i)) continue;
-		if (i == JOB_WEDDING || i == JOB_XMAS || i == JOB_SUMMER)
-			continue; //Classes that do not need exp tables.
-		j = pc->class2idx(i);
```
MishimaHaruna: if I'm not mistaken, now that this line is removed, the `j` variable is unused in this function?

4144: No, j variable used in other lines. For example in RENEWAL_DROP RENEWAL_EXP

MishimaHaruna: You're certainly right, I can indeed see it now. Inspecting code through this web application isn't the same as checking the actual file on a proper text editor.

MishimaHaruna:
Thank you!

PR #444, opened 2015-01-17
some fixes for different issues

4144:


src/char/char.c
```diff
@@ -1461,7 +1461,7 @@ bool char_char_slotchange(struct char_session_data *sd, int fd, unsigned short f
 	struct mmo_charstatus char_dat;
 	int from_id = 0;
 
-	if( from > MAX_CHARS || to > MAX_CHARS || ( sd->char_slots && to > sd->char_slots ) || sd->found_char[from] <= 0 )
+	if( from > MAX_CHARS || to >= MAX_CHARS || ( sd->char_slots && to > sd->char_slots ) || sd->found_char[from] <= 0 )
```
MishimaHaruna: How about `from`? I think that one should also be `< MAX_CHARS` (i.e. its check changed to `>=` as well)?

4144: fixed

shennetsind:
Thank you, your contribution is much appreciated.

PR #445, opened 2015-01-18
Command line arguments handling overhaul

MishimaHaruna:
- login_server, char_server, map_server as well as the tools (mapcache)
  now have a common command line arguments handling mechanism.
- All of them now accept `--help` (`-h`), `--version` (`-v`) and
  `--load-plugin`.
- login_server now accepts `--login-config` and `--lan-config` instead
  of relying on positional arguments to override those files. The old
  syntax will no longer work, please update your custom startup scripts.
- char_server now accepts `--char-config`, `--inter-config`,
  `--lan-config` instead of relying on positional arguments. The old
  syntax will no longer work, please update your custom startup scripts.
- mapcache now accepts `--grf-list`, `--map-list`, `--map-cache`,
  `--rebuild` in place of, respectively, `-grf`, `-list`, `-cache`,
  `-rebuild`.
- A new macro `CMDLINEARG()` is provided, to help defining new command
  line argument handlers (i.e. in plugins). the `addArg()` call is still
  required, but its syntax has changed. The `help` argument is now of type
  `const char *` rather than a function pointer, and it is supposed to
  contain the message to show in the `--help` screen. Pass `NULL` if no
  help message is desired.

Signed-off-by: Haru <haru@dotalux.com>

### DIFF
--- a/src/char/char.c
+++ b/src/char/char.c
@@ -50,10 +50,6 @@
 #endif
 
 // private declarations
-#define CHAR_CONF_NAME "conf/char-server.conf"
-#define LAN_CONF_NAME  "conf/subnet.conf"
-#define SQL_CONF_NAME  "conf/inter-server.conf"
-
 char char_db[256] = "char";
 char scdata_db[256] = "sc_data";
 char cart_db[256] = "cart_inventory";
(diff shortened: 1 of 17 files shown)


milky-milk:
Omg, this is awesome!

PR #446, opened 2015-01-18
Fixes direction in AOE skills

csnv:


shennetsind:
nice catch, so they were zero all this time

PR #450, opened 2015-01-21
Mapflags to plugin

dastgir:


src/map/map.h
```diff
@@ -803,6 +803,16 @@ struct map_cache_map_info {
 	int32 len;
 };
 
+// Mapflag Functions
+#define CHECK_MAPFLAG( cmd ) do { if ( map->list[ sd->bl.m ].flag.cmd ) clif->message(sd->fd,#cmd); } while(0)
+#define SET_MAPFLAG( cmd ) do { \
```
MishimaHaruna: I'm not sure about moving those macros into a .h file, since they rely on local variables of the original caller function, and weren't meant to be used elsewhere. I'd keep these two inside a .c file, and `#undef` once done, to restrict namespace cluttering (and provide some other way for plugins to do the same, or perhaps let plugins do it themselves)

dastgir: Maybe yes, let plugin do it themself,
will push a commit soon.

PR #453, opened 2015-01-25
Fix plugins loading in map server

4144:


shennetsind:
I don't think this is right, HPMDataCheck isn't meant to be exclusive to interfaces (and editing it directly doesn't solve anything, that file is script-generated)

4144:
HPMDataCheck was generated.

If you mean in HPMDataCheck must be different structures and not interfaces, i can remove this changes.

MishimaHaruna:
Whoops, my apologies for missing this. Thanks for the fix :3

PR #458, opened 2015-01-30
Missing break in skill

Nihadm89:
Lightning Spear of Ice

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -1509,6 +1509,7 @@ int battle_calc_skillratio(int attack_type, struct block_list *src, struct block
 					skillratio -= 30;
 					if (sd && sd->charm_type == CHARM_TYPE_WATER && sd->charm_count > 0)
 						skillratio += 5 * sd->charm_count;
+					break;
 #endif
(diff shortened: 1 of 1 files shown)


MishimaHaruna:
Judging by the comment right after, it seems like the missing break was intended, in order to fall through into the next case.

Nihadm89:
Intended or not it breaks the lightning spear skill. By putting the break in there the skill goes back to official damage.

MishimaHaruna:
Ah, I see. I'll reopen then, thank you!

MishimaHaruna:
This piece of code was added in 6dd08bef. @rud0lp20, can you confirm that the lack of `break` was an oversight in that commit?

rud0lp20:
yeah..missing break...

MishimaHaruna:
@rud0lp20 would you mind checking the other `/* Fall through */` comments in the same file if you can? They mark all the instances of omitted `break` (some of them might be intended, some might not perhaps)

rud0lp20:
ok...:D

PR #459, opened 2015-01-31
Add mob ai attribute what alow for slave stay alive after master was dead

4144:


shennetsind:
Is this mode official? What uses it?

4144:
No, not official.
I using it in own server.
And it not conflict with RO client.

PR #462, opened 2015-02-13
Fix wrong id usage in clif_addskill.

4144:
Was used skill id, but must be used skill idx.


shennetsind:
Thank you, your contribution is much appreciated.

PR #463, opened 2015-02-21
Improve performance with many mobs in small area

4144:
After both this commits and changed in configuration:
official_cell_stack_limit to 0
check_occupied_cells: 0
On test server in small room i spawned 7k mobs and CPU was used only near 20%


src/map/battle.c
```diff
@@ -6725,9 +6725,10 @@ static const struct battle_data {
 	{ "bone_drop",                          &battle_config.bone_drop,                       0,      0,      2,              },
 	{ "buyer_name",                         &battle_config.buyer_name,                      1,      0,      1,              },
 	{ "skill_wall_check",                   &battle_config.skill_wall_check,                1,      0,      1,              },
-	{ "official_cell_stack_limit",          &battle_config.official_cell_stack_limit,       1,      1,      255,            },
+	{ "official_cell_stack_limit",          &battle_config.official_cell_stack_limit,       1,      0,      255,            },
```
shennetsind: shouldn't this be set to 0 on conf/battle/misc.conf too?

4144: If set it to 0 will be not official default settings.
Or default settings can be not official?

shennetsind: Oh I see, my bad I counted the column wrong

shennetsind:
Your contribution is much appreciated.

PR #465, opened 2015-03-03
Updated item_db with New Costume's

dastgir:
Item Details are from itemInfo.lua (Provided by kRO)
More Detail of this file: http://hercules.ws/board/topic/8552-new-costume-headgear-db/
Contains almost all costumes from iteminfo.lua
Credits to @zackdreaver for sharing the db file on the board.


shennetsind:
Alright! Thank you both its much appreciated.

PR #470, opened 2015-03-17
Corrected homunculus stats for pre-re.

milky-milk:


MishimaHaruna:
Could you provide some info on where those values are taken from?

milky-milk:
Hello.
Those values was taken from eAthena and irowiki (http://irowiki.org/classic/Vanilmirth#Stats). Now in /pre-re/homunculus_db.txt all homunculus are with renewal stats. For example: in pre-re evolved Vanilmirth can't have more than 133 str.

shennetsind:
Thank you for your contribution is much appreciated.

PR #471, opened 2015-03-18
Fix sending 0x908 packet.

4144:


shennetsind:
Your contribution is much appreciated. Thank you

PR #473, opened 2015-03-19
Fix for "Solo at Sphinx Dungeon!" Quest

Inzianity:
Solo at Sphinx Dungeon Quest should require Zerom to be hunted.


dastgir:
Thanks

PR #475, opened 2015-03-22
Add channels.conf option irc_channel_autojoin for auto joining irc channel.

4144:


malufett:
what if instead of auto join for irc is make it also available to all channel(except maps)?

MishimaHaruna:
Thank you very much

PR #478, opened 2015-03-29
Avoid binding rental items.

emistry:
Prevent the NPC to show rental items when players try to bind an items.

Image


dastgir:
Maybe a source way to avoid binding??

emistry:
yes, perhap source way would be better. Anyway I just providing a temporary solution to fix the possible exploits. =D

dastgir:
Main reason is , I doubt if someone uses this npc, I guess many have custom scripts been made to suit their needs....
So maybe edit the scriptcommand of bind to not allow rental to be binded???

malufett:
@dastgir 
it can't be done since the script command 'getitembound2/getitembound' won't get the data of the item it has been passed it creates new one..so the best remedy is via scripting...

@emistry 
:+1: yup this does the trick....

PR #480, opened 2015-03-31
Added new items DB and corrected scripts (part 2)

zackdreaver:
Please check the script before merge.


zackdreaver:
Ok, i think its ready

dastgir:
OK, all seems good

PR #483, opened 2015-03-31
Allow sell items to market shop.

4144:
Not sure is gravity client support this, but for other clients it works.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/483)


hemagx:
Unfortunately Gravity client can't do it, and to not leave place for any possible issues we will close this for now and add support later if needed or president in gravity client.
Thank you.

4144:
Hm, hercules not only for gravity clients.
And this is only simple condition change

PR #484, opened 2015-04-01
Added more 'Old Glast Heim' instance items

zackdreaver:


dastgir:
Thanks :), seems like someone forgot to put items when script was put ...

PR #491, opened 2015-04-08
Fix some issues with the @help command.

k-py:
Fix help messages being trimmed due to an incorrect length value.
Fix help messages formatting: use clif_displaymessage2 to properly display newlines.


MishimaHaruna:
Merged, thank you very much!

PR #493, opened 2015-04-10
Hercules 'job_db1.txt' Redesign

malufett:


MishimaHaruna:
Holy, I didn't notice this until too late... This introduces some critical issues.

src/map/status.c
```diff
@@ -12086,33 +12054,178 @@ int status_get_sc_type(sc_type type) {
...
+		}
+
+		if ( (temp = libconfig->setting_get_member(jdb, "HPTable")) ) {
+			int level = 0, ave, total = 0;
+			config_setting_t *hp = NULL;
+			while ( (hp = libconfig->setting_get_elem(temp, level++)) ) {
+				status->HP_table[idx][level] = i32 = min(libconfig->setting_get_int(hp), battle_config.max_hp);
+				total += i32 - status->HP_table[idx][level - 1];
+			}
+			ave = total / (level - 1);
```
MishimaHaruna: This is a division by zero.

src/map/status.c
```diff
@@ -12086,33 +12054,178 @@ int status_get_sc_type(sc_type type) {
...
+		}
+
+		if ( (temp = libconfig->setting_get_member(jdb, "SPTable")) ) {
+			int level = 0, ave, total = 0;
+			config_setting_t *sp = NULL;
+			while ( (sp = libconfig->setting_get_elem(temp, level++)) ) {
+				status->SP_table[idx][level] = i32 = min(libconfig->setting_get_int(sp), battle_config.max_sp);
+				total += i32 - status->SP_table[idx][level - 1];
+			}
+			ave = total / (level - 1);
```
MishimaHaruna: Another division by zero.

src/map/status.c
```diff
@@ -12086,33 +12054,178 @@ int status_get_sc_type(sc_type type) {
...
+				}
+			}
+		}
+
+		if ( (temp = libconfig->setting_get_member(jdb, "HPTable")) ) {
+			int level = 0, ave, total = 0;
+			config_setting_t *hp = NULL;
+			while ( (hp = libconfig->setting_get_elem(temp, level++)) ) {
+				status->HP_table[idx][level] = i32 = min(libconfig->setting_get_int(hp), battle_config.max_hp);
+				total += i32 - status->HP_table[idx][level - 1];
```
MishimaHaruna: `total` is completely unnecessary, for it will always have the same value as `status->HP_table[idx][level]`.

src/map/status.c
```diff
@@ -12086,33 +12054,178 @@ int status_get_sc_type(sc_type type) {
...
+				status->HP_table[idx][level] = min(ave * level, battle_config.max_hp); /* some are still empty? then let's use the average increase */
+			}
+		}
+
+		if ( (temp = libconfig->setting_get_member(jdb, "SPTable")) ) {
+			int level = 0, ave, total = 0;
+			config_setting_t *sp = NULL;
+			while ( (sp = libconfig->setting_get_elem(temp, level++)) ) {
+				status->SP_table[idx][level] = i32 = min(libconfig->setting_get_int(sp), battle_config.max_sp);
+				total += i32 - status->SP_table[idx][level - 1];
```
MishimaHaruna: Same as above.

src/map/status.c
```diff
@@ -12086,33 +12054,178 @@ int status_get_sc_type(sc_type type) {
...
+				} else {
+					ShowWarning("status_read_job_db: unknown weapon type '%s'!\n", wname);
+				}
+			}
+		}
+
+		if ( (temp = libconfig->setting_get_member(jdb, "HPTable")) ) {
+			int level = 0, ave, total = 0;
+			config_setting_t *hp = NULL;
+			while ( (hp = libconfig->setting_get_elem(temp, level++)) ) {
```
MishimaHaruna: What if there are more entries in the libconfig file than the `MAX_LEVEL`? Considering the size of `HP_table[]` is `MAX_LEVEL+1`... Segmentation fault candidate?

src/map/status.c
malufett: I'm aware of it..thinking that no one will add HPTable:[]  in there db..XD

src/map/status.c
malufett: nope..cause what if the values are not same in divisor?

src/map/status.c
malufett: forgot to check on this one...XD

src/map/status.c
MishimaHaruna: How not? `total` starts at 0 and, at each iteration adds the difference of the current value with the previous value... so at the end of the loop, it'll contain:

```
HP_table[1] - HP_table[0] +
HP_table[2] - HP_table[1] +
HP_table[3] - HP_table[2] +
HP_table[4] - HP_table[3] +
HP_table[5] - HP_table[4] +
 ...
HP_table[X] - HP_table[X-1]
```

And with some basic maths (simplify away each `HP_table[n] - HP_table[n]`), it ends up being `HP_table[X] - HP_table[0]`... and since `HP_table[0] == 0`, then `total == HP_table[X]` (where `X` == `level`)

MishimaHaruna:
I'll commit a fix asap

MishimaHaruna:
Addressed issues in 7160872b900e0534b7355cc5f7d232c326a00cca. I also edited the HP/SP table guessing algorithm to take into account the SP or HP at level 1 (it now uses a formula such as `[value at level 1] + increment * (levels-1)`).

If you spot any issues, please let me know (or follow-up)

PR #495, opened 2015-04-10
Allow for plugins read custom fields from jobdb1.

4144:


malufett:
custom fields? I don't see any benefits from this..:D

dastgir:
@malufett , i see some benefits ,
Like what if I want to give extra bonus based on jobs, I can input extra fields in it, I also had a job faction, which would make it easier to implement using this plugin,  since I can just put Id here in extra custom field and avoid tons of if else and what not.

4144:
Yes for example i want add some kind of bonus. I creating in plugin array for each job, hooking functions status_read_job_db_sub with pre or post hook, and here i can read own bonuses and write to own array. After hook some function where this bonuses need to use. and done

malufett:
ow..I see..but isn't appropriate to use `job_db.conf` file for custom fields..well in fact you can just hook 'status_readdb' and make your own preference configuration file that will suited well for desired custom fields? cause this custom fields may conflict to the default functionality of `job_db.conf`...

4144:
@malufett and then need copy to plugin all db reading code for all fields, if i want add only two lines?

And who will update and maintaine this copy/paste code in plugin if something changed in server code?
Plugins need for split parts. Work what should do server is doing server. And some one can hook any small part.

Without splitting to small part enought to hook clif_parse, all other you can copy to plugin. But no one doing this :)

MishimaHaruna:
I'm for merging this pull request. I think it's the cleanest and more maintainable way. We allow plugins to add additional fields to other databases, why shouldn't we do the same for the job db?

@malufett what kind of conflicts are you thinking about? I'm not really sure what kind of problems it could generate o.o

src/map/status.c
```diff
@@ -12092,139 +12085,154 @@ void status_read_job_db(void) { /* [malufett/Hercules] */
...
+		ave = total / (level - 1);
+		for ( ; level <= pc->max_level[idx][0]; level++ ) {
+			status->SP_table[idx][level] = min(ave * level, battle_config.max_sp);
 		}
 	}
+}
+
+/*------------------------------------------
+* DB reading.
+* job_db1.txt    - weight, hp, sp, aspd
```
EPuncker: rename it to job_db.conf :P

MishimaHaruna:
Thank you very much

PR #501, opened 2015-04-21
Fixed a bug in Einbech "Oridecon/Mineral Quest"

michaelforge:
Because of some very old typo, this quest could never reach Minerals exchange dialog.

To save current server's progress, it's a good idea to rename `$einamanoama` to `$ein_amano` at **`mapreg`** table and make sure it's not > 60.


dastgir:
Thanks for contribution.

PR #502, opened 2015-04-21
Fix script function return always removing references from NPC variables.

k-py:
Bugreport:8642 http://hercules.ws/board/tracker/issue-8642-script-engine-issue-quo/

### DIFF
--- a/npc/dev/test.txt
+++ b/npc/dev/test.txt
@@ -86,6 +86,15 @@ function	script	F_TestNPCArrays	{
 	return getarraysize(.y);
 }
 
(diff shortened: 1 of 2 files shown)


k-py:
I removed an unnecessary check and rebased the branch.

k-py:
After further testing, I uncovered a crash condition caused by the incorrect usage of return outside the scope of a function or subroutine.
```
-	script	test	-1,{
	return getvariableofnpc(.x, "test2");
}
-	script	test2		-1,{
	end;
}
```
Already fixed.

MishimaHaruna:
Thank you very much

PR #505, opened 2015-04-26
Item Restrictions Updated to match info from kRO

dastgir:
Data from itemmoveinfov5.txt is been used as reference.
Thanks to Rytech for giving us the hint that the file is been put on client.
@MishimaHaruna @Streusel @malufett 
If anyone want to review it?
Else I will merge it in few days, if there's no problem with it.

### DIFF
--- a/db/re/item_db.conf
+++ b/db/re/item_db.conf
@@ -521,6 +521,9 @@ item_db: (
 	Type: 0
 	Buy: 1650
 	Weight: 50
+	Trade: {
+		noselltonpc: true
+	}
 	Script: <" itemheal rand(325,405),0; ">
 },
 {
@@ -1079,6 +1082,15 @@ item_db: (
 	Type: 0
 	Buy: 50
 	Weight: 10
+	Trade: {
+		nodrop: true
+		notrade: true
+		noselltonpc: true
+		nocart: true
+		nogstorage: true
+		nomail: true
+		noauction: true
+	}
 	Script: <" itemheal rand(45,65),0; ">
 },
 {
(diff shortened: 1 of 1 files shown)


dastgir:
Do not merge till I commit a next fix

dastgir:
Nodrop bug fixed.

dastgir:
Following file is converted from data.grf
And there's one on rdata.grf(which is having some other restrictions than data.grf), but I think rdata keeps changing as its sakray and we should prefer one on main server.
Which should be so preferred? @Streusel @malufett @MishimaHaruna

MishimaHaruna:
data.grf should be preferred, for sakray is an ever-changing beta-server, and not everything that is tested on sakray makes it to the main servers.

dastgir:
Will wait for a day before merging, if ever someone wants to leave their thoughts.

MishimaHaruna:
Please wait until someone can confirm that items like White_Slim_Potion, Old_Blue_Box, Old_Violet_Box, really can't be traded/dropped on official servers, while other items, such as the Brave/Valorous items, or even the wedding rings can. It looks very suspicious.

dastgir:
@MishimaHaruna 
Ok I confirmed via client: http://i.imgur.com/WAhDrGp.jpg 
The Format mentioned in client is wrong(the header says CanDrop, but 1 = Cannot drop)
So All thing will be reversed.
will commit a fix soon.

dastgir:
Now Its fine, I reversed the order in my tool(which is what was intended),
Just review it :)

MishimaHaruna:
Yup, seems much more reasonable now. I can't possibly go through each and every item, but I checked a reasonably large sample, and the restrictions seem correct. I believe it's good enough to merge.

Is there anything like this for pre-renewal from any official servers, by the way?

dastgir:
Its hard to get information from pre-re,
since we got this text file directly into client, so I guess pre-re should be untouched.

MishimaHaruna:
I agree. Better to leave pre-renewal untouched, since we don't know whether some restrictions were added or changed afterwards.

PR #506, opened 2015-04-27
Added Champion Monsters

dastgir:
Added MobDB/Skills/Spawn locations of champion monsters.
Credits to idAthena


dastgir:
@MishimaHaruna 
Any Problem if I merge this(since this info is mostly from idAthena and divine-pride

MishimaHaruna:
No problem from my side, but it'd be nice, for history-tracking purposes, to edit the commit message (i.e. git commit --amend) and include a line that goes along the lines of "Based on info from idAthena and Divine-Pride", in case a git-blame lands here in future.

PR #508, opened 2015-05-01
Extend sanity check configure flag with all sanity checks.

4144:
Also fix some issues what was found by new checks


MishimaHaruna:
Thank you very much

PR #509, opened 2015-05-05
Typo when setting JoinAnnounce for channels

Nihadm89:
The comment is right, but the wrong message number is there


dastgir:
Thanks for contribution

PR #510, opened 2015-05-05
Update script_commands.txt

MarvinLamart:
Actually, without "(" and ")" it triggers an error when script is loaded

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -5905,7 +5905,7 @@ controlling each of the spawned mobs with the unit* commands shown below.
 For example:
 
 	// We'll make a poring which will automatically attack invoking player:
-	.@mobGID = monster "Prontera",150,150,"Poring",PORING,1; // PORING is defined in the mob db and its value is 1002
+	.@mobGID = monster ("Prontera",150,150,"Poring",PORING,1); // PORING is defined in the mob db and its value is 1002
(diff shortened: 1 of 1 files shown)


dastgir:
Ohh wait, without () it shouldn't trigger error....

doc/script_commands.txt
```diff
@@ -5905,7 +5905,7 @@ controlling each of the spawned mobs with the unit* commands shown below.
 For example:
 
 	// We'll make a poring which will automatically attack invoking player:
-	.@mobGID = monster "Prontera",150,150,"Poring",PORING,1; // PORING is defined in the mob db and its value is 1002
+	.@mobGID = monster ("Prontera",150,150,"Poring",PORING,1); // PORING is defined in the mob db and its value is 1002
```
dastgir: Could you possible remove space between 'monster' and '('

dastgir:
So got your point, setting it to variable requires ()
But could you possibly remove the space between monster and ( as I mentioned in commit.

MishimaHaruna:
Also please change `"Prontera"` to `"prontera"` while you're editing that line. `"Prontera"` would be an invalid map ID.

MarvinLamart:
No problem, its done :)

dastgir:
Thanks for contribution :)

MarvinLamart:
It's a pleasure to help at my level x)

PR #512, opened 2015-05-06
WoE, Ur, Sapha, Nab, Rogue Treasure's combo effect

Inzianity:
-Fixed to make WoE Suit and WoE Boots get its set bonus when combined with WoE Manteau instead of WoE Muffler
-Ur Set, Nab Set, Sapha Set item combo effect fix
-Fixed Rogue of Treasure's combo effect with Cold Heart and Black Cat


EPuncker:
:+1:

PR #514, opened 2015-05-13
Changed close; into end; to try to fix bug 8666, since there is no di…

EPuncker:
Changed close; into end; to try to fix bug 8666, since there is no dialogue window a close; is not needed

http://hercules.ws/board/tracker/issue-8666-any-work-in-progress-npc-dialog-manufacturing-quit-and-try-again/


MishimaHaruna:
Thank you!

PR #517, opened 2015-05-14
Preview of new items DB and corrected scripts (part 5)

zackdreaver:
Based on AEGIS and official description (KRO preferable), Added : 
- Fixed bug#8625
- Added new items (Misc, equipments and costumes)
- Corrected some items script

In PR included RC_Player


db/re/item_db.conf
```diff
@@ -5855,9 +5855,11 @@ item_db: (
 	View: 3
 	Script: <"
 		bonus bFlee2,10;
-		if(JobLevel>=70) autobonus "{ bonus bBaseAtk,50; }",10,10000,BF_WEAPON,"{ specialeffect2 EF_POTION_BERSERK; }";
+		if(JobLevel=70)||(JobLevel>=50)&&(Class==Job_Rune_Knight||(JobLevel>=50)&&Class==Job_Rune_Knight_T) {
```
EPuncker: using JobLevel=70 instead of JobLevel>=70 will break this item for all servers that uses a higher max jlvl, and by the way that statement doesn't looks logic since it will always be true for both cases

jlvl >= 50 will trigger the jlvl =70 as well

I'd use something like:
if(JobLevel >= 70) && Class == Job_Rune_Knight_T) || (JobLevel >= 50 && Class == Job_Rune_Knight)

instead :+1:

MishimaHaruna: On a side note, `JobLevel=70` is in any case incorrect (in case, it would be `==`)

The statement is also incorrectly parenthesized. The version suggested by @EPuncker seems correct.

zackdreaver:
This PR is just a "preview", not to be merged

I will add more items while waiting for RC_Player and some other implementations,
if there any future.

zackdreaver:
... i will make new PR... this one will ready to merge after RC_Player implement

PR #521, opened 2015-05-21
Fix timer for Eden quest 110+

magimagica:
Fixed bug#8692
http://hercules.ws/board/tracker/issue-8692-eden-quests-111-120/


dastgir:
Thanks for contribution

PR #527, opened 2015-05-25
Dump stack in nullpo*/Assert* function.

4144:
Also enable in configure linker flag -rdynamic.


src/common/nullpo.c
```diff
@@ -32,6 +41,13 @@ void assert_report(const char *file, int line, const char *func, const char *tar
 
 	ShowError("--- %s --------------------------------------------\n", title);
 	ShowError("%s:%d: '%s' in function `%s'\n", file, line, targetname, func);
+#ifdef __GNUC__
+	size = backtrace (array, 10);
+	strings = backtrace_symbols (array, size);
+	for (i = 0; i < size; i++)
+		ShowError("%s\n", strings[i]);
+	free (strings);
```
MishimaHaruna: This requires including `<stdlib.h>`

src/common/nullpo.c
```diff
@@ -32,6 +41,13 @@ void assert_report(const char *file, int line, const char *func, const char *tar
 
 	ShowError("--- %s --------------------------------------------\n", title);
 	ShowError("%s:%d: '%s' in function `%s'\n", file, line, targetname, func);
+#ifdef __GNUC__
+	size = backtrace (array, 10);
+	strings = backtrace_symbols (array, size);
```
MishimaHaruna: According to the man page, the signature for this function is:

```
     char**
     backtrace_symbols(void* const* array, int size);
```

Having `size` as a `size_t`, causes a warning (-Wshorten-64-to-32). The variable should be changed to int, which is what `backtrace()` returns -- at least on BSD/Darwin. Please let me know if it's different on GNU/linux).

src/common/nullpo.c
4144: From man (http://linux.die.net/man/3/backtrace_symbols)
It need only #include <execinfo.h>

4144:
Updated

MishimaHaruna:
Thank you very much

PR #531, opened 2015-06-02
Plugin load failure now causes Fatal Error

Ancyker:
If you set a plugin to load and it fails for some reason, the server
will now exit. It will still warn when attempting to load a plugin
twice.


MishimaHaruna:
Thank you very much!

PR #558, opened 2015-06-14
Added Configuration in enabling/disabling monster hp bar on new clients.

Jedzkie:
Added Configuration in enabling/disabling monster hp bar on new clients.

corrected some white spaces and added missing EOL.


Jedzkie:
@MishimaHaruna please correct my pull request if you seen some wrong codings. thanks!

src/map/clif.c
```diff
@@ -18572,7 +18578,9 @@ void clif_defaults(void) {
 	clif->mvp_noitem = clif_mvp_noitem;
 	clif->changed_dir = clif_changed_dir;
 	clif->charnameack = clif_charnameack;
+#ifdef ENABLE_MONSTER_HP_BAR
```
dastgir: Ifdef should never be here, 
Ifdef should be on function, inside of here.

src/map/clif.h
```diff
@@ -724,7 +725,9 @@ struct clif_interface {
 	void (*mvp_noitem) (struct map_session_data* sd);
 	void (*changed_dir) (struct block_list *bl, enum send_target target);
 	void (*charnameack) (int fd, struct block_list *bl);
+#ifdef ENABLE_MONSTER_HP_BAR
```
dastgir: Same as above, ifdef shouldn't be here.
Function content should be enclosed in ifdef

src/config/core.h
```diff
@@ -68,6 +68,9 @@
 /// Comment to disable autotrade persistency (where autotrading merchants survive server restarts)
 #define AUTOTRADE_PERSISTENCY
 
+/// Comment to disable monster hp bar on new clients (2012-04-04aRagexeRE onwards). [Frost]
+#define ENABLE_MONSTER_HP_BAR
```
dastgir: I find it odd for this setting to have in core.
Maybe make a battle config?

src/map/clif.c
Jedzkie: So you prefered to remove this?

MishimaHaruna: Yes, this must be removed, because otherwise, if a plugin tries to use one of these functions, it'll crash.

- There should never be any ifdef inside the `xxx_defaults` function for an interface
- There should never be any ifdef inside the `struct xxx_interface` definition

src/map/clif.c
```diff
@@ -17107,6 +17110,7 @@ void clif_snap( struct block_list *bl, short x, short y ) {
 	clif->send(buf,packet_len(0x8d2),bl,AREA);
 }
 
+#ifdef ENABLE_MONSTER_HP_BAR
```
MishimaHaruna: The function must exist at all times, so the ifdef should go inside, rather than outside

Jedzkie: I see.. i update my pull request, kindly check it. thanks!

conf/battle/items.conf
```diff
@@ -79,13 +79,13 @@ autospell_stacking: no
...
 // 0: no
-item_restricted_consumption_type: 1
+item_restricted_consumption_type: yes
 
 // Enable all NPC to allow changing of equipments while interacting? (Note 1)
 // Script commands 'enable_items/disable_items' will not be override. (see doc/script_commands.txt)
 // 1 : yes(official)
 // 0 : no
-item_enabled_npc: 1
+item_enabled_npc: yes
```
EPuncker: the comment on both lines 80~81 and 86~86 says that it uses a integer parameter, yet you changed it to 'yes', well my point is: edit the comment as well and get rid off of integers :P

EPuncker:
my tip, make it accept bit values so user can choose if he wants to display on all mobs, only bosses, only normal mobs

dastgir:
wait, @MishimaHaruna should this be accepted?
client already have
/monsterhp

MishimaHaruna:
Yes, /monsterhp has a different use case. This allows the server administrator to decide whether players will be able to see the monsters' HP bars, while /monsterhp leaves the decision to the player.

src/map/clif.c
```diff
@@ -4083,10 +4083,10 @@ void clif_getareachar_unit(struct map_session_data* sd,struct block_list *bl) {
 				else if(md->special_state.size==SZ_MEDIUM)
 					clif->specialeffect_single(bl,421,sd->fd);
 #if PACKETVER >= 20120404
-				if( !(md->status.mode&MD_BOSS) ){
+				if(battle_config.show_monster_hp_bar) && (!(md->status.mode&MD_BOSS)){
```
MishimaHaruna: Too many parentheses,

```
if (battle_config.show_monster_hp_bar && !(md->status.mode&MD_BOSS)) {
```

src/map/mob.c
```diff
@@ -2063,13 +2063,13 @@ void mob_damage(struct mob_data *md, struct block_list *src, int damage) {
 		return;
 
 #if PACKETVER >= 20120404
-	if( !(md->status.mode&MD_BOSS) ){
+	if (battle_config.show_monster_hp_bar) && (!(md->status.mode&MD_BOSS)){
```
MishimaHaruna:
```
if (battle_config.show_monster_hp_bar && !(md->status.mode&MD_BOSS)) {
```

src/map/mob.c
```diff
@@ -2797,19 +2797,19 @@ int mob_class_change (struct mob_data *md, int class_)
...
-void mob_heal(struct mob_data *md,unsigned int heal)
+void mob_heal(struct mob_data *md, unsigned int heal)
 {
 	if (battle_config.show_mob_info&3)
 		clif->charnameack (0, &md->bl);
-	
+
 #if PACKETVER >= 20120404
-	if( !(md->status.mode&MD_BOSS) ){
+	if (battle_config.show_monster_hp_bar) && (!(md->status.mode&MD_BOSS)){
```
MishimaHaruna:
```
if (battle_config.show_monster_hp_bar && !(md->status.mode&MD_BOSS)) {
```

Jedzkie:
Bump?

dastgir:
Seems good now.

MishimaHaruna:
Would you be able to clean up the commits (split the unrelated, whitespace-only, changes to a separate commit, and squash the others so that the follow-ups won't appear in history)? Otherwise I will do it while merging the pull request

MishimaHaruna:
Rebased and merged, thank you!

Jedzkie:
Thanks haru!

PR #559, opened 2015-06-16
Sura quest fix

Nihadm89:
The quest worked fine on the first try. But the loop would break after that and only 1 person could do the quest per server restart. This small edit slightly adjusts the loop so it works as intended.

### DIFF
--- a/npc/re/jobs/3-2/sura.txt
+++ b/npc/re/jobs/3-2/sura.txt
@@ -636,10 +636,10 @@ OnTimer315000:
 	end;
 OnTimer320000:
 	mapannounce "sword_2-1","I guess you're not ready to meet the master yet...",bc_map;
+	mapwarp "sword_2-1","yuno_fild07",255,178;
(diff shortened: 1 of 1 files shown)


dastgir:
I really have no idea how this will fix it, since edit  do the same work,
mind explaining @Nihadm89 ?

Nihadm89:
The mapwarp portion I moved just for order of operations because it makes more sense to me to have the person warped first. I removed the stopnpc timer part because it was basically stopping the loop. Once the loop was stopped no one else could do anything. 
donpcevent "Buddy#Sura_Salon::OnDisable";
I put that part in order to prevent overlapping. 
The end result is that the script works so far with no complaints. 

Feel free to test the scripts the following way:

Before my edit, you can do the quest once with no issues, the second person would get stuck in the waiting room. Since the loop ends with stopnpc timer. So basically the script is done until next server restart.

After my edit the script continues, and due to the other checks it warps the players in the proper order. Without the other edits it would warp multiple players because order of operations.

PR #564, opened 2015-06-21
Add missing check to login and char server and remove useless checks from char and map servers

4144:
This checks was found by special tool. It cant find all issues, but can find some atleast.
Login and char server i checked before by hands.


4144:
Add commits for removing useless checks. This can fix some issues in char server

MishimaHaruna:
Merged, sorry for the delay!

PR #566, opened 2015-06-22
Fixing party creation failure condition

Ancyker:
mapif_party_created was incorrectly checking for a null pointer


MishimaHaruna:
Thank you!

PR #575, opened 2015-06-27
Crash fixes

Ancyker:
These commits fix 2 different crashes in Hercules.

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -2812,7 +2812,7 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
 			d->dmg_lv = ATK_BLOCK;
 			return 0;
 		}
-		if( sc->data[SC__MAELSTROM] && (flag&BF_MAGIC) && skill_id && (skill->get_inf(skill_id)&INF_GROUND_SKILL) ) {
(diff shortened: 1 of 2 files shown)


src/map/battle.c
```diff
@@ -2799,7 +2799,7 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
 			d->dmg_lv = ATK_BLOCK;
 			return 0;
 		}
-		if( sc->data[SC__MAELSTROM] && (flag&BF_MAGIC) && skill_id && (skill->get_inf(skill_id)&INF_GROUND_SKILL) ) {
+		if( sd && sc->data[SC__MAELSTROM] && (flag&BF_MAGIC) && skill_id && (skill->get_inf(skill_id)&INF_GROUND_SKILL) ) {
```
MishimaHaruna: This will prevent the skill from working when used by a clone or monster, is it intended?

In most cases like this (if you look through the codebase), we'd instead edit the line below to something like:

```
int sp = (sc->data[SC__MAELSTROM]->val1 * skill_lv + (sd ? sd->status.job_level / 5 : 0)) / 2;
```

Nihadm89: Can we approve his SC_GENSOU fix please.

MishimaHaruna:
Cherry-picked the remaining part (commit 5edf0c3). Thank you for your contribution

PR #581, opened 2015-07-04
Display NPC name when it has an invalid view data

mkbu95:
Having hundreds of NPCs, when that message shows up I knew trouble was ahead, because it was a pain finding the NPC with the invalid sprite.

This is just something that makes easier to do that.


dastgir:
+1

PR #582, opened 2015-07-04
Add checks for servers ip address in inter server connections.

4144:
If ip not in configured subnet, connection refused.
This can protect servers from brutforcing attacks.

### DIFF
--- a/conf/subnet.conf
+++ b/conf/subnet.conf
@@ -4,3 +4,4 @@
 // you can add more than one subnet 
 
 subnet: 255.0.0.0:127.0.0.1:127.0.0.1
+subnet: 0.0.0.0:127.0.0.1:127.0.0.1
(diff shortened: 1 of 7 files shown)


purityz:
after this merge I'm always get disconnected after selected server.

4144:
In your copy paste no any errors. This mean connection closed without errors. After you probably cant connect to char server.

This changes affect only inter server connections.
If you disconnected from server because this, it mean servers cant connect to each other.
Depend on your configuration you may be need use subnet.conf from repository, or if need edit it in correct way. It using address 127.0.0.1 as servers ip, may be you need change it to other

purityz:
this is my configuration.

subnet: 255.0.0.0:127.0.0.1:127.0.0.1
subnet: 0.0.0.0:127.0.0.1:127.0.0.1

untitled

PR #585, opened 2015-07-05
Fixes skill blocks overiding larger blocks

csnv:


malufett:
is there a scenario that this overriding happens? since you can't use the same skill if its still on cd...

csnv:
Not at the current base code.

I proposed this as a result of adding custom code for blocking asura for 5 minutes and realizing using Snap overrides these 5 minutes with its 2 seconds.

malufett:
ow..I see..thanks

PR #596, opened 2015-07-14
Fix for Issue 592

Milchdealer:
Exchanged message for its colored counterpart with DEFAULT_COLOR to prevent it from bugging out on special symbols.


4144:
Here one line change, but two useless commits. May be better clean it?

Milchdealer:
No
It's best practice to only merge on master and never commit directly.
Also I think they currently work on a bigger patch anyways and could just incorporate this

Milchdealer:
But it did add my synching the fork with the origin which I will remove later today

HerculesWSAPI:
Yep, best practice is to only merge on master, that's why it's recommended to start pull requests from specific branches (targeting Hercules:master), so they can be easily rebased before merging, as not to clutter the main repository history with reverse merges

MishimaHaruna:
Yep, best practice is to only merge on master, that's why it's recommended to start pull requests from specific branches (targeting Hercules:master), so they can be easily rebased before merging, as not to clutter the main repository history with reverse merges

Milchdealer:
Yes I'll clean it up after work

PR #601, opened 2015-07-18
Reduce map server memory usage for 3 MB with loaded HPM

4144:
Moved big arrays from interfaces.
Was RES for map-server 25500, after this changes 22548


xcitehub:
can we merge this ? i think it looks good

dastgir:
@MishimaHaruna

malufett:
looks good..nice this save compile time too... :+1:

PR #610, opened 2015-07-25
Dont stop npc scripts after warp.

4144:
At old times if npc what talking with player or player itself went in long distance or in other map, npc script terminated with error. Now it will not terminate after changes some months ago.

Current warp issue. If start talking with npc and warp, all npc related variables will be set to 0. And player cant continue talking to npc. This patch fix this issue.

More about issue and example code here: http://herc.ws/board/topic/10242-invalid-menu-selection-on-npc-after-warp/

Old patch what add this cleanup: https://github.com/HerculesWS/Hercules/commit/9babdbcbb4209a6ae861550eb662ac810d7ad886.

@MishimaHaruna what you can say?

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -4116,7 +4116,7 @@ will be seen to everyone around but not in chat window.
 
 ---------------------------------------
 
-*warp "<map name>",<x>,<y>;
+*warp "<map name>",<x>,<y>{,<flag>};
 
 This command will take the invoking character to the specified map, and if 
 wanted, specified coordinates too, but these can be random.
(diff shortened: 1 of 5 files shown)


emistry:
I think the termination script is added to avoid users being stuck by NPC script. Whenever char get stucked by script, you can alway release the char from being stucked simply by @warp @jump  @load etc...

dastgir:
@MishimaHaruna 
any bug if we remove this?

MishimaHaruna:
After a testing session with 4144, we figured out that this needs some changes.

The problem is that, when warping, the client forgets everything that happened before the warp packet was received (technically, it would appear this isn't as soon as it receives the warp packet, but rather, as soon as it starts loading the new map, or moves the character within the current map -- there may be some further packet processing in between, but it's not strictly relevant here),  so any open dialogs or menus are lost clientside. This means that the server will be waiting for user interaction (i.e. pressing Next or selecting a menu option), while the client is no longer showing any dialogs or menus, making the character effectively stuck.

4144:
Update pull request.

Now warp have optional parameter. If set it to 0, variable will be not cleaned after warp.

PR #619, opened 2015-08-03
New EquipPackets Support

dastgir:
Thanks to Rytech and 3CeAM
Uses 0 as data, till we know exactly know how item gets those optionData..

@MishimaHaruna Reviews?

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -2229,11 +2229,32 @@ void clif_addcards2(unsigned short *cards, struct item* item) {
 		cards[3] = item->card[i];
 }
 
+/**
+ * Fills in RandomOptions(Bonuses) of items into the buffer
+ *
+ * Dummy datais used since this feature isn't supported yet (ITEM_RDM_OPT).
+ * A maximum of 5 random options can be supported.
+ *
+ * @param buf[in,out] The buffer to write to. The pointer must be valid and initialized.
+ * @param item[in]    The source item.
+ */
+void clif_add_random_options(unsigned char* buf, struct item* item)
+{
+	int i;
+	for (i = 0; i < 5; i++){
+		WBUFW(buf,i*5+0) = 0;	// OptIndex
+		WBUFW(buf,i*5+2) = 0;	// Value
+		WBUFB(buf,i*5+4) = 0;	// Param1
+	}
+}
+
 
 /// Notifies the client, about a received inventory item or the result of a pick-up request.
 /// 00a0 <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.W <item type>.B <result>.B (ZC_ITEM_PICKUP_ACK)
 /// 029a <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.W <item type>.B <result>.B <expire time>.L (ZC_ITEM_PICKUP_ACK2)
 /// 02d4 <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.W <item type>.B <result>.B <expire time>.L <bindOnEquipType>.W (ZC_ITEM_PICKUP_ACK3)
+/// 0990 <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.L <item type>.B <result>.B <expire time>.L <bindOnEquipType>.W (ZC_ITEM_PICKUP_ACK_V5)
+/// 0a0c <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.L <item type>.B <result>.B <expire time>.L <bindOnEquipType>.W (ZC_ITEM_PICKUP_ACK_V6)
 void clif_additem(struct map_session_data *sd, int n, int amount, int fail) {
 	struct packet_additem p;
 	nullpo_retv(sd);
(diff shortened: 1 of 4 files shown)


src/map/clif.c
```diff
@@ -3948,6 +3981,19 @@ void clif_storageitemadded(struct map_session_data* sd, struct item* i, int inde
 	WFIFOB(fd,13) = i->refine; //refine
 	clif->addcards(WFIFOP(fd,14), i);
 	WFIFOSET(fd,packet_len(0x1c4));
+#else
+	WFIFOHEAD(fd,packet_len(0xa0a));
+	WFIFOW(fd, 0) = 0xa0a; // Storage item added
+	WFIFOW(fd, 2) = index+1; // index
+	WFIFOL(fd, 4) = amount; // amount
+	WFIFOW(fd, 8) = ( view > 0 ) ? view : i->nameid; // id
+	WFIFOB(fd,10) = itemdb_type(i->nameid); //type
```
MishimaHaruna: Does this not need to be passed through `itemtype()`? (see packet 0x1c4 above)

dastgir: Was a typo.

dastgir:
@Michieru 
Interfaced the function, some other optimization and fixed the warnings you told :+1:

src/map/clif.c
```diff
@@ -2229,13 +2229,30 @@ void clif_addcards2(unsigned short *cards, struct item* item) {
 		cards[3] = item->card[i];
 }
 
+void clif_add_random_options(unsigned char* buf, struct item* item)
```
MishimaHaruna: Since this is a new function, could you add a doxygen documentation comment for it?

Please reword as you see fit:
```
/**
 * Fills in (what are those random options exactly btw?) from the given item into a buffer.
 *
 * Dummy datais used since this feature isn't supported yet (ITEM_RDM_OPT).
 * A maximum of 5 random options can be supported.
 *
 * @param buf[in,out] The buffer to write to. The pointer must be valid and initialized.
 * @param item[in]    The source item.
 */
```

You can then remove the comment inside the function, since it's included in the above.

src/map/clif.c
```diff
@@ -2229,13 +2229,30 @@ void clif_addcards2(unsigned short *cards, struct item* item) {
...
 /// 00a0 <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.W <item type>.B <result>.B (ZC_ITEM_PICKUP_ACK)
 /// 029a <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.W <item type>.B <result>.B <expire time>.L (ZC_ITEM_PICKUP_ACK2)
 /// 02d4 <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.W <item type>.B <result>.B <expire time>.L <bindOnEquipType>.W (ZC_ITEM_PICKUP_ACK3)
+/// 0990 <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.L <item type>.B <result>.B <expire time>.L <bindOnEquipType>.W (ZC_ITEM_PICKUP_ACK_V5)
+/// 0a0c <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.L <item type>.B <result>.B <expire time>.L <bindOnEquipType>.W (ZC_ITEM_PICKUP_ACK_V6)
 void clif_additem(struct map_session_data *sd, int n, int amount, int fail) {
 	struct packet_additem p;
+#if PACKETVER >= 20150226
```
MishimaHaruna: Not really critical, but since you're editing the commit for other reasons, the scope of this variable could be minimized by moving this inside the `if (!fail) {` block below

src/map/clif.c
```diff
@@ -3754,11 +3789,7 @@ void clif_tradeadditem(struct map_session_data* sd, struct map_session_data* tsd
 {
 	int fd;
 	unsigned char *buf;
-#if PACKETVER < 20100223
-	const int cmd = 0xe9;
-#else
-	const int cmd = 0x80f;
-#endif
+	const int cmd = tradeaddType;
```
MishimaHaruna: Since you're at it, you could just remove this `cmd` constant now, and use tradeaddType directly below?

src/map/clif.c
```diff
@@ -5949,6 +5981,7 @@ void clif_cart_additem(struct map_session_data *sd,int n,int amount,int fail)
 {
 	int view,fd;
 	unsigned char *buf;
+	short i;
```
MishimaHaruna: Could you rename this to `offset` (more meaningful), and turn it into an `int` (possibly more performant)?

src/map/clif.c
```diff
@@ -5956,38 +5989,27 @@ void clif_cart_additem(struct map_session_data *sd,int n,int amount,int fail)
...
-	clif->addcards(WBUFP(buf,14), &sd->status.cart[n]);
-	WFIFOSET(fd,packet_len(0x1c5));
+	i = 1;
+#endif
+	WBUFB(buf,10+i)=sd->status.cart[n].identify;
+	WBUFB(buf,11+i)=sd->status.cart[n].attribute;
+	WBUFB(buf,12+i)=sd->status.cart[n].refine;
+	clif->addcards(WBUFP(buf,13+i), &sd->status.cart[n]);
+#if PACKETVER >= 20150226
+	clif->add_random_options(WBUFP(buf,22), &sd->status.cart[n]);
```
MishimaHaruna: For consistency sake, `22+offset` (even though it's not needed, it might save us from future issues)

src/map/packets_struct.h
```diff
@@ -256,6 +291,12 @@ struct NORMALITEM_INFO {
 #endif
 } __attribute__((packed));
 
+struct RndOptions {
+	short index;
+	short value;
+	unsigned char param;
+};
```
MishimaHaruna: It's a good idea to use the `__attribute((packed))` in this one as well, to keep it consistent with all compilers (i.e. the ones that don't understand the `#pragma pack` at line 256).

dastgir:
Updated , @MishimaHaruna

MishimaHaruna:
Thank you very much, merging

PR #624, opened 2015-08-07
Fix script loading via sql in itemdb.c

Smokexyz:
This fixes the script loading issue mentioned here -> http://herc.ws/board/topic/10249-issue-with-item-db-after-update/


dastgir:
Could you change the commit name according to the fix made?

Smokexyz:
how is it not describing the fix made?

dastgir:
Oops. I am blind. :x

dastgir:
Merged. Thanks for distribution

PR #629, opened 2015-08-10
Corrected search order in itemdb_searchname

MordekaiserGod:
It’ll now return the first result, rather than the last due to item2
being overwritten as it searches through the array.


MishimaHaruna:
Nice catch! Thank you very much

PR #637, opened 2015-08-12
Merging main.sql and logs.sql script and increasing database setup simplicity/redundancy.

GammaCrucis:
These modifications are made in the hopes of simplifying the database setup process by eliminating several steps from the installation guide on the wiki. As well as with the intent to render the SQL setup script more redundant in case of an incomplete initial run.

- Modified main.sql and merged it with logs.sql, so that the user just needs to run a single script to create the 'ragnarok' and 'log' schemas and have the proper tables created in their respective schemas. Users no longer have to manually create these schemas, assign them as default, and run the main.sql/logs.sql scripts separately. They can simply just run this single script instead.

- Modified the table creation statement for the login table/the insert statement for the default server communication 'user' to check for the prior existence of that data before inserting it with INSERT IGNORE statement.

- Did the same as above with the db_roulette table creation statement/insert statements.

- Modified db_roulette tables creation statement to have IF NOT EXISTS.

- Rewrote the db_roulette table creation statement (to fix a bizarre error I was having with it/the rewritten statement is identical) and moved it to the last table created in the 'ragnarok' schema as the client which supports the roulette feature (2014-08-22b) is still under stability testing.


dastgir:
I have a feeling that this was meant for rA , and you did PR for Hercules 
Or might be , you used rA SQL for this PR. Which is not acceptable

GammaCrucis:
My apologies, this version was indeed intended for rAthena. I'll submit a new PR with the proper SQL for Hercules.

ErMaqui:
If you do this, user cannot select the database to insert data. For example, you can have multiple map server instances, who requires multiple ragnarok / logs databases.

If you want to simplify the installation, you can make a script who question the user / admin for SQL host / root password, and then, with a text wizard, create the users to the database, create database names, pull all content and configure all files required on conf/import/ with the data.

anacondaqq:
### I am very much against these changes.
1. It should be 2 different database logs and the main database.
2. Main you can backup every day few times, without perfomance downgrade by crontab.
3. Logs database usually very big, and you can't do regular backups if you make main.sql & logs.sql in the same db, you just increase in multiple times a risk to lost of all your game database data. (or are you suggest me to remove detailed information about each items, events, when this information very critical to manage and administrate 1 game server)
4. This is required for multiple map-char-servers.
5. Separate logs database is easy manage, etc.

Please, dear devs, do not merge this with the main code...

PR #639, opened 2015-08-12
Merged main.sql and logs.sql script to increase database setup simplicity.

GammaCrucis:
These modifications were made in the hopes of simplifying the database setup process by eliminating several steps from the installation guide on the wiki. Additional modifications have been made to increase redundancy in case of an incomplete initial run of the script.

- Modified main.sql and merged it with logs.sql, so that the user just needs to run a single script to create the 'ragnarok' and 'log' schemas and have the proper tables created in their respective schemas. Users no longer have to manually create these schemas, assign them as default, and run the main.sql/logs.sql scripts separately.

- Modified the table creation statement for the login table/the insert statement for the default server communication 'user' to check for the prior existence of that data before inserting it with INSERT IGNORE statement.

- Did the same as above with the 'sql_updates' table creation/insert statements.


GammaCrucis:
My apologies once again for the improper PR earlier.

ErMaqui:
If you do this, user cannot select the database to insert data. For example, you can have multiple map server instances, who requires multiple ragnarok / logs databases.

If you want to simplify the installation, you can make a script who question the user / admin for SQL host / root password, and then, with a text wizard, create the users to the database, create database names, pull all content and configure all files required on conf/import/ with the data.

Sorry for duplicate comment, i haven't viewed the new PR.

dastgir:
Also not necessary , that one would use same database name.

It actually increases step, as changing from command line takes less time , rather than opening file and changing content inside it.

MishimaHaruna:
Thank you for your effort, but, as said above, and to add to it:

- The scripts shouldn't contain a database name, since that's not necessarily the same for everyone. Specifying a database name on the command line (`mysql -u user -ppassword databasename < sql-files/main.sql`) is way faster than opening a file and editing it.
- You can't/shouldn't expect the mysql user that will import the data to be able to create schemas/databases. You should expect the user to already have created that, since it is generally an operation that only a privileged user can do.
- It is not necessary that sql and main databases are on the same schema (and actually recommended against).

As such, we cannot accept this pull request.

GammaCrucis:
Yes I see that now. I mainly took a crack at this to simplify the process for Windows users (which is probably the biggest mistake I made in the first place).
I'm going to submit a separate pull request for the added redundancy in the SQL statements in main.sql. Then take a proper look at automating the process of creating the schema and tables.
I think it would be possible through a linux shell script/small .NET application to get the schemas and tables setup using user input. I'll post a release thread on the forum for that though.
Thank you for the input.
On Aug 12, 2015 10:51 AM, Haruna <notifications@github.com> wrote:Thank you for your effort, but, as said above, and to add to it:

The scripts shouldn't contain a database name, since that's not necessarily the same for everyone. Specifying a database name on the command line (mysql -u user -ppassword databasename < sql-files/main.sql) is way faster than opening a file and editing it.You can't/shouldn't expect the mysql user that will import the data to be able to create schemas/databases. You should expect the user to already have created that, since it is generally an operation that only a privileged user can do.It is not necessary that sql and main databases are on the same schema (and actually recommended against).

As such, we cannot accept this pull request.

PR #671, opened 2015-08-26
Conversion of refine_db to Libconfig refine_db.conf

Smokexyz:
This commit converts the current refine_db.txt to Hercules' awesome libconfig format. It provides legible configuration and an easy way to add more configuration if required. (currently only supports base refine success percentages, need more info for enriched/premium.)


4144:
Not sure is this good comment for pull requests, but still.
Style with brackets different in different lines.
Probably better follow style like this:
    if (expr) {
        ...
    } else {
        ....
    }

Here after if space. "{" on same line with if, and else in same line with "}" and "{"

Smokexyz:
Done as suggested in new commit, sorry i'm used to formatting legible code.

4144:
Another issue look like field "level" unused. It only checked for correct value but after not used.

Smokexyz:
it is needed for those checks, hence used.

4144:
It must be used or completly removed. I think better use it.
With current code this config will be wrong. In your code level 1 will works fine, but level 5 will be used as level 2:

{
Level: 1
Chance: 100
Bonus: 0
},
{
Level: 5
Chance: 100
Bonus: 0
}

4144:
Or other example:

Level 1
Level 3
Level 2

Level 2 and 3 will be wrong

Smokexyz:
I'm not sure how you're getting that here's my debug - 
[Debug]: Type 0 Level 1 Chance 100 Bonus 100
[Debug]: Type 0 Level 2 Chance 100 Bonus 100
[Debug]: Type 0 Level 3 Chance 100 Bonus 100
[Debug]: Type 0 Level 4 Chance 100 Bonus 100
[Debug]: Type 0 Level 5 Chance 60 Bonus 200
[Debug]: Type 0 Level 6 Chance 40 Bonus 200
[Debug]: Type 0 Level 7 Chance 40 Bonus 200
[Debug]: Type 0 Level 8 Chance 20 Bonus 200
[Debug]: Type 0 Level 9 Chance 20 Bonus 300
[Debug]: Type 0 Level 10 Chance 10 Bonus 300
[Debug]: Type 0 Level 11 Chance 8 Bonus 300
[Debug]: Type 0 Level 12 Chance 8 Bonus 300
[Debug]: Type 0 Level 13 Chance 8 Bonus 400
[Debug]: Type 0 Level 14 Chance 8 Bonus 400
[Debug]: Type 0 Level 15 Chance 7 Bonus 400
[Debug]: Type 0 Level 16 Chance 7 Bonus 400
[Debug]: Type 0 Level 17 Chance 7 Bonus 500
[Debug]: Type 0 Level 18 Chance 7 Bonus 500
[Debug]: Type 0 Level 19 Chance 5 Bonus 500
[Debug]: Type 0 Level 20 Chance 5 Bonus 500
[Debug]: Type 1 Level 1 Chance 100 Bonus 0
[Debug]: Type 1 Level 2 Chance 100 Bonus 0
[Debug]: Type 1 Level 3 Chance 100 Bonus 0
[Debug]: Type 1 Level 4 Chance 100 Bonus 0
[Debug]: Type 1 Level 5 Chance 100 Bonus 0
[Debug]: Type 1 Level 6 Chance 100 Bonus 0
[Debug]: Type 1 Level 7 Chance 100 Bonus 0
[Debug]: Type 1 Level 8 Chance 60 Bonus 0
[Debug]: Type 1 Level 9 Chance 40 Bonus 0
[Debug]: Type 1 Level 10 Chance 20 Bonus 0
[Debug]: Type 1 Level 11 Chance 18 Bonus 0
[Debug]: Type 1 Level 12 Chance 18 Bonus 0
[Debug]: Type 1 Level 13 Chance 18 Bonus 0
[Debug]: Type 1 Level 14 Chance 18 Bonus 0
[Debug]: Type 1 Level 15 Chance 18 Bonus 0
[Debug]: Type 1 Level 16 Chance 17 Bonus 300
[Debug]: Type 1 Level 17 Chance 17 Bonus 300
[Debug]: Type 1 Level 18 Chance 17 Bonus 300
[Debug]: Type 1 Level 19 Chance 15 Bonus 300
[Debug]: Type 1 Level 20 Chance 15 Bonus 300
[Debug]: Type 2 Level 1 Chance 100 Bonus 0
[Debug]: Type 2 Level 2 Chance 100 Bonus 0
[Debug]: Type 2 Level 3 Chance 100 Bonus 0
[Debug]: Type 2 Level 4 Chance 100 Bonus 0
[Debug]: Type 2 Level 5 Chance 100 Bonus 0
[Debug]: Type 2 Level 6 Chance 100 Bonus 0
[Debug]: Type 2 Level 7 Chance 60 Bonus 0
[Debug]: Type 2 Level 8 Chance 40 Bonus 0
[Debug]: Type 2 Level 9 Chance 20 Bonus 0
[Debug]: Type 2 Level 10 Chance 20 Bonus 0
[Debug]: Type 2 Level 11 Chance 18 Bonus 0
[Debug]: Type 2 Level 12 Chance 18 Bonus 0
[Debug]: Type 2 Level 13 Chance 18 Bonus 0
[Debug]: Type 2 Level 14 Chance 18 Bonus 0
[Debug]: Type 2 Level 15 Chance 18 Bonus 0
[Debug]: Type 2 Level 16 Chance 17 Bonus 600
[Debug]: Type 2 Level 17 Chance 17 Bonus 600
[Debug]: Type 2 Level 18 Chance 17 Bonus 600
[Debug]: Type 2 Level 19 Chance 15 Bonus 600
[Debug]: Type 2 Level 20 Chance 15 Bonus 600
[Debug]: Type 3 Level 1 Chance 100 Bonus 0
[Debug]: Type 3 Level 2 Chance 100 Bonus 0
[Debug]: Type 3 Level 3 Chance 100 Bonus 0
[Debug]: Type 3 Level 4 Chance 100 Bonus 0
[Debug]: Type 3 Level 5 Chance 100 Bonus 0
[Debug]: Type 3 Level 6 Chance 60 Bonus 0
[Debug]: Type 3 Level 7 Chance 50 Bonus 0
[Debug]: Type 3 Level 8 Chance 20 Bonus 0
[Debug]: Type 3 Level 9 Chance 20 Bonus 0
[Debug]: Type 3 Level 10 Chance 20 Bonus 0
[Debug]: Type 3 Level 11 Chance 18 Bonus 0
[Debug]: Type 3 Level 12 Chance 18 Bonus 0
[Debug]: Type 3 Level 13 Chance 18 Bonus 0
[Debug]: Type 3 Level 14 Chance 18 Bonus 0
[Debug]: Type 3 Level 15 Chance 18 Bonus 0
[Debug]: Type 3 Level 16 Chance 17 Bonus 900
[Debug]: Type 3 Level 17 Chance 17 Bonus 900
[Debug]: Type 3 Level 18 Chance 17 Bonus 900
[Debug]: Type 3 Level 19 Chance 15 Bonus 900
[Debug]: Type 3 Level 20 Chance 15 Bonus 900
[Debug]: Type 4 Level 1 Chance 100 Bonus 0
[Debug]: Type 4 Level 2 Chance 100 Bonus 0
[Debug]: Type 4 Level 3 Chance 100 Bonus 0
[Debug]: Type 4 Level 4 Chance 100 Bonus 0
[Debug]: Type 4 Level 5 Chance 60 Bonus 0
[Debug]: Type 4 Level 6 Chance 40 Bonus 0
[Debug]: Type 4 Level 7 Chance 40 Bonus 0
[Debug]: Type 4 Level 8 Chance 20 Bonus 0
[Debug]: Type 4 Level 9 Chance 20 Bonus 0
[Debug]: Type 4 Level 10 Chance 10 Bonus 0
[Debug]: Type 4 Level 11 Chance 8 Bonus 0
[Debug]: Type 4 Level 12 Chance 8 Bonus 0
[Debug]: Type 4 Level 13 Chance 8 Bonus 0
[Debug]: Type 4 Level 14 Chance 8 Bonus 0
[Debug]: Type 4 Level 15 Chance 7 Bonus 0
[Debug]: Type 4 Level 16 Chance 7 Bonus 1200
[Debug]: Type 4 Level 17 Chance 7 Bonus 1200
[Debug]: Type 4 Level 18 Chance 7 Bonus 1200
[Debug]: Type 4 Level 19 Chance 5 Bonus 1200
[Debug]: Type 4 Level 20 Chance 5 Bonus 1200
[Status]: Done reading '5' entries in 're/refine_db.conf'.

4144:
I mean what now field "Level" unused. But better use it.
If you try swap rate for level 1 and level 2 your code will read wrong data.
In other words for your code first must be level 1, then level 2, etc. But if rates unsurted or rate for some level missing in db, it will not works.

For level index in array you should use variable 'level' and not 'i'

Smokexyz:
This is viable but in this case here - 
if ( level > 0 ) status->dbs->refine_info[type].bonus[level] += status->dbs->refine_info[type].bonus[level-1];
if [level] is added before [level-1] then the value will be incorrect as the previous refine level bonus is added to the latter, therefore they should be in order. Also refine levels shouldn't be swapped because of this reason.

src/map/status.c
```diff
@@ -12325,6 +12291,127 @@ bool status_readdb_scconfig(char* fields[], int columns, int current) {
...
+			if (!libconfig->setting_lookup_int(tt, "Level", &level) || level <= 0 || level > MAX_REFINE) {
+				ShowError("status_readdb_refine_libconfig_sub: Invalid 'Level' configuration %d in entry #%d of \"%s\".", level, n, source);
+				return 0;
+			} else if (!libconfig->setting_lookup_int(tt, "Chance", &chance)) {
+				ShowWarning("status_readdb_refine_libconfig_sub: Missing 'Chance' configuration in entry #%d of \"%s\", defaulting to 0.", n, source);
+				status->dbs->refine_info[type].chance[i] = 0;
+			}
+			
+			if (!libconfig->setting_lookup_int(tt, "Bonus", &bonus))
+				status->dbs->refine_info[type].bonus[i] = 0;
```
4144: Here probably need use this status->dbs->refine_info[type].bonus[level - 1] = 0;

src/map/status.c
```diff
@@ -12325,6 +12291,127 @@ bool status_readdb_scconfig(char* fields[], int columns, int current) {
...
+			
+			if ( chance ) status->dbs->refine_info[type].chance[i] = chance;
```
4144: and here
if (chance) status->dbs->refine_info[type].chance[level - 1] = chance;

Same for other

Smokexyz: Yes I've tested this but if you add
Level: 5
Level: 4

Level 5 bonus will be wrong, because Level 5 bonus += Level 4 bonus
but in this case level 4 is not added when Level 5 is read.

4144:
You can add relative bonuses after reading all levels. inside reading for put number as is
After in next for you can add sum bonuses for current and all previous levels and put it to current level

In example like this:
for (i=0; i < MAX_REFINE; i++) {
  // reading db
  ....
  status->dbs->refine_info[type].bonus[i] = ...
  ....
}

for (i=1; i < MAX_REFINE; i++) {
  status->dbs->refine_info[type].bonus[i] += status->dbs->refine_info[type].bonus[i-1];
}

Some thing like this.

4144:
You can rebase to latest hercules master it will help to see build/test errors.

Smokexyz:
with option --enable-sanitize=full

clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=enum' option
clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=bool' option
clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=float-cast-overflow' option
clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=float-divide-by-zero' option
clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=object-size' option
clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=bounds' option
clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=signed-integer-overflow' option
clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=return' option
clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=null' option
clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=vla-bound' option
clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=unreachable' option
clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=integer-divide-by-zero' option
clang: error: '-fsanitize-undefined-trap-on-error' required with '-fsanitize=shift' option

4144:
I mean rebase this pull request. Now it based on old master and you cant see actual error why test failed.
Probably some kind or out of bound memory access

4144:
I you want do checks locally without rebasing pull request, need gcc 5.
build with --enable-sanitize=full and run servers. In map server console probably you will see issue

src/map/status.c
```diff
@@ -12325,6 +12291,151 @@ bool status_readdb_scconfig(char* fields[], int columns, int current) {
...
+	if ((t=libconfig->setting_get_member(r, "Rate")) && config_setting_is_list(t)) {
+		config_setting_t *tt = NULL;
+		bool duplicate[MAX_REFINE];
+		
+		memset(&duplicate,0,sizeof(duplicate));
+		
+		for(i=0; i < MAX_REFINE; i++) {
+			tt = libconfig->setting_get_elem(t, i);
+			
+			if (!tt)
```
MishimaHaruna: Could you explain the intent of this?  If an entry is missing, what happens? If we break out of the loop here (and the same way in the loop below), the server will report no errors

src/map/status.c
```diff
@@ -12325,6 +12291,151 @@ bool status_readdb_scconfig(char* fields[], int columns, int current) {
...
+				break;
+			else if (!config_setting_is_group(tt))
```
MishimaHaruna: What's the use case for this? When is `tt` not a group?

src/map/status.c
```diff
@@ -12325,6 +12291,151 @@ bool status_readdb_scconfig(char* fields[], int columns, int current) {
...
+				continue;
+			
+			int level=0, chance=0;
```
MishimaHaruna: This will fail in MSVC and other compilers not fully compliant with C99, please move this to the beginning of the for loop

src/map/status.c
```diff
@@ -12325,6 +12291,151 @@ bool status_readdb_scconfig(char* fields[], int columns, int current) {
...
+				status->dbs->refine_info[type].chance[i] = 0;
+			}
+			
+			level -= 1;
+			
+			if( duplicate[level] ) {
+				ShowWarning("status_readdb_refine_libconfig_sub: duplicate entry of Level %d in Type #%d of \"%s\".\n", level, type, source);
+			} else duplicate[level] = true;
+			
+			if ( chance ) status->dbs->refine_info[type].chance[level] = chance;
```
MishimaHaruna: There should be no need for this if/else. It yields the same result as if you assign `status->dbs->refine_info[type].chance[level] = chance` (since in the `else` case, `chance` would already be 0)

src/map/status.c
```diff
@@ -12325,6 +12291,151 @@ bool status_readdb_scconfig(char* fields[], int columns, int current) {
...
+			else status->dbs->refine_info[type].chance[level] = 0;
+		}
+		for(i=0; i < MAX_REFINE; i++) {
+			int bonus=0;
+			tt = libconfig->setting_get_elem(t, i);
```
MishimaHaruna: This loop assumes that the levels are in order (while the above does otherwise). Instead of calling once again `libconfig->setting_get_elem(t, i)`, you could just cache the Bonus value to a temporary array in the loop above, and here simply loop through that array

src/map/status.c
```diff
@@ -12325,6 +12291,151 @@ bool status_readdb_scconfig(char* fields[], int columns, int current) {
...
+		ShowError("can't read %s\n", filepath);
+		return 0;
+	}
+	
+	while( (r = libconfig->setting_get_elem(rdb,i++)) ) {
+		int type = status->readdb_refine_libconfig_sub(r, i-1, filename);
+		if( !type )
+			continue;
+		
+		if( duplicate[type] ) {
```
MishimaHaruna: You probably meant `type-1` (because `status_readdb_refine_libconfig_sub` returns `type+1`, line 12384 of this diff)

src/map/status.c
```diff
@@ -12325,6 +12291,151 @@ bool status_readdb_scconfig(char* fields[], int columns, int current) {
...
+		}
+	} else {
+		ShowWarning("status_readdb_refine_libconfig_sub: Missing Rate configuration in entry #%d of \"%s\", skipping.\n", n, source);
+		return 0;
+	}
+	
+	return type+1;
+}
+
+bool status_refinedb_lookup_const(const config_setting_t *r, const char *name, int *value) {
```
MishimaHaruna: Do we need constants here? Or numbers are enough?

Smokexyz:
Thanks Haru and 4414 I will fix all that in the next commit. 
Also using gcc 5.2 on OSX 10.10
configure: error: zlib library not found or incompatible, please specify the correct path with --with-zlib=DIR... stopping
zlib dir = /usr/include/zlib.h
also tried ./configure --enable-64bit --enable-sanitize=full --with-zlib=/usr
This doesn't happen with gcc 4.2 any ideas?

MishimaHaruna:
Indeed, I can reproduce the same issue with gcc 5.2.0 on OS X 10.10.

I bisected it, and it's caused by c5ff02958806ea5a672dcc9371602a6c090c6758

The issue isn't specific to zlib, but it causes all tests to fail while running the configure script, since it doesn't seem to recognize the `-rdynamic` option, i.e.:

```
configure:4897: checking whether gcc-5 produces 32bit code
configure:4915: gcc-5 -o conftest -g -O2 -pipe -ffast-math -fvisibility=hidden -Wall -Wextra -Wno-sign-compare  -DMAXCONN=16384 -I../common  -rdynamic conftest.c  >&5
gcc-5: error: unrecognized command line option '-rdynamic'
configure:4915: $? = 1
configure: program exited with status 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME ""
| #define PACKAGE_TARNAME ""
| #define PACKAGE_VERSION ""
| #define PACKAGE_STRING ""
| #define PACKAGE_BUGREPORT ""
| #define PACKAGE_URL ""
| #define STDC_HEADERS 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_MEMORY_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_UNISTD_H 1
| #define __EXTENSIONS__ 1
| #define _ALL_SOURCE 1
| #define _GNU_SOURCE 1
| #define _POSIX_PTHREAD_SEMANTICS 1
| #define _TANDEM_SOURCE 1
| /* end confdefs.h.  */
| 
| 		int main(int argc, char **argv){
| 			if(sizeof(void*) == 4) return 0;
| 			else return 1;
| 		}
| 
configure:4960: result: no
```

I'll commit a fix.

MishimaHaruna:
Fixed in 12c8ae2541b654a09bd6a2f264856f6fc0089a23

Smokexyz:
Thanks for the update of the configure file Haru. I will push a new PR with updated code, my git is messy at the moment. If you can please cancel this PR I will update shortly.

Smokexyz:
Added suggested changes in a cleaner single commit - #675

PR #674, opened 2015-08-27
Hotkey Rotate Packet Implemented

dastgir:
This is feature for client 2014-10-22 and greater..

@4144 @MishimaHaruna 
I have tested(in game), and it works,
just wanted to let you review the code, before its been pushed to master.


AnisotropicDefixation:
What exactly does this do?

dastgir:
In new clients, if you press ctrl+F12.
All items in shortcut key window (4 windows are there), all will shift one window up, so 4th window -> 3rd window, 3->2, 2->1 and 1st window to 4th window.
Nothing that great, but a feature of kRO. :)

PR #682, opened 2015-08-29
Convert refine_db.txt to refine_db.conf (libconfig)

Smokexyz:
Converts current refine_db.txt to be read using the libconfig format Hercules uses. Originally suggested by Emistry in this topic - http://herc.ws/board/topic/5101-overhaul-refine-db/

The following format is used for both pre-renewal and renewal data -

```
refine_db: (
{
	EquipmentLevel: (int)			// 0 for armors 1~4 for weapons
	StatsPerLevel: (int)				// Stats added per level of refinement
	RandomBonusStartLevel: (int)	// Start level for random bonus value
	RandomBonusValue: (int)		// Random bonus value (follows the usual formulation)
	Rate: (
	{
		Level: (int)				// Level of Refinement
		Chance: (int)				// Chance of success 100 = 100%
		Bonus: (int)				// Bonus for this level of refine (Optional Field), defaults to 0.
	},
	)
},
)
```
This format also allows easy and legible addition of new refine data without much editing.
Thanks to Haru and 4144 for pointing out inefficiency in code, edited
as suggested in the original PR #671


4144:
again fix commit :(

You can not reopen pull requests. just fix commits in your branch, and pull request will be automatically updated

Smokexyz:
Haha I'm sorry I forgot to add that one line when copying. Should I rebase?

4144:
You can squash commits. This you can do from interactive rebase

Smokexyz:
Done, sorry for the inconvenience.

Smokexyz:
Should I edit this to support enriched ore bonuses based on iRO? I think that's how kRO would use it if they have a separate refine db for premiums, and not rolling on the refine twice for enriched ores.
Or any other changes, let me know.

malufett:
your design is not quite comfortable for a user..having small entry in the db but yours making it too long..I prefer @emistry's design short and brief..

Smokexyz:
Feel free to cancel or add your own pull request man, In that design you won't know if you're reading the data for level 15 or 16 unless you add comments. Also if I was to add a premium bonus or enriched bonus field for each level, it's easily editable and legible here. Prey tell how you would try to understand more clustered data in a 100:100:100 format. The main purpose for this format was legibility, if that's not priority for your user you can go ahead and change it to exactly how emistry proposed the idea. I'm just trying to help out here.

Ancyker:
Ok, first of all, the example in the originally proposed idea isn't actually valid libconfig. 100:100 would be a string, requiring additional processing. The whole idea behind using libconfig is to get rid of comma separated values. Now you are saying the idea of colons is better? No. It's not. This pull request implements the conversion in the proper way because it's on the spirit of the whole reason for the switch to libconfig.

MishimaHaruna:
While being vertically very long, I fully agree with the two above points in favor of this format. If we want to make it more compact, the only thing that comes to mind would be to omit the fields that have a default value (for example, skip all the `Bonus: 0` and the `Chance: 100`).

Optionally, the level could be moved to outside the block as the key (changing the list into a group), but I'm fine either way.

Example
```
refine_db: {
Armor: {
	StatsPerLevel: 66
	RandomBonusStartLevel: 0
	RandomBonusValue: 0
	Rate: {
		Lv1: { },
		Lv2: { },
		Lv3: { },
		Lv4: { },
		Lv5: {
			Chance: 60
		},
		Lv6: {
			Chance: 40
		},
		Lv7: {
			Chance: 40
		},
		Lv8: {
			Chance: 20
		},
		Lv9: {
			Chance: 20
		},
		Lv10: {
			Chance: 10
		},
	}
},
WLv1: {
	StatsPerLevel: 200
	RandomBonusStartLevel: 8
	RandomBonusValue: 300
	Rate: {
		Lv1: { },
		Lv2: { },
		Lv3: { },
		Lv4: { },
		Lv5: { },
		Lv6: { },
		Lv7: { },
		Lv8: {
			Chance: 60
		},
		Lv9: {
			Chance: 40
		},
		Lv10: {
			Chance: 20
		},
	}
},
WLv2: {
	StatsPerLevel: 300
	RandomBonusStartLevel: 7
	RandomBonusValue: 500
	Rate: {
		Lv1: { },
		Lv2: { },
		Lv3: { },
		Lv4: { },
		Lv5: { },
		Lv6: { },
		Lv7: {
			Chance: 60
		},
		Lv8: {
			Chance: 40
		},
		Lv9: {
			Chance: 20
		},
		Lv10: {
			Chance: 20
		},
	}
},
WLv3: {
	StatsPerLevel: 500
	RandomBonusStartLevel: 6
	RandomBonusValue: 800
	Rate: {
		Lv1: { },
		Lv2: { },
		Lv3: { },
		Lv4: { },
		Lv5: { },
		Lv6: {
			Chance: 60
		},
		Lv7: {
			Chance: 50
		},
		Lv8: {
			Chance: 20
		},
		Lv9: {
			Chance: 20
		},
		Lv10: {
			Chance: 20
		},
	}
},
WLv4: {
	StatsPerLevel: 700
	RandomBonusStartLevel: 5
	RandomBonusValue: 1300
	Rate: {
		Lv1: { },
		Lv2: { },
		Lv3: { },
		Lv4: { },
		Lv5: {
			Chance: 60
		},
		Lv6: {
			Chance: 40
		},
		Lv7: {
			Chance: 40
		},
		Lv8: {
			Chance: 20
		},
		Lv9: {
			Chance: 20
		},
		Lv10: {
			Chance: 10
		},
	}
},
}
```

dastgir:
Why not make bonus: 0 and chance:100 as default value?

Smokexyz:
Bonus can be omitted currently, the default value is 0, for chance i'll add the default as 100. Is it possible to add a number value in key? As i tested this before it gave me a syntax error, I'll test this again.

Smokexyz:
It is also possible to change the rate format to -
```
	Rate: (
		[1,10,200],
		[2,20,200],
		[3,30,300],
		[4,40,400], // Level 4, Chance 40%, Bonus 400.
	),
```
using arrays instead of key names.

MishimaHaruna:
It's possible to use a number in key names, but they must begin with a letter (or an asterisk, apparently). That's why I suggested `Lv<N>`

Reference: http://www.hyperrealm.com/libconfig/libconfig_manual.html#Configuration-Files

malufett:
@Smokexyz 
lol..I don't mean to scrap your work..I'm just suggesting..:D

@Ancyker 
I know the format has syntax error...what I'm telling is the design flow...also I'm concern on the efficiency of the parsing..look at @MishimaHaruna design and compare...

MishimaHaruna:
@malufett there should be no concerns at all about the libconfig efficiency. We benchmarked it when we switched the item db to libconfig (we expected it to be horribly slow), and we figured out it was faster than the previous txt (csv) parser.

malufett:
@MishimaHaruna yup indeed but what I'm talking to is the efficiency of the design to be parsed if does it requires to much loops, to many function calls, can be easily code or fit to the current db structure, can it deliver its purpose and etc..hehehehehe

Ancyker:
I like the design suggested by haru, but is the Lv prefix really necessary?

Smokexyz:
@MishimaHaruna That's awesome, thanks for the reference, I'll update the format.

emistry:
well, I have never did any LibConfig stuff previously, so I dunno how effectively parsed into the server.
I am just suggesting a template that could provide a better readability compare to the current text files..

For the "Level" of refine, I think isn't really needed. 
True that it can help user clearly identify the level of refine when adjusting, but at the same time it could possible bring up another issue, where users are dumb enough to duplicate the contents and end up you have multiple refine bonus that point to same refine ? 
or does libconfig setting able to overwrite it ?  but in case it overwrite, user will still end up getting error for having incomplete entry.

I vote for -1 to add the "refine level".


```
    Rate: (
        [1,10,200],
        [2,20,200],
        [3,30,300],
        [4,40,400], // Level 4, Chance 40%, Bonus 400.
    ),
```
I like this too =D

MishimaHaruna:
@emistry the reason why I kept the level is so that we can throw warnings in case of duplicate or missing levels (actually, since the level is on the key, libconfig will throw a warning if it detects a duplicate key if I remember correctly)

@Ancyker Unfortunately the prefix is necessary. It's the same reason why in other libconfig files we prefixed item IDs with `ID` (i.e. item groups or item packages)

4144:
This configuration hard to read, and it cant be extended without pain.

    Rate: (
        [1,10,200],
        [2,20,200],
        [3,30,300],
        [4,40,400], // Level 4, Chance 40%, Bonus 400.
    ),

What if need for each level add some new parameter? I speaking not about plugins, but what if aegis did this? It already impossible to human understand what is 1, what is 10 and what is 200. Next will be 50, 80, ... etc. This named csv. And csv must be ripped form hercules.

Haru example is most readable version.

Smokexyz:
Done, the new format is as Haru designed. Also removing empty refine levels that use default values.
```
Armors/WeaponLevel1~4: { (0 - Armors, 1 - Level 1 Weapons, 2 - Level 2 Weapons, 3 - Level 3 Weapons, 4 - Level 4 Weapons)
						- For armors, values of 100 add 1 armor defense.
						- For weapons, values of 100 add 1 ATK&MATK.
	StatsPerLevel: (int)			This value is applied for ever level.
	RandomBonusStartLevel: (int)		This value specifies the start point for those levels that give a random bonus value (usually the first unsafe upgrade).
	RandomBonusValue: (int)			A random number between 0 and (Random bonus start level - Upgrade level + 1) * this value is applied for all upgrades past.
						Random bonus start level. This is only applied for weapons, and not displayed client-side.
	Rates: {				Per level configuration of the refine rates.
		Lv1~10: {			Level of refine
			Chance: (int)			Chance of successful refine (100 = 100%)
			Bonus: (int)			Bonus for this level of refine (Optional Field)
		}
	}
}
```

@4144 It's an option, but yes it's unreadable hence the first commit i made followed a legible format.

MishimaHaruna:
Starting review

db/pre-re/refine_db.conf
```diff
@@ -0,0 +1,125 @@
...
+ *			Pre-Renewal Refine Database				*
+ ********************************************************************************
+Armor/WeaponLevel1~4: { (0 - Armors, 1 - Level 1 Weapons, 2 - Level 2 Weapons, 3 - Level 3 Weapons, 4 - Level 4 Weapons)
+						- For armors, values of 100 add 1 armor defense.
+						- For weapons, values of 100 add 1 ATK.
+	StatsPerLevel: (int)			This value is applied for ever level.
+	RandomBonusStartLevel: (int)		This value specifies the start point for those levels that give a random bonus value (usually the first unsafe upgrade).
+	RandomBonusValue: (int)			A random number between 0 and (Random bonus start level - Upgrade level + 1) * this value is applied for all upgrades past.
+						Random bonus start level. This is only applied for weapons, and not displayed client-side.
+	Rates: {				Per level configuration of the refine rates.
```
MishimaHaruna: Could you add a mention about the lower entries being optional (and the default values in case an entry is omitted?)

db/pre-re/refine_db.conf
```diff
@@ -0,0 +1,125 @@
...
+		Lv1~10: {			Level of refine
+			Chance: (int)			Chance of successful refine (100 = 100%)
```
MishimaHaruna: From what I see below, chance also is optional (and defaults to 100) - it'd be nice to have it mentioned here

src/map/status.c
```diff
@@ -12278,38 +12279,118 @@ bool status_readdb_sizefix(char* fields[], int columns, int current)
...
+ *                validation errors.
+ * @return # of the validated entry, or 0 in case of failure.
+ */
+int status_readdb_refine_libconfig_sub(config_setting_t *r, const char *name, const char *source) {
+	config_setting_t *rate = NULL;
+	int i=0, type=0, bonus_per_level=0, rnd_bonus_v=0, rnd_bonus_lv=0;
+	
+	if(!strncmp(name, "Armors", 6))
+		type = 0;
+	else if(!strspn(&name[strlen(name)-1], "0123456789") || !(type = atoi(strtok(strdup(name), "WeaponLevel")))) {
```
MishimaHaruna: You're calling `strdup`, and never freeing its result. While this doesn't show up in the builtin memory leak detector (you're working around it by calling `strdup` directly instead of `aStrdup`), it still leaks memory.
You can avoid this entirely, by checking the string length: if the string is longer than the length of `"WeaponLevel"`, then you can `strncmp` it against `"WeaponLevel"` (using the appropriate max length), and then call `atoi` on the fist index after that.

src/map/status.c
```diff
@@ -12278,38 +12279,118 @@ bool status_readdb_sizefix(char* fields[], int columns, int current)
...
+		memset(&duplicate,0,sizeof(duplicate));
+		memset(&bonus_,0,sizeof(bonus_));
+		memset(&rnd_bonus,0,sizeof(rnd_bonus));
+		memset(&chance_,0,sizeof(chance_));
+		i=0;
+		while( (t = libconfig->setting_get_elem(rate,i++)) && config_setting_is_group(t) ) {
+			int level=0, chance=0, bonus=0;
+			char *rlvl = config_setting_name(t);
+			
+			if(!strspn(&rlvl[strlen(rlvl)-1], "0123456789") || (level = atoi(strtok(rlvl, "Lv"))) <= 0 ) {
```
MishimaHaruna: This doesn't need `strtok`. `strncmp` is what you're looking for.

By the way, from the `strtok` manual page:

> The `strtok()` function is used to isolate sequential tokens in a null-terminated string, `str`.  These tokens are separated in the string by **at least one of the characters in `sep`**.  [...]

In other words, `strtok(foo, "Lv")` uses as separator **either** `'L'` **or** `'v'`, not the `"Lv"` string.

src/map/status.c
```diff
@@ -12278,38 +12279,118 @@ bool status_readdb_sizefix(char* fields[], int columns, int current)
...
+		ShowError("can't read %s\n", filepath);
+		return 0;
 	}
-	return true;
+	
+	while( (r = libconfig->setting_get_elem(refine_db_conf.root,i++)) ) {
+		char *name = config_setting_name(r);
+		if((type=status->readdb_refine_libconfig_sub(r, name, filename))) {
+			if( duplicate[type-1] ) {
+				ShowError("status_readdb_refine_libconfig: duplicate entry for %s in \"%s\", skipping... \n", name, filename);
```
MishimaHaruna: The message mentions skipping, but the code isn't skipping it. `status_readdb_refine_libconfig_sub` overwrites any existing values, rather than skipping. Skipping might be a little tricky (you could move the code that retrieves the type here if you really want to skip it).

src/map/status.c
```diff
@@ -12278,38 +12279,118 @@ bool status_readdb_sizefix(char* fields[], int columns, int current)
...
+				continue;
+			} else duplicate[type-1] = true;
+			count++;
+		}
+	}
+    libconfig->destroy(&refine_db_conf);
```
MishimaHaruna: I believe this line has wrong indentation. It seems 4 spaces rather than a tab.

db/pre-re/refine_db.conf
```diff
@@ -0,0 +1,125 @@
...
+			Chance: 20
+		}
+		Lv9: {
+			Chance: 20
+		}
+		Lv10: {
+			Chance: 10
+		}
+	}
+}
```
MishimaHaruna: The file is missing a newline at eof

Smokexyz:
Thanks, updated with corrections. For the duplicate entry part I just left it with a warning since I think skipping would be unnecessary. The checks are corrected. Do let me know if i've missed something out, seems to be working fine otherwise.

MishimaHaruna:
Looks good now. Thank you very much for your contribution (and patience). Merged.

PR #690, opened 2015-09-03
Added exitwhile script syntax.

kisuka:
This small addition to the scripting syntax is for scenarios where you need to break out of a while loop from inside a switch that has been nested inside the loop.

Example:

```
while(1) {
    switch(select("Option 1", "Option 2", "Option 3")) {
    case 1:
        mes "Return to menu";
        next;
        break;
    case 2:
        exitwhile;
    case 3:
        close;
    }
}
mes "Option 2 selected";
```

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -1801,6 +1801,38 @@ const char* parse_syntax(const char* p)
 			return p;
 		}
 		break;
+	case 'E':
+	case 'e':
+		if( p2 - p == 9 && strncmp(p, "exitwhile", 9) == 0 ) {
+			char label[256];
+			int pos = script->syntax.curly_count - 1;
+
+			while(pos >= 0) {
+				if(script->syntax.curly[pos].type == TYPE_WHILE) {
+					sprintf(label, "goto __WL%x_FIN;", script->syntax.curly[pos].index);
+					break;
+				}
+				pos--;
+			}
+
+			if(pos < 0) {
+				disp_error_message("parse_syntax: unexpected 'exitwhile'", p);
+			} else {
+				script->syntax.curly[script->syntax.curly_count++].type = TYPE_NULL;
+				script->parse_line(label);
+				script->syntax.curly_count--;
+			}
+
+			p = script->skip_space(p2);
+
+			if(*p != ';')
+				disp_error_message("parse_syntax: need ';'", p);
+
+			p = script->parse_syntax_close(p + 1);
+
+			return p;
+		}
+		break;
 	case 'f':
 	case 'F':
 		if( p2 - p == 3 && strncmp(p, "for", 3) == 0 ) {


MishimaHaruna:
I like this, but I wonder whether we should have a more flexible approach. This doesn't allow you to jump out of the outer of two nested while loops for example, nor it does help if you have a for or do/while loop.

I was thinking of labeled blocks (see [perl](http://perldoc.perl.org/perlsyn.html#Basic-BLOCKs) or [Java](http://docs.oracle.com/javase/specs/jls/se7/html/jls-14.html#jls-14.7)), but with a syntax like the following (to make it easier on our parser):

```
while (true) OUTER: {
    debugmes "Outer loop beginning"; // [1]
    for (.@i = 0; .@i < foo; ++.@i) INNER: {
        debugmes "Inner loop beginning"; // [2]
        switch (.@bar[.@i]) {
        case 0:
            break INNER; // Jumps out of the for and executes [4]
        case 1:
            break OUTER; // Jumps out of the while and executes [5]
        case 2:
            continue INNER; // Continues the for and executes [2]
        case 3:
            continue OUTER; // Continues the while and executes [1]
        default:
            break; // Leaves the switch and executes [3]
        }
        debugmes "After switch"; // [3]
        while (foo) {
            --foo; // Unlabeled loops still work as usual
        }
    }
    debugmes "After inner loop"; // [4]
}
debugmes "After outer loop"; // [5]
```

Implementation shouldn't be too complicated, and it involves adding a .name field to the `script->syntax.curly[]` struct that is filled with the loop name (if labeled), and slight edit to the `break` and `continue` parsers so that they lookup by name if needed. Of course those labels wouldn't work with `goto` and the likes (i.e. they're not regular labels, they're just used to tag the loop)

kisuka:
I think that just complicates things and makes the already ugly syntax more ugly :/ sorry. This type of command is used very specifically in aegis and used very rarely. But it's required nonetheless for upcoming updates that use it. The only time exitwhile is used in aegis is during situation as described above. It's always only in a while loop that has a select inside it. No further nesting ever or above it.

MishimaHaruna:
Honestly, who cares about when it is used in AEGIS? We shouldn't copy their script structure 1:1, we should emulate their behavior (possibly optimizing the script)

kisuka:
This would require a huge rewrite to loops and backwards compatibility for the current scripts, right? It seems like a massive overhaul for something that I just need for some new scripts.

Ancyker:
Why not do what other languages do and add a parameter to break specifying how many loops/switches to break out of?

kisuka:
@Ancyker That would require having to figure out what type of loop is that many break outs above where the break happened. I'm not even sure if it's possible. I only made this pull request because this is a simple solution to something that is holding me back from releasing ep 14.2 files.

MishimaHaruna:
@kisuka This is *completely* backwards compatible. The label is optional. If no label is specified, it works exactly like it used to. If a label is specified, you can still use an unnamed break/continue (and you get the old behavior), or a named one (and you get the new behavior. It's just an extension to the old syntax.

@Ancyker Other languages do exactly this. We're adding a parameter to break/continue. Instead of a number, it's a name, so that you *can't* get it wrong and break out of the wrong number of loops (php does it the other way, keeps them unnamed. Perl and java do it this way).

Ancyker:
Wouldn't this do what you want:

While() {
Switch() {
Case 1: continue;
Case 2: break;
}
Break;
}

kisuka:
@MishimaHaruna would u be willing to code this? It's a bit beyond me, I wouldn't want to mess it up in some way or delay it because of my lack of skill.

@Ancyker In the scenario I listed above it would not work. the first option needs to go back to the switch() which happens because of it breaking out of the switch with break. The 2nd option needs to jump out of the switch. I can't put a break after the switch because the first option would then follow it and break out of the loop.

Ancyker:
while(1) {
    switch(select("Option 1", "Option 2", "Option 3")) {
    case 1:
        mes "Return to menu";
        next;
        continue;
    case 2:
        break;
    case 3:
        close;
    }
    break;
}
mes "Option 2 selected";

kisuka:
@Ancyker welp, I'm dumb. Thanks for the refresher. Nevermind this then. that did the trick.

Smokexyz:
variables could be used in the outer loop to trigger a break/continue after breaking from the inner switch/loop function. like so -
while(1) {
switch(select("1","2")) {
case 1:
 set .@breakwhile,1;
break;
case 2:
case 3:
break;
}
if(.@breakwhile) break;
}

jaBote:
Sorry for being late, but wouldn't this be even a better approach for what you're trying to do?

```
@menu = 0; // It would be convenient to note here that select sets @menu to your last choice
while(@menu != 2){
    switch(select("Option 1", "Option 2", "Option 3")) {
	case 1:
        mes "Return to menu";
        next;
        break;
    case 3:
        close;
    }
}
mes "Option 2 selected";
```

PR #696, opened 2015-09-05
Episode 14.2 Criatura Academy Release.

kisuka:


hemagx:
it's very good to have this update finally ! thanks for it kisuke, but i believe we should have a start point setting for RE PRE-RE as well cause this update just broke it

kisuka:
@hemagx working on it now.

PR #451, opened 2015-01-21
autotrade_vending struct to HPM

dastgir:
Updated the points mentioned in #373 (I could not reopen it, due to rebase done in the branch)
<a href='#crh-start'></a><a href='#crh-data-%7B%22processed%22%3A%20%5B%22https%3A//github.com/HerculesWS/Hercules/pull/451%23issuecomment-83663989%22%2C%20%22https%3A//github.com/HerculesWS/Hercules/pull/451%23issuecomment-83667691%22%5D%2C%20%22comments%22%3A%20%7B%22General%20Comment%22%3A%20%7B%22html_url%22%3A%20%22https%3A//github.com/HerculesWS/Hercules/pull/451%23issuecomment-83663989%22%2C%20%22comments%22%3A%20%5B%7B%22body%22%3A%20%22resolve%20merge%20conflicts%20%3AD%22%2C%20%22created_at%22%3A%20%222015-03-19T17%3A01%3A44Z%22%2C%20%22user%22%3A%20%7B%22avatar_url%22%3A%20%22https%3A//avatars.githubusercontent.com/u/4684880%3Fv%3D3%22%2C%20%22html_url%22%3A%20%22https%3A//github.com/EPuncker%22%7D%7D%2C%20%7B%22body%22%3A%20%22Would%20do%20it%20soon%20%2C%20%28wonder%20why%20it%20took%20so%20long%20to%20be%20merged%29%22%2C%20%22created_at%22%3A%20%222015-03-19T17%3A05%3A20Z%22%2C%20%22user%22%3A%20%7B%22avatar_url%22%3A%20%22https%3A//avatars.githubusercontent.com/u/2889931%3Fv%3D3%22%2C%20%22html_url%22%3A%20%22https%3A//github.com/dastgir%22%7D%7D%5D%2C%20%22title%22%3A%20%22General%20Comment%22%7D%7D%7D'></a>
<a href='https://www.codereviewhub.com/'></a>
- [x] <a href='#crh-comment-General Comment'></a> &nbsp;<b><a href='https://github.com/HerculesWS/Hercules/pull/451#issuecomment-83663989'>General Comment</a></b>
- <a href='https://github.com/EPuncker'></a> resolve merge conflicts :D
- <a href='https://github.com/dastgir'></a> Would do it soon , (wonder why it took so long to be merged)

[...]

PR #711, opened 2015-09-09
Autotrade Struct to HPM

dastgir:
Updated Autotrade Struct (Previous PR: #451 , it was difficult to update the old branch, so I just delete and created new)

also didn't saw that I have opened PR to update that minor thing :P


4144:
Compilation error

MishimaHaruna:
Thank you!

PR #725, opened 2015-09-19
Fix incorrect condition checks in attaching map of Sealed Shrine Instance.

QQfoolsorellina:


QQfoolsorellina:
original condition .@i < 2   will only check 1@cata

hemagx:
You used endless tower map

QQfoolsorellina:
@hemagx   this commit already use @cata ..... XD

PR #728, opened 2015-09-20
Fixed merge behaviour of map zones. Fixes #572

### LINKED ISSUE #572: map_zone_db and GvG mapflag
Refer to this:
http://herc.ws/board/topic/302-introducing-hercules-map-zone-database/page-5#entry57634

It goes like this:

Boot the server, everything's working, PvP-specific skills and gear aren't working in castles since WoE isn't active.

WoE begins, GvG mapflag is applied and now all that stuff can be used.

WoE ends, GvG mapflag is turned off, now this is where it breaks, the next time WoE starts, mapflag won't properly apply and all WoE-specific stuff will be disabled since maps are treated as normal zones. Setting it manually with @mapflag doesn't work either. Only solution is to reload scripts or reboot and having to do this after every WoE isn't exactly desirable.

This is a rather serious issue that's been around for quite a long time.

k-py:
- Both "Normal" and "PK Mode" default zones will now be overwritten (instead of merged) by "GvG", "PvP" and "Battlegrounds" zones.
- Refactored some code to clarify intent.

### DIFF
--- a/src/map/map.c
+++ b/src/map/map.c
@@ -3854,6 +3854,7 @@ struct map_zone_data *map_merge_zone(struct map_zone_data *main, struct map_zone
 
 	CREATE(zone, struct map_zone_data, 1);
 	safestrncpy(zone->name, newzone, MAP_ZONE_NAME_LENGTH);
+	zone->merge_type = MZMT_NEVERMERGE;
 	zone->disabled_skills_count = main->disabled_skills_count + other->disabled_skills_count;
 	zone->disabled_items_count = main->disabled_items_count + other->disabled_items_count;
 	zone->mapflags_count = main->mapflags_count + other->mapflags_count;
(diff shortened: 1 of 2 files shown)


src/map/map.c
```diff
@@ -5275,12 +5275,13 @@ void read_map_zone_db(void) {
 		ShowStatus("Done reading '"CL_WHITE"%d"CL_RESET"' zones in '"CL_WHITE"%s"CL_RESET"'.\n", zone_count, config_filename);
 		
 		/* post-load processing */
+		map->zone_all.merge_type = MZMT_NEVERMERGE;
```
MishimaHaruna: I only see `MZMT_NEVERMERGE` and `MAMT_MERGEABLE` being used. If this relies zeroed/allocated zones having an implicit `MZMT_NORMAL` (0), I believe it'd be better to explicitly initialize those explicitly.

k-py:
Indeed, that is a good idea. And after reviewing the code, it allowed me to realize that I had missed setting the merge_type for the "PK Mode" zone.

All done and branch rebased.

MishimaHaruna:
Looks good, thank you very much! Merged

PR #742, opened 2015-09-25
Fixed a bug in Lv4 Weapon Quest dialog

michaelforge:
Dialog with rolled rock/paper/scissors didn't appear because of variable typo.
An ancient one. Discovered by Aafemt.


dastgir:
Thanks for Contribution!

PR #751, opened 2015-09-28
Updated script_commands.txt

dastgir:
* Added Missing ScriptCommands in the doc.
* Corrected sleep/sleep2 parameter list.
* Changed bg_join_team return values.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -4457,6 +4457,14 @@ they will also have their skills reset upon 'changesex'.
 
 ---------------------------------------
 
+*changecharsex;
+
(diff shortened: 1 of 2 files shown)


dastgir:
Still have 1 missing 'mergeitem'
I don't know what exactly it does, and how it works, so haven't included it.

MishimaHaruna:
In case you want to add it, `mergeitem` opens the item merge window (see `npc/other/item_merge.txt`). Takes no arguments, returns nothing.

The item merge window shows any stackable items with different serial (i.e. coming from separate cash shop packages) that can be merged into one stack.

PR #752, opened 2015-09-28
Extra parameter for script command resetstatus.

Emistry:
Support extra 'type' parameters.
Required to disable battle_config.use_statpoint_table, to avoid status
point abuse.

Sample NPC for testing.
http://pastebin.com/raw.php?i=qxBLumHR

```
prontera,155,171,5  script  Reset#sample    861,{
    mes "Reset which status ?";
    .@i = select( "All","STR","AGI","VIT","INT","DEX","LUK" ) - 1;
    if ( .@i )
        .@i += 12;
    .@value = resetstatus(.@i);
    mes "Gained "+F_InsertComma( .@value )+" status points.";
    close;
}
```

> Edit by Haru -- inlined the code snippet here (as it's very short), to make it future proof

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -5438,11 +5438,22 @@ official rebirth scripts.
 
 ---------------------------------------
 
-*resetstatus;
+*resetstatus {<type>}; 
 
 This is a character reset command, which will reset the stats on the 
 invoking character and give back all the stat points used to raise them 
 previously. Nothing will happen to any other numbers about the character.
+Return the amount of statuspoint gained from reset.
+
+Valid type parameter are:
+	 0 - All (default)
+	// must disable battle_config.use_statpoint_table
+	13 - bSTR
+	14 - bAGI
+	15 - bVIT
+	16 - bINT
+	17 - bDEX
+	18 - bLUK
 
 Used in reset NPC's (duh!).
 
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -1388,7 +1388,7 @@ ACMD(baselevelup)
 		for (i = 0; i > -level; i--)
 			status_point += pc->gets_status_point(sd->status.base_level + i - 1);
 		if (sd->status.status_point < status_point)
-			pc->resetstate(sd);
+			pc->resetstate(sd,0);
 		if (sd->status.status_point < status_point)
 			sd->status.status_point = 0;
 		else
@@ -6257,7 +6257,7 @@ ACMD(users)
  *
  *------------------------------------------*/
 ACMD(reset) {
-	pc->resetstate(sd);
+	pc->resetstate(sd,0);
 	pc->resetskill(sd, PCRESETSKILL_RESYNC);
 	safesnprintf(atcmd_output, sizeof(atcmd_output), msg_fd(fd,208), sd->status.name); // '%s' skill and stats points reseted!
 	clif->message(fd, atcmd_output);
@@ -8002,7 +8002,7 @@ ACMD(allowks)
 
 ACMD(resetstat)
 {
-	pc->resetstate(sd);
+	pc->resetstate(sd,0);
 	safesnprintf(atcmd_output, sizeof(atcmd_output), msg_fd(fd,207), sd->status.name);
 	clif->message(fd, atcmd_output);
 	return true;
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -7173,66 +7173,89 @@ int pc_resetlvl(struct map_session_data* sd,int type)
 /*==========================================
  * /resetstate
  *------------------------------------------*/
-int pc_resetstate(struct map_session_data* sd)
+/*==========================================
+* /resetstate
+*------------------------------------------*/
+int pc_resetstate(struct map_session_data* sd, int type)
 {
 	nullpo_ret(sd);
+	int add = 0;
 
-	if (battle_config.use_statpoint_table) {
-		// New statpoint table used here - Dexity
-		if (sd->status.base_level > MAX_LEVEL) {
-			//pc->statp[] goes out of bounds, can't reset!
-			ShowError("pc_resetstate: Can't reset stats of %d:%d, the base level (%d) is greater than the max level supported (%d)\n",
-				sd->status.account_id, sd->status.char_id, sd->status.base_level, MAX_LEVEL);
-			return 0;
+	if (type == 0) {
+		if (battle_config.use_statpoint_table) {
+			// New statpoint table used here - Dexity
+			if (sd->status.base_level > MAX_LEVEL) {
+				//pc->statp[] goes out of bounds, can't reset!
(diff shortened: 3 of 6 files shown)


dastgir:
Maybe we should encourage constants??
changing
```
Valid type parameter are:
	 0 - All (default)
	// must disable battle_config.use_statpoint_table
	13 - bSTR
	14 - bAGI
	15 - bVIT
	16 - bINT
	17 - bDEX
	18 - bLUK
```
to
```
Valid type parameter are:
	0 - All (default)
	// battle_config.use_statpoint_table must be disable to use type.
	bStr
	bAgi
	bVit
	bInt
	bDex
	bLuk
```

emistry:
@dastgir 
but the "default - type 0" doesnt have constant.

dastgir:
that's why I haven't removed 0 from the list I have shown..

4144:
If no constant, you can always add it.
for example bAll and assign to it 0

emistry:
btw, shall we default the settings to "No" as well ?
https://github.com/HerculesWS/Hercules/blob/master/conf/battle/exp.conf#L91

I don't know if there exist any other work around for it,
but based on current Hercules setting, it's impossible to support the 'type' parameter.

if the statuspoint calculation was correct since the beginning, why don't we use it instead of the statspoint.txt ??

emistry:
now I found the video finally thx for ziu.
https://www.youtube.com/watch?v=d31q2m0i5lM

MishimaHaruna:
From what I can see, the Renewal statpoint.txt doesn't follow the formula (it stops doing so a at level 105->106, and grows a little slower instead). I believe that might be the reason why we use the stat point table.

emistry:
I tried using the `resetstatus();` to compare both status point return from the statpoint table and the auto generated statuspoint, it seem like both having same total amount at the end.

haven't tried with level 105 ~ 106.

AnnieRuru:
dig out this topic
https://rathena.org/board/topic/91297-wrong-stat-points-given-on-set-baselevel/?p=239424

tested, with renewal test server
`use_statpoint_table: yes`
level 175 rune knight has 3295 stat points upon `@reset`

`use_statpoint_table: no`
level 175 rune knight has 3545 stat points upon `@reset`
which follow the pre-renewal formula

@emistry 
where is the
`BUILDIN_DEF(resetstatus,"?"),`
I can't find it in the patch

MishimaHaruna:
Since we can't disable the use_statpoint_table setting, what do we want to do about this?

AnnieRuru:
1. close this pull request, and release a source/plugin release on the forum,
and tell them that only works for pre-renewal
2. try make out a formula for renewal :O I wonder is there anyone can make such complicated formula
3. put `ifdef RENEWAL` in the source code, and make this suggestion only for pre-renewal.
Also update the script_commands.txt to tell users the optional parameter only for pre-renewal

no, I'm not emistry, just my opinion

MishimaHaruna:
All three options sound good. I think 1 and 2 would be better than 3.

In case of option 1, we also have the StaffPlugins repository (which I intend on extending some time soon, and making its integration with the main repository easier).

In case of option 2... The formula might not be as hard as it seems... From what I see:

- At level 1, it adds a fixed amount of 48 points
- From level 2 to level 100, it adds `3 + (level-1)/5` points at each level (and this is the same as pre-renewal).
- From level 101 to level 150, it adds `3 + (101-1)/5 + (level-101)/10 => 3 + 20 + (level-101)/10 => 23 + (level-101)/10` points at each level.
- From level 151 to level 175, it adds `3 + (101-1)/5 + (151-101)/10 + (level-151)/7 => 3 + 20 + 5 + (level-151)/7 => 28 + (level-151)/7` points at each level.
- After level 175 the table is unknown (our current file goes up to 255, but if you put it in a spreadsheet, and calculate the difference between each line and the previous, you can clearly see that after 175 it makes no sense at all)

MishimaHaruna:
Here's a spreadsheet:  https://docs.google.com/spreadsheets/d/1qLBS3VmJuGCR9yHK-B7zR7i3E5Hgp58w18_PFrVYkhE/edit?usp=sharing

AnnieRuru:
>In case of option 2... The formula might not be as hard as it seems...

*bow down to the almighty Haru-sama*

the formula for pre-re, defined in the source code is here
https://github.com/HerculesWS/Hercules/blob/master/src/map/pc.c#L10928

if you are interested in optimizing the formula for pre-renewal,
try take a look on this topic on eathena
https://www.eathena.ws/board/index.php?s=&showtopic=192349&view=findpost&p=1329753

the npc script used for checking is
```
prontera,150,181,5	script	show stats#1	1_F_MARIA,{
	for ( .@i = 0; .@i < BaseLevel; ++.@i )
		.@stat += .@i /5 + 3;
	.@stat += ( Upper == 1 )? 97:45;
//	insert stuff here for renewal formula !
	dispbottom strnpcinfo(2) +": "+ .@stat;
	end;
}
```
I guess its time for playing algorithm script again XD

```
prontera,155,181,5	script	show stats#2	1_F_MARIA,{
	//3 + (level-1)/5
	if ( BaseLevel < 100 ) {
		for ( .@i = 0; .@i < BaseLevel; ++.@i )
			.@stat += .@i /5 + 3;
	}
	//23 + (level-101)/10
	else if ( BaseLevel < 150 ) {
		for ( .@i = 0; .@i < 100; ++.@i )
			.@stat += .@i /5 + 3;
		for ( .@i = 100; .@i < BaseLevel; ++.@i )
			.@stat += (.@i-100) /10 + 23;
	}
	//28 + (level-151)/7
	else if ( BaseLevel <= 175 ) {
		for ( .@i = 0; .@i < 100; ++.@i )
			.@stat += .@i /5 + 3;
		for ( .@i = 100; .@i < 150; ++.@i )
			.@stat += (.@i-100) /10 + 23;
		for ( .@i = 150; .@i < BaseLevel; ++.@i )
			.@stat += (.@i-151) /7 + 28;
	}
	.@stat += ( Upper == 1 )? 97:45;
	dispbottom strnpcinfo(2) +": "+ .@stat;
	end;
}
```
erm ... its ok until level 157, but after level 158 and above, the value seems different ...
are you sure is `28 + (level-151)/7` ?
or maybe my script is wrong ...

MishimaHaruna:
> 5a5a327#diff-14ed211b89b5f27543e7901daff24590L157
> I think the reason why it jump after level 175 is because

Ohh I see, makes sense. So the remaining values still use the previous formula.

Maybe we can consider getting rid of the table altogether?

HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/752)

Emistry:
I am closing this PR, since the intention to add this script command was to allow one to reset their status and gain back the status points that used to raise the stat.
I believe it could be done using the update from this PR https://github.com/HerculesWS/Hercules/pull/2246 already.
Feel free to re-open if anyone feel like this is needed, I will try update it.

PR #758, opened 2015-09-29
Fix in academy.txt

Atemo:


dastgir:
Thanks for Contribution :+1:

PR #764, opened 2015-10-01
Fixed 2 Sealed Shrine Bugs

Ancyker:
So these were fixed a long long time ago, and then broken again a long time ago.


Nihadm89:
The maelstrom fix I committed a few days ago. 
...
https://github.com/HerculesWS/Hercules/commit/d52dde700d0f1a598ebea87844007ff610352114
...
It's a fix based on Haru's idea. It does resolve the issue, confirmed on my server.

...
Info here:
https://github.com/HerculesWS/Hercules/pull/575
...

PR #772, opened 2015-10-06
Added Packet DB Entrys for 2015-09-16 RagexeRE

ossi0110:
Added Packet DB Entrys for 2015-10-01 RagexeRE
Fixed Chat Packet for  2015-09-XX RagexeRE Clients


ossi0110:
Commit will be aded in a upcomming PR

PR #778, opened 2015-10-08
Convert mob_db from txt to conf format.

4144:
Converting mob_db.txt and mob_db2.txt to mob_db.conf and mob_db2.conf

Renamed some items to correct format. Was not supported dots and started from number.
Also removed sql mob_db loading.
Converter included.


4144:
Still need testing

kyeme:
i can't see cards in drops?

4144:
Cards you mean field?
It useless, cards can be added to normal drops

4144:
Ok travis passed, this mean it ready for testing

kyeme:
I mean this
from example Poring
	Drops: {
		Jellopy: 7000
		Knife_: 100
		Sticky_Mucus: 400
		Apple: 1000
		Empty_Bottle: 1500
		Apple: 150
		Unripe_Apple: 20
	}

no Poring_Card

4144:
@kyeme ah ok, thanks will add to converter

db/re/item_db.conf
```diff
@@ -43668,7 +43668,7 @@ item_db: (
 },
 {
 	Id: 5012
-	AegisName: "Ph.D_Hat"
```
MishimaHaruna: We need to figure out a way to avoid editing constants. I'll see what I can do with libconfig, otherwise we can fall back to the list-of-lists approach in the drops list of the mob db

src/map/mob.c
```diff
@@ -4630,13 +4866,10 @@ void mob_load(bool minimal) {
 	}
 	sv->readdb(map->db_path, "mob_item_ratio.txt", ',', 2, 2+MAX_ITEMRATIO_MOBS, -1, mob->readdb_itemratio); // must be read before mobdb
 	mob->readchatdb();
-	if (map->db_use_sql_mob_db) {
```
MishimaHaruna: This removes the SQL mob database, without removing it entirely (config files, config parser). I have a commit in the `deprecationwarning` branch that gets rid of teh config files entries as well, I shall merge that before merging this pull request

src/map/mob.c
```diff
@@ -3646,77 +3646,354 @@ void item_dropratio_adjust(int nameid, int mob_id, int *rate_adjust)
...
-	db->lv = atoi(str[4]);
-	db->lv = cap_value(db->lv, 1, USHRT_MAX);
-	mstatus->max_hp = atoi(str[5]);
-	mstatus->max_sp = atoi(str[6]);
+	if (!libconfig->setting_lookup_string(mobt, "Name", &str) || !*str ) {
+		ShowWarning("mob_read_db_sub: Missing Name in mob %d of \"%s\", skipping.\n", class_, source);
+		aFree(entry);
+		return false;
+	} else {
+		safestrncpy(entry->name, str, sizeof(entry->sprite));
```
MishimaHaruna: Wrong sizeof

src/map/mob.c
```diff
@@ -3646,77 +3646,354 @@ void item_dropratio_adjust(int nameid, int mob_id, int *rate_adjust)
...
+		safestrncpy(entry->jname, str, sizeof(entry->sprite));
```
MishimaHaruna: Is there a reason why we want to keep name and jname? If not, we can completely remove one of the two

src/map/mob.c
```diff
@@ -3646,77 +3646,354 @@ void item_dropratio_adjust(int nameid, int mob_id, int *rate_adjust)
...
+	}
+
+	if ((t = libconfig->setting_get_member(mobt, "Stats"))) {
+		if (config_setting_is_group(t)) {
+			mob->read_db_stats_sub(entry, mstatus, class_, t);
+		} else if (mob->lookup_const(mobt, "Stats", &i32) && i32 >= 0) {
+			mstatus->str = mstatus->agi = mstatus->vit = mstatus->int_ = mstatus->dex = mstatus->luk =
+				mob_parse_dbrow_cap_value(class_, UINT16_MIN, UINT16_MAX, i32);
+		}
+	}
```
MishimaHaruna: Original stat-parsing code was ensuring these values wouldn't go below 1, to avoid divisions by zero and such (at least in pre-renewal builds). If we remove that, we should investigate to make sure it's safe to.

src/map/mob.c
```diff
@@ -3733,74 +4010,126 @@ bool mob_parse_dbrow(char** str) {
 #endif
 
 	//Tests showed that chase range is effectively 2 cells larger than expected [Playtester]
-	if (db->range3 > 0)
-		db->range3 += 2;
+	if (entry->range3 > 0)
```
MishimaHaruna: It's funny how this check (not the fault of this commit, it was already like this) is done before actually reading the value...

src/map/mob.c
```diff
@@ -3733,74 +4010,126 @@ bool mob_parse_dbrow(char** str) {
...
-		if (db->range3 < db->range2)
-			db->range3 = db->range2;
+		entry->range3 = entry->range3 * battle_config.chase_range_rate / 100;
+		if (entry->range3 < entry->range2)
+			entry->range3 = entry->range2;
 	}
 
-	mstatus->size = atoi(str[22]);
-	mstatus->race = atoi(str[23]);
+	if (mob->lookup_const(mobt, "Size", &i32) && i32 >= 0) {
```
MishimaHaruna: Since we're at it, we should check that this is a valid SIZE_ value

src/map/mob.c
```diff
@@ -3733,74 +4010,126 @@ bool mob_parse_dbrow(char** str) {
...
+		mstatus->size = i32;
+	} else {
+		mstatus->size = 1;
+	}
+
+	if (mob->lookup_const(mobt, "Race", &i32) && i32 >= 0) {
```
MishimaHaruna: Same here, better check that it is a valid race

src/map/mob.c
```diff
@@ -3733,74 +4010,126 @@ bool mob_parse_dbrow(char** str) {
...
+		mstatus->race = i32;
+	}
 
-	i = atoi(str[24]); //Element
-	mstatus->def_ele = i%10;
-	mstatus->ele_lv = i/20;
+	if (mob->lookup_const(mobt, "ElementType", &i32) && i32 >= 0) {
```
MishimaHaruna: Suggestion about `ElementType` / `ElementLevel`:

```
Element: ("Fire", 4)
```

src/map/mob.c
```diff
@@ -3646,77 +3646,354 @@ void item_dropratio_adjust(int nameid, int mob_id, int *rate_adjust)
...
+			ShowWarning("Item %s not found\n", name);
+			i ++;
+			continue;
+		}
+		if (mob->lookup_const(t, name, &i32) && i32 >= 0) {
+			value = i32;
+		}
+		if (value <= 0)
+		{
+			ShowWarning("Wrong drop chanse %d for item %s\n", value, name);
```
MishimaHaruna: Umm, *chance* (or *rate*)

tools/mobdbconverter.py
```diff
@@ -0,0 +1,269 @@
...
+    itemDb = readItemDB(startPath + "db/item_db2.conf", itemDb)
+    convertFile(startPath + "db/re/mob_db.txt", startPath + "db/re/mob_db.conf", itemDb)
+    convertFile(startPath + "db/mob_db2.txt", startPath + "db/mob_db2.conf", itemDb)
+elif sys.argv[1] == "pre-re":
+    itemDb = readItemDB(startPath + "db/pre-re/item_db.conf", itemDb)
+    itemDb = readItemDB(startPath + "db/item_db2.conf", itemDb)
+    convertFile(startPath + "db/pre-re/mob_db.txt", startPath + "db/pre-re/mob_db.conf", itemDb)
+    convertFile(startPath + "db/mob_db2.txt", startPath + "db/mob_db2.conf", itemDb)
+else:
+    printHelp();
```
MishimaHaruna: it'd be nice if this worked like all other converters out there (parses stdin, outputs to stdout), so that it can be used with arbitrary files (or as a CGI script)

MishimaHaruna:
Two final notes I'd like to add for future works (or for integration here)

- Since we're converting the mob db, it's worth considering an integration with the mob_skill_db (would help decreasing redundancy and inconsistencies across files, and make it easier to edit a mob's behavior in just one file rather than having to search bits of it all over the place). If we're going to do it, we should considering doing it now, so that the users will have to convert their db only once (rather than writing a more complicated converter later to merge them)

- The item db has a nice `Inherit` optional behavior (for item_db2) entries. It'd be nice to add that here as well, so that it's easier to override just some fields from the mob_db2 without having to copypaste everything over (and risk for the copies to diverge).

src/map/mob.c
4144: I copy validation from old code. Actual validation happens below

src/map/mob.c
4144: If fix this code, will be no more range3 equal to 1 or 2. May be better remove this code?

src/map/mob.c
4144: I not touched variables related to sql mob db. Will see what will left after your commits

src/map/mob.c
4144: size check added in updated commits

src/map/mob.c
MishimaHaruna: Oh, I did not notice that, my wrong

src/map/mob.c
MishimaHaruna: Considering how this code has never been in use (it was misplaced since the beginning), removing it is the less impacting option I guess, especially considering how much of an undocumented hack it is. Oh, why didn't I start earlier to reject those large and messy, I wonder.
Whether it really is needed or not, only @Michieru can tell, since he's the original committer.

tools/mobdbconverter.py
```diff
@@ -0,0 +1,279 @@
...
+# -*- coding: utf8 -*-
+#
+# Copyright (C) 2015  Andrei Karas (4144)
+
+# This Script converts mob_db.txt to mob_db.conf format
+
+import os
+import re
+import sys
+from sets import Set
```
MishimaHaruna: The module `sets` looks like it was deprecated in python 2.6 and removed in python 3. Removing this line allows compatibility with all recent version of python (python 2.6 was released 7 years ago)

tools/mobdbconverter.py
```diff
@@ -0,0 +1,279 @@
...
+    print("\t{0}: \"{1}\"".format(name, value))
+
+def startGroup(name):
+    print("\t{0}: {{".format(name))
+
+def endGroup():
+    print("\t}")
+
+def printHelp():
+    print "MobDB converter from txt to conf format"
```
MishimaHaruna: Adding parentheses helps with python 3 compatibility

kyeme:
The Inherit s now aavailable?.

kyeme:
Tested - Inherit is not working..

PR #780, opened 2015-10-09
Show the last killed mob quest

vykimo:
This change display on the top of the head of the character the last killed mob in case of quest with multi-target.


src/map/clif.c
```diff
@@ -15594,12 +15594,13 @@ void clif_quest_delete(struct map_session_data *sd, int quest_id) {
 
 /// Notification of an update to the hunting mission counter (ZC_UPDATE_MISSION_HUNT).
 /// 02b5 <packet len>.W <mobs>.W { <quest id>.L <mob id>.L <total count>.W <current count>.W }*3
-void clif_quest_update_objective(struct map_session_data *sd, struct quest *qd)
+void clif_quest_update_objective(struct map_session_data *sd, struct quest *qd, int mob_id)
```
MishimaHaruna: You added an extra argument here, but you didn't add it to the places where this function is called or referenced

dastgir:
@vykimo  update?

vykimo:
I failed to add other files in the commit from github...

PR #781, opened 2015-10-09
Show the last killed mob quest

vykimo:


dastgir:
Merge all commits into 1 please ...

PR #790, opened 2015-10-12
Error message when talking to a pre-defined channel and you aren't on channel

ErMaqui:
This error message are created for #ally channel, when a user not on guild / alliance tries to talk to them.


conf/messages.conf
```diff
@@ -1416,6 +1416,7 @@
 1402: You're not in that channel, type '@join <#channel_name>'
 1435: You're now in the '#%s' channel for '%s'
 1436: You're already in the '%s' channel
+1500: You're not allowed to talk on this channel
```
MishimaHaruna: The max value you can use here is 1499 (see `MAX_MSG` in atcommand.h).

To avoid increasing the value, if only one message needs to be added, you can re-use some old free IDs (search for 'FREE' in messages.conf, there are some, i.e. 35, 1202, etc)

ErMaqui:
@MishimaHaruna can I do something more with this PR?

MishimaHaruna:
Looks good to merge now, thank you

MishimaHaruna:
Merged, thank you for your contribution!

PR #796, opened 2015-10-13
Conversion Tool, Warps and HP/SP Tables fixed

Vennren:
Made modifications to the conversion tool for the main database changes when coming from rA to Herc.
Renamed identical tables from rA to Herc format, dropped tables that only rA uses and converted tables that were used differently.


npc/warps/cities/prontera.txt
```diff
@@ -3,46 +3,48 @@
...
 //===== Additional Comments: =================================
 //= 1.3 prt19 warp fix [shadow]
 //= 1.4 Added warps for the sign quest [MasterOfMuppets]
 //= 1.5 Corrected NPC names to fall within proper restrictions. [L0ne_W0lf]
+//= 1.6 Corrected Warps for New Prontera. [Tranquility]
 //============================================================
 
 //= Prontera City ============================================
-prontera,107,215,0	warp	prt01	2,2,prt_in,240,139
+prontera,105,214,0	warp	prt01	2,2,prt_in,240,139
```
dastgir: not many uses new prontera, maybe make it optional in new file? what say @MishimaHaruna ?

MishimaHaruna: This is a bit controversial I guess.

Yes, I believe not many people use the new prontera map, so the old warps must stay around. It can be either another file (or at least, a commented out part in the current file, but that makes it harder to toggle)

Considering that we still ship the mapcache for the old version, the old warps should be enabled by default.

dastgir: Also we don't support 2015 client fully, so enabling new prontera warp by default is somewhat faulty.

Nihadm89:
It might be ugly but we could leave a commented out version of the new coordinates and leave a comment letting everyone know. Just an option.

Vennren:
The new prontera map works from 2012 clients onwards, made possible by Tokeiburu. Personally using original kRO with 20130807 client without any issues. Should be usuable by the masses.

sql-files/upgrades/rAthena-main-upgrade.sql
```diff
@@ -132,20 +131,15 @@ ALTER TABLE `storage` MODIFY `nameid` INT(11) NOT NULL DEFAULT '0',
...
-  `base_death` TINYINT( 4 ) UNSIGNED NOT NULL DEFAULT '100',
+  `base_exp` SMALLINT(6) UNSIGNED NOT NULL DEFAULT  '100',
+  `base_drop` SMALLINT(6) UNSIGNED NOT NULL DEFAULT '100',
+  `base_death` SMALLINT(6) UNSIGNED NOT NULL DEFAULT '100',
   PRIMARY KEY (`account_id`)
 ) ENGINE=MyISAM;
 
--- Saving bank_vault data from rAthena's login table
--- to our account_data table. There may be some not working cases.
-INSERT INTO `account_data` (`account_id`, `bank_vault`) SELECT `account_id`, `bank_vault` FROM `login` WHERE `bank_vault` > 0 ;
```
MishimaHaruna: It would probably be wise to import this from their acc_reg_num table now.

Smokexyz:
I'm going to split some of these commits into separate PRs and commits, for a lot of reasons. Then we can bury this PR in the grave.

HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/796)

PR #813, opened 2015-10-22
Fix crash in instance.c

4144:


src/map/instance.c
```diff
@@ -94,9 +94,13 @@ int instance_create(int owner_id, const char *name, enum instance_owner_type typ
 	}
 
 	if( type != IOT_NONE && *icptr ) {
-		ARR_FIND(0, *icptr, i, strcmp(instance->list[iptr[i]].name,name) == 0 );
```
dastgir: What about
		ARR_FIND(0, *icptr, i, iptr[i] != -1 && strcmp(instance->list[iptr[i]].name,name) == 0 );

PR #820, opened 2015-10-25
Fix a warning when loading instance duplicates (#812)

### LINKED ISSUE #812: Warning when creating Instance
```
[Warning]: npc_parseview: Use of numeric NPC view IDs is deprecated and may be removed in a future update. Please use NPC view constants instead. ID '111' specified in file 'INSTANCING', line '1'.
```

MishimaHaruna:
This changes the way duplicates are created when they're needed in an instance.

Previously, duplicates were created through an opportunely forged fake script line, parsed as if it was a normal duplicate NPC loaded from a script. Now, it's done without formatting and parsing a script line.

This fixes #812.

### DIFF
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -142,15 +142,11 @@ void mvptomb_create(struct mob_data *md, char *killer, time_t time)
 	if ( md->tomb_nid )
 		mob->mvptomb_destroy(md);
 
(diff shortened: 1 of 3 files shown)


MishimaHaruna:
Added four commits to address the plugin flexibility request

PR #827, opened 2015-10-27
Replaced getpartymember / getguildmember parameter type with constant.

emistry:
constant for getpartymember ( <party id>,<type> )
constant for getguildmember ( <guild id>,<type> )


emistry:
lol.. =='' how come all previous changes was commited.. ==''

db/const.txt
```diff
@@ -3431,3 +3431,14 @@ SEPTEMBER	9
...
+
+// getpartymember( <party id>,<type> );
+PARTYMEMBER_NAME	0
+PARTYMEMBER_CID	1
+PARTYMEMBER_AID	2
+
+// getguildmember( <guild id>,<type> );
+GUILDMEMBER_NAME	0
+GUILDMEMBER_CID	1
+GUILDMEMBER_AID	2
```
Smokexyz: Cant they both share member name , cid and aid constants since the types are the same?

dastgir: For these 2 commands, they can... Have a command name maybe ?

emistry: personally i would prefer
```
MEMBER_AID
MEMBER_CID
MEMBER_NAME
```
or something even shorter ... maybe =D

Jedzkie:
@emistry suggestion

PMEM_
GMEM_

Issue #834, opened 2015-10-28
min_npc_vendchat_distance setting shouldn't apply to invisible NPC sprite


emistry:
https://github.com/HerculesWS/Hercules/blob/master/conf/battle/player.conf#L149
```
// Minimum distance a vending/chat room must be from a NPC in order to be placed
// Default: 3 (0: disabled).
min_npc_vendchat_distance: 3
```
I think this option shouldn't apply to any "invisible" NPC.

This settings implemented to avoid players vending block the NPC visibility. I think it shouldn't apply to any NPC sprite that is invisible, 

__Example Sprites__ :
* -1
* HIDDEN_NPC (111)
* HIDDEN_WARP_NPC (139)

or at least it shouldn't apply to NPC with sprite of -1

MishimaHaruna:
Are there actual use cases for NPCs with view ID `-1`, but placed on a map (as opposed to floating NPCs)?

As for `HIDDEN_NPC`, while it doesn't have a sprite, it still shows a clickable bubble, and as such, I think the distance setting should apply to it (making it otherwise very hard to click, just like the case where a sprite is visible)

I'm not familiar with the in-game appearance of `HIDDEN_WARP_NPC`, so I can't say anything about that one.

emistry:
owh ya the `HIDDEN_NPC` does have the clickable bubble.
but `HIDDEN_WARP_NPC` doesn't have the clickable bubble.
the NPC with `-1` as sprite work exactly the same like `HIDDEN_WARP_NPC`.

I think it should apply to `HIDDEN_WARP_NPC` or `-1` then.

the `-1` previously vastly used in floating npc, but sometime users do use it to create an invisible sprite, which exactly the same with `HIDDEN_WARP_NPC`.

AnnieRuru:
@emistry , if you mean your vending script, why not just add a few lines in the source for it ?
just like my koe script also has a few line patch, completely custom stuff

I think we shouldn't do this kind of change
just because 1 person script doesn't work, we shouldn't change the whole root system

for your edit: HIDDEN_WARP_NPC is used in official script
changing the behavior might cause official script bug

EDIT: yeah, I fixed your script again
http://herc.ws/board/topic/4509-vendor-control-rewrite/

hemagx:
actually i remember in official anything that hidden are just fine to do something near them, need test tho, if i got some time later.

emistry:
[...]

PR #837, opened 2015-10-29
Add new config to allow player vend/chat room nearby hidden npc (#834)

AnnieRuru:
tested with
```
prontera,145,187,5	script	test1	-1,{}
prontera,155,187,5	script	test2	HIDDEN_NPC,{}
prontera,165,187,5	script	test3	HIDDEN_WARP_NPC,{}
```
apparently when using -1 with coordinates, the server doesn't throw error

### DIFF
--- a/src/map/npc.c
+++ b/src/map/npc.c
@@ -137,7 +137,7 @@ int npc_isnear_sub(struct block_list* bl, va_list args) {
 	if( nd->option & (OPTION_HIDE|OPTION_INVISIBLE) )
 		return 0;
 
-	if( battle_config.vendchat_near_hiddennpc && ( nd->class_ == -1 || nd->class_ == 139 ) )
+	if( battle_config.vendchat_near_hiddennpc && ( nd->class_ == FAKE_NPC || nd->class_ == HIDDEN_WARP_CLASS ) )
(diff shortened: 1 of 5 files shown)


src/map/npc.c
```diff
@@ -137,6 +137,9 @@ int npc_isnear_sub(struct block_list* bl, va_list args) {
 	if( nd->option & (OPTION_HIDE|OPTION_INVISIBLE) )
 		return 0;
 
+	if( battle_config.vendchat_near_hiddennpc && ( nd->class_ == -1 || nd->class_ == 139 ) )
```
MishimaHaruna: We have a constant for this, you could use it here: `HIDDEN_WARP_CLASS`

AnnieRuru: then how about -1 ?
```
enum actor_classes {
	WARP_CLASS = 45,
	HIDDEN_WARP_CLASS = 139,
	WARP_DEBUG_CLASS = 722,
	FLAG_CLASS = 722,
	INVISIBLE_CLASS = 32767,
};
```
I didn't see the FAKE_NPC available

MishimaHaruna:
Merged, thank you!

PR #839, opened 2015-10-30
Add *pcblockattack *pcblockskill *pcblockchat *pcimmune script commands

AnnieRuru:
reference
http://herc.ws/board/topic/11020-

tested with
```
prontera,154,185,5	script	can attack	1_F_MARIA,{
	pcblockattack getcharid(3), 0;
}
prontera,157,185,5	script	no attack	1_F_MARIA,{
	pcblockattack getcharid(3), 1;
}
prontera,154,181,5	script	can chat	1_F_MARIA,{
	pcblockchat getcharid(3), 0;
}
prontera,157,181,5	script	no chat	1_F_MARIA,{
	pcblockchat getcharid(3), 1;
}
prontera,154,177,5	script	vulnerable	1_F_MARIA,{
	pcimmune getcharid(3), 0;
}
prontera,157,177,5	script	immune	1_F_MARIA,{
	pcimmune getcharid(3), 1;
}
prontera,154,173,5	script	can skill	1_F_MARIA,{
	pcblockskill getcharid(3), 0;
}
prontera,157,173,5	script	no skill	1_F_MARIA,{
	pcblockskill getcharid(3), 1;
}
```
btw I feel *pcimmune name doesn't fit
any better names ?

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -5860,11 +5860,13 @@ Examples:
 ---------------------------------------
 
 *pcblockmove <id>,<option>;
-	
-Prevents the given ID from moving when the option != 0, and 0 enables the 
-ID to move again. The ID can either be the GID of a monster/NPC or account 
-ID of a character, and will run for the attached player if zero is 
-supplied.
+*pcblockattack <id>,<option>;
+*pcblockskill <id>,<option>;
+*pcimmune <id>,<option>;
+
+Prevents the given player to do following actions when the option > 0,
+and 0 enables it again. The ID only works for player, and will run for
+the attached player if zero is supplied.
 
 Examples:
 
@@ -5874,6 +5876,8 @@ Examples:
 // Enables the current char to move again.
 	pcblockmove getcharid(3),0;
 
+// Makes the current player invulnerable, same as @monsterignore
+	pcimmune getcharid(3),1;
 
 ---------------------------------------
 //=====================================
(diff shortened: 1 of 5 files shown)


AnnieRuru:
AHHHHH I forgot to add documentations for script command.txt

MishimaHaruna:
I agree with @Ancyker that it'd look better if the arguments were swapped and the ID was optional. The problem, though, is that it'd be a dangerous change in pcblockmove (a backwards incompatible change that wouldn't even cause a warning). If we were to do that, we'd need to at least rename pcblockmove to something else to avoid the change being silent.

Ancyker:
@MishimaHaruna Why not force true/false (1/0) and error on anything else. Trying to pass an rid then would fail and throw an error.

Ancyker:
For instances where ID=0, you can see if RID is specified and is 0 or 1, detecting they likely did not intend what's happening. In the new system, RID 0 and 1 don't exist, so it should error anyway.

So someone doing this post-update:

    pcblockmove 0,0;
    pcblockmove 0,1;

Would throw an error, because the 2nd param has data and the data is zero or one, and those don't exist.

Ancyker:
Another option is force the true/false constant, then you know for sure they intended it to be that way.

AnnieRuru:
@Ancyker that's good idea, I can do that
the account ID always start from 200000 anyways, its already defined in the source
so just 1 or 0 wouldn't hurt

AnnieRuru:
@Ancyker, errr ... how to document this ?
`*pcblockmove <option|ID>{, <option };`
doesn't look write

MishimaHaruna:
@Ancyker While you can check at runtime the value of an argument to be 1/0, you can't check it at parse time. For an API change of this kind, a runtime check doesn't cut it. It must error out at server startup.

AnnieRuru:
closing this one
ima going to write just 1 new *pcblock with types
*pcblock <type>, <value>{, <iD> };
eg: pcblock PCBLOCK_MOVE, true;
pcblock PCBLOCK_ATTACK, false, getcharid(3);

PR #842, opened 2015-10-31
Add new *pcblock script command

AnnieRuru:
This is NOT what I had in mind when I did the previous pull request
but a lot was going on, on the forums, and somehow it turn out this way
... now this has become quite complex
1. I seriously think PCBLOCK_ is a better constant, but the community want it as BLOCK
   because I fear maybe in the future there will be a real BLOCK constants coming from aegis ? I dunno
   that's why it was PCBLOCK_ in the source, BLOCK as scripts
   need Haru opinion on this
2. I don't really like to remove others work, like removing the name [Valaris] in the source code <.<
   but members like the bit-wise method, so meh ...
   so I removed those 2 sd->state.
3. I'm wondering the enum block_action should belongs to script.h or pc.h...
4. we have suggested so many types
   here we go ...

BLOCK_MOVE => its a must
- keep the pcblockmove for backward compatibility, but add this in for a better management control

BLOCK_ATTACK => its a must
BLOCK_SKILL => its a must
- I don't like using OPTION_XMAS or SC_XMAS to prevent the player to attack/skill, need a proper script command

BLOCK_CHAT => its a must
- I don't like using SC_BERSERK to prevent player to chat, need a proper script command

BLOCK_IMMUNE => its a must
- I don't like using atcommand "@monsterignore" in scripts, it will create unnecessary log

BLOCK_USEITEM => look like a must
- although I have a noitem mapflag to do the same, but preventing player to attack, then skill, its logical to prevent item next

BLOCK_STORAGE => redundant
- maybe should code a nostorage mapflag ?

BLOCK_SENDMAIL => no...
- the idea is that, some private server has a special map where players can try out godly item in a closed, nowarp+nowarpto map
  however there is an exploit where players can send item away into another account using friend list-> send mail
  I will make a pull request to prevent this exploit by adding a battle config

BLOCK_EQUIPMENT => no...
- doesn't make sense, should use map_zone_db.conf or my noitem mapflag

BLOCK_TELEPORT => no..
- already have noteleport mapflag

BLOCK_SITSTAND => looks ok
- we have script command *sit, *stand and *issit(),
  means we can actually script it to make the player sit, then pcblocksit to prevent them standing up

BLOCK_VENDING => no ..
- already has novending mapflag and cell_novending

BLOCK_TRADING => no...
- I think *getitembound already make this useless.
- Since the special created item for event is already bound, there is no way for them to trade anymore
- also already has notrade mapflag

BLOCK_PICKITEM => no
- already has nomobloot and nomvploot mapflag

BLOCK_COMMAND => 50% yes 50% no
- nocommand mapflag exist, I know, but there are times maybe they need to talk while having`@command` disabled

BLOCK_GSTORAGE
- I seriously think this should belongs to nostorage mapflag
  but we don't have one

last but not least, tested with this script

```
prontera,155,184,5  script  sdfhdskfj   1_F_MARIA,{
Ontest:
    for ( .@i = 0; .@i < 8; .@i++ )
        .@menu$ = .@menu$ + .name$[.@i] + .flag$[ !!( getpcblock() & (1 << .@i) ) ] +":";
    .@s = select( .@menu$ ) -1;
    pcblock 1 << .@s, !( getpcblock() & (1 << .@s) );
    close;
OnInit:
    bindatcmd "@pcblock", strnpcinfo(0)+"::Ontest";
    setarray .name$, "BLOCK_MOVE", "BLOCK_ATTACK", "BLOCK_SKILL", "BLOCK_USEITEM",
        "BLOCK_CHAT", "BLOCK_IMMUNE", "BLOCK_SITSTAND", "BLOCK_COMMAND";
    setarray .flag$, " [^ff0000OFF^000000]", "[^0000ffON^000000]";
    end;
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/842)

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -6184,12 +6184,15 @@ Examples:
 
 ---------------------------------------
 
-*pcblockmove(<id>, <option>)
+*pcblockmove(<account id>, <option>)
 
-Prevents the given ID from moving when the optionis true , and false
-enables the ID to move again. The ID can either be the GID of a
-monster/NPC or account ID of a character, and will run for the attached
-player if zero is supplied.
+    @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+    @ /!\ This command is deprecated @
+    @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+
+Prevents the player from moving when the option != 0, and 0 enables the 
+player to move again. The player has to be the account ID of a character,
+and will run for the attached player if zero is supplied.
 
 Examples:
 
@@ -6199,6 +6202,53 @@ Examples:
 // Enables the current char to move again.
 	pcblockmove(getcharid(CHAR_ID_ACCOUNT), false);
 
+---------------------------------------
+
+*setpcblock(<type>,<option>)
+*checkpcblock()
+
+Prevents the player from doing the following action.
+
+For setpcblock, when the <option> is true(1) will block them, and false(0)
+will allow those actions again.
+
+The checkpcblock command returned value is a bit mask of the currently
+enabled block flags (or PCBLOCK_NONE when none is set).
+
+The <type> listed are a bit mask of the following:
+	PCBLOCK_NONE (only used by checkpcblock)
+	PCBLOCK_MOVE
+	PCBLOCK_ATTACK
+	PCBLOCK_SKILL
+	PCBLOCK_USEITEM
+	PCBLOCK_CHAT
+	PCBLOCK_IMMUNE
+	PCBLOCK_SITSTAND
+	PCBLOCK_COMMANDS
+
+Examples:
+
+// Make the current attached player invulnerable, same as @monsterignore
+	setpcblock(PCBLOCK_IMMUNE, true);
+
+// Prevents the current char from attacking or using skills
+	setpcblock(PCBLOCK_ATTACK|PCBLOCK_SKILL, true);
+
+// Re-enables attack, skills and item use
+	setpcblock(PCBLOCK_ATTACK|PCBLOCK_SKILL|PCBLOCK_ITEM, false);
+
+// checkpcblock related checks
+	if ((checkpcblock() & PCBLOCK_IMMUNE) != 0)
+		mes "You are invulnerable!";
+
+	if ((checkpcblock() & (PCBLOCK_MOVE|PCBLOCK_SITSTAND)) == (PCBLOCK_MOVE|PCBLOCK_SITSTAND))
+		mes "You can't walk or sit";
+
+	if ((checkpcblock() & (PCBLOCK_ATTACK|PCBLOCK_SKILL)) == PCBLOCK_NONE)
+		mes "You can attack and use skills";
+
+	if ((checkpcblock() & PCBLOCK_CHAT) == PCBLOCK_NONE)
+		mes "You can't chat";
 
 ---------------------------------------
 //=====================================
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -7522,11 +7522,11 @@ ACMD(sizeguild)
  *------------------------------------------*/
 ACMD(monsterignore)
 {
-	if (!sd->state.monster_ignore) {
-		sd->state.monster_ignore = 1;
+	if (!sd->block_action.immune) {
+		sd->block_action.immune = 1;
 		clif->message(sd->fd, msg_fd(fd,1305)); // You are now immune to attacks.
 	} else {
-		sd->state.monster_ignore = 0;
+		sd->block_action.immune = 0;
 		clif->message(sd->fd, msg_fd(fd,1306)); // Returned to normal state.
 	}
 
@@ -10090,6 +10090,8 @@ bool atcommand_exec(const int fd, struct map_session_data *sd, const char *messa
 			return false;
 		}
 	}
+	if (sd->block_action.commands) // *pcblock script command
+		return false;
 
 	if (*message == atcommand->char_symbol)
 		is_atcommand = false;
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -10488,6 +10488,9 @@ void clif_parse_ActionRequest_sub(struct map_session_data *sd, int action_type, 
 				return;
 			}
 
+			if (sd->block_action.sitstand) // *pcblock script command
+				break;
+
 			if (sd->ud.skilltimer != INVALID_TIMER || (sd->sc.opt1 && sd->sc.opt1 != OPT1_BURNING ))
 				break;
 
@@ -10514,6 +10517,9 @@ void clif_parse_ActionRequest_sub(struct map_session_data *sd, int action_type, 
 				clif->standing(&sd->bl);
 				return;
 			}
+
+			if (sd->block_action.sitstand) // *pcblock script command
+				break;
 
 			pc->update_idle_time(sd, BCIDLE_SIT);
 
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -18062,9 +18062,88 @@ BUILDIN(pcblockmove) {
 	else
 		sd = script->rid2sd(st);
 
-	if (sd != NULL)
-		sd->state.blockedmove = flag > 0;
+	if (!sd)
+		return true;
 
+	if (flag)
+		sd->block_action.move = 1;
+	else
+		sd->block_action.move = 0;
+
+	return true;
+}
+
+BUILDIN(setpcblock)
+{
+	struct map_session_data *sd = script->rid2sd(st);
+	enum pcblock_action_flag type = script_getnum(st, 2);
+	int state = (script_getnum(st, 3) > 0) ? 1 : 0;
+
+	if (sd == NULL)
+		return true;
+
+	if ((type & PCBLOCK_MOVE) != 0)
+		sd->block_action.move = state;
+
+	if ((type & PCBLOCK_ATTACK) != 0)
+		sd->block_action.attack = state;
+
+	if ((type & PCBLOCK_SKILL) != 0)
+		sd->block_action.skill = state;
+
+	if ((type & PCBLOCK_USEITEM) != 0)
+		sd->block_action.useitem = state;
+
+	if ((type & PCBLOCK_CHAT) != 0)
+		sd->block_action.chat = state;
+
+	if ((type & PCBLOCK_IMMUNE) != 0)
+		sd->block_action.immune = state;
+
+	if ((type & PCBLOCK_SITSTAND) != 0) 
+		sd->block_action.sitstand = state;
+
+	if ((type & PCBLOCK_COMMANDS) != 0)
+		sd->block_action.commands = state;
+
+	return true;
+}
+
+BUILDIN(checkpcblock)
+{
+	struct map_session_data *sd = script->rid2sd(st);
+	int retval = PCBLOCK_NONE;
+
+	if (sd == NULL) {
+		script_pushint(st, PCBLOCK_NONE);
+		return true;
+	}
+
+	if (sd->block_action.move != 0)
+		retval |= PCBLOCK_MOVE;
+
+	if (sd->block_action.attack != 0)
+		retval |= PCBLOCK_ATTACK;
+
+	if (sd->block_action.skill != 0)
+		retval |= PCBLOCK_SKILL;
+
+	if (sd->block_action.useitem != 0)
+		retval |= PCBLOCK_USEITEM;
+
+	if (sd->block_action.chat != 0)
+		retval |= PCBLOCK_CHAT;
+
+	if (sd->block_action.immune != 0)
+		retval |= PCBLOCK_IMMUNE;
+
+	if (sd->block_action.sitstand != 0)
+		retval |= PCBLOCK_SITSTAND;
+
+	if (sd->block_action.commands != 0)
+		retval |= PCBLOCK_COMMANDS;
+
+	script_pushint(st, retval);
 	return true;
 }
 
@@ -24520,7 +24599,9 @@ void script_parse_builtin(void) {
 		BUILDIN_DEF(rid2name,"i"),
 		BUILDIN_DEF(pcfollow,"ii"),
 		BUILDIN_DEF(pcstopfollow,"i"),
-		BUILDIN_DEF(pcblockmove,"ii"),
+		BUILDIN_DEF_DEPRECATED(pcblockmove,"ii"), // Deprecated 2018-05-04
+		BUILDIN_DEF(setpcblock, "ii"),
+		BUILDIN_DEF(checkpcblock, ""),
 		// <--- [zBuffer] List of player cont commands
 		// [zBuffer] List of mob control commands --->
 		BUILDIN_DEF(getunittype,"i"),
@@ -25093,6 +25174,17 @@ void script_hardcoded_constants(void)
 	script->set_constant("MST_AROUND3", MST_AROUND3, false, false);
 	script->set_constant("MST_AROUND4", MST_AROUND4, false, false);
 	script->set_constant("MST_AROUND", MST_AROUND , false, false);
+ 
+	script->constdb_comment("pc block constants, use with *setpcblock* and *checkpcblock*");
+	script->set_constant("PCBLOCK_NONE",     PCBLOCK_NONE,     false, false);
+	script->set_constant("PCBLOCK_MOVE",     PCBLOCK_MOVE,     false, false);
+	script->set_constant("PCBLOCK_ATTACK",   PCBLOCK_ATTACK,   false, false);
+	script->set_constant("PCBLOCK_SKILL",    PCBLOCK_SKILL,    false, false);
+	script->set_constant("PCBLOCK_USEITEM",  PCBLOCK_USEITEM,  false, false);
+	script->set_constant("PCBLOCK_CHAT",     PCBLOCK_CHAT,     false, false);
+	script->set_constant("PCBLOCK_IMMUNE",   PCBLOCK_IMMUNE,   false, false);
+	script->set_constant("PCBLOCK_SITSTAND", PCBLOCK_SITSTAND, false, false);
+	script->set_constant("PCBLOCK_COMMANDS", PCBLOCK_COMMANDS, false, false);
 
 	script->constdb_comment("Renewal");
 #ifdef RENEWAL
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -6649,7 +6649,7 @@ int battle_check_target( struct block_list *src, struct block_list *target,int f
 			if (t_bl == s_bl)
 				break;
 
-			if( sd->state.monster_ignore && flag&BCT_ENEMY )
+			if (sd->block_action.immune && flag&BCT_ENEMY)
 				return 0; // Global immunity only to Attacks
 			if (sd->status.karma && s_bl->type == BL_PC && BL_UCCAST(BL_PC, s_bl)->status.karma)
 				state |= BCT_ENEMY; // Characters with bad karma may fight amongst them
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -5162,6 +5162,9 @@ int pc_useitem(struct map_session_data *sd,int n) {
 	if( sd->status.inventory[n].nameid <= 0 || sd->status.inventory[n].amount <= 0 )
 		return 0;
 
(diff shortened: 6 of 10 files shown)


dastgir:
What if we have like
struct {
    unsigned int move: 1;
    unsigned int mail : 1;
    ....
} block;

Would change those 4 bytes'into 1 byte, and would save more on 64bit system

AnnieRuru:
hmm ... not sure about saving memory space
but I was thinking that sometime in the future,
if somebody wants to add more BLOCK_ it will be very easy with bits

if haru is for this change, then I change it

Ancyker:
Worth noting map zone db cannot be reloaded or otherwise changed dynamically.

AnnieRuru:
@Ancyker 
there are a few ways of using skills to move the player
like champion snap, thief back sliding, teakwon jump ... etc
BLOCK_SKILL will prevent those already
so it will be BLOCK_MOVE|BLOCK_SKILL|BLOCK_COMMAND

and the idea that priest bring potions
already has notrade mapflag too

MishimaHaruna:
I'd also prefer to have those as a bitmask struct, as Dastgir suggests. I'm not worried about a long sequence of similar checks in the script command (and in fact it's a good idea, since it decouples the script-side bit positions from the source-side ones, allowing greater flexibility whenever we need to change anything in source, without affecting the script API). Whether they should be inside `sd->state` or in a separate struct is up to you, I don't have any strong opinion vouching for either.


Now, to go into detail about each `PCBLOCK*` type, I'll quote you bit by bit

> I seriously think PCBLOCK_ is a better constant, but the community want it as BLOCK because I fear maybe in the future there will be a real BLOCK constants coming from aegis ? I dunno that's why it was PCBLOCK_ in the source, BLOCK as scripts need Haru opinion on this

I also prefer  `PCBLOCK_*`, because it's more specific. It'd make sense for it to be `BLOCK_*` if it was something more generic, not affecting only PCs. Having a more specific name makes it less likely to conflict with other constants later, indeed (regardless of whether they're from Aegis or anywhere else).

> I don't really like to remove others work, like removing the name [Valaris] in the source code <.< but members like the bit-wise method, so meh ... so I removed those 2 sd->state.

Honestly, signing every bit of code someone writes is kind of stupid, so it's perfectly fine to remove those names whenever the related code is edited (it's not fine to just remove the names when the code isn't otherwise edited). The only things that should be signed are perhaps TODO/FIXME comments, where it makes sense to know who wrote it, in case others have questions or want to handle the issue, or similar matters. So, don't worry about removing those signed comments, they were useless to begin with.

> I'm wondering the enum block_action should belongs to script.h or pc.h...

Yes, once the variables are grouped into a bitmask struct, those constants will be only used in script-related code, and can be moved to script.h. The enum name could then be changed to `pcblock_action`, since they're meant to be used by the `pcblock*` commands.

> BLOCK_MOVE => its a must
> keep the pcblockmove for backward compatibility, but add this in for a better management control

Sounds good. I'll take care of marking the old command as deprecated, and eventually remove it. It's fine to have backwards compatibility for a while, but it's a maintenance burden to keep it indefinitely.

> BLOCK_ATTACK => its a must

Okay

> BLOCK_SKILL => its a must
> I don't like using OPTION_XMAS or SC_XMAS to prevent the player to attack/skill, need a proper script command

I agree.

> BLOCK_CHAT => its a must
> I don't like using SC_BERSERK to prevent player to chat, need a proper script command

Yep, agreed.

> BLOCK_IMMUNE => its a must
> I don't like using atcommand "@monsterignore" in scripts, it will create unnecessary log

Yes, calls to `atcommand` from scripts should be well avoided. It's only provided for flexibility in edge cases.

> BLOCK_USEITEM => look like a must
> although I have a noitem mapflag to do the same, but preventing player to attack, then skill, its logical to prevent item next

Okay. Be sure to make it very clear in documentation with regards to what happens when an item is normally auto-used (i.e. skills that consume items). If those still work, it's better to specify it in the documentation, to avoid surprises.

> BLOCK_STORAGE => redundant
> maybe should code a nostorage mapflag ?

Redundant, yep. Having it as a mapflag might be also seen as redundant, since the standard, non-GM means to open storage are through scripts (and as such, you can block it without needing a mapflag). It can be discussed elsewhere though.

> BLOCK_SENDMAIL => no...
> the idea is that, some private server has a special map where players can try out godly item in a closed, nowarp+nowarpto map however there is an exploit where players can send item away into another account using friend list-> send mail I will make a pull request to prevent this exploit by adding a battle config

Okay. I don't see a reason to block this at a player-by-player level either.

> BLOCK_EQUIPMENT => no...
> doesn't make sense, should use map_zone_db.conf or my noitem mapflag

Yep. I don't currently see a need for this.

> BLOCK_TELEPORT => no..
> already have noteleport mapflag

Yep.

> BLOCK_SITSTAND => looks ok
> we have script command *sit, *stand and *issit(), means we can actually script it to make the player sit, then pcblocksit to prevent them standing up

I don't see much of a need for this, but it won't hurt. I can see how it can be used.

> BLOCK_VENDING => no ..
> already has novending mapflag and cell_novending

Yep, redundant.

> BLOCK_TRADING => no...
> I think *getitembound already make this useless.
> Since the special created item for event is already bound, there is no way for them to trade anymore also already has notrade mapflag

Notrade mapflag seems enough, unless there are other reasons this was asked.

> BLOCK_PICKITEM => no
> already has nomobloot and nomvploot mapflag

Rather than pickitem, I'd see more use in a dropitem counterpart, completing or integrating the notrade flag.

> BLOCK_COMMAND => 50% yes 50% no
> nocommand mapflag exist, I know, but there are times maybe they need to talk while having@command disabled

I think this is reasonable. The way I'd see it implemented is separating BLOCK_CHAT and BLOCK_COMMAND so that the first only blocks chat (keeping commands enabled) and the second only blocks commands, keeping chat enabled.

> BLOCK_GSTORAGE
> I seriously think this should belongs to nostorage mapflag but we don't have one

Yep. Definitely not at a player level in any case.

MishimaHaruna:
As for the map zone db reload capabilities, that's a separate matter, and it shouldn't affect decisions taken for this issue. I'd rather have a reloadable map zone db, than a workaround.

doc/script_commands.txt
```diff
@@ -5874,6 +5873,47 @@ Examples:
...
+'pcblock' command. Additionally, you may insert account ID within the bracket
+to check on another player.
+
+Examples:
+
+	if (getpcblock() & BLOCK_IMMUNE)
+		mes "You are invulnerable!";
+
+	if (getpcblock() & (BLOCK_ATTACK|BLOCK_SKILL))
+		mes "You are unable to attack and use any skills.";
```
MishimaHaruna: Nitpick (because otherwise potentially misleading for a reader): this isn't accurate, the "and" should be an "or", since this condition is true when the player is unable to attack OR use any skills.

src/map/script.c
```diff
@@ -16294,9 +16294,56 @@ BUILDIN(pcblockmove) {
...
+	if(!sd)
+		return true;
+
+	if(!type) {
+		ShowError("buildin_pcblock: Invalid type 0.\n");
+		return false;
+	}
+
+	if(flag)
+		sd->block_action |= type;
```
MishimaHaruna: Unsafe. You don't know what value will be passed here, it can be an invalid value. (But this is irrelevant after `block_action` is changed to a bit field as suggested.)

src/map/atcommand.c
```diff
@@ -9813,6 +9813,8 @@ bool atcommand_exec(const int fd, struct map_session_data *sd, const char *messa
 			clif->message(fd, msg_fd(fd,143));
 			return false;
 		}
+		if ( sd->block_action & PCBLOCK_COMMAND ) // *pcblock script command
```
MishimaHaruna: In cases like this that are a response to a direct action from the player, I think we should show an informative message explaining why they can't do it (similarly to what happens with the nocommand mapflag).

PCBLOCK_SITSTAND and possibly some others, would be nice if they had similar messages.

src/map/clif.c
```diff
@@ -9963,6 +9963,9 @@ void clif_parse_ActionRequest_sub(struct map_session_data *sd, int action_type,
 				return;
 			}
 
+			if( sd->block_action & PCBLOCK_SITSTAND ) // *pcblock script command
```
MishimaHaruna: Just a nitpick, but when you edit this to use a bitfield, could you please make your whitespace consistent with newly written code? We're trying to stick to a spacing conform to `if (condition) {` in all new code.

AnnieRuru:
>Sounds good. I'll take care of marking the old command as deprecated, and eventually remove it. It's fine to have backwards compatibility for a while, but it's a maintenance burden to keep it indefinitely.
I have been using this command since eathena period
google search 'annieruru pcblockmove' will have my old gvg/battleground scripts
so if you remove them, I have to update all my previous scripts ?

>Okay. Be sure to make it very clear in documentation with regards to what happens when an item is normally auto-used (i.e. skills that consume items). If those still work, it's better to specify it in the documentation, to avoid surprises.
you are right
when BLOCK_USEITEM = false, BLOCK_SKILL = true
I can use potions, but cannot use fire bolt scroll

>Redundant, yep. Having it as a mapflag might be also seen as redundant, since the standard, non-GM means to open storage are through scripts (and as such, you can block it without needing a mapflag). It can be discussed elsewhere though.
no, what I mean is, this only happens on high rate server
on high rate server, their focus is more on pvp aspect, not on farming items
so some of them allows player to use `@storage` atcommand,
so their players don't have to go back to towns to restock stuffs or ...
when their inventory is full, just `@storage` and store their junks

the problem is, although they allow them to use `@storage` when farming or even on event maps
but they dislike players to spam white potions when pvp'ing

that's why my noitem mapflag is getting popular
because hercules map_zone_db.conf only block single item, but cannot block an item type

nvm about this, should discuss else where, not related to this patch whatsoever

>Rather than pickitem, I'd see more use in a dropitem counterpart, completing or integrating the notrade flag.
oh yeah BLOCK_DROPITEM
just the same ... already have nodrop mapflag
already has nodrop, notrade, nomvploot, nomobloot mapflags ...

>I think this is reasonable. The way I'd see it implemented is separating BLOCK_CHAT and BLOCK_COMMAND so that the first only blocks chat (keeping commands enabled) and the second only blocks commands, keeping chat enabled.
the problem is, when BLOCK_CHAT enabled, BLOCK_COMMAND is automatically enabled
that's why many server use SC_BERSERK to block players from using atcommand
and I don't see a need to let players to use command while chatting disable
... unless IF you really want it, I have to code a separate function,
it wont be inside `pc_can_talk` function anymore

>Just a nitpick, but when you edit this to use a bitfield, could you please make your whitespace consistent with newly written code? We're trying to stick to a spacing conform to if (condition) { in all new code.
which type ?
`if ( sd->block_action & PCBLOCK_COMMAND )`
`if( sd->block_action & PCBLOCK_COMMAND )`
`if (sd->block_action & PCBLOCK_COMMAND)`

doc/script_commands.txt
AnnieRuru: Ooops !

AnnieRuru:
>In cases like this that are a response to a direct action from the player, I think we should show an informative message explaining why they can't do it (similarly to what happens with the nocommand mapflag).
PCBLOCK_SITSTAND and possibly some others, would be nice if they had similar messages.

BLOCK_MOVE
BLOCK_ATTACK
BLOCK_SKILL => must not show the message.
- the player already knew how the event works, so they understood why they can't move and so on
- especially for BLOCK_ATTACK, if the player has `/nc` on, the chat log will get spam

BLOCK_USEITEM => 20%
- in related to BLOCK_ATTACK|BLOCK_SKILL, its the same for this one
but there are some cases when the event only block from using items ?
at the moment, I think there's no need to display message ... yet ?

BLOCK_CHAT => its a must
- this one must add

BLOCK_IMMUNE => no...
- the player already know they are invulnerable, so there's no need a message for this

BLOCK_SITSTAND => how to do ?
- this is meant for certain types of events when the gm wants to block players from standing up
yes I mean, I saw some private server GM host an event ask the player to line up and count them
its quite fun to make a script to force these players to line up  ... take photo and so on
so when display a message for this, they should know that this is a temporary
`You are prevented from standing up at the moment`
`You have been prevented from standing up`
`You can't stand up right now`
`You can't stand up`
... which message should be display ? I guess should make a poll LOL

BLOCK_COMMAND => ok
- no problem for this one, very simple msg

==============================================================

BUT HERE COMES THE PROBLEM

I noticed that most of message display are using `clif->message` commands
which makes the player shows a message in the log

but we recently have a new `clif_ShowScript`
which it doesn't shows a message in the log

if I switched to `clif_ShowScript`, then every command above can be display a message,
...with the exception of BLOCK_IMMUNE

so ? use `clif->message` or `clif_ShowScript` ?

MishimaHaruna:
> I have been using this command since eathena period google search 'annieruru pcblockmove' will have my old gvg/battleground scripts so if you remove them, I have to update all my previous scripts ?

I understand it's annoying (I have a very large private scripts library myself, so I know the pain), but yes, eventually. We're not going to keep duplicate commands, or it'll grow into a dinosaur we'll be unable to maintain (if it already isn't).
By marking the command as deprecated, the scripts will still load, and throw a warning when loaded, so that it's easy to spot where to edit, before the command is eventually removed.
Whenever a command is marked as deprecated, I usually do a folder-wide search in both the official and the private scripts (I keep all of my scripts in a private git repository), and edit where needed.

> no, what I mean is, this only happens on high rate server on high rate server, their focus is more on pvp aspect, not on farming items so some of them allows player to use @storage atcommand, so their players don't have to go back to towns to restock stuffs or ... when their inventory is full, just @storage and store their junks
> 
> the problem is, although they allow them to use @storage when farming or even on event maps but they dislike players to spam white potions when pvp'ing
> 
> that's why my noitem mapflag is getting popular
> because hercules map_zone_db.conf only block single item, but cannot block an item type
> 
> nvm about this, should discuss else where, not related to this patch whatsoever

I see, makes sense. Thanks for the clarification. Yep, let's discuss this elsewhere

> oh yeah BLOCK_DROPITEM
> just the same ... already have nodrop mapflag
> already has nodrop, notrade, nomvploot, nomobloot mapflags ...
Yup, I agree

> the problem is, when BLOCK_CHAT enabled, BLOCK_COMMAND is automatically enabled
> that's why many server use SC_BERSERK to block players from using atcommand and I don't see a need to let players to use command while chatting disable
> ... unless IF you really want it, I have to code a separate function,
> it wont be inside pc_can_talk function anymore

I think it's more flexible if they're completely separated. Are you sure that `SC_BERSERK` blocks commands right now? I see that all the calls to `pc->can_talk()` (where the check for `SC_BERSERK` occurs) are *after* the calls to `atcommand->exec()`, and as such, commands shouldn't be affected (unless I'm missing something?)

In both `clif_parse_WisMessage` and `clif_parse_GlobalMessage` there's a block similar to:

```
	if ( atcommand->exec(fd, sd, message, true) )
		return;

	// Statuses that prevent the player from whispering
	if( !pc->can_talk(sd) )
		return;
```

> which type ?
```
if ( sd->block_action & PCBLOCK_COMMAND )
if( sd->block_action & PCBLOCK_COMMAND )
if (sd->block_action & PCBLOCK_COMMAND)
```

The third one please. The rationale behind it, is that we're planning (in the long term) to convert the codebase to the [Linux Kernel Coding Style](https://www.kernel.org/doc/Documentation/CodingStyle) (with the exception that we aren't enforcing the 80-character line length -- let's face it, screens are wider nowadays -- although a line length of no more than 120 characters would be a good idea)

> EDIT: and how about this
```
if( sd->block_action&PCBLOCK_IMMUNE && flag&BCT_ENEMY )
```

For consistency with the coding style I linked, this would be:
```
if (sd->block_actions & PCBLOCK_IMMUNE && flag & BCT_ENEMY)
```
(although I'm often guilty of removing the spaces around the binary `&` and `|` operators)

> `BLOCK_MOVE`
> `BLOCK_ATTACK`
> `BLOCK_SKILL` => must not show the message.
> the player already knew how the event works, so they understood why they can't move and so on especially for `BLOCK_ATTACK`, if the player has /nc on, the chat log will get spam

I agree about move and attack (those would be annoyingly spammy), but why also skills? (if there's a reason, I'm fine with your decision - I'm just curious to know the reason)

> BLOCK_USEITEM => 20%
> in related to `BLOCK_ATTACK|BLOCK_SKILL`, its the same for this one
> but there are some cases when the event only block from using items ?
> at the moment, I think there's no need to display message ... yet ?

I think it's the same as the skills. Unlike the much spammier move and attack actions that can be repeatedly triggered (/nc or autofollow), items and skills are the effect of a direct command by the player.

> BLOCK_CHAT => its a must
> this one must add

Yep

> BLOCK_IMMUNE => no...
the player already know they are invulnerable, so there's no need a message for this

Agreed. That'd be really annoying.

> BLOCK_SITSTAND => how to do ?
> this is meant for certain types of events when the gm wants to block players from standing up yes I mean, I saw some private server GM host an event ask the player to line up and count them its quite fun to make a script to force these players to line up ... take photo and so on so when display a message for this, they should know that this is a temporary
* You are prevented from standing up at the moment
* You have been prevented from standing up
* You can't stand up right now
* You can't stand up
> ... which message should be display ? I guess should make a poll LOL

Umm, entirely up to you (but the *are prevented* ones are awkward to read). My personal vote goes to the *You can't stand up right now* / *You can't sit right now* variant.

> BLOCK_COMMAND => ok
> no problem for this one, very simple msg

Okies

> BUT HERE COMES THE PROBLEM
> I noticed that most of message display are using clif->message commands which makes the player shows a message in their chat log
> but we recently have a new clif_ShowScript which it doesn't shows a message in their chat log
> if I switched to clif->ShowScript, then every command above can be display a message, ...with the exception of BLOCK_IMMUNE
> so ? use clif->message or clif->ShowScript ?

Hmm, good question. From a player's perspective (but perhaps my opinion isn't indicative enough, I'm an extremely casual player), I'd say that showing a message above the character's head is almost the same as not showing it (I mean, I'd need to be looking there, where there's only useless information, like the name of the skill that I just cast, which I already know). A (red?) message in the chat window would be much more meaningful, with the purpose to catch my attention and tell me that my command/action was rejected by the server for some reason.

kyeme:
Hi @annieruru, can you add "BLOCK_INVITE" to disable add as friend on a player? (good for gvg script that have spectator mode) thank you!

AnnieRuru:
> (good for gvg script that have spectator mode)

I am quite sure in gvg script, should have `mf_guildlock` mapflag
so should have block guild invitation

BLOCK_PARTYINVITE - no need, already has `mf_partylock` mapflag
BLOCK_GUILDINVITE - no need, already has `mf_guildlock` mapflag

but you mean friend invite ... yes, should add BLOCK_FRIENDINVITE

kyeme:
Yes I mean BLOCK_FRIENDINVITE, because its annoying when someone suddenly add you as a friend when you're in GVG. Anyway, I think this isnt too important because they can use @noask so, you can forget I requested for this one. ^^

MishimaHaruna:
Is this complete? If so, I'll take ownership and rebase/merge it

dastgir:
Seems to be coded... but would need test

Ridley8819:
Bumping this @MishimaHaruna @hemagx

Smokexyz:
Why does everything in hercules take a year to process? 🤣

Ridley8819:
It would need a block all tho, this is the most important one

Asheraf:
This has been already updated by haru on HerculesWS/Hercules@121a0778ab94a4a7e0f3c5dc0f4ce208a8fe8086 in request_06 branch
maybe this could be closed.

src/map/script.c
```diff
@@ -18062,9 +18062,88 @@ BUILDIN(pcblockmove) {
 	else
 		sd = script->rid2sd(st);
 
-	if (sd != NULL)
-		sd->state.blockedmove = flag > 0;
+	if (!sd)
+		return true;
```
4144: probably if pc not assigned better return false?

src/map/script.c
```diff
@@ -18062,9 +18062,88 @@ BUILDIN(pcblockmove) {
...
+}
+
+BUILDIN(setpcblock)
+{
+	struct map_session_data *sd = script->rid2sd(st);
+	enum pcblock_action_flag type = script_getnum(st, 2);
+	int state = (script_getnum(st, 3) > 0) ? 1 : 0;
+
+	if (sd == NULL)
+		return true;
```
4144: probably if pc not assigned better return false?

src/map/script.c
4144: ok

PR #847, opened 2015-11-04
Fix *queueremove should return true if player is not in the queue

AnnieRuru:
```
	if ( !queueadd( $@q, getcharid(3) ) )
		dispbottom "success";
	else
		dispbottom "already join";
	end;
}
	if ( !queueremove( $@q, getcharid(3) ) )
		dispbottom "success";
	else
		dispbottom "not yet join";
	end;
}
	if ( queuesize($@q) ) {
		.@it = queueiterator($@q);
		do {
		} while ( qicheck(.@it) );
		qiclear .@it;
	}
	else {
		dispbottom "no entry";
	}		
	end;
OnInit:
	$@q = queue();
	end;
}
```
even if the player is not in the queue, click on the *queueremove will still return success
this pull request will fix it

*PS: yes there is another bug about iterator return -1, will pull request another one*

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -8966,15 +8966,15 @@ Returns the amount of entries in the queue instance of <queue_id>.
 
 *queueadd(<queue_id>,<var_id>);
 
-Adds <var_id> to queue of <queue_id>, returning 1 if <var_id> is already
-present in the queue, otherwise returning 0.
+Adds <var_id> to queue of <queue_id>, returning 0 if <var_id> is already
+present in the queue, otherwise returning 1.
 
 ---------------------------------------
 
 *queueremove(<queue_id>,<var_id>);
 
-Removes <var_id> from queue of <queue_id>, returning 1 if <var_id> is not
-present in the queue, otherwise returning 0.
+Removes <var_id> from queue of <queue_id>, returning 0 if <var_id> is not
+present in the queue, otherwise returning 1.
 
 ---------------------------------------
 
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -18923,6 +18923,9 @@ bool script_hqueue_remove(int idx, int var) {
 			}
 
 		}
+		else
+			return true;
+
 	}
 	return false;
 }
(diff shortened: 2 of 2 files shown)


dastgir:
I guess queue bugs were reported via irc, and this https://github.com/HerculesWS/Hercules/tree/scriptqueue?files=1 was gonna fix it... Haru would be able to enlighten if that branch fixes this too..(mostly it will,)

AnnieRuru:
eh ? I already fixed all the queue bugs on my test server ....
atm, pull request 2nd one

dastgir:
The one on branch totally refactors the script queue(uses vector with it), since problem was something else if I remember well..

AnnieRuru:
oh man ... I can't understand what Haru doing there
I was just putting ShowDebug here and there by trial and error
then bam ! my script works

AnnieRuru:
I've ran through a forum search on the queue iterator script command
and I agree with them saying `queueadd` and `queueremove` should flip their true/false statement

because currently the script will looks like
```
	if ( queueadd( $@q, getcharid(3) ) == false )
		dispbottom "success";
```
when the condition is false, it join successfully ?
doesn't make sense

AnnieRuru:
if want to do the change, do it now
currently nobody using queue iterator script commands yet because of the popular -1 bug

but once all 4 bugs are fixed (and merge)
I'll start spreading this technique on the forum
at that time it will be harder to change things anymore

dastgir:
Well, its vastly used in my eBG, which is gonna release soon, @MishimaHaruna , could you possibly tell if that branch is to be merged or this fixes?

MishimaHaruna:
My branch (as pull request #857) doesn't flip the return value of `queueadd` on the script engine API.
If we decide we want to, then merge the relevant parts of this pull request there (it'll need some small changes, since some functions were refactored)

AnnieRuru:
you are the one that's decide things now, not me haha
whatever decision you make, we will all follow

I think you can fix this simple fix yourself
just leave a mention `(#847)` like this enough

PR #848, opened 2015-11-04
Fix *queueiterator return same amount of *queuesize upon *queueremove

AnnieRuru:
```
	if ( !queueadd( $@q, getcharid(3) ) )
		dispbottom "success";
	else
		dispbottom "already join";
	end;
}
	if ( !queueremove( $@q, getcharid(3) ) )
		dispbottom "success";
	else
		dispbottom "not yet join";
	end;
}
	if ( queuesize($@q) ) {
		.@it = queueiterator($@q);
		do {
		} while ( qicheck(.@it) );
		qiclear .@it;
	}
	else {
		dispbottom "no entry";
	}		
	end;
OnInit:
	$@q = queue();
	end;
}
```
player A join the queue, check will display as
`1. 2000000`
player B join the queue, check will display as
`1. 2000000`
`2. 2000001`
now player A removed from the queue, check will display as
`1. -1`
`2. 2000001`

this is definitely wrong
the `queueiterator` should have the same amount as `queuesize`

this pull request should solve it
*I think there's no need to do what complex index move, just simply list accordingly enough I guess ?*

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -19078,8 +19078,8 @@ BUILDIN(queueiterator) {
 
 	RECREATE(script->hqi[ idx ].item, int, queue->size);
 
-	for ( i = 0; i < queue->size; i++ )
-		if ( queue->item[i] != -1 )
+	for (i = 0; i < queue->size; i++)
+		if (queue->item[i] != -1)
 			script->hqi[idx].item[count++] = queue->item[i];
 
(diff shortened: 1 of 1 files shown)


AnnieRuru:
AHHH forgotten about that if (abc) format ROFL

MishimaHaruna:
This should be fixed in #857 (by refactoring the way those arrays are handled)

Issue #846, opened 2015-11-04
HQO_OnLogout still exist even if the script already performed *queuedel


AnnieRuru:
```
prontera,155,185,5	script	djhfksjf	1_F_MARIA,{
	.qid = queue(); // create new queue
	queueadd .qid, getcharid(3); // add player to the queue
	warp "guild_vs2", 50,50; // play the event !
	queueopt .qid, HQO_OnLogout, strnpcinfo(0)+"::OnQuit"; // if the player logout, automatically forfeit

	sleep 5000; // ok lets say the event already ends

	queuedel .qid; // remove the queue
	mapwarp "guild_vs2", "prontera", 155,185; // warp player back to town
	end; // what ? the player still has HQO_OnLogout ? isn't that *queuedel command should've remove player's HQO_OnLogout ?
OnQuit:
	queueremove .qid, getcharid(3);
	announce strcharinfo(0)+" has quit !", bc_all;
	end;
}
```
after the event ends, if the player had joined this event logout
the server will spam the error
`[Warning]: script_hqueue_remove: unknown queue id 5 (used with var 2000000)`

I'm quite sure this is happening because the qiclear only clear the server side memory.
but it doesn't loop the player's side to remove HQO_Onxxxx stuff

it happens to all 3 HQO_Onxxxx

[s]upon reading `BUILDIN(queuedel)` from script.c further strengthen my claim[/s]
sry wrong info

AnnieRuru:
errr wait ... 
or issit that I have to loop every member's in the loop manually in scripts ?
```
prontera,155,185,5	script	djhfksjf	1_F_MARIA,{
	.qid = queue(); // create new queue
	queueadd .qid, getcharid(3); // add player to the queue
	warp "guild_vs2", 50,50; // play the event !
	queueopt .qid, HQO_OnLogout, strnpcinfo(0)+"::OnQuit";

	sleep 3000; // ok lets say the event already ends

	// ADD THIS
	queueiterator .qid;
	queueremove .qid, qiget(.qid);
	while ( qicheck(.qid) )
		queueremove .qid, qiget(.qid);
	qiclear .qid;
	//
	queuedel .qid; // remove the queue
	mapwarp "guild_vs2", "prontera", 155,185; // warp player back to town
	end; // what ? the player still has HQO_OnLogout ? isn't that *queuedel command should've remove player's HQO_OnLogout ?
OnQuit:
	queueremove .qid, getcharid(3);
[...]

PR #849, opened 2015-11-04
Fix *queuedel doesn't remove player's side queue

AnnieRuru:
https://github.com/HerculesWS/Hercules/issues/846
I found out the reason

sd->queues[j] is the queue ID from the player
script->hq[idx].item[i] is the account ID of the queue in that array
so it was comparing wrong information

it should be comparing to idx instead

I also just check the `script_hqueue_clear`,
that function IS comparing to idx too


AnnieRuru:
damn ... I listed wrong name ...

MishimaHaruna:
This should be fixed in #857 (by having `script_hqueue_del` call `script_hqueue_clear`)

AnnieRuru:
ok, close this one once you merge your vector style ... thing

PR #850, opened 2015-11-04
Fix @QMapChangeTo$ not usable when HQO_OnMapChange triggered

AnnieRuru:
- also add documentation for it
- and change HQO_onLogOut into HQO_OnLogOut in source code

well this is the final one
now my Last Man Standing script working perfectly fine without anymore errors !
http://upaste.me/a93221862ed4bf9a4
I mean, you have to patch all these 4 patches, then only this script doesn't bug

EDIT: this patch contains all 4 bug fix
http://upaste.me/65dc218636607ad7a

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -8987,6 +8987,9 @@ Modifies <queue_id>'s <optionType>. When <option val> is not present
 Currently 3 options are available:
 HQO_OnDeath (1), HQO_OnLogout (2), HQO_OnMapChange (3)
 
+When HQO_OnMapChange event is triggered, it sets a temp char var
+@QMapChangeTo$ with the destination map name.
+
 Note: The constant names are not final.
 
 Example:
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -5452,7 +5452,7 @@ int pc_setpos(struct map_session_data* sd, unsigned short map_index, int x, int 
 		for( i = 0; i < sd->queues_count; i++ ) {
 			struct hQueue *queue;
 			if( (queue = script->queue(sd->queues[i])) && queue->onMapChange[0] != '\0' ) {
-				pc->setregstr(sd, script->add_str("QMapChangeTo"), map->list[m].name);
+				pc->setregstr(sd, script->add_str("@QMapChangeTo$"), map->list[m].name);
 				npc->event(sd, queue->onMapChange, 0);
 			}
 		}
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -18959,7 +18959,7 @@ BUILDIN(queueopt) {
 				else
 					script->hq[idx].onDeath[0] = '\0';
 				break;
-			case HQO_onLogOut:
+			case HQO_OnLogOut:
 				if( script_hasdata(st, 4) )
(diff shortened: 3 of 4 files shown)


AnnieRuru:
maybe should change `@QMapChangeTo$` into `@mapchangeto$`
while nobody using these commands yet ...

Aeromesi:
Curious AnnieRuru, if no one has used these commands, how do you know about them? By looking into Hercules's SRC? and if so how come none of this is documented? First time I've seen anything about such commands as:

queuesize/queueadd/queueopt, things like this I never knew existed and it irritates me if it's already existed, because it would help with my scripting abilities =/

dastgir:
It existed like from past year...

MishimaHaruna:
The commands exist and are documented, but are still at an alpha stage and quite broken (returning wrong values or crashing). I have an ongoing branch where I made some fixes (and some more may be necessary once I look into the latest issues that AnnieRuru opened).
Considering the situation, I agree that if we need to change something in the commands API, it's better to do it now, than when the use of these commands will be widespread.

AnnieRuru:
@Aeromesi 
these script commands has been announced to public for years
http://herc.ws/board/topic/928-memory-slasher-may-30-patch/

there are 2 reasons why people doesn't know about this

1. because when it made known to public,
there are 10 script commands you need to learn in order to write a hercules battleground script
and of course, like haru said. its bugged, and sometimes crash
(its bg_queue crash, not from queue script command itself )
so even some of them want to learn, only to find out these script commands aren't really usable

2. I switched side from rathena to hercules
and I brought this createbgid/setbgid/getbgusers custom script commands from rathena
http://herc.ws/board/topic/4570-battleground-system-without-waitingroom/
and wrote them into plugin in hercules
that's when hercules community found out mine are way easier to learn
mine just need you to learn 3 script commands, but hercules' want you to learn 10 script commands
furthermore, my modification has already spread in rathena forum for at least 1 year before I join hercules,
so I can guarantee mine are 100% bug free
I also refused to learn hercules queue iterator system at that time because I always thought
mine just work the same as hercules queue iterator system when it come to writing bg scripts

it wasn't until recently Ancyker said something about OnPCxxxxEvent cost a lot of server resources
only made me rediscover these queue iterator script command can also be used for events
only just 3 days ago I started to learn how to use these script commands

MishimaHaruna:
@AnnieRuru Do you mind if I merge this into #857, so that both are merged at the same time?

Also, if we are to rename constants, or change anything on the script side of the queue system, we should do so now (and merge everything at the same time, as one pull request). Brainstorming needed.

AnnieRuru:
well, when I was playing around with this queue iterator script commands,
I didn't know you already started it way before me

I think you should disregard my pull request, and only fix whatever I have reported

your branch, doesn't fix the `QMapChangeTo`,
it need `@QMapChangeTo$` the symbols in order to work
try read getpartymember or something else similar, they also return `$@partymemberaid`
```
mapreg->setreg(reference_uid(script->add_str("$@partymemberaid"), j),p->party.member[i].account_id);
```

I still think we should rename `@QMapChangeTo$` into `@mapchangeto$` ... 
though, this just my personal idea anyway

MishimaHaruna:
This was merged as part of PR #857 (I cherry-picked your commit and kept the non-conflicting part). Thank you!

PR #857, opened 2015-11-09
Scriptqueue

MishimaHaruna:
This fixes various issues with the script queue commands, while refactoring the code to comply with the latest Hercules standards.


AnnieRuru:
https://github.com/HerculesWS/Hercules/blob/master/db/const.txt#L3359
should change `HQO_OnLogout` into `SQO_OnLogout` in constant file ?

PR #858, opened 2015-11-09
Update *dispbottom script command to allow 0xRRGGBB color

AnnieRuru:
this has been added into rathena already
https://github.com/rathena/rathena/blob/master/doc/script_commands.txt#L3811

there is no reason we left this one out

and, I just compare hercules and rathena script_commands.txt
I found that hercules missing so many script commands from rathena, its not even funny
be prepare I do a lot of pull request from now on

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -4091,18 +4091,20 @@ Note: rid2name may not produce correct character names since rid means
 
 ---------------------------------------
 
+*message <account ID>,"<message>";
 *message "<character name>","<message>";
 
 That command will send a message to the chat window of the character 
-specified by name. The text will also appear above the head of that 
-character. It will not be seen by anyone else.
+specified by account ID or name. The text will also appear above the head
+of that character. It will not be seen by anyone else.
 
 ---------------------------------------
 
-*dispbottom "<message>";
+*dispbottom "<message>"{,<color>};
 
-This command will send the given message into the invoking character's 
-chat window.
+This command will send the given message into the invoking character's
+chat window. The color format is in RGB (0xRRGGBB), and default to green
+if <color> field is left out.
 
 ---------------------------------------
 
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -13612,14 +13612,24 @@ BUILDIN(atcommand) {
 
 /*==========================================
  * Displays a message for the player only (like system messages like "you got an apple" )
+ * dispbottom "<message>"{,<color>};
  *------------------------------------------*/
 BUILDIN(dispbottom)
 {
-	TBL_PC *sd=script->rid2sd(st);
+	TBL_PC *sd = script->rid2sd(st);
 	const char *message;
-	message=script_getstr(st,2);
-	if(sd)
-		clif_disp_onlyself(sd,message,(int)strlen(message));
+	int color = 0;
+
+	message = script_getstr(st,2);
+	if (script_hasdata(st,3))
+		color = script_getnum(st,3);
+
+	if (sd) {
+		if (script_hasdata(st,3))
+			clif->messagecolor_self(sd->fd, color, message);
+		else
+			clif_disp_onlyself(sd, message, (int)strlen(message));
+	}
 	return true;
 }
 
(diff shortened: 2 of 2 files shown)


src/map/script.c
```diff
@@ -13573,14 +13573,28 @@ BUILDIN(atcommand) {
...
+	TBL_PC *sd = NULL;
+	int color = 0;
 	const char *message;
-	message=script_getstr(st,2);
-	if(sd)
-		clif_disp_onlyself(sd,message,(int)strlen(message));
+
+	message = script_getstr(st,2);
+	if (script_hasdata(st,3))
+		color = script_getnum(st,3);
```
MishimaHaruna: Does this even work? I can see an overflow if you use a color with red value greater than `0x7f`.

src/map/script.c
```diff
@@ -13573,14 +13573,28 @@ BUILDIN(atcommand) {
...
+	if (script_hasdata(st,4))
+		sd = map->id2sd(script_getnum(st,4));
+	else
+		sd = script->rid2sd(st);
+
+	if (sd) {
+		if (color > 0)
```
MishimaHaruna: This condition is incorrect. It can't handle the case where you want the message to be printed in black (`0x000000`)

doc/script_commands.txt
```diff
@@ -4083,18 +4083,20 @@ Note: rid2name may not produce correct character names since rid means
...
 That command will send a message to the chat window of the character 
-specified by name. The text will also appear above the head of that 
-character. It will not be seen by anyone else.
+specified by account ID or name. The text will also appear above the head
+of that character. It will not be seen by anyone else.
 
 ---------------------------------------
 
-*dispbottom "<message>";
+*dispbottom "<message>",{<color>,{<account ID>}};
```
MishimaHaruna: To be honest, I'd prefer it if we left out the 'account ID' optional argument, for various reasons:

- There is no way to specify the account ID without specifying the color, making it half useless.
- Either all commands or no commands should be able to act on a separate character (and I'd prefer no commands *).

*: The reason why I said I prefer that no commands had this option, is that:
- You can already attachrid (and there's very little reason for a script to act on a non-attached character in any case)
- If we were to do it, we'd need an unified way for all commands, to report whether the given character was online or not .Using attachrid instead, you can check whether the attach command was successful.

src/map/script.c
AnnieRuru: hmm... rathena did it with if (script_hasdata(st,3)) condition
so this is my fault

doc/script_commands.txt
AnnieRuru: >Using attachrid instead, you can check whether the attach command was successful.

completely agree with this one
that means our command will be different than rathena
but no matter, the RRGGBB color is more important

AnnieRuru:
just you know, *getitem script command will return true; when the <account ID> player is offline
and I thought it is intended, so even if the player offline,
can always give the rewards to everybody in the array without worry about isloggedin check

AnnieRuru:
what the hell ?
I just test this 
```
prontera,159,185,5	script	kjdsfkshfs	1_F_MARIA,{
	dispbottom "test black", 0x000000;
	dispbottom "actual black", 0x000001;
	dispbottom "test white", 0xffffff;
	dispbottom "test red", 0xff0000;
	dispbottom "test blue", 0x0000ff;
	dispbottom "test yellow", 0xffff00;
	end;
}
```
the 0x000000 actually display white color
0x000001 actually display black color ??

screenshots here
http://herc.ws/board/topic/3899-suggestion-displaybottom-message-colors/?p=65567

MishimaHaruna:
That's very interesting o.o
Client glitch perhaps?

MishimaHaruna:
About getitem returning true: It's true that it doesn't abort the script... But you have no way to know whether the getitem was successful or not... I can only call a command like that, broken by design...

AnnieRuru:
>Client glitch perhaps?

maybe o.o

>About getitem returning true: It's true that it doesn't abort the script... But you have no way to know whether the getitem was successful or not... I can only call a command like that, broken by design...

it always successful
```
prontera,155,185,5	script	kjdhfksjfhs	1_F_MARIA,{
	.pid = getcharid(1);
	getpartymember .pid, 0;
	getpartymember .pid, 2;
	for ( .@i = 0; .@i < $@partymembercount; .@i++ ) {
		message $@partymembername$[.@i], "Your team won the event";
		getitem 501, 1, $@partymemberaid[.@i];
	}
	end;
}
```
not just `getitem` and `message`, there are a few more like `warpchar`, `getcharip` ... 
interestingly, `delitem` does halt the script o.o
`sit`, `stand` `issit` just halt the script without telling no rid attached ...

however I don't really mind if you want to make these changes
this will promote a better and proper scripting habit

*that also means I need to update most of my event scripts haha...hahaha....*

so, do you want to add `script_rid2sd` function check for `*message` command ?

AnnieRuru:
in the reference, Haru said all script commands shouldn't use optional argument for account ID
so this one need a little extra check ...

AnnieRuru:
force commit push, sry for making the comment on the commit disappear
I don't know how to just update to latest revision

so, now this pull request use the latest `script->id2sd` and `script->nick2sd`
I hope Haru doesn't complain anymore XD

tested with
```
prontera,159,185,5	script	kjdsfkshfs	1_F_MARIA,{
    dispbottom "test black", C_BLACK;
    dispbottom "actual black", C_BLACK +1;
    dispbottom "test white", C_WHITE;
    dispbottom "test red", C_RED;
    dispbottom "test blue", C_BLUE;
    dispbottom "test yellow", C_YELLOW;

	message strcharinfo(0), "test message";
	message getcharid(3), "test message";
	detachrid;
	message "AnnieRuru", "test message";
	message 2000000, "test message";
    end;
}
```
it really throw warning if player not found XD

MishimaHaruna:
Merged, thank you!

b814cbd, 19d83e6, 91c12a9

PR #863, opened 2015-11-15
Fix *qicheck can return true, can list 1 extra empty entry

AnnieRuru:
```
	if ( queueadd( $@q, getcharid(3) ) )
		dispbottom "success";
	else
		dispbottom "already join";
	end;
}
	if ( queueremove( $@q, getcharid(3) ) )
		dispbottom "success";
	else
		dispbottom "not yet join";
	end;
}
	if ( queuesize($@q) ) {
		.@it = queueiterator($@q);
//		while ( qicheck(.@it) )
//			dispbottom ( ++.@i )+". "+ qiget(.@it);
		do {
		} while ( qicheck(.@it) );
//		for ( qiget(.@it); qicheck(.@it); qiget(.@it) )
//			dispbottom ( ++.@i )+". "+ qiget(.@it);
		qiclear .@it;
	}
	else {
		dispbottom "no entry";
	}       
	end;
OnInit:
	$@q = queue();
	end;
}
```
using the do { <statement> } while (qicheck(.@it)) method
the script will listed 1 extra entry as 0
`1. AnnieRuru`
`2. 0`

the problem, is everywhere is telling members to use
`for ( qiget(.@it); qicheck(.@it); qiget(.@it) )`
I tried that, but it never return any result

maybe should tell everyone to use do {} while () ?

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -19213,7 +19213,7 @@ BUILDIN(qicheck)
 
 	it = &VECTOR_INDEX(script->hqi, idx);
 
-	if (it->pos <= 0 || it->pos > VECTOR_LENGTH(it->entries)) {
+	if (it->pos <= 0 || it->pos >= VECTOR_LENGTH(it->entries)) {
 		script_pushint(st, 0);
 	} else {
 		script_pushint(st, 1);


AnnieRuru:
yes, its very funny,
http://upaste.me/3f432197164d59389
I already ShowDebug and can tell you value is correct
`VECTOR_INDEX(it->entries, it->pos)`
is shows 2000000 in the debug message, but 
`script_pushint(st, VECTOR_INDEX(it->entries, it->pos++));`
it just display 0. Even I removed the ++, it also shows 0

in that patch, that one works for` ( while(qicheck(.@it) ) <dispbottom>` method

I just can't figure out how to make it works for the
`for ( qiget(.@it); qicheck(.@it); qiget(.@it) )` method

dastgir:
You are using qiget inside for , so that would obviously return 0.
Using for loop will be like this
```
for ( .@entry = qiget(.@it); qicheck(.@it); .@entry = qiget(.@it) )
          dispbottom ( ++.@i )+". "+ .@entry;
```

AnnieRuru:
pfff ... disregard this report then
ok this is not documented everywhere lol

PR #866, opened 2015-11-15
Not allow send too big variable names to char server from map server …

### LINKED ISSUE #865: long variable name cause server crash
while I was amazed by the long variable name `@Queue_Destination_Map$` <- *yes it still works*
so I test this
```
	.@1234567890123456789012345678901234567890$ = "ok";
	dispbottom .@1234567890123456789012345678901234567890$; // it works ?!?!?!
//	a1234567890123456789012345678901234567890$ = "ok"; // server crash if uncomment
	end;
OnInit:
	$a1234567890123456789012345678901234567890$ = "ok"; // after this line load once, then comment this line and @reloadscript will server crash
	announce $a1234567890123456789012345678901234567890$, bc_all;
	end;
}
```
it looks like everywhere,
documentation and wiki told us that variable names cannot longer than 23 characters
but it seems the server doesn't mind how long the variable name ... 

BUT
the problem comes when it needs to save into sql
currently almost all fields use `VARCHAR(32)`, which is 32 characters long
unlike what we have been told all the time

so the server crash when we save the variable name length more than 32
```
[SQL]: DB error - Data too long for column 'key' at row 1
[Debug]: at d:\ragnarok\hercules\src\char\inter.c:588 - REPLACE INTO `char_reg_str_db` (`char_id`,`key`,`index`,`value`) VALUES ('150000','a1234567890123456789012345678901234567890$','0','ok')
```

so, fix all variable name length at 23 or 32 ?

4144:
…(Closes #865)

### DIFF
--- a/src/map/intif.c
+++ b/src/map/intif.c
@@ -333,6 +333,11 @@ int intif_saveregistry(struct map_session_data *sd) {
 		if( varname[0] == '@' ) /* @string$ can get here, so we skip */
 			continue;
 
+		if (strlen(varname) > 31)
+		{
+			ShowError("Variable name too big: %s\n", varname);
+			continue;
+		}
 		src = DB->data2ptr(data);
 
 		/* no need! */


AnnieRuru:
seriously ? o.o
but the seconds one that I said, save into mapreg isn't fixed
```
OnInit:
	$a1234567890123456789012345678901234567890$ = "ok";
	announce $a1234567890123456789012345678901234567890$, bc_all;
	sleep 1000;
	$a1234567890123456789012345678901234567890$ = "";
	announce $a1234567890123456789012345678901234567890$, bc_all;
	end;
}
```

PR #867, opened 2015-11-15
Throw error when variable name length too long

AnnieRuru:
@dastgir 
shouldn't it be something like this ? o.o
so it throw error on what ever kind of variable type, to make it consistent

and ... its `VARCHAR(32) BINARY` ... so ... not sure if it is exactly 32 length ...

#865 
awwww ....


src/map/script.c
```diff
@@ -2680,6 +2680,13 @@ struct script_data *get_val(struct script_state* st, struct script_data* data) {
 	prefix = name[0];
 	postfix = name[strlen(name) - 1];
 
+	if (strlen(name) > 32) {
+		ShowError("script_get_val: variable name too long. '%s'\n", name);
```
4144: not 32, but 31.
Also probably it should be some where in parser if you want prevent using it.
Same for second check

AnnieRuru: already tested it can save in SQL with 32 characters of $zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz$
but maybe because it is `VARCHAR(32) BINARY` and not `VARCHAR(32)`

4144:
must be 31.
Your server not crashing, but corrupting memory only. 32 is wrong. in char server buffer is 32 and it mean 31 char + 1 null char. strlen count only non null chars.

AnnieRuru:
o.o ... mapreg doesn't corrupt, but char-server can corrupt ... this is funny
yeah, my char-server crash when saving 32 characters
```
OnInit:
    a2345678901234567890123456789012 = 123456;
    announce a2345678901234567890123456789012 +"", bc_all;
    end;
}
```

AnnieRuru:
ok I got an entry in my char_reg_str_db table,
`a234567890123456789012345678901$`
the moment I log in, map-server crash

now this is not funny anymore, serious bug

dastgir:
Travis check failed because you haven't updated main.SQL
It should have
```
INSERT INTO `sql_updates` (`timestamp`) VALUES (1447620161);
```

PR #868, opened 2015-11-16
Fix various ShowWarning that doesn't tell the source of the error

AnnieRuru:
`return true;`
means its ok.

`return false;`
means the script will still run, nothing wrong, but it just report the source of the error

if want to halt the script, use `st->state = END;`

if want to report the source of error without using return false, use `script->reportsrc(st)`

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -18854,7 +18854,7 @@ BUILDIN(queuesize)
 	if (idx < 0 || idx >= VECTOR_LENGTH(script->hq) || !VECTOR_INDEX(script->hq, idx).valid) {
 		ShowWarning("buildin_queuesize: unknown queue id %d\n",idx);
 		script_pushint(st, 0);
-		return true;
+		return false;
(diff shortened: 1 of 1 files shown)


dastgir:
Seems good to be merged.
@4144

4144:
Why not add "return false" to other functions if error happend?
For example in strcharinfo.

AnnieRuru:
`return false;` only shows the debug message of the source npc
it doesn't tell `Player not Attached!` error
only `script->rid2sd(st)` has the ShowError message on it, `map->id2sd(id)` don't

for `(strcharinfo)` case,
the `script->rid2sd(st)` already has `script->reportsrc(st);` inside the `*script_rid2sd` function
if change `if (!sd) return true;` into `return false;` the source npc will show twice

AnnieRuru:
err ... bump ?

MishimaHaruna:
Merged, thank you!

PR #871, opened 2015-11-16
Add *getunittype and update *getmapxy

AnnieRuru:
rathena has this custom stuff
https://github.com/rathena/rathena/commit/2cee5b6
and this makes hercules members jealous
of course, I think hercules should add them as well

however, when I review their code, its SUPER SUCKS
I will tell how bad of their code later, 1 by 1
and I do this very very slowly to ensure nothing goes wrong

so this particular one,
... ok let me laugh 1st ...
their `getunittype` seems to insert some random value
BUT we will follow using the value from existing `getmapxy`
of course, CONSTANTS !

this has been tested with
```
prontera,155,188,5	script	sdkfjshdkfd	1_F_MARIA,{
	skill 238, 1, 0; // alchemist skills
	skill 243, 1, 0;
	skill 244, 1, 0;
	skill 247, 1, 0;
	.@id = monster( "this", -1,-1, "--ja--", PORING, 1, "" );
	getmapxy .@map$, .@x, .@y, 3, .@id;
	dispbottom .@map$ +" "+ .@x +" "+ .@y;
	end;
OnPCDieEvent:
	if ( !killerrid ) end;
	getmapxy .@map$, .@x, .@y, getunittype(killerrid), killerrid;
	dispbottom .@map$ +" "+ .@x +" "+ .@y;
	end;
}
prontera	mapflag	pvp
```
I used homunculus, pets, monsters, etc to kill that player
that player can shows coordinates of the killer object without failure

in this case, their getunitdata that return coordinate is useless
I already coded setmobdata/getmobdata
http://upaste.me/ed6c2196382229bc7
but I still need to finalize the code
when I'm ready, I'll pull request that one


AnnieRuru:
can merge now ?
I'm writing something that touches `map->nick2sd` now
if I pull request that now without merging this, will create conflict with that patch later

Emistry:
just curious, does it able to retrieve the coordinate X and Y for dead monster that triggered the event ??
```
OnKill:
	getmapxy( .@map$,.@x,.@y, getunittype(killerrid), killerrid );
	dispbottom "Die location - "+ .@map$ +" "+ .@x +" "+ .@y;
	end;
```

dastgir:
If monster is dead, possibly all its info is free'd, so maybe not?

AnnieRuru:
@Emistry 
OnPCDieEvent -> `killerrid` is set to the **GID** of the monster
https://github.com/HerculesWS/Hercules/blob/master/src/map/pc.c#L7501 <- `src->id` is GID
OnNPCKillEvent -> `killedrid` set to the **mob_id** of the monster
https://github.com/HerculesWS/Hercules/blob/master/src/map/mob.c#L2601 <- `md->class` is mob_id
monster with event label -> nothing is set
since you know which monster by manipulating event label

yes, dastgir is correct, for that kind of information, have to do it by `*mobevent` script command
by pushing a custom variable `$@mob_x` and `$@mob_y`,
monster with MOBEVENT_DEAD will be another `else if` condition,
which doesn't trigger with OnNPCKillEvent or having an event label
since monster with event label cannot trace the ID, so have do [something like this](https://www.eathena.ws/board/index.php?s=&showtopic=273450&view=findpost&p=1503966)
... I always say I plan to code `*mobevent`, by never start yet LOL

well I guess after I finish up `setmobdata/getmobdata`, `*mobevent` will be next
but this pull request have to merge 1st !! since they come in parts

MishimaHaruna:
Merged, thank you!

Issue #881, opened 2015-11-20
Assassin Critical Rate


godmike:
I don't know if this is bug in my server, but one of my player insisting that this is bug when wearing any type of Katar the critical rate should doubled the value in the status window.

FYI- I am using the Pre-Renewal Settings(2013-12-23)

AnnieRuru:
https://github.com/HerculesWS/Hercules/issues/573

just like the others (item's sex doesn't matter, hidden npc can be detected ... )
if the official server do something that doesn't make sense,
can always do the correct behavior in battle config
and the value will always follow the official behavior

godmike:
You mean the default settings is the doubled value doesn't show in the stats window. right?
Actually I am trying to implement the code they shared on the #573, but the code is already existing in my source. And the doubled value doesn't show when I wear Katar.

dastgir:
Solved by #882

PR #882, opened 2015-11-20
*Show the katar critical bonus in player status window

AnnieRuru:
- this is unofficial behavior, so it has to be in battle config
#881


MishimaHaruna:
Merged, thank you!

Issue #883, opened 2015-11-20
Autobonus not working


Jedzkie:
Hi! it seems *autobonus, *autobonus2, *autobonus3, and *autobonus4 is not working. It does not call the script even the chance is 100%, tried in ID: 5342

Kindly check this @MishimaHaruna @4144 @AnnieRuru

AnnieRuru:
found the culprit

PR #886, opened 2015-11-20
Fix documentation error for autobonus trigger rate (#883)

AnnieRuru:
#883
@Jedzkie , try this ?

the value should be 10000 according to documentation
and check all items in item_db.conf, does it use 1000 as 100% or 10000 as 100% ?

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -5699,7 +5699,7 @@ as expected, refer only to an invoking character.
 What these commands do is 'attach' a script to the player which will get
 executed on attack (or when attacked in the case of autobonus2).
 
-Rate is the trigger rate of the script (10000 = 100%). 
+Rate is the trigger rate of the script (1000 = 100%). 
 
 Duration is the time that the bonus will last for since the script has 
 triggered.


AnnieRuru:
now that I read carefully
https://github.com/HerculesWS/Hercules/blob/master/doc/script_commands.txt#L5738
```
//Grants a 1% chance of starting the state "all stats +10" for 10 seconds 
//when using weapon or misc attacks (both melee and ranged skills) and 
//shows a special effect when the bonus is active.
	autobonus "{ bonus bAllStats,10; }",10,10000,BF_WEAPON|BF_MISC,"{ specialeffect2 EF_FIRESPLASHHIT; }";
```
if it is 10000, 1% should be 100
but in the example script, it is 10

now I start to wonder this is wrong on the documentation instead ?

AnnieRuru:
yes
I don't have official info, and the client translation always say *chances to trigger ...*
they never tell what are the chances in percent

need official confirmation =/

MishimaHaruna:
I can probably do some comparison with official data (likely against a RWC 2012 special.sc and Zone server)

MishimaHaruna:
I tried to find the code responsible for the `SetAutoWeapon`/`SetAutoMweapon`/`SetAutoAttacked`/`SetSkillActivate` Aegis commands (equivalent to our `autobonus` variants), but it's a bit more complex than expected. I found some parts I suspect to be related, but I might be completely off, so I'd rather take another route.

I examined several item scripts, and no value higher than 1000 seems to have ever been used to date (at least in the scripts I could find). This might either mean that 1000 is equivalent to 100%, or that they always use very low values for balance reasons.

The item with the highest value that I found is `Krieger_Launcher1` (an outstanding 200). If our code is correct, it'd mean that, in Aegis, the item has a 20% to trigger its effect on each attach (which seems ridiculously high).
Would anyone be able to test that in an aegis server? Unlike many other items, the difference (between 2% and 20%) should be very noticeable, so it should be enough to clear up the doubt.

AnnieRuru:
talking about reading rathena source code, I just found their documentation is at 1000
https://github.com/rathena/rathena/blob/master/doc/script_commands.txt#L5331
since this typo is passed down since eathena period, I think we can trust in their information

hemagx:
If I can go home soon, will test on aegis

dastgir:
@hemagx still not at home? xD

hemagx:
Yep, I'm moving. But I will ask my friend to do this test, hopefully he
will have enough free Time.

hemagx:
Test Character Status:
Job: Gunslinger
Level: 99
ASPD: 190
Weapon: Krieger_Launcher1
Refine: +10

Test Results:

1- With Effect duration changed to 1 second instead of 20 cause Aegis doesn't repeat under effect
Hits: 500
Triggers: 41

2- With Effect Duration changed to 0 second instead of 20
Hits: 500
Triggers: 95 ~ 105 ( a chance of 10 wrong counts )

3- With Effect Duration changed to 0 seconds instead of 20 and chance to 1000 instead of 200
Hits: 500
Triggers: 500 ( Happened with each bullet ! )

Here's my report, Tested on Aegis.

hemagx:
@MishimaHaruna i would like to check all random stuff from items after this test, we may-be failing other stuff too.

AnnieRuru:
@hemagx 
try check this one #895 while you are at it
that one also mess up

hemagx:
@AnnieRuru
pull #895  doesn't require any checks, it just require a serious cleanup, i discussed the way with haru and maybe i will do it.
HOWEVER, for formulas i can check later but still i prefer to not play with skills formulas >.> or generally battle.c/skill.c/status.c before serious rewrite.

MishimaHaruna:
Thank you, merged in 53abb8cdf92cd3d153b0a3d5b99b55cbd631f36c

theultramage:
Ref https://github.com/HerculesWS/Hercules/commit/d0c501816214cd604e7f7efddb8debff03c514e8 (https://github.com/rathena/rathena/commit/99de37788255b7205d2ae93249499e510df3aab5). This @jaBote made an un-auditable commit, which included ```Corrected an error in the rate of autobonus command (100% is 10000, not 1000 as it was before).```, but at the time of the commit, the code was using 1000 like usual. Cydh fixed it in rA this summer: https://github.com/rathena/rathena/commit/290bd622f302b90217867f4479216b5fb5fc6941.

Issue #703, opened 2015-09-07
ForceSerial not working


Jedzkie:
Hi All!

I tried to put force serial in this item ID#13921 (Antique_Pipe_Box) and by using getitem, the items stacked, it does not separate the items one by one,  Im using 2010-06-16aRagexeRE and pre-re configuration, using the latest herc version.

Thanks!

MishimaHaruna:
Is this still a problem? If so, was your getitem specifying an amount greater than 1? Does it stack between separate getitem calls?

Jedzkie:
@MishimaHaruna yes its greater than 1, no it does not stack in seperate getitem calls.

AnnieRuru:
how to reproduce ?
if you mean this...
```
{
	Id: 13921
	AegisName: "Antique_Pipe_Box"
	Name: "Gentleman's Pipe Box"
	Type: 18
	Buy: 20
	Weight: 10
	Trade: {
		nodrop: true
		notrade: true
		noselltonpc: true
		nocart: true
		nogstorage: true
		nomail: true
		noauction: true
	}
	Script: <" getitem 5377,5; ">
},
```
I got 5 Gentleman's Pipe in separated, non stack

Jedzkie:
Nope, not that. I mean the box itself not the items that get in the box. For example getitem Antique_Pipe_Box,5;

AnnieRuru:
does all cash shop item (type 18) give separate items ?
if it is, then the fix is very easy, I can do it

AnnieRuru:
that will be my last time to fix stuffs on the source code
Ind gave me repo access to fix npc scripts and documentations only .....

AnnieRuru:
tested again
item_db.conf, having the red potion force serial
```
{
	Id: 501
	AegisName: "Red_Potion"
	Name: "Red Potion"
	Type: 0
	Buy: 50
	Weight: 70
	ForceSerial: true
	Script: <" itemheal rand(45,65),0; ">
},
```
then the script
```
prontera,155,185,5	script	give split	1_F_MARIA,{
	getitem 501, 1;
	getitem 501, 2;
	getitem 501, 3;
	getitem 501, 4;
	getitem 502, 1;
	getitem 502, 2;
	getitem 502, 3;
	getitem 502, 4;
	end;
}
prontera,158,185,5	script	merge	1_F_MARIA,{
	mergeitem;
	end;
}
```
it seems to be working correctly already

@Jedzkie try again ?

[...]

PR #887, opened 2015-11-20
Fix getitem/getitem2 give stacked items when ForceSerial on

AnnieRuru:
#703

haiz ......

### DIFF
--- a/src/map/itemdb.c
+++ b/src/map/itemdb.c
@@ -492,8 +492,11 @@ int itemdb_isequip2(struct item_data *data) {
  *------------------------------------------*/
 int itemdb_isstackable(int nameid)
 {
-	int type=itemdb_type(nameid);
-	switch(type) {
+	struct item_data *data = itemdb->exists(nameid);
+	nullpo_ret(data);
+	if (data->flag.force_serial)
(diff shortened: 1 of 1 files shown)


MishimaHaruna:
I'm not sure we should handle it this way. With this patch, items with the force-serial option enabled, become non-stackable.

I believe there are some cases where we want them to stackable (please correct me if I'm wrong):
- Items with a serial can be merged (and have their serial removed), i.e. through the "Mergician" NPC (i.e. `mergeitem` script command)
- A `getitem` with `amount > 0` of a force-serial item should ideally generate a stack of items (with the same serial)

AnnieRuru:
>A getitem with amount > 0 of a force-serial item should ideally generate a stack of items (with the same serial)

ehh ... ?
then you mean me and Jedzkie has misunderstood the behavior of force serial ?

if what you say there is right, then this patch is garbage
the original code from main branch is already correct

hemagx:
if the item is stackable then it serialize the whole stack with same serial, if other stack came goes with another serial, it's how it works :)

AnnieRuru:
close this one then, I'll identify the real threat

PR #913, opened 2015-12-07
Random Start Point

KirieZ:
Fix issue #007750 (from old tracker)

### DIFF
--- a/conf/char-server.conf
+++ b/conf/char-server.conf
@@ -94,9 +94,10 @@ autosave_time: 60
 // Display information on the console whenever characters/guilds/parties/pets are loaded/saved?
 save_log: yes
 
-// Start point, Map name followed by coordinates (x,y)
-start_point: iz_int,97,90
-start_point_pre: new_1-1,53,111
+// Start points, Map name followed by coordinates (x,y)
+// Note: It's possible to add up to 5 points, it will be chosen radomly when character is created
+start_point: iz_int,97,90,iz_int01,97,90,iz_int02,97,90,iz_int03,97,90,iz_int04,97,90
+start_point_pre: new_1-1,53,111,new_2-1,53,111,new_3-1,53,111,new_4-1,53,111,new_5-1,53,111
 
 // Starting items for new characters
 // Format is: id1,quantity1,stackable1,idN,quantityN,stackableN
--- a/src/char/char.c
+++ b/src/char/char.c
@@ -137,10 +137,11 @@ struct fame_list taekwon_fame_list[MAX_FAME_LIST];
 
 // Initial position (it's possible to set it in conf file)
 #ifdef RENEWAL
-	struct point start_point = { 0, 97, 90 };
+	struct point start_point[MAX_START_POINTS] = { { 0, 97, 90 } };
 #else
-	struct point start_point = { 0, 53, 111 };
+	struct point start_point[MAX_START_POINTS] = { { 0, 53, 111 } };
 #endif
+int start_point_num;
 
 unsigned short skillid2idx[MAX_SKILL_ID];
 
@@ -1645,6 +1646,7 @@ int char_make_new_char_sql(struct char_session_data* sd, char* name_, int str, i
 	char name[NAME_LENGTH];
 	char esc_name[NAME_LENGTH*2+1];
 	int char_id, flag, k, l;
+	int point_id = rand() % start_point_num;
 
 	nullpo_retr(-2, sd);
 	nullpo_retr(-2, name_);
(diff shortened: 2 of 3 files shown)


KirieZ:
@Jedzkie  to be honest I don't know which symbol to use haha I think colon will not work very well (maybe config file will parse it incorrectly? not sure how the parser works) but I was thinking about semi-colon, so:
mapname,x,y;mapname2,x2,y2;

AnnieRuru:
http://herc.ws/board/tracker/issue-7750-official-server-startpoint/

I remember I read somewhere that strtok is not a safe operation
Ctrl+F inside src folder, we don't use them

however I saw strtok_r enclosed in ifdef WIN32, I found this funny though

its better to rewrite this stuffs like *explode script command, using a declared delimiter properly

KirieZ:
@AnnieRuru i've used strtok because start_items uses it (in the same file)

AnnieRuru:
https://github.com/HerculesWS/Hercules/blob/master/conf/char-server.conf#L106
```
// Starting items for new characters
// Format is: id1,quantity1,stackable1,idN,quantityN,stackableN
// stackable:
//		0 - Not stackable (weapon, armor, egg, pet armor)
//		1 - Stackable
start_items: 1201,1,0,2301,1,0
```
oh yeah ... that kind of rathena-like format ... speechless
*nothing to say*
well still have to review by other dev, since **this is an official feature**

dastgir:
@AnnieRuru we would be on that format if https://github.com/HerculesWS/Hercules/tree/char-server_update branch was merged.

hemagx:
You don't start in random one, you start in diffente one if previous is
full ( 250 player in it ) we need to add option for limit player and actual
check for all maps

AnnieRuru:
https://github.com/HerculesWS/Hercules/commit/c6482e9870645ffe59a6a059b819574d4ac79fd9
ah yes, this looks more like *hercules*

but what hemagx said is quite interesting, if this is true then this pull request void

a few things that I am thinking about
1. what happens if all maps are full ? which shouldn't happen I guess, 1250 new players ...
2. then what happens if just create a character, then leave it like that, and only login on a later date ?
for example, when the server just open, I immediately create 9 characters, and only login the 1st one to play
when I login the 1st char already pushed me to 3rd map (already got 500+ new players ?)
so what will happen to the other 8 chars if login on a later date when the server already less full?

hemagx:
I'm not home yet sadly to check my files, but I believe character server
management this part live time.

KirieZ:
about the format, certainly this new format is a lot better, @Dastgir do you know if this will be merged soon?
if what hemag said is true, wouldn't be better to do this through script? we can set a fixed spawn and warp on login

hemagx:
Nope, a script will not be executed till map loaded wich will cause warp
effect ( pretty annoying ) also you will Litterly need many scripts and
checks in every single cell, player should be moved if busy place ( not
introduced in aegis but best logic ) finally scripts slower :p

KirieZ:
What I was thinking was to put it at OnPcLogin, if map has more than a number of users try next one, but as you said, it'll probably do the warp effect.

So this check have to be done when character is logging into the game world, right? if player is at start map and current map is full, try the next one until a free map is found (like getmapusers script command). But in this case, I believe the code would have to be made in map-server, while at least one spawn point is kept in char-server (for character creation).

If I understood correctly aegis only does these moves in the first login, right? so if I login and I'm moved to new_1-2, and logout, whenever I login again I'll be in this map? I believe this would be the best, as it would be strange that guys around you will disappear if you just logout and login again :P

hemagx:
It's good way, but the other way seem more logical.   I would say to add
option btw none will have this issue, seriously no Ro server official or
private had this numbers expect Japan server, this system can be helpful in
case of custom train system when you actually need to cap to low number 10
~5, of course there's other options but it would be lovely to have train
map cap option or even as general option, I'm not sure if it's possible to
be in map server as client should get his first map from character I
believe ? I don't remember this part tbh

Smokexyz:
@HerculesWS/developers This PR has conflicts and is outdated. Shall we close this?
Or would @KirieZ like to perform an update?
The concept that this PR is tackling is unclear on the official behaviour. It seems like a "hack" to attempt to emulate something similar.

KirieZ:
I will update it if you wantl, but as you said, it's not completely clear if that's the official behavior.

Mysteries:
Do we have confirmation of where you start in kRO? @Ridley8819 if you create a new char... are you in random locations or consistently placed in the same location?

Smokexyz:
@Asheraf can you also confirm if the start point is different or the same? I don't remember starting on different maps with a novice on kRO.

PR #917, opened 2015-12-09
Update max packet size according to client date

dastgir:


src/common/socket.c
```diff
@@ -238,7 +238,11 @@ fd_set readfds;
 
 // Maximum packet size in bytes, which the client is able to handle.
 // Larger packets cause a buffer overflow and stack corruption.
-static size_t socket_max_client_packet = 24576;
+#if PACKETVER >= 20131223
+static size_t socket_max_client_packet = 0x10000;
```
MishimaHaruna: Are you sure this is 0x10000, rather than 0xffff or 0x7fff?

Rytech2: When I looked in the client through ollydbg, I saw the hex value as 0x10000 and just added this as 65536. Older clients showed a value of 0x5000.

MishimaHaruna: But there is a difference between `if (len < 0x10000) { /* accept the packet */ }` and `if (len > 0x10000) { /* reject the packet */ }`. Both have a 0x10000, but accept different ranges. 0x10000, in our case, would mean a buffer of 0x10001.

Rytech2: So your going to set the buffer higher then the max the client can support? I wouldn't do it. When Ai4rei made this config in eA, she set it to 20480 which im guessing she has the source setup to reject anything larger. I just assumed if she put the exact value of what the old clients used, then using the exact value of what the new ones use would be safe. Unless dev's in Herc changed something on this.

dastgir: Then, Why it was 0x6000 default till now ?

Rytech2: Thats something I would like to know. Did someone in the Herc team change it, or did someone in rA do it and then it got merged in here? I noticed that rA is applied the new size limit as well a day or 2 ago.

dastgir:
Changed to 0xFFFF,

src/common/socket.c
dastgir: I just got someone who pointed me to that commit, and what's the issue, So I searched up in hercules forum, and found out your post saying about max packet size..

4144:
Why removing configuration option from code?
Probably better comment it in config, but if some one want, it can be uncommented.

dastgir:
Oh, that way it's better, will commit it soon

dastgir:
Done @4144

4144:
@dastgir plaease rebase this pr, and i will merge it

dastgir:
What to rebase? Isn't there only 1 commit?

4144:
yes, but if merger it as is it will be not at top, but some where in history

dastgir:
Done @4144

PR #918, opened 2015-12-10
Script CleanUp

dastgir:
* [x] Changes ItemID to Constants
* [ ] set to Direct Assignment


dastgir:
Opening other, since this one have many items constant containing "." which gives error when loading script.

PR #921, opened 2015-12-10
ItemDB Updates:

Jedzkie:
- Added 'Malang Cat Can' item.
 - Added 'Ru Blue Spear' item.
 - Added specialeffect on GX's Poison items
 - Modified some of new Cards from kRO (http://forums.irowiki.org/showpost.php?p=1412481&postcount=6)
 - Change the item id of 'RWC Silver Brooch' from 2998 to 2999.
 - Follow up https://github.com/HerculesWS/Hercules/commit/e511dcf5940cee7a6e43744543a5ca96bc23cf84, thanks to zackdreaver.
 - item_db.conf renewal minor clean-up xD


dastgir:
Seems good...
@MishimaHaruna

db/re/item_db.conf
```diff
@@ -42242,8 +42307,9 @@ item_db: (
 	Weight: 10
 	Loc: 2
 	Script: <"
-		bonus2 bSkillAtk,GC_CROSSIMPACT,100;
-		if (getrefine() >= 10) bonus2 bSkillAtk,GC_CROSSIMPACT,200;
+		bonus2 bSkillAtk,GC_CROSSIMPACT,50;
+		if (getequipid(EQI_HAND_R) == 4) bonus bSkillAtk,GC_CROSSIMPACT,50;
```
dastgir: bonus2

db/re/item_db.conf
```diff
@@ -42242,8 +42307,9 @@ item_db: (
 	Weight: 10
 	Loc: 2
 	Script: <"
-		bonus2 bSkillAtk,GC_CROSSIMPACT,100;
-		if (getrefine() >= 10) bonus2 bSkillAtk,GC_CROSSIMPACT,200;
+		bonus2 bSkillAtk,GC_CROSSIMPACT,50;
+		if (getequipid(EQI_HAND_R) == 4) bonus2 bSkillAtk,GC_CROSSIMPACT,50;
```
AnnieRuru: `if (getequipid(EQI_HAND_R) == 4)`
means, if the right hand equipment item ID is 4
is there any Item ID 4 ??
I thought it start from 501, and its Red_Potion !

something wrong on this line

hemagx:
is it correct to add items without AEGIS data ?
some items are safe to be added cause they have no secret info ( like rates etc ) and so simple and direct in the description, however we still have issue the AEGIS Name of them item.
as soon as the real data come to us the Aegis name will have to be changed and then we can break so many stuff and we know that kRO love to make weird Aegis names sometimes so guessing it is not an option too.

db/re/item_db.conf
Jedzkie: Sorry typo its getequipweaponlv

Jedzkie:
For now, we are getting information thru item description, its not a problem about the item name or the rates, we can commit it again until we get the correct information for that, my goal for now is to get the item information as fast as possible even i dont have enough data for that item, so that we are not far from the development of aegis in regards of item database and mob database.

hemagx:
Aegis name is constant so when it get changed later it's a constant change same bad as if we change item id later.
as soon as we push this data to master branch it means it should be usable so for sure some will use it.
using Aegis names in scripts are much better than IDs for various reasons one of it is being readable without the need of check item_db all time.
now let's say some started use it and then we had item database changes to fix Aegis name, they should check for every update for sure but in item database updates it's hard to spot such slight changes so then we will turn all them scripts to be broken cause it will try to getitem 0, cause our consent doesn't exist anymore. that scenario. 
other one is if we implement official scripts includes this items and then later we change the aegis names again to correct one, but forget to change scripts it will get broken.

right now such updates for constant are not safe with our script engine that never detected the issue till it run the script for real.

Jedzkie:
@dastgir ? any news for this? is this good to merge now?

MishimaHaruna:
Thank you, merged as f00431e

PR #923, opened 2015-12-10
Change the sequence of Status ID's according to its number.

Jedzkie:
Minor clean-up.


MishimaHaruna:
Manually merged as 412e377b6e0eb28ad01712808773d0c7effb8676, thank you!

PR #926, opened 2015-12-10
Implemented NoViewID

dastgir:
(http://herc.ws/board/topic/10020-official-mapflag-to-disable-viewid/)


src/map/atcommand.c
```diff
@@ -3873,6 +3873,8 @@ ACMD(mapinfo) {
 		strcat(atcmd_output, msg_fd(fd,1096)); // PartyLock |
 	if (map->list[m_id].flag.guildlock)
 		strcat(atcmd_output, msg_fd(fd,1097)); // GuildLock |
+	if (map->list[m_id].flag.guildlock)
```
4144: copy/paste error?

dastgir: Indeed

src/map/map.h
```diff
@@ -716,6 +718,7 @@ struct map_data {
 		unsigned noknockback : 1;
 		unsigned notomb : 1;
 		unsigned nocashshop : 1;
+		unsigned noviewid : 22;
```
AnnieRuru: where did you get the value 22 ?
is there a constant ? EQI_MAX ?

AnnieRuru:
we don't have a doc\mapflag.txt like rathena do
https://github.com/rathena/rathena/blob/master/doc/mapflags.txt

AnnieRuru:
I got a feeling this whole patch looks funny 
in `setmapflag` script command, we have this
`map->list[m].flag.noviewid = (val <= 0) ? 0 : val; break;`
then when checking the condition, it is
`if (!map_no_view(sd->bl.m,EQP_COSTUME_GARMENT) && .... `
means we have to know the actual bitmask of EQP_COSTUME_GARMENT, which is 0x002000
but the limit seems to be just 22, something not right

src/map/map.h
dastgir: Max Value is EQP_SHADOW_ACC_L = 0x200000, i.e 22 bit,
the : 22 means 22 Bit, and not value 22

dastgir:
22 is the Bit, and not the Maximum value

AnnieRuru:
hahaha, come laugh at me for someone who doesn't take class in C
ok so, we should define a `EQP_MAX` in pc.h,
will it give compile error ? hmm ...

src/map/map.h
```diff
@@ -227,6 +227,8 @@ enum {
 #define map_flag_gvg2(m) (map->list[m].flag.gvg || map->list[m].flag.gvg_castle)
 // No Kill Steal Protection
 #define map_flag_ks(m) (map->list[m].flag.town || map->list[m].flag.pvp || map->list[m].flag.gvg || map->list[m].flag.battleground)
+// No ViewID
+#define map_no_view(m, view) (map->list[m].flag.noviewid & view)
```
MishimaHaruna: You might want to add parentheses around `view`, to make this safer.

```
map_no_view(m, foo|bar);  // (map->list[m].flag.noview & foo | bar)
```

dastgir:
Done @MishimaHaruna

MishimaHaruna:
Merged, thank you!

theultramage:
The '22' bits thing is really awful, please change it to a simple integer!
Might be useful to add ```// enum equip_pos bitmask``` next to the declaration in map.h.
Also, please fix the tabs-instead-of-spaces indentation used in script.c in that patch.

MishimaHaruna:
Hmm, good points indeed.

PR #929, opened 2015-12-10
Changed avg_increment formula

dastgir:


dastgir:
@hemagx

src/map/status.c
```diff
@@ -12358,7 +12358,7 @@ void status_read_job_db_sub(int idx, const char *name, config_setting_t *jdb)
 			avg_increment = 1;
 		}
 		for (++level; level <= pc->max_level[idx][0]; level++ ) {
-			status->dbs->SP_table[idx][level] = min(base + avg_increment * level, battle_config.max_sp);
+			status->dbs->SP_table[idx][level] = min(status->dbs->SP_table[idx][level-1] + avg_increment, battle_config.max_sp);
```
MishimaHaruna: You're no longer using the `base` variable, but you're still calculating and assigning to it.

Also, if `level` is 1 here (meaning an empty `SPTable` entry in the file), here you're accessing the array entry `status->dbs->SP_table[idx][0]`, which I don't see being assigned to anywhere.

dastgir:
How about this @MishimaHaruna ?

MishimaHaruna:
Tagging as in-progress for now (for the reasons previously discussed)

HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/929)

PR #937, opened 2015-12-12
Update mute checks mess

hemagx:
Change all chat mute checks to use pc->can_talk instead of direct check.
Put new macro pc_ismuted and change all other kind of mutes to use it

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -6111,9 +6111,7 @@ ACMD(npctalk)
 	bool ifcolor=(*(info->command + 7) != 'c' && *(info->command + 7) != 'C')?0:1;
 	unsigned int color = 0;
 
-	if (sd->sc.count && //no "chatting" while muted.
-		(sd->sc.data[SC_BERSERK] || (sd->sc.data[SC_DEEP_SLEEP] && sd->sc.data[SC_DEEP_SLEEP]->val2) ||
-		 (sd->sc.data[SC_NOCHAT] && sd->sc.data[SC_NOCHAT]->val1&MANNER_NOCHAT)))
+	if (!pc->can_talk(sd))
 		return false;
 
 	if(!ifcolor) {
@@ -6160,9 +6158,7 @@ ACMD(pettalk)
 		return false;
 	}
 
-	if (sd->sc.count && //no "chatting" while muted.
-		(sd->sc.data[SC_BERSERK] || (sd->sc.data[SC_DEEP_SLEEP] && sd->sc.data[SC_DEEP_SLEEP]->val2) ||
-		 (sd->sc.data[SC_NOCHAT] && sd->sc.data[SC_NOCHAT]->val1&MANNER_NOCHAT)))
+	if (!pc->can_talk(sd))
 		return false;
 
 	if (!*message || sscanf(message, "%99[^\n]", mes) < 1) {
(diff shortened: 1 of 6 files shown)


dastgir:
Yay, finally.

AnnieRuru:
looks ok to me

AnnieRuru:
after I do some testing, I found that, the player has already muted, right ?
so whats the point of having another `pc->cantalk` check in atcommand ?
since the player already mute, they also can't use any atcommand

hemagx:
Because there's status mute you too without the being actually muted?

AnnieRuru:
example ?
I know SC_NOCHAT and SC_BERSERK definitely block player do atcommand
```
	sc_start SC_NOCHAT,10000,1,10000,SCFLAG_NOAVOID|SCFLAG_FIXEDRATE|SCFLAG_FIXEDTICK;
//	sc_end SC_ALL;
//	sc_start2 SC_PHI_DEMON, 1200000, 10, 10;
//	sc_start4 SC_BUCHEDENOEL, 600000, 3, 3, 3, 7;
//	sc_start SC_INCMHPRATE,600000,3;
//	sc_start SC_INCMSPRATE,600000,3;
//	sc_start SC_INCHITRATE,600000,3;
//	sc_start SC_CRITICALPERCENT,600000,7;
OnInit:
	attachrid 2000000;
	sc_end SC_ALL;
}
```

hemagx:
Players under normal mute/status mute should be still able to execute commands, source been always like this, unless it's been bugged, that other case but that how it was and how being suggested to be.
the pc_can_talk is to ensure the pc can actually talk under different status please check the follow code

``` C
/**
 * Determines whether a player can talk/whisper based on status changes
 * Called from clif_parse_GlobalMessage and clif_parse_WisMessage
 * @retval true Can talk
 **/
bool pc_can_talk( struct map_session_data *sd ) {
	nullpo_retr(false, sd);

	if( sd->sc.data[SC_BERSERK] ||
		(sd->sc.data[SC_DEEP_SLEEP] && sd->sc.data[SC_DEEP_SLEEP]->val2) ||
		(sd->sc.data[SC_NOCHAT] && sd->sc.data[SC_NOCHAT]->val1&MANNER_NOCHAT) )
		return false;

	return true;
}
```

hemagx:
The talk commands however, blocked to use under mute.
forgot to mention.

AnnieRuru:
LOL of course I know about pc->can_talk ...
>Players under normal mute/status mute should be still able to execute commands
Haru also said the same thing before here (#842)

```
	sc_start SC_BERSERK, 1,1;
	end;
OnInit:
	attachrid 2000000;
	sc_end SC_ALL;
}
```
try test it, I can't execute atcommands

I don't simply believe in reading source code
I only believe in my test server

MishimaHaruna:
When in SC_BERSERK, commands are accepted... The behavior you're seeing is because the client normally doesn't send chat messages to the server when it knows it's in berserk (but of course a hexed client, or a bot, or a third party client could send them).

If you want to test it, trick the client into thinking you're not in berserk state (skip the berserk status change packet), while still setting it in source, and test atcommands then (you'll see you're still able to use them)

Code you can use to do that test (added lines marked with `+`):

```
 void clif_status_change(struct block_list *bl,int type,int flag,int tick,int val1, int val2, int val3) {
 	struct packet_status_change p;
 	struct map_session_data *sd;
 
 	if (type == SI_BLANK)  //It shows nothing on the client...
 		return;
 
 	nullpo_retv(bl);
 
 	if (!(status->type2relevant_bl_types(type)&bl->type)) // only send status changes that actually matter to the client
 		return;
 
 	if ( tick < 0 )
 		tick = 9999;
 
 	sd = BL_CAST(BL_PC, bl);
 
+	if (type == SI_BERSERK) {
+		// Let's trick the client into thinking there's no berserk status :D
+		clif->message(sd->fd, flag ? "****Berserk started****" : "****Berserk ended****");
+		return;
+	}
+
 	p.PacketType = status_changeType;
 	p.index = type;
 	p.AID = bl->id;
 	p.state = (unsigned char)flag;
 
 #if PACKETVER >= 20120618
 	p.Total = tick; /* at this stage remain and total are the same value I believe */
 #endif
 #if PACKETVER >= 20090121
 	p.Left = tick;
 	p.val1 = val1;
 	p.val2 = val2;
 	p.val3 = val3;
 #endif
 	clif->send(&p,sizeof(p), bl, (sd && sd->status.option&OPTION_INVISIBLE) ? SELF : AREA);
 }
```

Issue #723, opened 2015-09-19
Plugin addBattleConf


SamuelHercules:
I have added battleconf settings into the plugin which is loaded perfectly.
 
The problem is, the value can't be detected by the script command getbattleflag
 
Although setbattleflag command can affect the custom battle_conf setting I have made with the plugin.

I have checked the getbattleflag and it seems to just check the battle_data which is limited to the list in battle.h? or is there a way getbattleflag can read custom battleconf settings added in a plugin?

Thanks!

MishimaHaruna:
Confirmed. `getbattleflag` doesn't check for plugin-owned configuration settings.

Samuel23:
bump?

PR #940, opened 2015-12-12
HPM Config embading update

hemagx:
-- Fixed Possible Crash when null parse function pointer passed to HPMi->addConf
-- Fixed issue #723 now it's possible to retrieve Battle Config Settings from plugins into scripts
-- Now it's possible to use same parse function for all config entries
-- Now Battle Config entries must have a return function


MishimaHaruna:
Thank you! Merged as f102c91

PR #943, opened 2015-12-13
Fix SC_STOMACHACHE to its official behavior.

Jedzkie:


MishimaHaruna:
Thank you, merged as 6476db2

PR #945, opened 2015-12-14
eathena_upgrade.sql - Adding a missing parse on line 59

maqc1:
Self explanatory.


MishimaHaruna:
Thank you very much, merged

PR #950, opened 2015-12-14
Script commands with optional parameter look for player will now ShowWarning

AnnieRuru:
@MishimaHaruna , I just finished bake your favorite chocolate pie :smile:
now script commands that optionally look for player char_id/account_id/name,
will ShowWarning on map-server if the player is not found

no longer promote bad scripting habit, like it ?

well the ingredient was actually taken from rathena script engine ...
https://github.com/rathena/rathena/blob/master/src/map/script.c#L120

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -5942,7 +5942,7 @@ BUILDIN(warpchar) {
 	y=script_getnum(st,4);
 	a=script_getnum(st,5);
 
-	sd = map->charid2sd(a);
+	sd = script->charid2sd(st, a);
 	if( sd == NULL )
 		return true;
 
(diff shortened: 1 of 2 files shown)


MishimaHaruna:
Looks good, I'll be reviewing (and hopefully merging) this today :)

I like how this standardizes the error messages through the code (although I'd still advise against using any commands that directly target a not-attached player - the reason is that the calling script won't know whether the command succeeded or not, and won't be able to act accordingly)

MishimaHaruna:
`ispartneron` should be exempted from this, shouldn't it? (it's meant to check if the partner of the currently attached character is online, and shouldn't show a warning when it isn't). I'll edit that one out while merging.

MishimaHaruna:
Merged, thank you!

AnnieRuru:
> (although I'd still advise against using any commands that directly target a not-attached player - the reason is that the calling script won't know whether the command succeeded or not, and won't be able to act accordingly)

@MishimaHaruna , if you really want to make the script halt like `script->rid2sd` function do
just add `st->state = END;` inside those 3 functions, very easy :D
though, I still prefer to make it just ShowWarning without the END state
otherwise some live server event script broken

MishimaHaruna:
No, I don't think I want it to END... The command would have no way to recover from that. It's better to have it continue with a warning. The reason why I advise using attachrid instead of this is that attachrid returns false if it fails to attach (so that the script will know that it failed, and can inform the user, or take any necessary action).
If a failure to execute the command isn't important, it's of course fine to use this method instead.

PR #953, opened 2015-12-14
Correct item type 11 to be Restricted usable item type.

hemagx:
Now this type gonna make the usable item not usable on sit/mount and not
delayed consumer anymore.
to make an item delayed consumer you have to use KeepAfterUse option.

Current Tasks to be done before merge :

- [x] Source Edit
- [ ] Check all items in database to make sure all items that should be delayed updated
- [ ] Update Wiki  And inform community


AnnieRuru:
wait wait wait ... you mean you really gonna change the type 11 behavior
so all our customs has to add `KeepAfterUse` flag ?

hemagx:
yes.

AnnieRuru:
then you have to fix `*itemskill` and `*pet` script command as well
I just tested firebolt scroll (690) not working after this patch

hemagx:
@MishimaHaruna 

at pc.c ( pc_useitem )
``` C
	if (sd->inventory_data[n]->flag.keepafteruse)
		clif->useitemack(sd,n,amount,true);
	else {
		if (sd->status.inventory[n].expire_time == 0) {
			clif->useitemack(sd, n, amount - 1, true);
			removeItem = true;
		} else {
			clif->useitemack(sd, n, 0, false);
		}
	}
```

would not be better for customize to only keep items with flag (keepafteruse) ? 
you can still assign it to rental item

src/map/battle.c
```diff
@@ -4047,7 +4047,8 @@ struct Damage battle_calc_misc_attack(struct block_list *src,struct block_list *
 #endif
 			md.damage = ( (sd?pc->checkskill(sd,NC_MAINFRAME):10) + 8 ) * ( skill_lv + 1 ) * ( status_get_sp(src) + sstatus->vit );
 			RE_LVL_MDMOD(100);
-			md.damage += status_get_hp(src) - totaldef;
+
+			md.damage += ((status_get_hp(src) < totaldef) ? totaldef : status_get_hp(src)) - totaldef;
```
dastgir: Split this into another commit?
Since this is not with item type, but some underflow check...

src/map/skill.c
```diff
@@ -13048,7 +13047,7 @@ int skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_id
 			sd->itemid = sd->itemindex = -1;
 			if( skill_id == WZ_EARTHSPIKE && sc && sc->data[SC_EARTHSCROLL] && rnd()%100 > sc->data[SC_EARTHSCROLL]->val2 ) // [marquis007]
 				; //Do not consume item.
-			else if( sd->status.inventory[i].expire_time == 0 ) // Rental usable items are not consumed until expiration
+			else if( sd->status.inventory[i].expire_time == 0 && sd->inventory_data[i]->flag.keepafteruse) // Rental usable items are not consumed until expiration
```
dastgir: So what would happen if item is rental and not with KeepAfterUse tag?

src/map/battle.c
hemagx: That was mistake. my files a bit messy.

src/map/skill.c
hemagx: being deleted ? a consumable rental should never exist so, if you wanna keep it use the flag.

hemagx:
Current Commit should be last one for source hopefully.

hemagx:
@AnnieRuru @Jedzkie Please help me in check items as i have no knowledge about it

AnnieRuru:
if you wanna change the name, there are a few places left 
https://github.com/HerculesWS/Hercules/blob/master/conf/battle/party.conf#L28
https://github.com/HerculesWS/Hercules/blob/master/src/plugins/db2sql.c#L76

check items ? example ?
most item type that not using *itemskill or *pet need to add KeepAfterUse flag, that is ...

hemagx:
pet taming are not even type 11, tho it should be type 11.
for itemskill, well it's custom change to those items.

kyeme:
@hemagx Upon double clicking the scroll it already consumes 1 scroll and upon clicking it to the target another 1 is consumed. Im not just sure maybe there is something wrong when i applied the code?

AnnieRuru:
nope I don't get that error ...
how did you apply the code ? manually ?

http://stackoverflow.com/questions/6188591/download-github-pull-request-as-unified-diff
always download the pull request patch by adding `.diff` on it
like this pull request is
https://github.com/HerculesWS/Hercules/pull/953
add `.diff` at the end
https://github.com/HerculesWS/Hercules/pull/953.diff

kyeme:
@AnnieRuru I will apply again, thanks for the tip.

hemagx:
the source seem to had enough testing, however i can't inform community and edit wiki myself cause i suck in this.
@MishimaHaruna @Mysteries @dastgir @4144

src/map/skill.c
AnnieRuru: >a consumable rental should never exist 

I think I found a few consumable rental official item
```
{
	Id: 17226
	AegisName: "C_Center_Potion_Box"
	Name: "Infinite Concentration Potion Box"
	Type: 18
	Buy: 10
	Weight: 10
	Trade: {
		nodrop: true
		notrade: true
		noselltonpc: true
		nocart: true
		nogstorage: true
		nomail: true
		noauction: true
	}
	Script: <" rentitem C_Center_Potion,604800; ">
},
```
```
{
	Id: 12884
	AegisName: "C_Center_Potion"
	Name: "Infinite Concentration Potion"
	Type: 11
	Trade: {
		nodrop: true
		notrade: true
		nostorage: true
		nocart: true
		noselltonpc: true
		nomail: true
		noauction: true
		nogstorage: true
	}
	Script: <"
		sc_start SC_ATTHASTE_POTION1,1800000,4;
		specialeffect2 EF_POTION_CON;  /* NOTE: No special effect when used. */
	">
},
```

hemagx: I need more information, i'm sure Aegis can't rent items for sure.

AnnieRuru: what information do you need ?
I found this item inside hercules' `item_db.conf` and also rathena `item_db.txt`
I think this item should be official isn't it ?
just test this item in Aegis

hemagx: consumable items in Aegis can't go rental cause actually, in the item struct there's no place to put rental end date.
so rent it is impossible so this look very weird, i need to know the source of this item.

hemagx: wait wait, there's case it's possible.
but then it's not about the rental item, it's hardcoded for sure.
item type 11, forgot it.
so we should just mark it as keepafteruse

kyeme:
~Up~

kyeme:
Boarding halter <- rental.

hemagx:
Yep, rental, but if you have it not rental on Aegis it never gone :) so i believe it's hard coded ^^

MishimaHaruna:
The db2sql plugin will no longer compile after this.

```
StrBuf->Printf(&buf, "'%u',", it->flag.delay_consume?IT_DELAYCONSUME:it->type);
```
I think we can change it to just this?
```
StrBuf->Printf(&buf, "'%u',", it->type);
```

hemagx:
i believe so haruka

kyeme:
~ Up ~

MishimaHaruna:
Is this ready to merge (except for the edit suggested above)?

hemagx:
@MishimaHaruna exporting list from Aegis

dastgir:
@hemagx what's left?

HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/953)

MrKeiKun:
Requesting for rebase thanks~

PR #954, opened 2015-12-14
Fixes Shadow/Costume Items on older clients

### LINKED ISSUE #947: Costume System Enhancement
Hi! I just found out that if i uses clients that doesn't support costume system like (2010-06-16aRagexeRE), i can still wear it but i can't unequip it, maybe adding checks that if PACKETVER is LOWER than 2010-07-07aRagexeRE, you cannot wear costume items. @hemagx  :)

hemagx:
Costume and Shadow items are not equip-able anymore on older clients that doesn't support it.

Closes #947


MishimaHaruna:
Merged, thank you

PR #955, opened 2015-12-14
Replaced several numeric constants with enums

MishimaHaruna:
This extends the Mob ID and item ID enums with all the currently hardcoded mob/item IDs, and adds similar enums for homunculus IDs, mercenary IDs and elemental IDs.

Some missing assertions have been added, and minor cleanup was done to the edited code, wherever it seemed unnecessarily complex or convoluted.


MishimaHaruna:
This is ready to merge, if anyone can review it

AnnieRuru:
compile - ok
test homunculus - ok
4144 can merge it now

this patch breaks my koe plugin lol, I have to fix that custom plugin later

4144:
@MishimaHaruna what you think about moving hardcoded ids to configs?
This will allow to change mobid/homunid/etcid without recompilation. Because now possible to change mob_db but need rebuild server

MishimaHaruna:
@4144: Ideally, we might want to remove some of those (and improve the item database, for example to mark the emperium or the treasure boxes or the manuk/splendide/mora mobs, rather than hardcoding their IDs.
Once we reduce the IDs that need to be magically treated in a special way, then sure, we can move them to a configuration file...

PR #956, opened 2015-12-15
Added support for the period sign in the script engine constants

### LINKED ISSUE #830: Unable to use AegisName as constant  if it contain full stop symbol
```
prontera,156,178,5  script  sdkfdsjf    1_F_MARIA,{
    dispbottom "Mob ID: "+ DEVIRUCHI +" | Egg ID: "+ Deviruchi_Egg; // OK
    dispbottom "Mob ID: "+ BAPHOMET_ +" | Egg ID: "+ Bapho_Jr._Egg; // error
}
```

immediately get an error from map server

.

use Regular Expression to find with Notepad++ inside item_db.conf

```
AegisName: "[a-z0-9_]*\.
```

return about 7 results

MishimaHaruna:
- Fixes #830
- AegisNames are now validated, and only [A-Za-z0-9_'.] are allowed
- All other symbols have been replaced with `_`.

---

This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/956)

### DIFF
--- a/db/pre-re/item_db.conf
+++ b/db/pre-re/item_db.conf
@@ -56241,7 +56241,7 @@ item_db: (
 },
 {
 	Id: 6375
-	AegisName: "Amatsu_Bead_!"
+	AegisName: "Amatsu_Bead__"
 	Name: "Amatsu Bead !"
 	Type: 3
 },
--- a/db/re/item_db.conf
+++ b/db/re/item_db.conf
@@ -23787,7 +23787,7 @@ item_db: (
 },
 {
 	Id: 2023
-	AegisName: "Thanatos_Two-Handed_Staff"
+	AegisName: "Thanatos_Two_Handed_Staff"
 	Name: "Thanatos Two-Handed Staff"
 	Type: 4
 	Buy: 20
@@ -49323,7 +49323,7 @@ item_db: (
 },
 {
 	Id: 4688
-	AegisName: "True_Howard_Alt-Eisen_Card"
+	AegisName: "True_Howard_Alt_Eisen_Card"
 	Name: "True Howard Alt-Eisen Card"
 	Type: 6
 	Buy: 20
@@ -71626,7 +71626,7 @@ item_db: (
 },
 {
 	Id: 6375
-	AegisName: "Amatsu_Bead_!"
+	AegisName: "Amatsu_Bead__"
 	Name: "Amatsu Bead !"
 },
 {
@@ -123945,7 +123945,7 @@ item_db: (
 },
 {
 	Id: 15073
-	AegisName: "Anti-Magic_Suit"
+	AegisName: "Anti_Magic_Suit"
 	Name: "Anti-Magic Suit"
 	Type: 5
 	Buy: 20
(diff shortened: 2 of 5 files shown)


AnnieRuru:
after I apply this patch
```
	set .as.d'f'', 4;
	dispbottom .as.d'f'' +"";
	end;
}
```
this script no longer throw error

you mean now we are allow to have `.` and `'` in our variables too ?

MishimaHaruna:
Although not officially, `'` was accepted for a while already. This patch only adds `.`... I agree with you that we shouldn't encourage it, but I'm not sure about disallowing it. When variables or constants are parsed, it doesn't know yet whether it'll be a constant or not, so we can't cause a parse-time error (it's the same limitation of this current scripting language/engine that doesn't let us error out when the user tries to use a nonexisting constant, the reason why I want to change the syntax and the parser).

How about we keep it as accepted, but not documented (or in other words, use at your own risk, but we don't officially allow it, so if you use it and it breaks, you get to keep the pieces)

AnnieRuru:
the reason why `'` is accepted is because it is an instance variable
I just know today we can actually do it at the middle on other types of variables

why not just add an extra check at `set_reg` function ?
since the constant doesn't run that function ...
so do a similar check like `skip_word` function inside `set_reg` function
only skip the check for prefix symbols, not allow to do them at the middle of their variables

tested, although it can't detect at parse time, but at least able to detect at run time

MishimaHaruna:
Actually, the real reason why ' is allowed are item constants (think of `Monster's_Feed`)... Instance variables are handled a few lines above (line 751 in the edited file). The prefix check is separate from the check for internal characters (@ and $ are only allowed in the prefix)

Your suggestion of testing it in set_reg would work, I guess it's better than nothing. It'll be almost undetectable though (similar to typos in constant names, you'll never know it happened until too late).

The real fix (parse time check) will be a feature of the new engine when we get to work on it

AnnieRuru:
the build failed because there is conflict in item_db.conf, nvm that
```
prontera,155,185,5  script  kdjsfhkshf  1_F_MARIA,{
//  set ..as.d'f'', 4;
    dispbottom ..as.d'f'' +"";
    end;
OnInit:
    ..as.d'f'' = 4;
    end;
}
```
this script, the map-server keep crashing
I tried to ShowDebug
```
	if (!validate_varname(name)) {
ShowDebug("1");
		if (is_string_variable(name)) {
ShowDebug("2");
			ShowError("script:set_reg: failed to set value for invalid variable name '%s' to \"%s\"\n", name, (const char *)value);
		} else {
ShowDebug("3");
			ShowError("script:set_reg: failed to set value for invalid variable name '%s' to '%d'\n", name, (int)h64BPTRSIZE(value));
		}
ShowDebug("4");
		script->reportsrc(st);
ShowDebug("5");
		st->state = (sd->state.dialog) ? CLOSE : END;
ShowDebug("6");
		if (st->state == CLOSE)
			clif->scriptclose(sd, st->oid);
		return 0;
	}
```
it show until number 5
I am guessing because script running on OnInit doesn't have `sd`
it is NULL, and when it read `(sd->state.dialog)` the server crash

MishimaHaruna:
Indeed, my wrong! I'll edit it to also check that `sd != NULL`.

AnnieRuru:
related to #1003 
https://github.com/HerculesWS/Hercules/blob/master/src/map/script.c#L15716
currently we can still make variable name longer than 32 characters with `*setd`/`*getd` script commands

its good that you made a `validate_varname` function here
try use that function in setd/getd as well

PR #957, opened 2015-12-15
ItemDB Updates:

Jedzkie:
- Change ID to Constants of some items both pre-re and re.
 - Corrected some item names and scripts.
 - Added 'Mysterious_Can2', 'Mysterious_PET_Bottle2', '2009_Rice_Cake_Soup', 'Vulcan Bullet Magazine', 'Unripe_Yggdrasil', 'Iggdrasilberry_Box' effect.

### DIFF
--- a/db/pre-re/item_db.conf
+++ b/db/pre-re/item_db.conf
@@ -63179,7 +63179,7 @@ item_db: (
 	Buy: 2
 	Weight: 350
 	BuyingStore: true
-	Script: <" getitem 13204,500; ">
+	Script: <" getitem Lightning_Sphere, 500; ">
 },
 {
 	Id: 12145
@@ -63189,7 +63189,7 @@ item_db: (
 	Buy: 2
 	Weight: 350
 	BuyingStore: true
-	Script: <" getitem 13206,500; ">
+	Script: <" getitem Blind_Sphere, 500; ">
 },
 {
 	Id: 12146
@@ -63199,7 +63199,7 @@ item_db: (
 	Buy: 2
 	Weight: 350
 	BuyingStore: true
-	Script: <" getitem 13205,500; ">
+	Script: <" getitem Poison_Sphere, 500; ">
 },
 {
 	Id: 12147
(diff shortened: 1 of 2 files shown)


db/re/item_db.conf
```diff
@@ -84724,7 +84744,7 @@ item_db: (
 {
 	Id: 12642
 	AegisName: "Fruit_Of_Mastela_Box2"
-	Name: "Fruit Box (100)"
+	Name: "Mastela Fruit 100 Box"
```
dastgir: Mastela Fruit Box 100 maybe?

db/pre-re/item_db.conf
```diff
@@ -66942,7 +66952,7 @@ item_db: (
 	Job: 0x00000400
 	Upper: 8
 	BuyingStore: true
-	Script: <" //madoheal rand(300,400),0; ">
+	Script: <" if(checkmount() == MOUNT_MADO) itemheal rand(300,400), 0; ">
```
dastgir: Mado in pre-renewal?
Seems doubtful

Jedzkie: 3rd jobs are enabled in pre-re, what if server admin uses 3rd job in pre-renewal? Mechanics cannot use that item.

AnnieRuru: I remember I read a blog in rathena saying ... 3rd job never exist in pre-renewal
https://rathena.org/board/blog/23/entry-87-pre-renewal-rathena/
if we want to follow it officially, yeah pre-renewal shouldn't even have any 3rd job stuff
I don't play official server anymore so don't ask  me about official things

Jedzkie: yes, in official, 3rd job never exist in pre-renewal. but in our System, it is enabled but it has limitations.

dastgir: @MishimaHaruna

MishimaHaruna:
Merging, thank you!

PR #960, opened 2015-12-15
Found some static-assignment overflows -> fixed.

wilkemeyer:
Just found while fixing code to be able to be compiled as C++ src.

Overflow in Static Assignemt -> the declared buffer was too small for "\0", as "" will implicitly terminate the string.

### DIFF
--- a/src/map/map.c
+++ b/src/map/map.c
@@ -3912,7 +3912,7 @@ struct map_zone_data *map_merge_zone(struct map_zone_data *main, struct map_zone
 }
 
 void map_zone_change2(int m, struct map_zone_data *zone) {
-	char empty[1] = "\0";
+	char *empty = "";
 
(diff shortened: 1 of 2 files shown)


MishimaHaruna:
Please note that this is perfectly acceptable C code, while it's not valid C++ code, and compiling Hercules in C++ is a non-goal of this project (C++ uses a too small subset of C to be useful, and modifies it in too many restrictive and incompatible ways).

Regardless of that, this pull request will be merged (with small edits), because using literal character strings with an embedded `\0` is poor code style.
Initializing a non-const char pointer with a static string, though, is bad practice (modifying the content of such a character string yields undefined behavior), so I'll edit those to be const char where possible.

MishimaHaruna:
Thank you very much, merged as 8a6f7e1f79146f6989bc921152f2fc539d44aa0f

PR #958, opened 2015-12-15
ItemDB Updates:

Jedzkie:
* [x] Implement official effect of MVP Scrolls.
* [x] Change ID to Constants
* [x] Dropping 'SC_ITEMSCRIPT' (SC_ITEMSCRIPT is not present in aegis script commands and status list)

PR #962, opened 2015-12-16
Added official effect for Buche De Noel item. (SC_BUCHEDENOEL)

Jedzkie:


AnnieRuru:
no good, previously it added the value like this
```
	sc_end SC_ALL;
	sc_start4 SC_BUCHEDENOEL, 600000, 3, 3, 3, 7;
//	sc_start SC_INCMHPRATE,600000,3;
//	sc_start SC_INCMSPRATE,600000,3;
//	sc_start SC_INCHITRATE,600000,3;
//	sc_start SC_CRITICALPERCENT,600000,7;
}
```
in the comment one, I saw my Critical in player status screen from 1 into 8
but when I use yours, the value went from 1 into 2

something not right

Jedzkie:
Thanks for the Report, Going to fix this.

Jedzkie:
@AnnieRuru Done.

AnnieRuru:
still no good
```
	if (sc->data[SC_BUCHEDENOEL]) {
		ShowDebug( "%d", sc->data[SC_BUCHEDENOEL]->val4 );
		critical += sc->data[SC_BUCHEDENOEL]->val4 *10;
	}
```
I tested it needs to *10 the value in order for the value to display properly in status window

hmm ... ok now have to recheck all status in `status_calc_critical` function
because I found that the *10 is from crusader spear quicken

Jedzkie:
Omggggggggg... -.- whyyyyy!?!?!?!?!?!

AnnieRuru:
because there is no documentation, or comments on top of that function ;-)
so obviously everyone did wrong
I also noticed this line
`return (short)cap_value(critical,10,SHRT_MAX);`
it cap the minimum value at 10,
it can't mean you have 10 critical at minimum, right ?

dastgir:
What I suggest is, if 2 PR changes SC, make them into 1 PR(but not 1 commit)
, like if #958 gets merged before this, then this one will be conflicted, else that one will be conflicted.

Just my thought

AnnieRuru:
we can always force commit in PR to resolve conflicts ^.^

Jedzkie:
@dastgir okay next time.

MishimaHaruna:
What's the status of this? I can resolve the merge conflicts, but is the implementation correct?

AnnieRuru:
nope, @MishimaHaruna , this PR is bug
also all the status change in `status_calc_critical` need to review again
https://github.com/HerculesWS/Hercules/blob/master/src/map/status.c#L4888
there are no comment on top of the function to remind others to `*10` the value for the critical

in order to add 1 critical from a status change, has to do 
`critical += sc->data[SC_WHATEVER]->val1 *10;`
but in that function, some did the *`10`, some didn't
in this PR, Jedzkie didn't do the `*10`

dastgir:
This is what happens when we lack proper documentation...

Jedzkie:
Found this @AnnieRuru in status.c

case SC_PLUSAVOIDVALUE:
case SC_CRITICALPERCENT:
	val2 = val1*10; //Actual boost (since 100% = 1000)
	break;

AnnieRuru:
its useless to tell me LOL, I'm scripting mod, not a developer
I only write custom stuffs

... I am soo speechless, why not just `*10` inside `status_calc_critical`
but have to `*10` inside `status_change_start` ... ?
and confirm only some of them has `*10` inside `status_change_start`
which makes this a culprit

I don't have aegis or official information about these kinda thing, I can't do this =/

PR #965, opened 2015-12-16
Fix battle config Hp/Sp overflow value

AnnieRuru:
while I was playing with `max_hp` value in `\battle\player.conf`, and set the value very high
I noticed my char_server.exe spam this error message
```
[SQL]: DB error - Out of range value for column 'hp' at row 1
[Debug]: at d:\ragnarok\hercules\src\char\char.c:479 - UPDATE `char` SET `base_l
evel`='150', `job_level`='60',`base_exp`='0', `job_exp`='0', `zeny`='100000000',
`max_hp`='30959',`hp`='100030959',`max_sp`='1821',`sp`='100002121',`status_point
`='2597',`skill_point`='0',`str`='130',`agi`='130',`vit`='130',`int`='130',`dex`
='130',`luk`='1',`option`='0',`party_id`='0',`guild_id`='1',`pet_id`='0',`homun_
id`='2',`elemental_id`='0',`weapon`='14',`shield`='0',`head_top`='167',`head_mid
`='0',`head_bottom`='0',`last_map`='prontera',`last_x`='155',`last_y`='189',`sav
e_map`='prontera',`save_x`='156',`save_y`='191', `rename`='0',`delete_date`='0',
`robe`='0',`slotchange`='0', `char_opt`='0', `font`='0', `uniqueitem_counter` ='
1919',`hotkey_rowshift`='0' WHERE  `account_id`='2000000' AND `char_id` = '15000
0'
```
after play around more and found my hexed client can only accept value to 21474836

hmm ... I wonder should assign to @4144 or @dastgir  ?


hemagx:
after testing the maximum value is 0x7FFFFFFF ( 2147483647 )
How have you done your test @AnnieRuru

AnnieRuru:
screenshot tell one thousand words
battle config
`max_hp: 21474836`
`max_sp: 21474836`
https://github.com/AnnieRuru/customs/blob/master/screenshots/screenRuru_Server034.jpg
battle config
`max_hp: 21474837`
`max_sp: 21474837`
https://github.com/AnnieRuru/customs/blob/master/screenshots/screenRuru_Server035.jpg
and I no need to explain the maximum value of mediumint unsigned in SQL, right ?

hemagx:
if you wanna test max client capity of anything, you should do it by  directly play with packets and not server settings that may-be affected by bugs/other caps.
http://i.imgur.com/qCF5l07.jpg
this require further investigation anyway.

AnnieRuru:
yes, the hp/sp bar can support up to that value,
but Alt+End shows your Hp/Sp doesn't

com'on I knew people gonna say that,
if that person really knows how to play with packet,
they also know how to edit the battle.c file

dastgir:
@AnnieRuru , not really,
I have seen many people(mostly server owners) which only touch one part of the source, and only knows one part :p


And this is about SQL type and not hexed client.
as suggested by hemagx, if we wanna see client limit, play with packets to know to what extent er can reach

Rytech2:
Here's what I see in the character_info packet....
  /* this+0x2a */ int hp
  /* this+0x2e */ int maxhp
  /* this+0x32 */ short sp
  /* this+0x34 */ short maxsp

But this is from a dated aegis server. Im seeing homunculus, mercenary, and elemental packets that use INT values for SP and seeing your screenshot on how high the SP is appearing as, SP for players is sending as INT as well. The ZC_PAR_CHANGE packet which updates things shown on the GUI also sends everything it updates in a INT value. All of this means that the max possible HP and SP is limited to the max value a Signed INT value can give.

AnnieRuru:
hmm I just noticed this part
look at my screenshot, the HP value is `21474837`, but it show **-99%**
and hemagx's screenshot, the HP value is `2147483647`, but it shows 0% ... or maybe haven't regen his hp yet

I start to think this is not about packet, but client side ... coding limitation ...
@Rytech2 can you try whats the % value of the HP if gone over `21474836` ?
does it loop over into negative percent too ?

I'm guessing the server did actually send an int type packet already
but my client can't actually show the percent properly

hemagx:
yea yea, in Aegis it's limited to 21474836 if your hp going higher Aegis will hardcode it to 1.
btw for the -99% it seem like you're sending something wrong. but yes the HP bar is bugged.

MishimaHaruna:
So is it confirmed to be limited to 21474836 to avoid client glitches? Can I merge this?

AnnieRuru:
I admit I'm bad with client ....
I only check with my own hexed client with hercules/rathena

probably need someone to actually check with aegis too

MishimaHaruna:
Confirmed. Merged, thank you!

PR #971, opened 2015-12-16
Added official effect for Ancient Spirit Agimat item. (SC_PHI_DEMON)

Jedzkie:


AnnieRuru:
still no good, 20% increase damage I got

I seriously think this part
```
		if (sc->data[SC_PHI_DEMON]) {
			sd->right_weapon.addrace[RC_DEMON] += sc->data[SC_PHI_DEMON]->val1;
			sd->left_weapon.addrace[RC_DEMON] += sc->data[SC_PHI_DEMON]->val1;
		}
```
should belongs to battle.c
`damage += damage * sc->data[SC_PHI_DEMON]->val1 / 100;`

Jedzkie:
Weird.. I tried this again, and i got 10% only. o_O

AnnieRuru:
I tried again, I got 34944 into 41866
which is 1.198x times increase
*(please make sure your dex is high enough so it wont do random damage)*

need 3rd person to test ...

dastgir:
By the commit , it seems good, I will check ingame now though...

dastgir:
SC conflict with recent commits, update it and it will be merged

dastgir:
@AnnieRuru the method Jedzkie used is right,
(I was using renewal)
The calculation of bonuses in renewal/pre-renewal is kinda at fault..
@malufett and @Michieru knows those stuff better, and would be good if they could fix the calculations

PR #973, opened 2015-12-16
Fix Cash Shop Item Type

### LINKED ISSUE #704: Force Serial
When you opened a box (type:18) and its contents has a ForceSerial: true, the items are not stackable. This is normal but the bug arises when you reconnect your character and you open another Box, the first one you opened are now stacked to the old one and to your next box opening, it will not be stack-able. So in the first place, the bug happens when you reconnect.

hemagx:
Now Cash Shop item type is not stackable with option to make it stackable
in config.
also items coming out from cash shop item type items ( aka boxes ) get a
unique id
Closes #704

### DIFF
--- a/conf/battle/items.conf
+++ b/conf/battle/items.conf
@@ -93,3 +93,9 @@ item_enabled_npc: 1
 // 2 : disabled equipments are nullify, disabled cards will caused the equipment to unequip
 // 3 : disabled equipments are unequip, disabled cards will caused the equipment to unequip (1+2)
 unequip_restricted_equipment: 0
+
+// Stack Cash Shop Type items
+// Officially this items doesn't stack and unique_id is guaranteed for each
+// 1 : yes
+// 0 : no(official)
+cash_item_stack: 0
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -7153,6 +7153,7 @@ static const struct battle_data {
 	{ "feature.roulette",                   &battle_config.feature_roulette,                1,      0,      1,              },
 	{ "show_monster_hp_bar",                &battle_config.show_monster_hp_bar,             1,      0,      1,              },
 	{ "fix_warp_hit_delay_abuse",           &battle_config.fix_warp_hit_delay_abuse,        0,      0,      1,              },
+	{ "cash_item_stack",                    &battle_config.cash_item_stack,                 0,      0,      1,              },
 };
 #ifndef STATS_OPT_OUT
 /**
--- a/src/map/battle.h
+++ b/src/map/battle.h
@@ -512,6 +512,7 @@ struct Battle_Config {
 	int show_monster_hp_bar; // [Frost]
 
 	int fix_warp_hit_delay_abuse;
+	int cash_item_stack;
 };
 
 /* criteria for battle_config.idletime_critera */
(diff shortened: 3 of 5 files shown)


dastgir:
Just a suggestion,
Can you also make config for IT_CASH to be stackable?

hemagx:
Check the commit please ? :D i already done this.

conf/battle/items.conf
```diff
@@ -93,3 +93,9 @@ item_enabled_npc: 1
 // 2 : disabled equipments are nullify, disabled cards will caused the equipment to unequip
 // 3 : disabled equipments are unequip, disabled cards will caused the equipment to unequip (1+2)
 unequip_restricted_equipment: 0
+
+// Stack Cash Shop Type items
+// Officially this items doesn't stack and unique_id is guaranteed for each
+// 1 : yes
+// 0 : no(official)
+cash_item_stack: 0
```
AnnieRuru: `cashbox_open_item_stackable`
this name makes more sense :o

src/map/battle.c
```diff
@@ -7153,6 +7153,7 @@ static const struct battle_data {
 	{ "feature.roulette",                   &battle_config.feature_roulette,                1,      0,      1,              },
 	{ "show_monster_hp_bar",                &battle_config.show_monster_hp_bar,             1,      0,      1,              },
 	{ "fix_warp_hit_delay_abuse",           &battle_config.fix_warp_hit_delay_abuse,        0,      0,      1,              },
+	{ "cash_item_stack",                    &battle_config.cash_item_stack,                 0,      0,      1,              },
```
AnnieRuru: move this line to directly below `unequip_restricted_equipment`
it will be easier to find

AnnieRuru:
2 problems I found
1. when I set to `yes`, to keep the old behavior
nope, it still same behavior as `no`

2. when I already have 100 units of item inside my inventory (MAX_INVENTORY)
I click to open the box,
the error message come out
`You can't have this item because you will exceed the weight limit`
which is correct, but I didn't receive the item, and the cash box disappear !

conf/battle/items.conf
hemagx: the option is to stack IT_CASH and not the items inside the IT_CASH

hemagx:
1. i tested it even again now, works fine, please remember you need to merge the old boxes after change the settings using the merger NPC

2. Please open issue for this, it's not part of this commit or PR

AnnieRuru:
1st one
oh LOL, I just `@reloadbattleconf` that's the problem
but still the same, it just like force serial the cash type 18 items
the old behavior include when open a box, it should stack, but doesn't

this line in script.c
`if( ( !itemdb->isstackable2(data) || data->flag.force_serial || data->type == IT_CASH || (originalitem && originalitem->type == IT_CASH) ) && !item_data->unique_id )`
missing the battle_config check

btw, refer to this issue https://github.com/HerculesWS/Hercules/issues/703

2nd one
since you already know, just make another PR =/

hemagx:
1- The ForceSerial behavior on item coming out of IT_CASH Item type doesn't need to be configured neither cancelled through battle config, you can always merge using the merger ( that why it exist... )
also the battle config setting is for stacking or not.

2- Yes i know, but i use the issues list assigned to me and other website to make my TODO list, i will remember this one, but maybe too late

hemagx:
Change to in progress, an improvement to logs in the merger is going to be added.

Smokexyz:
This is inactive since 2015, suggesting that it's not really in progress and also deprecated. Closing.

PR #908, opened 2015-12-02
Add EQP_ = Equipment Position constants

AnnieRuru:
the VAR_ constant added on this date
https://github.com/HerculesWS/Hercules/commit/f4278f36db32f35e535a0ea8feb1cf83ca95019e
then the LOOK_ constant added on this date
https://github.com/HerculesWS/Hercules/commit/a25c820fba6bd86fe2fb50e2b5b512295583522d

I compare the value, which both of them exist on the *setlook documentation,
with this script I made some times ago
https://github.com/AnnieRuru/customs/blob/master/scripts/Dressing%20Girl.txt
and found the one made by lone_wolf is wrong, brian is actually correct
so have to remove the one made by lone_wolf

also, the F_GetArmorType global function
https://github.com/HerculesWS/Hercules/blob/master/npc/other/Global_Functions.txt#L312
doesn't seem right, if the upper headgear cover mid, or lower, it will return default case
I wonder maybe should update it ? or better just remove it

PR #974, opened 2015-12-17
Added Support for BodyStyle

dastgir:


AnnieRuru:
I didn't see an entry for mob_avail
does it means we can't create a monster disguise as player using custom body palette ?

dastgir:
Woah, need to test that thing....

AnnieRuru:
also don't have `2015-12-17--15-58.sql` file
and the `index.txt`
https://github.com/HerculesWS/Hercules/tree/master/sql-files/upgrades

and, this commit will conflict with 1 of my pull request, for the LOOK_ constant #908 
we have to decide which one to go 1st :O

dastgir:
This PR should be merged after #978 is merged(entity packets)

dastgir:
About SQL file, I forgot to 'git add' :x

dastgir:
@AnnieRuru Added the SQL file.

this PR is done too...(need #978 merged first, so can edit some of the location in it)

dastgir:
Done @Jedzkie

src/map/status.h
```diff
@@ -1923,6 +1923,7 @@ enum scb_flag
 	SCB_RANGE   = 0x10000000,
 	SCB_REGEN   = 0x20000000,
 	SCB_DYE     = 0x40000000, // force cloth-dye change to 0 to avoid client crashes.
+	SCB_BODY    = 0x80000000, // Force bodysStyle change to 0
```
MishimaHaruna: This is where it gets dangerous. `0x80000000` no longer fits an int32 (but only an uint32). We currently have a mix of signed an unsigned integer for this. Considering it's a bitmask, we can turn it into an uint32 in all places where it's used (for example, `status_calc_*`), but it'll be soon time to change it to 64 bit, since the next value we add will go there.

dastgir: Actually, this was not needed currently, since no SC does this, so maybe its best to leave that commented with proper explanation?

dastgir:
Rebased and commented those @MishimaHaruna

src/map/atcommand.c
```diff
@@ -1704,6 +1704,32 @@ ACMD(model)
...
 /*==========================================
+ * @bodystyle [Rytech]
+ *------------------------------------------*/
+ACMD(bodystyle)
+{
+	int body_style = 0;
+
+	memset(atcmd_output, '\0', sizeof(atcmd_output));
+
+	if (!message || !*message || sscanf(message, "%d", &body_style) < 1) {
```
MishimaHaruna: `message` is guaranteed by contract to be non-NULL. You should remove the `!message` condition from here

src/map/atcommand.c
```diff
@@ -1704,6 +1704,32 @@ ACMD(model)
...
+		sprintf(atcmd_output, "Please, enter a body style (usage: @bodystyle <body ID: %d-%d>).", MIN_BODY_STYLE, MAX_BODY_STYLE);
+		clif->message(fd, atcmd_output);
+		return false;
+	}
+
+	if (body_style >= MIN_BODY_STYLE && body_style <= MAX_BODY_STYLE) {
+			pc->changelook(sd, LOOK_BODY2, body_style);
```
MishimaHaruna: Please reindent these two lines (it's 2 levels instead of 1)

src/map/status.c
```diff
@@ -9868,6 +9875,17 @@ int status_change_start(struct block_list *src, struct block_list *bl, enum sc_t
 		}
 		calc_flag&=~SCB_DYE;
 	}
+	/* Currently No SC's use this
```
MishimaHaruna: I'd prefer an `#if 0` / `#endif` to (temporarily) disable code, because it nests correctly

dastgir: Done

dastgir:
Edited and added #if 0

MishimaHaruna:
Merged, thank you!

PR #977, opened 2015-12-17
Dressroom Packet implemented

dastgir:


dastgir:
@hemagx @MishimaHaruna 
this one is small, could review it quickly.

PR #993, opened 2015-12-20
Mob DB support for the db2sql plugin

MishimaHaruna:
This adds support to rebuild the mob database SQL script through the db2sql plugin (I'll update the server-side infrastructure to automatically run it on each push, by the time this pull request is merged).

One notable difference in the way the db2sql plugin works after this pull request is that, while it'd previously regenerate both the renewal and the pre-renewal sql scripts, now it only regenerates the one for which the server is compiled (the item_db2 and mob_db2 are always regenerated). This change is necessary due to incompatibilities in how the mob database is processed. This change should not be a concern for most users, since it's normal to only edit the pre-renewal or the renewal (or the db2) version, and not both.

Included with this pull request there is a number of more or less unrelated commits that I'd like to be merged before or alongside the db2sql changes:
- Several floating point operations were replaced with integer operations. Floating point values suffer precision / rounding issues, and it is a bad idea to use them to avoid overflows in integer operations (unless it's in a place where the loss of precision is of no concern).
- The JName field was re-added, and is now optional. If not specified, it defaults to the Name.
- The Mob DB parsing function was partly rewritten to mimic the way the Item DB parser works. This was changed in order to make it possible for the db2sql plugin to operate without duplicating code.
- Element, Race and Size numerical values were replaced with their `Ele_*`, `RC_*`, `Size_*` constants counterparts.


jTynne:
Can we get a merge?

MishimaHaruna:
Branch is ready for code-review (but please notify me before merging, since it requires some changes to the API bot)

MishimaHaruna:
Rebased. Calling @HerculesWS/developers for review (When you think it's ready, please mark it as codereview:accepted instead of merging, because I need to edit the HerculesWS API bot the moment this is merged)

MishimaHaruna:
@HerculesWS/developers can I get a review of this?

PR #1000, opened 2015-12-21
define FAKE_NPC as -1 in const.txt

AnnieRuru:
is there a reason why FAKE_NPC is not constant ?
in the source code, there is a FAKE_NPC
https://github.com/HerculesWS/Hercules/blob/master/src/map/npc.h#L128
and we are allow to do -1 as sprite ID in npc script
but why not FAKE_NPC ?

`[Warning]: npc_parseview: Invalid NPC constant 'FAKE_NPC' specified in file 'npc/zzz.txt', line'1'. Defaulting to INVISIBLE_CLASS.`


MishimaHaruna:
There's absolutely no reason (other than I hadn't done it yet and/or forgot to)!

Thank you for the reminder (and the pull request), oneesama :)

After this, we can remove the special exception we have for the `-1` value, since it no longer has a reason to exist.

dastgir:
@AnnieRuru anything remaining?
this PR seems good so far

AnnieRuru:
yes, still one
wait Haru do that remove `-1` support from the source

MishimaHaruna:
I pushed two commits to your pull request, I hope you don't mind :)

MishimaHaruna:
Merged, thank you!

AnnieRuru:
haha, sure no problem, my branch already has permission to let you edit since long ago

PR #1002, opened 2015-12-21
Follow-up to fd78190a5ec4f19f10027f7a7ebe0e6ce023f672.

kisuka:


npc/mapflag/zone.txt
```diff
@@ -131,3 +131,19 @@ umbala	mapflag	zone	Towns
...
+te_prtcas04	mapflag	zone	WoE TE
+te_prtcas05	mapflag	zone	WoE TE
+te_alde_gld	mapflag	zone	WoE TE
+te_aldecas1	mapflag	zone	WoE TE
+te_aldecas2	mapflag	zone	WoE TE
+te_aldecas3	mapflag	zone	WoE TE
+te_aldecas4	mapflag	zone	WoE TE
+te_aldecas5	mapflag	zone	WoE TE
+teg_dun01	mapflag	zone	WoE TE
+teg_dun02	mapflag	zone	WoE TE
```
AnnieRuru: suffer the rathena script language symptom again ? XD
`[Warning]: npc_parse_mapflag: Invalid zone 'WoE TE'! removing flag from te_prt_gld in file 'npc/zzz.txt', line '1'.`
I believe we use `map_zone_db.conf`
https://github.com/HerculesWS/Hercules/blob/master/db/re/map_zone_db.conf#L116

AnnieRuru:
ops ! sry I comment without actual test
now I download the diff and test
hopefully my answer will be more accurate

`skill "ECL_PEONYMAMY",1;`
`skill "ECL_SADAGUI",1;`
`skill "ECL_SEQUOIADUST",1;`
use constants

all the `TE_` items ...
I remember I read somewhere that these items only give the bonus in TE_ castle right ?
but when I tried, it gives bonus when I am in prontera town

MishimaHaruna:
Merged 2a16362 (mostly because of the RC_All fix), anything else can be done as a further follow-up

kisuka:
@AnnieRuru yeah, the descriptions say that but the official files didn't even have the limits on it. So we have to fix it ourselves once we actually get the TE scripts added.

I purely was just adding these to get them out of the way for the upcoming scripts / actual update.

AnnieRuru:
> the descriptions say that but the official files didn't even have the limits on it. So we have to fix it ourselves once we actually get the TE scripts added.

ok understood
I guess that is Aegis behavior, that is

PR #1003, opened 2015-12-22
Merge of #866 and #867

### LINKED ISSUE #865: long variable name cause server crash

MishimaHaruna:
This is a merge of the pull requests #866 and #867, with a change through the source, to change the maximum variable size from 31+1 to 32+1 (to hold variable names up to 32 character long). This is in order to avoid having to delete rows from the SQL tables.

Closes #865 as fixed, closes #866 as merged and closes #867 as merged.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -525,7 +525,8 @@ If a variable was never set, it is considered to equal zero for integer
 variables or an empty string ("", nothing between the quotes) for string 
 variables. Once you set it to that, the variable is as good as forgotten 
 forever, and no trace remains of it even if it was stored with character 
-or account data.
+or account data. The maximum length of variable name including prefix and
+suffix is 32.
 
 Some variables are special, that is, they are already defined for you by 
 the scripting engine. You can see the full list somewhere in 
--- a/src/char/inter.c
+++ b/src/char/inter.c
@@ -1186,7 +1186,7 @@ int mapif_parse_Registry(int fd)
 
 	if( count ) {
 		int cursor = 14, i;
-		char key[32], sval[254];
+		char key[SCRIPT_VARNAME_LENGTH+1], sval[254];
 		bool isLoginActive = sockt->session_is_active(chr->login_fd);
 
 		if( isLoginActive )
@@ -1194,8 +1194,9 @@ int mapif_parse_Registry(int fd)
 
 		for(i = 0; i < count; i++) {
 			unsigned int index;
-			safestrncpy(key, (char*)RFIFOP(fd, cursor + 1), RFIFOB(fd, cursor));
-			cursor += RFIFOB(fd, cursor) + 1;
+			int len = RFIFOB(fd, cursor);
+			safestrncpy(key, (char*)RFIFOP(fd, cursor + 1), min((int)sizeof(key), len));
+			cursor += len + 1;
 
 			index = RFIFOL(fd, cursor);
 			cursor += 4;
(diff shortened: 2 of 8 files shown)


AnnieRuru:
so, you are not going to change the sql files ... 
well then how about this ?
```
prontera,159,185,5	script	kjdsfkshfs	1_F_MARIA,{
	for ( .@i = 0; .@i < 25; ++.@i )
		.@tmp$ += "1234567890";
	.@tmp$ += "12345";
	dispbottom "len = "+ getstrlen( .@tmp$ );
	a1234$ = .@tmp$;
}
```
`[SQL]: DB error - Data too long for column 'value' at row 1
[Debug]: at d:\ragnarok\hercules\src\char\inter.c:604 - REPLACE INTO `char_reg_s
tr_db` (`char_id`,`key`,`index`,`value`) VALUES ('150000','a1234$','0','╠╠╠╠╠╠╠╠
╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠
╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠
╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠
╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠╠a1234$')`
I know its different issue but,
there is a different with `varchar` and `varchar binary` for SQL

MishimaHaruna:
Umm o.o that's a different issue, I didn't check the maximum length of the value, this pull request only handles long keys. I can do the same for values (and perhaps increase the limit) as well.

The difference between `varchar` and `varchar binary`, from the documentation I found, is just in how it handles the upper/lower case changing functions (they're not handled in binary), and how it handles leading and trailing whitespace, which aren't a concern here.

AnnieRuru:
ops, my bad, I misunderstood the behavior of `binary`, thx for clarification

erm, yes
[mapreg save in varchar(255)](https://github.com/HerculesWS/Hercules/blob/master/sql-files/main.sql#L591)
and the rest like [acc_reg_str_db](https://github.com/HerculesWS/Hercules/blob/master/sql-files/main.sql#L54) [char_reg_str_db](https://github.com/HerculesWS/Hercules/blob/master/sql-files/main.sql#L228) [global_acc_reg_str_db](https://github.com/HerculesWS/Hercules/blob/master/sql-files/main.sql#L324) are save in varchar(254)
yes its different issue

PR #1012, opened 2015-12-23
Add *swap script command

AnnieRuru:
https://github.com/HerculesWS/Hercules/blob/master/src/common/cbasetypes.h#L329
in the source, there is a swap command
but we don't have script command for it

so I make one XD

tested with all these values
```
prontera,155,187,5	script	test1	1_F_MARIA,{
	.@var1 = 111;
	.@var2 = 222;
	swap .@var1, .@var2;
	dispbottom "var1 = "+ .@var1;
	dispbottom "var2 = "+ .@var2;
}

prontera,157,187,5	script	test2	1_F_MARIA,{
	$var1[99] = 111;
	var2[33] = 222;
	swap $var1[99], var2[33];
	dispbottom "var1 = "+ $var1[99];
	dispbottom "var2 = "+ var2[33];
}

prontera,159,187,5	script	test3	1_F_MARIA,{
	detachrid;
	.@var1 = 111;
	.@var2 = 222;
	swap .@var1, .@var2;
	announce "var1 = "+ .@var1, bc_all;
	announce "var2 = "+ .@var2, bc_all;
}

prontera,161,187,5	script	test4	1_F_MARIA,{
	swap 32132, C_RED; // expect to throw error
}
```
EDIT: allow to swap string will be cool
```
prontera,155,185,5	script	test5	1_F_MARIA,{
	.@var$[0] = "abc";
	.@var$[1] = "xyz";
	swap .@var$[0], .@var$[1];
	dispbottom "var1 = "+ .@var$[0];
	dispbottom "var2 = "+ .@var$[1];
}
```

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -15710,7 +15710,7 @@ BUILDIN(md5)
 
 BUILDIN(swap)
 {
-	TBL_PC *sd = NULL;
+	struct map_session_data *sd = NULL;
 	struct script_data *data1, *data2;
 	const char *varname1, *varname2;
 	int64 uid1, uid2;
@@ -15718,9 +15718,19 @@ BUILDIN(swap)
 	data1 = script_getdata(st,2);
 	data2 = script_getdata(st,3);
 
-	if (!data_isreference(data1) || !data_isreference(data2)) {
+	if (!data_isreference(data1) || !data_isreference(data2) || reference_toconstant(data1) || reference_toconstant(data2)) {
 		st->state = END;
 		return true; // avoid print error message twice
+	}
+
+	if (reference_toparam(data1) || reference_toparam(data2)) {
+		ShowError("script:swap: detected parameter type constant.\n");
+		if (reference_toparam(data1))
+			script->reportdata(data1);
+		if (reference_toparam(data2))
+			script->reportdata(data2);
+		st->state = END;
+		return false;
 	}
 
 	varname1 = reference_getname(data1);
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -15708,6 +15708,66 @@ BUILDIN(md5)
 	return true;
 }
 
+BUILDIN(swap)
+{
+	TBL_PC *sd = NULL;
+	struct script_data *data1, *data2;
+	const char *varname1, *varname2;
+	int64 uid1, uid2;
+
+	data1 = script_getdata(st,2);
+	data2 = script_getdata(st,3);
+
+	if (!data_isreference(data1) || !data_isreference(data2)) {
+		st->state = END;
+		return true; // avoid print error message twice
+	}
+
+	varname1 = reference_getname(data1);
+	varname2 = reference_getname(data2);
+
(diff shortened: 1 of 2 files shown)


src/map/script.c
```diff
@@ -15708,6 +15708,66 @@ BUILDIN(md5)
...
+{
+	TBL_PC *sd = NULL;
+	struct script_data *data1, *data2;
+	const char *varname1, *varname2;
+	int64 uid1, uid2;
+
+	data1 = script_getdata(st,2);
+	data2 = script_getdata(st,3);
+
+	if (!data_isreference(data1) || !data_isreference(data2)) {
```
MishimaHaruna: It's a good idea to add ` || reference_toconstant(data1) || reference_toconstant(data2)` to the check (even if it appears to be useless at the moment) - see `BUILDIN(__setr)` for comparison

src/map/script.c
```diff
@@ -15708,6 +15708,66 @@ BUILDIN(md5)
...
+	else {
+		int value1, value2;
+
+		value1 = script_getnum(st,2);
+		value2 = script_getnum(st,3);
+
+		if (value1 != value2) {
+			script->set_reg(st, sd, uid1, varname1, (void*)h64BPTRSIZE(value2), script_getref(st,3));
+			script->set_reg(st, sd, uid2, varname2, (void*)h64BPTRSIZE(value1), script_getref(st,2));
+		}
```
AnnieRuru: actually, I thought about changing this part to use only 1 temporary variable, -> value
```
	if ( script_getnum(st,2) != script_getnum(st,3) ) {
		int value = script_getnum(st,2);
		script->set_reg(st, sd, uid1, varname1, (void*)h64BPTRSIZE(script_getnum(st,3)), script_getref(st,3));
		script->set_reg(st, sd, uid2, varname2, (void*)h64BPTRSIZE(value), script_getref(st,2));
	}
```
but I wonder using 2 temporary variable better ? which has better readability ?
or just 1 variable, but run more functions

src/map/script.c
AnnieRuru: after I tested with `report_data` script command,
`data_isreference` actually return 0 on both `reference_getconstant` or `data_isint`
which is, yeah, useless

but I'll add it just for additional check, no harm

src/map/script.c
MishimaHaruna: I think the current way is more readable... Using a few more variables is fine, if the purpose is to improve readability

AnnieRuru:
After I re-read my `report_data` script command
it seems the `reference_toparam` is nested inside `data_isreference`

tested just now
```
prontera,161,187,5	script	test4	1_F_MARIA,{
	swap BaseLevel, JobLevel; // expect to throw error
}
```
was expected to throw error, but it wasn't
my base level 60 and job level1, became base level 1, job level 60 LOL

so I added additional check

MishimaHaruna:
Oh, wow, nice catch about the parameter thing - I didn't think of that.

Okay, I'll be merging this, seems all good now

MishimaHaruna:
Merged, thank you!

PR #1022, opened 2015-12-26
Add *setgroupid script command.

Emistry:
Useful for scripts that temporary adjust the group id of players.

## Example 
```
prontera,155,181,5	script	setgroupid	757,{
	mes "Input new Group ID";
	input .@groupid,0,99;

	dispbottom "Current Group ID : "+getgroupid();
	.@ouput = setgroupid( .@groupid );
	// .@ouput = setgroupid( .@groupid,"name" );
	// .@ouput = setgroupid( .@groupid,2000000 );
	// .@ouput = setgroupid( .@groupid,strcharinfo(0) );
	// .@ouput = setgroupid( .@groupid,getcharid(3) );
	dispbottom "> Update : "+ ( .@ouput ? "SUCCESS":"FAILED" );
	dispbottom "Latest Group ID : "+getgroupid();
	close;
}
```

## Result
Image of Yaktocat

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -9068,6 +9068,30 @@ BUILDIN(getgmlevel)
 	return true;
 }
 
+/// set the group ID of the player.
+/// setgroupid(<new group id>{,"<character name>"|<account id>})
+/// return 1 on success, 0 if failed.
+BUILDIN(setgroupid) {
+	struct map_session_data* sd = NULL;
+	int new_group = script_getnum(st, 2);
+
+	if (script_hasdata(st, 3)) {
+		if (script_isstringtype(st, 3))
+			sd = script->nick2sd(st, script_getstr(st, 3));
+		else
+			sd = script->id2sd(st, script_getnum(st, 3));
+	}
+	else
+		sd = script->rid2sd(st);
+
+	if (sd == NULL)
+		return true; // no player attached, report source
+
+	script_pushint(st, !pc->set_group(sd, new_group));
+
+	return true;
+}
+
 /// Returns the group ID of the player.
 ///
 /// getgroupid() -> <int>
@@ -20253,7 +20277,8 @@ void script_parse_builtin(void) {
 		BUILDIN_DEF(getskilllv,"v"),
 		BUILDIN_DEF(getgdskilllv,"iv"),
 		BUILDIN_DEF(basicskillcheck,""),
-		BUILDIN_DEF(getgmlevel,""),
+		BUILDIN_DEF(getgmlevel, ""),
+		BUILDIN_DEF(setgroupid, "i?"),
 		BUILDIN_DEF(getgroupid,""),
 		BUILDIN_DEF(end,""),
 		BUILDIN_DEF(checkoption,"i"),
--- a/conf/battle/player.conf
+++ b/conf/battle/player.conf
@@ -120,6 +120,10 @@ show_hp_sp_drain: no
 // Display the gained hp/sp values from killing mobs? (Ie: Sky Deleter Card)
 show_hp_sp_gain: yes
 
+// Show the critical bonus for katar class weapon in player status window?
+// On official server, the critical bonus from katar class weapon isn't display. (Default: no)
+show_katar_crit_bonus: no
+
 // If set, when A accepts B as a friend, B will also be added to A's friend
 // list, otherwise, only A appears in B's friend list.
 // NOTE: this setting only enables friend auto-adding; auto-deletion does not work yet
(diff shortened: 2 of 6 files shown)


src/map/script.c
```diff
@@ -9013,6 +9013,33 @@ BUILDIN(getgmlevel)
...
 }
 
+/// set the group ID of the player.
+/// setgroupid( <new group id>,{<account ID> or <Player Name>} )
+/// return 1 on success, 0 if failed.
+BUILDIN(setgroupid) {
+	TBL_PC* sd;
+	int new_group = 0;
+
+	if (script_hasdata(st, 2))
```
dastgir: No need of hasdata, since 2nd parameter is compulsory

src/map/script.c
```diff
@@ -9013,6 +9013,32 @@ BUILDIN(getgmlevel)
...
+	TBL_PC* sd;
+	int new_group = 0;
+
+	new_group = script_getnum(st, 2);
+
+	if (script_hasdata(st, 3)) {
+		if (script_isstringtype(st, 3))
+			sd = map->nick2sd(script_getstr(st, 3));
+		else
+			sd = map->id2sd(script_getnum(st, 3));
```
AnnieRuru: use the new `script->nick2sd` and `script->id2sd` function
https://github.com/HerculesWS/Hercules/pull/950

src/map/script.c
```diff
@@ -9013,6 +9013,32 @@ BUILDIN(getgmlevel)
...
+	}
+	else
+		sd = script->rid2sd(st);
+
+	if (sd == NULL)
+		return false; // no player attached, report source
```
AnnieRuru: `if (sd == NULL) return true;`
since the script->rid2sd already has `script->reportsrc(st);` function

src/map/script.c
```diff
@@ -9013,6 +9013,32 @@ BUILDIN(getgmlevel)
 	return true;
 }
 
+/// set the group ID of the player.
+/// setgroupid( <new group id>,{<account ID> or <Player Name>} )
+/// return 1 on success, 0 if failed.
+BUILDIN(setgroupid) {
+	TBL_PC* sd;
```
AnnieRuru: `TBL_PC *sd == NULL;`
yes, the standard for pull request is strict ...
I also prefer to write plugins rather than pull request ...

Emistry: erm...
```
TBL_PC *sd == NULL;
```
i check the other script commands but most of it doesn't initialize it to `NULL`.

hemagx: please use struct map_session_data *sd = NULL;
and not == btw, anyway, usage of TBL_PC is not allowed out of BL_CAST

src/map/script.c
Emistry: lol ..weird .. probably my branch missed some update ==''
the commit from #950 isn't inside my branch. lol

src/map/script.c
hemagx: a NULL check should always be there, if a player not attached the script will cause server crash.

hemagx:
seem good to merge after fix the little comments, please squash it too after finishing.

src/map/script.c
AnnieRuru: like I said before, `return false;` just print the source location of the npc name
> since the script->rid2sd already has script->reportsrc(st); function

`return true;` means end of the function without displaying an error
`return false;` means end of the function + display the location and name of the source npc

hemagx: ah i thought you meant remove the check, nvm then.

src/map/script.c
AnnieRuru: oops, yes, its `TBL_PC *sd = NULL;`, thx

yes its possible to use `TBL_PC` in script.c file because it calls from map.h
https://github.com/HerculesWS/Hercules/blob/22d3fac0c9a9c39a4f3b544867ba1de4d8d2ecaa/src/map/map.h#L838

hemagx: possible to use doesn't mean we should use it, Hercules is under Unix Kernel syntax write rules, which doesn't allow this :)

AnnieRuru: https://github.com/HerculesWS/Hercules/commit/8382ee1cea443ac62a16c21c750423f3a2a624af
oops ... disregard about this
but when Haru review my commit, he didn't change my `TBL_PC` into `struct map_session_data`

hemagx: probably because he did not release the documents about it yet :)
but the more PRs/Commits follow it, the less work for the next commits where all syntax fix happens.

AnnieRuru:
just in case Emistry close this pull request to update his branch ... don't have to do that
what I did was,
1. download this pull request
https://github.com/HerculesWS/Hercules/pull/1022.diff
2. delete the branch in your computer
3. switch to main branch
4. pull latest git
5. reapply the that diff
6. commit to the same branch name
7. push commit with force
Under Options:
`Force: May discard     [x] Known changes      [ ] Unknown changes`
check the `Known changes` box

Emistry:
lol ... i guess i have to delete the branch and recreate it... messed up. :(

AnnieRuru:
LOL
you almost got it though

PR #1024, opened 2015-12-26
Change all TBL_* to actual struct to follow unix kernel syntax

hemagx:
Change all TBL_* to actual struct to follow unix kernel syntax

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -640,7 +640,7 @@ ACMD(who) {
 	StrBuf->Init(&buf);
 
 	iter = mapit_getallusers();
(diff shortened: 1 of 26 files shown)


dastgir:
So for what purpose is TBL_PC?

AnnieRuru:
I'm sensing my plugins get broken if this is merge ...
once the hercules bot update HPM_HookingMap, that is ...

hemagx:
@dastgir
it's for BL_CAST macro and should never used anywhere

@AnnieRuru 
no it wont get broken

dastgir:
@Annieruru
TBL_PC is just typedef
(You still can use TBL_PC)
It's like TBL_PC = struct map_session_data,
And TBL_PC won't be removed..
So none of your plugins will break

PR #1025, opened 2015-12-26
Added support for MC_CARTDECORATE skill.

Jedzkie:
This skill is only supported by 2015-08-05aRagexeRE clients onwards.


dastgir:
+1

src/map/clif.c
```diff
@@ -18572,6 +18589,25 @@ void clif_dressroom_open(struct map_session_data *sd, int view)
 	WFIFOSET(fd,packet_len(0xa02));
 }
 
+void clif_cartdeco(struct map_session_data *sd)
+{
+#if PACKETVER >= 20150805
+	int i = 0, cart_deco = 3, fd;
```
MishimaHaruna: Could you explain the purpose of the `cart_deco` variable? (it doesn't look like it can be changed anywhere, is it always 3?)

Jedzkie: Yes, Official servers only have 3 kinds of carts in cart deco skill.

db/pre-re/skill_db.txt
```diff
@@ -1024,7 +1024,7 @@
 //2541,0,0,0,0,0,0,??,0,no,0,0,0,none,0,	NPC_UGLYDANCE,Ugly Dance 2
 //2542,0,0,0,0,0,0,??,0,no,0,0,0,none,0,	ALL_TETANY,Tetany
 //2543,0,0,0,0,0,0,??,0,no,0,0,0,none,0,	ALL_RAY_OF_PROTECTION,Ray of Protection
-//2544,0,0,0,0,0,0,??,0,no,0,0,0,none,0,		MC_CARTDECORATE,Decorate Cart
+2544,0,6,4,0,0x1,0,1,1,no,0,0x1,0,none,0,		MC_CARTDECORATE,Cart Decorate 2
```
MishimaHaruna: Why the 2 here?

Jedzkie: Already change it, in kRO, they named it as Change Cart 2.

MishimaHaruna:
Merged, thanks!

PR #1029, opened 2015-12-27
changed some condition checks on storage.c

dastgir:


src/map/storage.c
```diff
@@ -269,10 +269,9 @@ int storage_storageget(struct map_session_data* sd, int index, int amount)
 	if( amount < 1 || amount > sd->status.storage.items[index].amount )
 		return 0;
 
-	if( (flag = pc->additem(sd,&sd->status.storage.items[index],amount,LOG_TYPE_STORAGE)) == 0 )
-		storage->delitem(sd,index,amount);
-	else
-		clif->additem(sd,0,0,flag);
+	if ( storage->delitem(sd,index,amount) == 0 )
```
MishimaHaruna: If you delete before knowing whether there is enough room in the inventory, the item will be lost, no?

dastgir: hmm.. need to do some good checks then, 
else if item is added and from storage its not deleted, it would be exploit too

MishimaHaruna: We have a `pc_checkadditem` function that could be used here. I'm not sure it's complete, but it claims to be doing exactly what we need here

dastgir: It doesn't exactly do what we need here

PR #1030, opened 2015-12-27
Move declaration of login config struct to login interface

hemagx:
Move declaration of login config struct to login interface


MishimaHaruna:
Merged, thanks!

PR #1031, opened 2015-12-27
Added OneClickIdentify packet

dastgir:


dastgir:
@hemagx 
@MishimaHaruna

src/map/clif.c
```diff
@@ -11416,20 +11416,39 @@ void clif_parse_ItemIdentify(int fd,struct map_session_data *sd) __attribute__((
 ///     -1 = cancel
 void clif_parse_ItemIdentify(int fd,struct map_session_data *sd)
 {
-	short idx = RFIFOW(fd,2);
+	short idx = RFIFOW(fd, 2) - 2;
 
 	if (sd->menuskill_id != MC_IDENTIFY)
 		return;
-	if( idx == -1 ) {// cancel pressed
+	if( idx < 0 ) {// cancel pressed
```
MishimaHaruna: Just to make sure, what value does it send when pressing cancel?

Before this commit, cancel was when `RFIFOW(fd, 2) == -1`. After this commit, it is when `RFIFOW(fd, 2) == 1` (because you moved the -2 to line 11419)

dastgir: Will move it to different commit later

dastgir:
Fixed @MishimaHaruna

src/map/clif.c
```diff
@@ -11484,6 +11484,25 @@ void clif_parse_ItemIdentify(int fd,struct map_session_data *sd)
...
 
+/// Identifying item with right-click (CZ_REQ_ONECLICK_ITEMIDENTIFY).
+/// 0A35 <index>.W
+void clif_parse_OneClick_ItemIdentify(int fd,struct map_session_data *sd)
+{
+	int cmd = RFIFOW(fd,0);
+	short idx = RFIFOW(fd, packet_db[cmd].pos[0]) - 2;
+	int n;
+
+	if ( idx < 0 )
```
MishimaHaruna: I'd add one more sanity check here: if the packet is altered so that the idx value is greater than the inventory size, we have an out-of-bounds array access a few lines below.

src/map/clif.c
```diff
@@ -11484,6 +11484,25 @@ void clif_parse_ItemIdentify(int fd,struct map_session_data *sd)
...
+		return;
+	
+	if ( sd->status.inventory[idx].identify )
```
MishimaHaruna: Also make sure that there's an item in that inventory slot

PR #1032, opened 2015-12-28
ItemDB Update

Jedzkie:
- Added official effect for Thai Perfume CAST, Thai Perfume ASPD, Thai Perfume ATK and Thai Perfume MATK items.


src/map/status.c
```diff
@@ -5953,8 +5967,10 @@ short status_calc_aspd_rate(struct block_list *bl, struct status_change *sc, int
 		aspd_rate += sc->data[SC_BATTLESCROLL]->val1 * 10;
 	if (sc->data[SC_STEAMPACK])
 		aspd_rate += sc->data[SC_STEAMPACK]->val2 * 10;
+	if (sc->data[SC_SKF_ASPD])
+		aspd_rate += sc->data[SC_SKF_ASPD]->val1 * 10;
```
dastgir: + means lower the aspd
- means increase the aspd
What's the behavior of this status?
I guess it should increase?

hemagx:
the commit does include many syntax fixing, this is required indeed however it make it hard to be able to read the actual changes.
please separate syntax fixing in other PRs

Jedzkie:
@hemagx okay, in my next pr.

src/map/status.c
AnnieRuru: yes
that also means `SC_BATTLESCROLL` should be `-=` as well

PR #1040, opened 2015-12-29
Balance Patch

dastgir:
Thanks to @Michieru 

Skill and Status Updates

@Michieru  could you review if I have included all the patch and is correct?


Jedzkie:
HOLY MOLY <3

src/map/skill.c
```diff
@@ -5317,23 +5337,25 @@ int skill_castend_nodamage_id(struct block_list *src, struct block_list *bl, uin
...
-						continue;
-					if(range && (psd->bl.x < x0 || psd->bl.y < y0 ||
-						psd->bl.x > x1 || psd->bl.y > y1))
-						continue;
-					if(!skill->check_unit_range(bl,psd->bl.x,psd->bl.y,skill_id,skill_lv))
-						ret |= skill->castend_pos2(bl,psd->bl.x,psd->bl.y,skill_id,skill_lv,tick,flag);
+				clif->skill_nodamage(src,bl,skill_id,skill_lv,1);
+ 				elemental->delete(sd->ed,0);
+				skill->unitsetting(src,MG_SAFETYWALL,skill_lv + 5,src->x,src->y,0);
+				skill->unitsetting(src,AL_PNEUMA,1,src->x,src->y,0);
```
AnisotropicDefixation: Wasn't this the bugged behavior on release? As far as I know the skill only casts SW now.

src/map/skill.c
```diff
@@ -9129,7 +9133,7 @@ int skill_castend_nodamage_id(struct block_list *src, struct block_list *bl, uin
 			break;
 
 		case MI_HARMONIZE:
-			clif->skill_nodamage(src,bl,skill_id,skill_lv,sc_start2(src,bl,type,100,skill_lv,(sd?pc->checkskill(sd,WM_LESSON):1),skill->get_time(skill_id,skill_lv)));
+			clif->skill_nodamage(src,bl,skill_id,skill_lv,sc_start(src,bl,type,100,skill_lv,skill->get_time(skill_id,skill_lv)));
```
AnisotropicDefixation: Should apply the same status on the caster as well.

src/map/status.c
```diff
@@ -7749,6 +7773,25 @@ int status_change_start(struct block_list *src, struct block_list *bl, enum sc_t
...
+			status_change_end(bl, SC_CURSE, INVALID_TIMER);
+			status_change_end(bl, SC_CONFUSION, INVALID_TIMER);
+			status_change_end(bl, SC__CHAOS, INVALID_TIMER);
+			status_change_end(bl, SC_ILLUSION, INVALID_TIMER);
+			status_change_end(bl, SC_SILENCE, INVALID_TIMER);
+			status_change_end(bl, SC_BURNING, INVALID_TIMER);
+			status_change_end(bl, SC_COLD, INVALID_TIMER);
+			status_change_end(bl, SC_DEEP_SLEEP, INVALID_TIMER);
+			status_change_end(bl, SC_FEAR, INVALID_TIMER);
+			status_change_end(bl, SC_MANDRAGORA, INVALID_TIMER);
```
AnisotropicDefixation: Apparently also removes masquerades now, need a confirmation.

src/map/status.c
```diff
@@ -8807,29 +8856,29 @@ int status_change_start(struct block_list *src, struct block_list *bl, enum sc_t
...
-				val4 = tick / 4000;
 				tick_time = 4000; // [GodLesZ] tick time
+				val4 = tick / tick_time;
 				break;
 			case SC_PYREXIA:
-				status->change_start(src, bl,SC_BLIND,10000,val1,0,0,0,30000,SCFLAG_NOAVOID|SCFLAG_FIXEDTICK|SCFLAG_FIXEDRATE); // Blind status that last for 30 seconds
-				val4 = tick / 3000;
 				tick_time = 3000; // [GodLesZ] tick time
+				val4 = tick / tick_time;
+				status->change_start(src, bl,SC_BLIND,10000,val1,0,0,0,30000,SCFLAG_NOAVOID|SCFLAG_FIXEDTICK|SCFLAG_FIXEDRATE); // Blind status that last for 30 seconds
```
AnisotropicDefixation: This still needs work, the blind and hallucination statuses should be unremovable, hallucination it causes isn't normal one that flips the screen but only changes damage numbers. There's also some strange additional effect (taken from iROwiki): "If the affected entity takes damage while under the effects of this poison, it will remain in flinching motion, which will interrupt the skill casting".

PR #1043, opened 2015-12-29
TraderShop

dastgir:
Saving amount and ID's in array


AnnieRuru:
since the `npc_buylist_sub` also doing this same operation, why not having it as its own function ?

also, you forgotten to tell it at script_commands.txt and npc_trader_sample.txt to reflect this change

HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1043)

dastgirp:
@mekolat @4144

src/map/npc.c
```diff
@@ -1761,7 +1762,22 @@ static bool npc_trader_pay(struct npc_data *nd, struct map_session_data *sd, int
...
+		// Clear the Array
+		script->cleararray_pc(sd, "@bought_nameid", (void*)0);
+		script->cleararray_pc(sd, "@bought_quantity", (void*)0);
+	
+		// save list of bought items
+		for (i = 0; i < VECTOR_LENGTH(*item_list); i++) {
+			struct itemlist_entry *entry = &VECTOR_INDEX(*item_list, i);
+			intptr_t nameid = entry->id;
+			intptr_t amount = entry->amount;
+			script->setarray_pc(sd, "@bought_nameid", i, (void *)nameid, &key_nameid);
```
mekolat: I'm not a fan of char-bound variables, we should rather run the script and inject the variables as scope variables. I'll have to rewrite #1821 though, because I'm not satisfied with the way I implemented it 😕

PR #1044, opened 2015-12-29
Enchantment Stones Update:

Jedzkie:
- In Official, when you use a Enchantment stone, you can compound them in all of the General items that have slots.


dastgir:
could be split into 2.
1: Changes Loc: of all enchantments
2; Added new enchantments

Jedzkie:
Done. @dastgir @AnnieRuru please do review this, thanks!

Issue #140, opened 2013-09-14
Bug #7418 / Rental End


shennetsind:
Regarding http://hercules.ws/board/tracker/issue-7418-setmount-boarding-hatler/
I'd like to put up a discussion on how to best handle this.
I find Xgear's proposal (within the report) to not be flexible, while on this particular case the problem would be solved other rentals might not deal exclusively with stauses, if possible I'd rather use a more circumspect alternative, for example (I'm not really infatuated about this particular one but alas):
Items be given a source-code-controlled flag, via for example a script command, to sign whether it should run something upon rental ending, for example, just as we have a file for cash functions:

```
npc/other/RentalFunctions.txt
```

```
OnInit:
rentalevent(Boarding_Halter); //horrible function name.
end;

OnRentalEnd12622://run to attached player
sc_end SC_ALL_RIDING;
end;
```

Or:

```
OnRentalExpire://yet another dreadful name. not a fan of triggering the event for any item expiral, flagging would be most efficient
//attached to player; has a .@itemid variable created storing what item triggered it.
if( .@itemid == Boarding_Halter )
sc_end SC_ALL_RIDING;
end;
```

PR #1046, opened 2015-12-29
Added 'OnRentalExpireEvent' trigger.

Emistry:
Enable users to create custom events to control the expired items.
Example: return cards/enchantment that compounded within rental items.
## Example NPC Script:

https://pastebin.com/RtxifmSv
capture
## Screenshot

capture
I didnt save the value of `amount` because I believe rental items are removed one by one.

Extra : https://github.com/HerculesWS/Hercules/issues/140

Since I have no idea how to limit the server to only load only one event label, so I have to force the script to clear the variables just in case they loaded more than one OnRentalExpireEvent which might caused exploitations depend on how they setup it.

### DIFF
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -472,6 +472,7 @@ int pc_inventory_rental_clear(struct map_session_data *sd)
 /* assumes i is valid (from default areas where it is called, it is) */
 void pc_rental_expire(struct map_session_data *sd, int i) {
 	short nameid = sd->status.inventory[i].nameid;
+	int j;
 
 	/* Soon to be dropped, we got plans to integrate it with item db */
 	switch( nameid ) {
@@ -533,6 +534,17 @@ void pc_rental_expire(struct map_session_data *sd, int i) {
 			}
 			break;
 	}
+
+	pc->setreg(sd, script->add_str("@expire_itemid"), nameid);
+	pc->setreg(sd, script->add_str("@expire_refine"), sd->status.inventory[i].refine);
+	pc->setreg(sd, script->add_str("@expire_identify"), sd->status.inventory[i].identify);
+	pc->setreg(sd, script->add_str("@expire_attribute"), sd->status.inventory[i].attribute);
+	pc->setreg(sd, script->add_str("@expire_bound"), sd->status.inventory[i].bound);
+	// pc->setreg(sd, script->add_str("@expire_uniqueid"), sd->status.inventory[i].unique_id);
+	for (j = 0; j < MAX_SLOTS; j++) {
+		pc->setreg(sd, reference_uid(script->add_str("@expire_card"), j), sd->status.inventory[i].card[j]);
+	}
+	npc->script_event(sd, NPCE_RENTALEXPIRE);
 
 	clif->rental_expired(sd->fd, i, sd->status.inventory[i].nameid);
 	pc->delitem(sd, i, sd->status.inventory[i].amount, 0, DELITEM_NORMAL, LOG_TYPE_OTHER);
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -1102,6 +1102,41 @@ field for cash windows. It is used to remove items that are set as currency.
 Should be used along with OnCountFunds, see /doc/sample/npc_trader_sample.txt
 for more information.
 
+OnRentalExpireEvent:
+
+This special label is triggered when a rental item expired. It also return a set of 
+temporary variable that stored the information of expired item.
+Example :
+	
+	-	script	rental_expired	-1,{
+		
+		OnRentalExpireEvent:
+			// do whatever you want here when item expired.
+			dispbottom " ---- Item Expired ---- ";
+			dispbottom " > @expire_itemid = "+@expire_itemid;
+			dispbottom " > @expire_refine = "+@expire_refine;
+			dispbottom " > @expire_identify = "+@expire_identify;
+			dispbottom " > @expire_attribute = "+@expire_attribute;
+			dispbottom " > @expire_bound = "+@expire_bound;
+			dispbottom " > @expire_uniqueid = "+@expire_uniqueid;
+			dispbottom " > @expire_card[0] = "+getitemname( @expire_card[0] );
+			dispbottom " > @expire_card[1] = "+getitemname( @expire_card[1] );
+			dispbottom " > @expire_card[2] = "+getitemname( @expire_card[2] );
+			dispbottom " > @expire_card[3] = "+getitemname( @expire_card[3] );
+			
+			// must clear variables to avoid possible exploit !!
+			// especially with multiple NPCs running same label
+			@expire_itemid = 0;
+			@expire_refine = 0;
+			@expire_identify = 0;
+			@expire_attribute = 0;
+			@expire_bound = 0;
+			@expire_uniqueid = 0;
+			deletearray @expire_card;
+			end;
+	}
+
+
 On<label name>:
 
 These special labels are used with Mob scripts mostly, and script commands 
--- a/src/map/npc.c
+++ b/src/map/npc.c
@@ -4542,14 +4542,15 @@ void npc_read_event_script(void)
 		char *name;
 		const char *event_name;
 	} config[] = {
-		{"Login Event",script->config.login_event_name},
-		{"Logout Event",script->config.logout_event_name},
-		{"Load Map Event",script->config.loadmap_event_name},
-		{"Base LV Up Event",script->config.baselvup_event_name},
-		{"Job LV Up Event",script->config.joblvup_event_name},
-		{"Die Event",script->config.die_event_name},
-		{"Kill PC Event",script->config.kill_pc_event_name},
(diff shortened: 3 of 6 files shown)


AnnieRuru:
no, we don't want anymore OnPCxxxxEvent stuffs in the repo
every one of these will take off more memory load on the server

its better to leave this kind of stuffs as a plugin
just like OnPCStatCalcEvent https://github.com/HerculesWS/Hercules/pull/351

src/map/pc.c
```diff
@@ -534,6 +535,17 @@ void pc_rental_expire(struct map_session_data *sd, int i) {
 			break;
 	}
 
+	pc->setreg(sd, script->add_str("@expire_itemid"), nameid);
+	pc->setreg(sd, script->add_str("@expire_refine"), sd->status.inventory[i].refine);
+	pc->setreg(sd, script->add_str("@expire_identify"), sd->status.inventory[i].identify);
+	pc->setreg(sd, script->add_str("@expire_attribute"), sd->status.inventory[i].attribute);
+	pc->setreg(sd, script->add_str("@expire_bound"), sd->status.inventory[i].bound);
+	// pc->setreg(sd, script->add_str("@expire_uniqueid"), sd->status.inventory[i].unique_id);
```
AnnieRuru: `pc->setreg(sd, script->add_str("@expire_uniqueid"), (int)sd->status.inventory[i].unique_id);`
just add `(int)` can convert uint64 into int
but yes, will data lost if it overflow the value

hemagx: It doesn't convert it's like lying on the compiler and force it to compile.
The lowest unique number is already higher than int 32

AnnieRuru: I always feel its weird
why it want to use unique_id something like `644253684334719` instead of start from `1` ?
that number is found from my `inventory` table
and the number progress linearly , 1 by 1 with each equipment created ...

MishimaHaruna: The reason why it doesn't start from 1 is that it also encodes the character ID of the current character at the time of creation of the item. It was implemented to solve an issue where you could have the same unique ID generated twice, if two mapservers were running at the same time (the same character can't be connected to two different map servers at the same time).

```
return ((uint64)sd->status.char_id << 32) | sd->status.uniqueitem_counter++;
```

The upper 32 bits contain the character ID, while the lower 32 contain a sequential number starting from 1 (a separate counter for each character).

If you need to use those in a script (for example if you want to implement this in a plugin, or something), you could still split it in two variables, and store the two halves separately:

```
pc->setreg(sd, script->add_str("@expire_uniqueid_lo"), (int)(sd->status.inventory[i].unique_id&0xffffffff));
pc->setreg(sd, script->add_str("@expire_uniqueid_hi"), (int)((sd->status.inventory[i].unique_id>>32)&0xffffffff));
```

hemagx: We had to do it multiple map server safe, so the serial is based on formula
First four bytes is character id.
Second four bytes is a numeric number increased by one each time, however
when you read as one value you get this huge number

AnnieRuru: ahhh ... so its better to split into 2 variables
```
SELECT char_id, unique_id, unique_id >>32, unique_id & ((1<<32)-1) FROM inventory where unique_id <> 0;
SELECT char_id, unique_id, unique_id >> 32 & 0xffffffff, unique_id & 0xffffffff FROM inventory where unique_id <> 0;
```
I guess rathena implementation is off then
https://github.com/rathena/rathena/blob/67f8cb2/src/map/script.c#L8460
we could do better on these

AnnieRuru:
I just think up the best method
since hercules item_db.conf can add as many flag as possible,
I think its better to have another field something like this
```
{
	Id: 12622
	AegisName: "Boarding_Halter"
	Name: "Halter Lead"
	Type: 11
	Buy: 20
	Delay: 3000
	Trade: {
		nodrop: true
		notrade: true
		noselltonpc: true
		nocart: true
		nostorage: true
		nogstorage: true
		nomail: true
		noauction: true
	}
	Script: <" setcashmount(); ">
	OnRentalExpire: <" if (hascashmount()) setcashmount(); ">
},
```
like this we don't have to let it run any npc at all
and also make developer life easier without having to open another file

Emistry:
I agree that these event should only limit it to run only once. I was intend to make it only run once, but i have no idea to work on that, failed.

> no, we don't want anymore OnPCxxxxEvent stuffs in the repo
> every one of these will take off more memory load on the server

Well, if want to enforce doing this way, maybe all other OnPCxxxEvent should follow this also ?? revamp all of it ?

for the `item_db.conf` extra field `OnExpire: <" if (hascashmount()) setcashmount(); ">` I believe it's just the same with normal script. It's still running one instance per item I guess.

The idea I create the event was to enable renting of slotted items.
- You rent out slotted items
- Players compound their own card sets into the rental items.
- When expired, the rental items will be removed and they get their own card sets back.
This is one of the current limitation from rental items system. Long ago rAthena does planned to work on storing the cards id that compounded in the rental items, but it seem like alot of work to do.

At the moment, creating plugins is still out of my capabilities, and I am not in a favor to create a hardcoded event label within the source, just like how the `MOTD` did or eAmod security script. It just dont feel right.

AnnieRuru:
`OnPCStatCalcEvent` can't be avoided, it has to global
so the idea that every player on the server can have regen extra hp
*for those who doesn't know how to touch source code`
just do `OnPCStatCalcEvent: bonus bHPrecovRate,100;`

`OnPCUseSkillEvent`, honestly if skill_db.txt has turn into format like item_db.conf and mob_db.conf
I would have use this same format
and in case you wondering, my OnPCUseSkillEvent doesn't trigger on every skill
it only trigger from the skill in the list, and each skill has their own event label

`OnPCPickUpEvent`, I think I can convert to use this format
`OnPick: "npcname::Onlabel"`
going to code this later

`OnPCAttackEvent` and `OnPCIdleEvent` can become `queueopt_idle` and `queueopt_attack`
so these 2 labels wont trigger globally, only for those in the queue/event map
I want to do this very soon, as a plugin

and all these stuffs are custom, only available as plugin
_______________________________________________

yes, those OnPCKillEvent and OnPCDieEvent
I already started to convert all my custom scripts to use queue iterator script command
once I made `*mobattach/*getmasterid`, maybe can start to flag it as outdated and eventually remove

one of the most resource intensive label we have right now, is `OnNPCKillEvent`
which trigger on every monster that doesn't have event label
if we can work out a less resource intensive way similar to quest log system
maybe can remove that too,
but the current quest log system still has a lot limitation,
for example, still can't set up the same quest with randomize mob ID, like hunting mission script

the rest I think can keep it, like OnPCLoginEvent and OnPCLogoutEvent, which did the job fine

_______________________________________________

>for the item_db.conf extra field OnExpire: <" if (hascashmount()) setcashmount(); "> I believe it's just the same with normal script. It's still running one instance per item I guess.

I prefer the way `OnRentalExpire12622:` rather than yours, which run on every single item

______________________________________________

>The idea I create the event was to enable renting of slotted items.
>
>Long ago rAthena does planned to work on storing the cards id that compounded in the rental items, but it seem like alot of work to do.

I actually know the way how to solve this
just add `OnExpireReturnVar: true` flag ... or some similar names
then can return a set of variables that is similar to your defined variable
its for those who wants to get their cards back

______________________________________________

>At the moment, creating plugins is still out of my capabilities, and I am not in a favor to create a hardcoded event label within the source, just like how the MOTD did or eAmod security script. It just dont feel right.

it also doesn't mean everything you made can go inside repo ...
creating plugins, I think you should learn it,
almost every hercules server out there started to throw out patches to use plugins

honestly I also think MOTD should run in OnPCLoginEvent,
maybe the only 1 reason I can think of,
is to prevent MOTD collide with server's own OnPCLoginEvent
because MOTD always run before OnPCLoginEvent

Emistry:
> Comparing to OnPCUseSkillEvent or OnPCxxxEvent I do feel like it's less resources intensive.

Because I believe that `OnRentalExpire` occurence rate is much less compare to these, even if your `OnPCUseSkillEvent` is meant for only 1 skill. 

> I prefer the way OnRentalExpire12622: rather than yours, which run on every single item

I would rather have only one script with hundred of `cases` within switch-statement instead of hundred of event label within one script. 
Beside, I believe switch-statement could provide better readability.

> I actually know the way how to solve this
> just add OnExpireReturnVar: true flag ... or some similar names

Haha I like this idea too if it was implemented in this way.

> it also doesn't mean everything you made can go inside repo ...
Yes, I understand about this.

> honestly I also think MOTD should run in OnPCLoginEvent,
> maybe the only 1 reason I can think of,
> is to prevent MOTD collide with server's own OnPCLoginEvent
> because MOTD always run before OnPCLoginEvent

Well, I actually hope that MOTD is disabled or removed. Rarely seen any server still using the MOTD. The MOTD is just basically serve like a reminder telling that "Yes, you're using Hercules Emulator." and nothing else.

Smokexyz:
RIP 29 Dec 2015 - 10 April 2017.

HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1046)

PR #1049, opened 2015-12-30
Add new 'c' type variant to constant check on bonus script command

AnnieRuru:
related to http://herc.ws/board/topic/9332-warning-pc-bonusx-display-itemid/

there is a much simpler way to report this error
enforce the 1st field in the bonus script command for constant check
```
{
	Id: 1950
	AegisName: "Rope"
	Name: "Rope"
	Type: 4
	Buy: 2500
	Weight: 400
	Atk: 45
	Range: 2
	Slots: 3
	Job: 0x00080000
	Loc: 2
	WeaponLv: 1
	EquipLv: 3
	View: 14
	Script: <" bonus bAddMaxWaight, 1000; ">
},
```
will print error as
```
[Warning]: Unexpected type for argument 1. Expected constant.
[Debug]: Data: variable name='bAddMaxWaight' index=0
[Debug]: Function: bonus
[Warning]: pc_bonus: unknown type 0 1000 !
```
looks neat ? no, it isn't, there is a problem
I tried with bonus bMaxHP, for some stupid reason, it throw errors
_____________________________
so I made a plugin `report_data` script command
```
#include "common/hercules.h"
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include "map/script.h"
#include "common/HPMDataCheck.h"

HPExport struct hplugin_info pinfo = {
	"reportdata",
	SERVER_TYPE_MAP,
	">.<",
	HPM_VERSION,
};

BUILDIN(report_data) {
	if ( !script_hasdata(st,2) ) {
		ShowWarning( "script:report_data: a field is needed to run this command.\n" );
		return false;
	}
	else {
		struct script_data* data = script_getdata(st,2);
		if ( reference_toconstant(data) )
			ShowDebug( "constant -> %d", reference_getconstant(data) );
		if ( reference_toparam(data) )
			ShowDebug( "parameter -> %d", reference_getparamtype(data) );
		if ( reference_tovariable(data) )
			ShowDebug( "variable -> %s", reference_getname(data) );
		script->reportdata(data);
	}
	return true;
}

HPExport void plugin_init (void) {
	addScriptCommand( "report_data", "?", report_data);
}
```
check with this script
```
prontera,155,185,5	script	kjdhsfsjhf	1_F_MARIA,{
	report_data bMaxHP;
	report_data MaxHp;
	report_data bStr;
	report_data SC_BLESSING;
	dispbottom bMaxHP +"";
}
```
it print as
```
[Debug]: Data: number value=6
[Debug]: parameter -> 6[Debug]: Data: param name='MaxHp' type=6
[Debug]: constant -> 2[Debug]: Data: number value=13
[Debug]: constant -> 19[Debug]: Data: number value=30
```
What ??? bStr can be read as constant, but bMaxHP isn't  ??
I have no idea how to solve this issue

this pull request cannot be merge until someone find out why bMaxHP is being read as int
otherwise all server out there will getting this warning message

and this is on a main branch, anybody feels like doing this, go ahead


dastgir:
Maybe you have a variable named bMaxHP?
Check SQL table for it?

MishimaHaruna:
Before using `reference_toconstant()`/`reference_toparam()`/`reference_tovariable()`, you need to check that the data you have is a reference (`data_isreference()`).

The bad news is that `data_isreference()` returns false for the constants, since they're converted to int while parsing...

AnnieRuru:
ah ... I did that check on swap script command, but forgotten to do it here LOL

so some of the constant are being parse as C_INT for no apparent reason
but, bAddMaxWaight DEFINITELY parse as C_NAME

maybe that 'c' type check just add `data_isint(data)` check ...
hmm ... it looks like working ...

AnnieRuru:
yup, now its working
it still can't tell where the error come from, and it still detect at run-time
but it can tell the name of the typo
so you can just search the name of the typo with notepad++

the status change constants, however, I did like this before
```
prontera,155,185,5	script	kjdhsfsjhf	1_F_MARIA,{
	bonus bStr, 10;
	bonus bMaxHP, 10;
	bonus bAddMaxWeight, 10;
//	bonus bAddMaxWaight, 10;
	sc_start SC_BLESSING, 100, 1;
//	sc_start SC_BLESING, 100, 1;
	for ( .@i = 0; .@i < 10; ++.@i )
		sc_end .@i;
}
```
using loop with `.@i`, so must not enforce the 'c' rule on status change

I thought about using another 'b' rule or something
but it can stone the player -> SC_STONE = 0
so better hard-coded in these script commands

AnnieRuru:
yup that's all

the job constant, if enforce 'c' rule, our jobmaster script will broke
https://github.com/HerculesWS/Hercules/blob/master/npc/custom/jobmaster.txt#L158
and that thing can easily debug, very easy to find the error

only the bonus and special effect is very hard to find, and nobody use variables on them

MishimaHaruna:
How about we make this optional? Because I can see some cases where one would want to use variables. Since this is just a workaround until we improve the engine to be able to report those issues at parse time, I think it should at least be optional (even just an ifdef, as long as it's easy to disable)

AnnieRuru:
hmm ?
my 9 years scripting experience, I never use variables on bonus and special effect
the bonus only works in item script, so its fine
special effect ... I never do more than 3 types ...

yes this thing is kinda a work around ... until your new script engine ...
but I guess your new script engine still has very long ETA ...

so you mean I should release this as a source release instead ? LOL

ok the moment you mark code review rejected, I release on the forum =/

MishimaHaruna:
I'm closing this pull request, since I can't think of a non-hackish way to implement it.

The current code for type 'c' won't work as intended because it accesses `reference_to*()` without checking if `data_isreference()`, which seems to work, but just by chance (it uses the variable value as index in the string buffer, and chance is that the lookup fails, making it seem it's really working).

We'll fix this with other kinds of improvements to the script engine, I believe it's the best solution.

Issue #906, opened 2015-12-02
"Please don't forget me" skill affect self


AnnieRuru:
can somebody confirm this ?
DC_DONTFORGETME skill shouldn't affect self, but on hercules it can affect self

there are also some other problem like .. `@allskill` doesn't give all skills for dancer

kyeme:
but on hercules it can affect self <- because of Bard and Dancer Spirit ?

AnnieRuru:
on rathena, cast "Please don't forget me" will have 50% movement penalty
and when soul link, gain back 100% movement speed

on hercules, no matter soul link or not,
cast "Please don't forget me" will reduce movement speed in insane amount


Btw, now I'm unable to use Hercules test server, I just pull latest git
```
[Warning]: HPMDataCheck:HPMHooking: 'npc_chat_interface' size mismatch 1 != 36
[Fatal Error]: HPM:plugin_load: 'plugins/HPMHooking_map.dll' failed DataCheck, out of sync from the core (recompile plugin)!

Map-Server has terminated abnormally.
Restarting in 15 seconds, press Ctrl+C to cancel.
```

Jedzkie:
@AnnieRuru is this official behavior in renewal and pre-re? If yes, i can check this

AnnieRuru:
just setup a rathena test server and compare with hercules, I believe rathena is doing the right thing
well, just a hunch

and, I haven't debug the value yet

Jedzkie:
@AnnieRuru just tested the skill Dont Forget Me in Aegis, and we got equal behavior. Im not sure in Pre-renewal if the speed affects the caster.

PR #1056, opened 2016-01-03
Skill tree fix

MishimaHaruna:
Rewritten skill_tree parser in a more robust way

- Fixes an issue that prevented skills with more than 4 pre-requisites
  or more than 3 pre-requisites and a minimum level from being parsed
  correctly (and without any warning or error messages).
- Removes the limit on 5 pre-requisites (replaced a fixed size array
  with a VECTOR)
- Reduces memory usage of skill_tree from 794kB to 440kB (32 bit) or
  523kB (64 bit).
- Fixes an issue that prevented multiple inheritance from working
  correctly in rare cases (incorrect definition order), without any
  warning or error messages. Now a warning is displayed if a job is
  inherited before being defined.
- Fixes an issue that prevented skills inherited from being correctly
  merged with the skills defined for the current job.
- Prevents a job from inheriting itself by accident.
- Correctly detects skills defined twice for the same job.

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -2,7 +2,7 @@
  * This file is part of Hercules.
  * http://herc.ws - http://github.com/HerculesWS/Hercules
  *
- * Copyright (C) 2012-2015  Hercules Dev Team
+ * Copyright (C) 2012-2016  Hercules Dev Team
  * Copyright (C)  Athena Dev Teams
  *
  * Hercules is free software: you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


AnnieRuru:
looks ok to me
but the dancer `@allskill` still doesn't work, though

it still detect as bard and trying to give bard skills, but its dancer ...

MishimaHaruna:
Ah, I'll check that o.o

Does it have an open bugreport btw?

AnnieRuru:
only mention partially #906
>there are also some other problem like .. `@allskill` doesn't give all skills for dancer

MishimaHaruna:
Since I still can't reproduce the dancer issue, let's leave that for later :x

PR #1057, opened 2016-01-03
Add *makeitem2 script command.

Emistry:
- Added makeitem2 script command.
- Updated makeitem script command.
- Both makeitem/makeitem2 script command will return 1 on success 0 if fail.

## Sample NPC:
```
prontera,155,175,5	script	Sample#makeitem	1_M_04,{
	makeitem2 5353,2,"prontera",160,170,1,-1,1,4403,4403,4403,4403;
	end;
}
```


src/map/script.c
```diff
@@ -7437,52 +7437,136 @@ BUILDIN(grouprandomitem) {
...
 
-	if( script_isstringtype(st, 2) ) {
+	if (script_isstringtype(st, 2)) {
 		const char *name = script_getstr(st, 2);
 		struct item_data *item_data = itemdb->search_name(name);
 		if (item_data)
 			nameid = item_data->nameid;
 		else
 			nameid = UNKNOWN_ITEM_ID;
-	} else {
```
hemagx: this is correct syntax, please revert this line.

src/map/script.c
```diff
@@ -7437,52 +7437,136 @@ BUILDIN(grouprandomitem) {
...
-	memset(&item_tmp,0,sizeof(item_tmp));
+	memset(&item_tmp, 0, sizeof(item_tmp));
 	item_tmp.nameid = nameid;
-	item_tmp.identify=1;
+	item_tmp.identify = 1;
+
+	if (!itemdb->isstackable(nameid))
+		amount = 1;
+
+	script_pushint(st, map->addflooritem(NULL, &item_tmp, amount, m, x, y, 0, 0, 0, 0) > 0);
```
hemagx: value of true/false returned by this check should be 1/0 however still it's not safe value please change this line to something like (map->addflooritem(NULL, &item_tmp, amount, m, x, y, 0, 0, 0, 0) == 1)? 1 : 0

Emistry: hmm, isn't that is the same ??

src/map/script.c
```diff
@@ -7437,52 +7437,136 @@ BUILDIN(grouprandomitem) {
...
+	}
+	else
+		m = map->mapname2mapid(mapname);
+
+	if (m == -1) {
+		ShowError("makeitem2: non-existing map '%s'!\n", mapname);
+		return false;
+	}
+
+	struct item_data *item_data = itemdb->exists(nameid);
```
hemagx: C99 rules are not fully supported in MSVC, please move definition of this struct to the first of the function.

src/map/script.c
```diff
@@ -7437,16 +7437,15 @@ BUILDIN(grouprandomitem) {
...
 /*==========================================
- *
- *------------------------------------------*/
-BUILDIN(makeitem)
-{
-	int nameid,amount;
-	int x,y,m;
+* makeitem <item id or name>,<amount>,"<map name>",<X>,<Y>;
+*------------------------------------------*/
+BUILDIN(makeitem) {
```
dastgir: according to Coding Style Chapter 3(https://github.com/HerculesWS/Hercules/wiki/Coding-Style#chapter-3-placing-braces-and-spaces)
It should be
```
BUILDIN(makeitem)
{
```

src/map/script.c
```diff
@@ -7455,34 +7454,115 @@ BUILDIN(makeitem)
...
+		m = map->mapname2mapid(mapname);
+
+	if (m == -1) {
+		ShowError("makeitem2: non-existing map '%s'!\n", mapname);
+		return false;
+	}
+
+	item_data = itemdb->exists(nameid);
+	if (item_data == NULL) {
+		return -1;
```
dastgir: -1?
I think it should be false here?

src/map/script.c
```diff
@@ -7455,34 +7455,127 @@ BUILDIN(makeitem)
...
+	if (strcmp(mapname, "this") == 0) {
 		TBL_PC *sd;
 		sd = script->rid2sd(st);
-		if (!sd) return true; //Failed...
-		m=sd->bl.m;
-	} else
-		m=map->mapname2mapid(mapname);
+		if (sd == NULL) {
+			script_pushint(st, 0);
+			return false;
```
MishimaHaruna: if sd is NULL, it means that `script->rid2sd()` failed. rid2sd already reports the failure source (so you should return true here, if you don't want it to be reported twice). No need to push anything either, since `script->rid2sd()` already aborts the script on failure, so no return value will be examined.

src/map/script.c
```diff
@@ -7455,34 +7455,127 @@ BUILDIN(makeitem)
...
+	identify = script_getnum(st, 7);
+	refine = script_getnum(st, 8);
+	attribute = script_getnum(st, 9);
+	c1 = (short) script_getnum(st, 10);
+	c2 = (short) script_getnum(st, 11);
+	c3 = (short) script_getnum(st, 12);
+	c4 = (short) script_getnum(st, 13);
+
+	if (strcmp(mapname, "this") == 0) {
+		TBL_PC *sd;
```
MishimaHaruna: Usage of TBL_PC is now forbidden (sorry - it wasn't yet when the pull request was made). Could you please replace this with:

```
struct map_session_data *sd;
```

src/map/script.c
```diff
@@ -7455,34 +7455,127 @@ BUILDIN(makeitem)
...
+		sd = script->rid2sd(st);
+		if (sd == NULL) {
+			script_pushint(st, 0);
+			return false;
```
MishimaHaruna: Same as above (makeitem)

src/map/script.c
```diff
@@ -7475,16 +7476,102 @@ BUILDIN(makeitem)
...
+	y = script_getnum(st, 6);
+	identify = script_getnum(st, 7);
+	refine = script_getnum(st, 8);
+	attribute = script_getnum(st, 9);
+	c1 = (short) script_getnum(st, 10);
+	c2 = (short) script_getnum(st, 11);
+	c3 = (short) script_getnum(st, 12);
+	c4 = (short) script_getnum(st, 13);
+
+	if (strcmp(mapname, "this") == 0) {
```
AnnieRuru: I just know that `*makeitem` script command can use "this" flag

from my past scripting experience, `*makeitem` used in event script usually runs without a RID attached
so "this" flag just plain stupid
also, the `*monster` support "this" flag is because it also support the -1 coordinate

`monster "this", -1, -1, "--ja--", PORING, 1, "";`
this will summon a poring NEARBY the attached player, not on the player standing on

your implementation, which copy from `*makeitem`, doesn't support this -1 coordinate input

its either you remove the "this" flag (which you also didn't mention in the document)
or make it support for the -1 input to spawn items AROUND the player without using getmapxy

Emistry:
lol...it seem like i messed it up when try to rebase ==''

PR #1058, opened 2016-01-03
Fix *getitem2 and @item2 refine bug.

Emistry:
- Negative refine value isn't allowed.

## Example
Create a NPC that give item using getitem2 script command.
```
prontera,155,175,5	script	Sample#getitem2	1_M_04,{
	getitem2 5353,2,1,-1,1,4403,4403,4403,4403 );
	end;
}
```

## Result
capture


dastgir:
Thanks for Contribution

PR #1061, opened 2016-01-04
ItemDB Format Update

dastgir:
* Changes Job Field in item_db.conf
* It is backward compatible (Or should I remove backward Compatibility?)
* Changes Job field to long long int to support more values


hemagx:
Usage of long long int should be unsafe, I believe you should use uint64
instead

dastgir:
@hemagx
libconfig accepts long long int variable,
so take it as long long int and save to uint64?

hemagx:
Long long int can be smaller but not bigger than uint64

dastgir:
@hemagx ,
But it leaves incompatible data type warning

hemagx:
Let me go home and see

dastgir:
@hemagx , changed to uint64

Jedzkie:
@dastgir I see a little bug here, maybe you put specific Third Job Fields, on our current settings, the id#2480 (Black Wing Boots) cannot be worn by Hunter/Sniper Job even the level is 100+, in this new setting hunters can wear this if they reached 100+.

dastgir:
Strange @Jedzkie (I haven't touched upper code), I will try this now and let you know what is the result

dastgir:
@Jedzkie unable to reproduce

dastgir:
@MishimaHaruna

MishimaHaruna:
Merged via 7868637, thank you very much!

PR #1062, opened 2016-01-04
NPC Updates

Jedzkie:
All of these changes are compared to kRO (Baphomet Server).


npc/jobs/1-1e/taekwon.txt
```diff
@@ -32,10 +32,10 @@
 //================= Description ===========================================
 //= Official iRO TaeKwon Job Change Quest
 //================= Current Version =======================================
-//= 1.4
+//= 1.5
```
dastgir: Changing version just for location change?

npc/merchants/refine.txt
```diff
@@ -50,7 +50,7 @@
 //================= Description ===========================================
 //= Refining NPCs and Metal Salesmen.
 //================= Current Version =======================================
-//= 3.3
+//= 3.5
```
dastgir: Changing version just for location change?
Applies to all remaining npc's.

npc/jobs/1-1e/taekwon.txt
Jedzkie: For me its okay, every changes in file must change version.

dastgir: @hemagx @MishimaHaruna 
Need your thoughts here

MishimaHaruna: I'd go with @kisuka 's idea (not sure if it's written anywhere yet, but it was already discussed once).

Version numbers made of three numbers: major.minor.patch

Script rewrite -> Changes the major number
Important changes -> changes the minor number
Small fixes -> changes the patch number

Not sure about this specific file, since it's not really a trivial fix (it changes a NPC's position), so it might deserve a 1.4->1.5 version bump.
A change like the one below, in `npc/merchants/refine.txt` is just a sprite change and a facing direction change, so it would be enough to bump it from 3.3 to 3.3.1 (not 3.5 in any case o.o)

Calling @kisuka 's opinion on this

Jedzkie:
@dastgir i only undo the incorrect version changes.

MishimaHaruna:
What's the status of this?

dastgir:
I am just against version change for this thing, everything else is ready

MishimaHaruna:
Merged via 540b4b0a29cbfbab4df73e4be4f86d9024a48959, thank you!

I edited the scripts versions a little, using these criteria:

- If a script was completely rewritten, then , the major version is increased by 1 (this wasn't the case for any scripts in this pull request)
- If a script adds, removes or renames NPCs (unless the rename is only visual, and the `::name` stays the same), then the minor version is increased by 1
- All other changes increase the patch version by 1

where:

version x.y.z means major version x, minor version y and patch version z
If z is equal to 0, then the version is written as x.y.

Issue #745, opened 2015-09-26
Missing packet at arrows de-equip


kontownik:
Main problem: when I equip arrow it's good I got state change packet, but on de-equip I didn't get state 0.

rev. e6d9b557c979954d49f80d180696d0fa50ff1db0  Date:   Mon Aug 31 16:30:03 2015 +0800
packet version: 20130807

At equip I got:
[Network] Recv: PACKET_ZC_MSG_STATE_CHANGE4 {index: 695, AID: 2000001, state: 1, TotalMS: 9999, RemainMS: 9999…}

which is good but when I try to de-equip it...

[Network] Send: PACKET_CZ_REQ_TAKEOFF_EQUIP {index: 6}
[Network] Recv: PACKET_ZC_ACK_TAKEOFF_EQUIP_V5 {index: 6, wearLocation: 32768, result: true}

There should be additional packet like:
PACKET_ZC_MSG_STATE_CHANGE4 of index 695 with state 0

kontownik:
none can confirm???

kontownik:
bump

MishimaHaruna:
Does aegis send that packet?

Jedzkie:
@MishimaHaruna Aegis sends this packet.

malufett:
this a renewal changed..where ammunition requirements are now being checked when changing states, equipment and job..confirmed in aegis..same with carts, bullets, equip and mounts...

Jedzkie:
@malufett please reponse on my pm's senpai <3 <3 <3

kontownik:
em my first post explains exactly whats missing (its logic that if we change state once we should change it again when it's off)

PR #1079, opened 2016-01-09
Bow UnEquip Official Behavior

Jedzkie:
- When unequip a bow if there's arrow equipped, it will also unequips the arrow.
- Added battle configuration.

Ref: #745

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1079)

### DIFF
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -9789,6 +9789,15 @@ int pc_unequipitem(struct map_session_data *sd,int n,int flag)
 		status_change_end(&sd->bl, SC_BENEDICTIO, INVALID_TIMER);
 		status_change_end(&sd->bl, SC_ARMOR_RESIST, INVALID_TIMER);
 	}
+	
+#ifdef RENEWAL
+	if (battle_config.bow_unequip_arrow) {
+		if (pos & EQP_ARMS) {
+			if (sd->equip_index[EQI_AMMO] > 0)
+				pc->unequipitem(sd, sd->equip_index[EQI_AMMO], PCUNEQUIPITEM_FORCE);
+		}
+	}
+#endif
 
 	if( sd->state.autobonus&pos )
 		sd->state.autobonus &= ~sd->status.inventory[n].equip; //Check for activated autobonus [Inkfish]
--- a/conf/battle/items.conf
+++ b/conf/battle/items.conf
@@ -93,3 +93,7 @@ item_enabled_npc: 1
 // 2 : disabled equipments are nullify, disabled cards will caused the equipment to unequip
 // 3 : disabled equipments are unequip, disabled cards will caused the equipment to unequip (1+2)
 unequip_restricted_equipment: 0
+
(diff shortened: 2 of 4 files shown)


kyeme:
Need info for pre-renewal

zackdreaver:
as far i remember,
in pre-re arrows will not unequipped when bow is put off (idRO).

Jedzkie:
@kyeme @zackdreaver Maybe they change it recently, tested in RWC Aegis (14.1) and the arrow still equipped even the arrow is unequipped.

zackdreaver:
"the arrow still equipped even the arrow is unequipped"
so it's a bug? heh

Jedzkie:
sorry, its bow xD just a type LOOOOOOOL!

zackdreaver:
hmm in renewal,
I played ranger in iRO (around 2013) as far i remember, it will not unequipped the arrows.
But, because it has a config, i think it's fine

Jedzkie:
In kRO and iRO today, this is now the behavior.

src/map/pc.c
```diff
@@ -9789,6 +9789,15 @@ int pc_unequipitem(struct map_session_data *sd,int n,int flag)
 		status_change_end(&sd->bl, SC_BENEDICTIO, INVALID_TIMER);
 		status_change_end(&sd->bl, SC_ARMOR_RESIST, INVALID_TIMER);
 	}
+	
+#ifdef RENEWAL
+	if (battle_config.bow_unequip_arrow) {
+		if (pos & EQP_ARMS) {
+			if (sd->equip_index[EQI_AMMO] > 0)
```
dastgir: you could merge this and above if

MishimaHaruna:
Just a random thought - It'd be nice if this wasn't hidden behind an ifdef (since it can be enabled and disabled from the configuration). The tricky part would be to set the default value differently for renewal and pre-renewal.

malufett:
there should be a general settings for this one..its not only arrow has this behavior also bullets, cannon balls, carts and bombs..I already implemented those before..you should take considerations on that...

EPuncker:
:+1:

Ridley8819:
well, nearly a year... /sob
anyways, I'm confirming this (again) for kro, but I don't know about pre-re

skyleo:
@Ridley8819 
If you uneqip a Bow while having Arrows equipped, you'll still have the arrows equipped on pre-RE.
Tested on iRO Classic with Hunter Bow and Silver Arrows on Hunter.

Smokexyz:
If this is not official behaviour, we can close it unless someone would like add it as a battle config and rebase.

EPuncker:
@Smokexyz this is official

Ridley8819:
Yes it is, so we need it for renewal, but not for pre-re. However, I talked with Jedzkie and he won't update this right now so maybe it's easier to open a new pull request.

Smokexyz:
Rebased in #1663, therefore closing.

PR #1080, opened 2016-01-09
Fixes #1014

### LINKED ISSUE #1014: Relax
The skill Relax is not working according to the aegis, the icone is disappearing in about 10 seconds what should be up to the player to get up.

source: http://herc.ws/board/tracker/issue-8627-relax/?gopid=24976#entry24976

Jedzkie:
~ credits to csnv.

### DIFF
--- a/src/map/status.c
+++ b/src/map/status.c
@@ -3518,11 +3518,20 @@ void status_calc_regen_rate(struct block_list *bl, struct regen_data *regen, str
 		regen->flag &=~RGN_SP; //No natural SP regen
 	}
 
+
+	// Tension relax allows the user to recover HP while overweight
+	// at 1x speed. Other SC ignored? [csnv]
 	if (sc->data[SC_TENSIONRELAX]) {
-		regen->rate.hp += 2;
-		if (regen->sregen)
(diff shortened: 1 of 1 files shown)


src/map/status.c
```diff
@@ -12409,8 +12418,10 @@ int status_natural_heal(struct block_list* bl, va_list args) {
 		}
 	}
 
-	if (flag && regen->state.overweight)
-		flag=0;
+	// SC_TENSIONRELAX allows HP to be recovered even when overweight. [csnv]
+	if (flag && regen->state.overweight && (sc && !sc->data[SC_TENSIONRELAX])) {
```
MishimaHaruna: Did you mean `if (flag && regen->state.overweight && (sc == NULL || sc->data[SC_TENSIONRELAX] == NULL))`? (if sc is `NULL`, it means that `SC_TENSIONRELAX` is inactive, so it should enter the if)

Jedzkie:
@MishimaHaruna

dastgir:
Seems good to be mergef

dastgir:
Merged,

PR #1081, opened 2016-01-09
PR_MAGNIFICAT Skill Fix.

Jedzkie:
- PR_MAGNIFICAT HP Regeneration applies only in PRE-RENEWAL.


zackdreaver:
I thought it was a modification of servers beside kRO and iRO.
idRO has HP regeneration bonus,
but afaik it was special mod (you know, like ASPD in renewal)

MishimaHaruna:
If this is the correct behavior, it seems good to merge (otherwise, it should be tested on Aegis)

dastgir:
its official behavior as of kRO. Merging it . Thank You.

PR #1083, opened 2016-01-09
Fixes #695

### LINKED ISSUE #695: Impositio Manus
Original topic: from Atemo https://github.com/rathena/rathena/issues/610

http://irowiki.org/wiki/Impositio_Manus

Lower levels of Impositio Manus will overwrite higher levels. For example, someone with level 5 IM activated will have +25 ATK from the skill, but if level 1 IM were used on them their ATK would drop to only +5. Therefore, this skill can be used offensively.
Confirmed on aegis.

Jedzkie:


MishimaHaruna:
I'd say it should also act the other way around, no? (if a higher level is used, the bonus should be updated). #695 doesn't seem to mention this case, but it feels natural if it does.

Jedzkie:
@kyeme

MishimaHaruna:
Tests on official servers (iRO classic) show that if you have Impositio Manus level 1 active and you receive Impositio Manus level 5, it gets overwritten. To my understanding this means that the line `if (sc && sc->data[SC_IMPOSITIO] && sc->data[SC_IMPOSITIO]->val1 > val1)` should be removed.

HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1083)

PR #1086, opened 2016-01-09
Fixes #907

### LINKED ISSUE #907: Cloaking exceed: too fast movement speed bonus
Hmm, if i remember correctly, in official servers (iRO and idRO)
Cloaking exceed gave 140% movement speed bonus,
tested in rAthena, they aren't that fast too.

but in Herc, it feels like 200% ?
(no speed modifier, like inc agi or gears that gives movement bonus)

Jedzkie:
GC_CLOAKINGEXCEED Speed must 140% rather than 200+%

### DIFF
--- a/src/map/status.c
+++ b/src/map/status.c
@@ -8977,8 +8977,8 @@ int status_change_start(struct block_list *src, struct block_list *bl, enum sc_t
 				tick_time = 3000; // [GodLesZ] tick time
 				break;
 			case SC_CLOAKINGEXCEED:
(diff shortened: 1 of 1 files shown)


dastgir:
Should give better commit description, so one can identify what was the issue without a need to open issue tracker

dastgir:
Merged, Thank you,
@Jedzkie  why don't you write a documentation of proper value, some seems to confuse about how the function takes the value

PR #1088, opened 2016-01-10
Constdb revamp

MishimaHaruna:
This converts the constants db (const.txt) to libconfig format (constants.conf). The reason behind the conversion is that it allows implementing new features more easily.

This pull request also implements a new 'Deprecated' field in the converted database. This is useful for example for #908, so that the removed constants may be deprecated instead (and removed at a later time). A silent and immediate removal, as we know, would cause the script parser not to choke on them, and interpret them as a char-reg instead.


MishimaHaruna:
Asking @HerculesWS/developers for feedback (and review if possible)

4144:
what about converter?

MishimaHaruna:
Updated to add a converter, addressing @4144 's concerns

AnnieRuru:
Yay ~ 2 of my plugins broken ... well its kinda expected
erm, I was busy with something else ... oh well

anyway, why not having those value can parse with the source constant ?
so like ... the MAX_LEVEL for example
`MAX_LEVEL: MAX_LEVEL`
the left side is declare as script constant
and the right side is declare from the source
so there is no need to look up in the script.c file for those constants hidden there

there are still many members doesn't know OPTION_XMAS is actually a constant ...
because it is hidden in the source file

MishimaHaruna:
How about we generate (and automatically update through the HerculesWS API bot) a file with all the constant names, in the documentation folder? That way all the constants would be listed in a file, that doesn't need manual maintenance

4144:
good idea

PR #1089, opened 2016-01-10
Added epoll as event-dispatcher on linux (instead of select())

wilkemeyer:
This feature/improvement will enable epoll by default on linux (if its supported, which should be by default by the most distributions since at least 10 years) 

epoll is a lot faster compared to select, as select  has to traverse over all monitored sockets for each fdset on each call; where as epoll uses a per-socket notification list in kernel so if there's an event which we're monitoring on - the kernel adds the corresponding socket to the applicaiton's epoll-instance watch-list, so the next call to epoll_wait will catch this events up.

Additionally the epoll 'result' is an array - more like the windows select() result, we can iterate over;
 no more 'check all fd's up to fd_max if its in the reported event set'...

I've added also a configuration variable at socket's conf (conf/packet.conf) calld epoll_maxevents
Which's set by default to (FD_SETSIZE/2);
 
This setting allows to change how many events are processed at max at once per call to epoll_wait() / cycle.

In past projects (non-ragnarok) it has been turned out that for a single-threaded application a value of half the maxmimum concurrent connections(= max conc. read events) is the most balanced value.

If we'll get reports from heavily loaded servers (i'd assume ~ 1k+ players) about massive delays, we may should change the default value to FD_SETSIZE.


MishimaHaruna:
Thanks for this pull request, but to be honest, I would have preferred to see poll() instead of epoll()...

- poll is portable to all POSIX OSes, epoll is linux-only
- there are of course some benefits in epoll compared to poll, but they're outweighed by the higher overheads of epoll (many more system calls), especially in single-threaded applications, and/or applications with less than thousands of concurrent connections (the average for Hercules is in the tens/hundreds)

That being said, I'll review this pull request (but I'll only merge it when this is disabled by default)

MishimaHaruna:
Merged through #1347

Issue #998, opened 2015-12-21
merge rAthena's Unit Controller Script Commands


AnnieRuru:
https://github.com/rathena/rathena/commit/2cee5b6ff1bf53c4ae53bc1278b09ae84b8a0a76
this one commit has a lot of stuffs in it, if wanna break it up, there are
- [x] Add `*getunittype` and update `*getmapxy` (#871)
- [ ] extend `*monster` and `*areamonster` script command to support array
- [ ] add `MD_NORANDOM_WALK` and `MD_NOCAST_SKILL` monster flag
- [ ] split `*unitstop` into `*unitstopwalk` and `*unitstopattack`
- [ ] extend `*unitskilluseid`/`*unitskillusepos` for fixed cast time
- [ ] `*setmobdata`/`*getmobdata`
- [ ] extend homun/pet/merc to return GID

and these are missing in rathena but I feel like add in
- [ ] `*mobattach`/`*getmasterid` script command
- [ ] `*mobevent` script command
- [ ] `*setunitsize`/`*getunisize` (#895) <- I already tried but failed to do it

also means I have to make 10 total pull request
before we have ourselves a full capable unit-controller commands like rathena has =/

hemagx:
make all of them in same PR and rebase later when it's done.

AnnieRuru:
no way
I seriously don't like the way rathena do it all at once
its better to split out 1 by 1, do it slowly

when I did that getunittype script command, I never thought about using existing getmapxy
until I suddenly found out can use getmapxy to retrieve the coordinate of the monster by using GID
and Haru also update all getmapxy script commands to use UNITTYPE_ constant,
I never thought about that before too

if we do it slowly, examine the system part by part, we wont suffer the same fate like rathena
rathena getunittype is fuck up
rathena setunitdata has `30 - for slave to copy master's mode` but doesn't have mobattach script command
rathena UMOB_MAPID return map index, OMFG ??
rathena UNPC_DMGIMMUNE <-- isn't npc already immune to attack ??
and some others I find it very ridiculous

wanna know how fuck up rathena system is ? read here
http://herc.ws/board/topic/8974-script-command-setmobdata-getmobdata/?p=65553
https://rathena.org/board/topic/103785-add-parameters-for-setunitdata/?p=293176
https://rathena.org/board/topic/104023-setunitdata-parameters-what-are-these-for/

PR #1093, opened 2016-01-11
NoKnockback mode

dastgir:


4144:
I remember last time new mode was rejected if it was non official.
May be better create plugin?

4144:
from other look, if it removing some hardcoded value this is good

src/map/skill.c
```diff
@@ -2089,7 +2089,7 @@ int skill_blown(struct block_list* src, struct block_list* target, int count, in
 		case BL_MOB:
 		{
 			const struct mob_data *md = BL_UCCAST(BL_MOB, target);
-			if (md->class_ == MOBID_EMPELIUM)
+			if (md->status.mode&MD_NOKNOCKBACK)
```
MishimaHaruna: +1 for allowing removal of hardcoded IDs

AnnieRuru:
>I remember last time new mode was rejected if it was non official.

there are 2 monster mode is definitely official #998 
add `MD_NORANDOM_WALK` and `MD_NOCAST_SKILL` monster flag
I remember I saw a video about kro instance
that monster walk through a bridge and players have to defeat them before they pass over
these 2 monster flag is definitely official, its already in rathena

btw, can change that unsigned short mode into int ? for future expansion ...
its already in my todo list ...
or wait until my pull request turn then only change into int ?

MishimaHaruna:
Merged via 3305409b7208befa9f5d26620d30b3490365fc21.

I also changed the mob mode field to uint32, to allow third party mods and future expandability.

PR #1097, opened 2016-01-13
Official behavior when reached max level

dastgir:


4144:
may be move 999999999 into configuration options?

hemagx:
i believe having something like
#define FAKE_MAX_EXP 999999999
will be better than hardcode the number :)

hemagx:
i believe having something like
#define FAKE_MAX_EXP 999999999
will be better than hardcode the number :)

dastgir:
Added battle_config

hemagx:
i believe having something like
#define FAKE_MAX_EXP 999999999
will be better than hardcode the number :)

cydh:
merge this! :p

dastgir:
@MishimaHaruna @hemagx

EPuncker:
:+1:

Smokexyz:
@dastgir Why is there no description to this ? This PR also has conflicts.

MrKeiKun:
Requesting to rebase this PR, since its already ready for merged IMHO.

PR #1098, opened 2016-01-13
Fix 3rd jobs BaseLevel requirement

AnnieRuru:
continue on the adventure on my friends server...
he increased the BaseLevel on all jobs on db\re\exp.txt
although his server has [job changer npc](https://github.com/HerculesWS/Hercules/blob/master/npc/custom/jobmaster.txt),
but I found out the official script gives 3rd job exclusive items, so I do the 3rd class quest

what a weird thing is, my gypsy can do the wanderer job change quest on level100+, no problem
but my high priest can't start the quest .... what a stupid check

on a side note, I noticed warlock check with [BaseLevel > 94](https://github.com/HerculesWS/Hercules/blob/master/npc/re/jobs/3-1/warlock.txt#L61)
is this seriously official ?
although the npc only start the quest on level99+

### DIFF
--- a/npc/re/jobs/3-1/archbishop.txt
+++ b/npc/re/jobs/3-1/archbishop.txt
@@ -64,7 +64,7 @@ prt_church,103,88,3	script	Praying Minister#arch	1_M_PASTOR,{
 		mes "I'm going to make them ministers worthy of serving Odin.";
 		close;
 	}
-	if ((BaseLevel == 99) && (JobLevel > 49)) {
+	if ((BaseLevel >= 99) && (JobLevel > 49)) {
 		if (job_arch == 0) {
 			if (SkillPoint) {
 				mes "You can't change jobs without using all your skill points. Please use all of your skill points before applying to change jobs~";
@@ -775,7 +775,7 @@ job3_arch01,29,34,3	script	Valkyrie#arch	4_F_VALKYRIE2,{
 		mes "- and come for the challenge again. -";
 		close;
 	}
-	if ((BaseLevel != 99) || (JobLevel < 50) || (BaseJob != Job_Priest)) {
+	if ((BaseLevel < 99) || (JobLevel < 50) || (BaseJob != Job_Priest)) {
 		warp "odin_tem02",282,263;
 		end;
 	}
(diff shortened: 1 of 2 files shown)


hemagx:
For warlock, yes it's valid check from official.
and i'm against merge a PR that just made to aid a custom change ~

MishimaHaruna:
Well, let's put it in another way... While the pull request describes a custom situation (which we may decide to or not to support), the changes made in it are actually sensible, and make the script more robust, so why not :)

AnnieRuru:
>and i'm against merge a PR that just made to aid a custom change ~

custom ? seriously ? I just said my gypsy can change job at level100+
https://github.com/HerculesWS/Hercules/blob/master/npc/re/jobs/3-2/wanderer.txt#L63
it should be a `BaseLevel == 99` check, but it isn't, so I can still change at level100+

all other official job changer use  `>` or `<` check,
only arch bishop and guillotine cross use the `!=` check

so, you want to say follow the aegis, and left only arch bishop and guillotine cross bug ?
this doesn't make sense

hemagx:
I'm not sure if you ever played official server, but a gypsy level 100 never exist.
change is fine, I'm just against the idea of do it only to aid a custom change even if it's a "good change".
in other words, against your PR purpose, but not your PR.

AnnieRuru:
I don't like that way of thinking
if everyone only fix this **custom** sort of things only in their own server,
this will later promote nobody going to report bugs, even if it is official one
(because I know how to fix this myself so I don't have to report anymore buahaha)

and yes, I play official server before, about 11 years ago
from the introduction of lutie town, until the introduction of niflheim town
http://ragnarok.wikia.com/wiki/Episode_Updates_%28RO%29
which means from episode 2 until episode 8

AnnieRuru:
@hemagx 
here you go,
I found these Ragnarok Online access card in my precious childhood memory storage box
I just scanned them this morning, now you can have a look

12 hours reload

daily access

monthly access

I'm sure @Emistry saw them before, since we live in same country

btw that server already shut down
http://ragnarok.wikia.com/wiki/Thread:7948
https://forum.lowyat.net/topic/2985769/all
same reason, I stop playing because bots run rampant on every server

hemagx:
@AnnieRuru 
yea i know them shutdown cause i had backup of them servers before it.

dastgir:
Thank you for the contribution

PR #1100, opened 2016-01-14
more meaningful names for fields of enum guild_permission

theultramage:
I came across a custom patch that extended the guild_permission logic with additional features, controlled via atcommands. While cleaning it up to use enum constants instead of hexadecimal values, I found that the current form of GPERM_BOTH does not extend very well. So I tweaked the naming a bit. I also noticed that the same name, GPERM_BOTH, is used in two unrelated contexts, so I split off a more meaningfully named alias.


MishimaHaruna:
Merged, thank you

PR #1106, opened 2016-01-16
Updated Entity Packets

dastgir:


src/map/clif.c
```diff
@@ -1065,6 +1065,7 @@ void clif_set_unit_idle(struct block_list* bl, struct map_session_data *tsd, enu
 #endif
 #if PACKETVER >= 20150513
 	p.body = vd->body_style;
+	memcpy(p.name, status->get_name(bl), NAME_LENGTH);
```
4144: what about move "memcpy(p.name, status->get_name(bl), NAME_LENGTH);" into separate function?
This will allow plugins for example translate name or do any other things with it.

4144:
also memcpy using here is error.
If it try copy memory from null pointer, server will crash

src/map/clif.c
dastgir: Oops, didn't notice, was tired when i was doing PR.

this is character name, so doing for plugin is useless (or just simply hook status->get_name)

dastgir:
Updated @MishimaHaruna

src/map/clif.c
```diff
@@ -1065,6 +1065,7 @@ void clif_set_unit_idle(struct block_list* bl, struct map_session_data *tsd, enu
 #endif
 #if PACKETVER >= 20150513
 	p.body = vd->body_style;
+	safestrncpy(p.name, status->get_name(bl), NAME_LENGTH);
```
4144: safestrncpy will crash too if one of parameters is null pointer.
Here possible null pointer in status->get_name(bl)

src/map/clif.c
4144: status->get_name() not have target sd or bl paramerter. Because this do any actions with name in get_name() function is useless.
This need because now in one packet combined two old packets. With old packets function can be hooked. But here need hook all spawn/move functions for change only name what was before in one separate packet. I suggest move it strsafecpy with name to separate function.

src/map/clif.c
dastgir: @4144 we have nullpo_retv(bl) so how would it cause null pointer?

dastgir:
Updated @4144

PR #1110, opened 2016-01-17
Script CleanUp Phase 2

dastgir:
Changed those ugly select using colon, to use comma


dastgir:
I have missed some select 
the one's which use variables, like
```
.@menu$ = "A:B:C";
select("1:2:3:"+ .@menu$);
```
Since I wouldn't like to have hybrid syntax for these kind of select's

dastgir:
@AnnieRuru @Jedzkie @Streusel , review?

dastgir:
@MishimaHaruna @hemagx

PR #1115, opened 2016-01-21
more meaningful variable names in battle_calc_damage()

theultramage:
The original eathena source mapped ```sd``` to ```bl``` (the target), and ```sc``` to bl's status. Later on, people found the need to add equivalents for ```src```, and instead of first renaming the existing variables, just dropped in the mappings ```tsd``` to ```src``` (the source) and ```tsc``` to src's status. These were really poor choices, since the 't' prefix refers to the target in all other code. I'd like this inconsistency resolved.

Things get a bit more complicated when considering existing custom code, because if the names are just swapped around, all that code would silently break, acting on the source instead of the target and vice versa.

Here is an attempt to recover from this situation. I introduce a top-level variable for src's sd to eliminate four locally-scoped duplicates, and I rename ```ssd``` and ```sd``` to something different and unique. This will cause all custom code to fail to compile, and hopefully people resolving the conflict will do the right thing. Once the dust settles, the variables can once again be renamed, this time to fit the standard naming convention of ```sd``` and ```tsd```.

The same goes for the ```sc``` / ```tsc``` mixup. Here though, I did not do the fourth commit to rename ```sc``` to ```tsc``` because it involved a ridiculous amount of changes. Let me know if I should add that one as well. If it is not done, then custom code containing ```sc``` will get broken once the names are switched around.

An alternative solution would be to go with the non-conventional naming of ```s_sd``` + ```sd``` and ```s_sc``` + ```sc```. The advantage would be less differences, and better diff compatibility with eA/rA.

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -2717,7 +2717,7 @@ void battle_calc_skillratio_weapon_unknown(int *attack_type, struct block_list *
  * After this we apply bg/gvg reduction
  *------------------------------------------*/
 int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Damage *d,int64 damage,uint16 skill_id,uint16 skill_lv) {
-	struct map_session_data *s_sd, *sd = NULL;
+	struct map_session_data *s_sd, *t_sd;
 	struct status_change *sc, *tsc;
 	struct status_change_entry *sce;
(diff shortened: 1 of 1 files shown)


MishimaHaruna:
Merged, thank you!

PR #1123, opened 2016-01-25
Add Constant for Gender.

Emistry:
Improve readability of scripts when using variable `Sex`.

### DIFF
--- a/db/constants.conf
+++ b/db/constants.conf
@@ -3806,4 +3806,9 @@ constants_db: {
 	C_CHOCOLATE: 0xD2691E
 	C_GOLD:      0xFFD700
 	C_VIOLET:    0xEE82EE
+	
+	/** Gender */
+	SEX_FEMALE:	0
+	SEX_MALE:	1
+	
 }
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -553,6 +553,7 @@ Weight      - Amount of weight the character currently carries.
 MaxWeight   - Maximum weight the character can carry.
               Display as in MaxWeight/10.
 Sex         - 0 if female, 1 if male.
+			  Additionally, you could also use constants: SEX_MALE or SEX_FEMALE to match the gender.
 Class       - Character's job.
 Upper       - 0 if the character is normal class, 1 if advanced, 2 if baby.
 BaseClass   - The character's 1-1 'normal' job, regardless of Upper value.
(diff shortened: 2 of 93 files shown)


dastgir:
MALE / FEMALE are too generic name..
Maybe have some different name?
@MishimaHaruna

4144:
Maybe G_MALE/G_FEMALE?

MishimaHaruna:
I'd suggest `SEX_MALE` / `SEX_FEMALE` (bonus: it even matches the enum we have in mmo.h)

EPuncker:
update script_commands as well :+1:

Emistry:
@EPuncker  any suggestion for the template or structure ? https://github.com/HerculesWS/Hercules/blob/master/doc/script_commands.txt#L555
```
Sex         - 0 if female, 1 if male.
			  Additionally, you could also use constants: SEX_MALE or SEX_FEMALE to match the gender.
```
or 
```
Sex         - return types:
				0 - SEX_FEMALE		- Female Character
				1 - SEX_MALE		- Male Character
```

dastgir:
If it matches the source , why don't we move this constant to source? So these constants will always be in sync?

Mysteries:
I agree to add the constants of SEX_MALE and SEX_FEMALE. Looks better / seems better xD

db/constants.conf
```diff
@@ -3806,4 +3806,9 @@ constants_db: {
 	C_CHOCOLATE: 0xD2691E
 	C_GOLD:      0xFFD700
 	C_VIOLET:    0xEE82EE
+	
+	/** Gender */
+	FEMALE:	0
```
MishimaHaruna: Still not renamed here

doc/script_commands.txt
```diff
@@ -553,6 +553,7 @@ Weight      - Amount of weight the character currently carries.
 MaxWeight   - Maximum weight the character can carry.
               Display as in MaxWeight/10.
 Sex         - 0 if female, 1 if male.
```
dastgir: Encourage constant and remove numbers from documentation?

dastgir:
Also, a better commit message

Emistry:
> Also, a better commit message

hmm... Example ?

MishimaHaruna:
Taking care of this - I'll rebase, split in commits, merge.

MishimaHaruna:
Merged via 12a85a011c2008ea92b9197e79646aaa6ece5a41

Issue #1104, opened 2016-01-15
Soul Destroyer Range


zackdreaver:
Reference:
https://github.com/rathena/rathena/issues/905
kRO Renewal Client's = \data\luafiles514\lua files\skillinfoz\skillinfolist.lub

```
    [SKID.ASC_BREAKER] = {
        "ASC_BREAKER",
        SkillName = "Soul Destroyer",
        MaxLv = 10,
        SpAmount = { 20, 20, 20, 20, 20, 30, 30, 30, 30, 30 },
        bSeperateLv = true,
        AttackRange = { 9, 9, 9, 9, 9, 9, 9, 9, 9, 9 },
        _NeedSkillList = {
            { SKID.TF_DOUBLE, 5 },
            { SKID.TF_POISON, 5 },
            { SKID.AS_CLOAKING, 3 },
            { SKID.AS_ENCHANTPOISON, 6 }
        }
    },
```

It's says the range maxed at 9 cells,
but in Herc is 7:

```
379,7,6,1,-1,0x40,0,10,1,yes,0,0,0,misc,0,
```

Jedzkie:
@zackdreaver please confirm also in pre-re

hemagx:
i believe the client uses circles and server-side uses squares ?
can you test if there's actual issue?

zackdreaver:
@Jedzkie 
I don't remember how long in pre-renewal, it's been 6 yers
and i dont have aegis server

Playtester:
I tested on a pre-renewal Aegis server to confirm it's 9 range on pre-renewal, so I'm pretty sure about it.

Also yes, it uses a circle range for players.

Michieru:
Test on renewal and Pre-renewal, both are 10 cell

zackdreaver:
Michi, can't we just stick with 9 cells as stated in skillinfos?
thank you for testing it.

Playtester:
Maybe Hercules has a different implementation for ranges, but rAthena uses Aegis ranges.

Range = 9 means "9 cells in between you and the target if you stand on a linear line"

Michieru:
Range is 10 cells between you and the target as I show on aegis screen.

Playtester:
It's hard to see but that looks like 9 cells and not 10 cells.

AnisotropicDefixation:
It does look like 10 cells to me.

Playtester:
[...]

PR #1125, opened 2016-01-26
Skill Fixes

Jedzkie:
1.)  Fixes Soul Destroyer skill range. #1104

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1125)


iRagno:
In latest Hercules it's used .conf file for skill_db.

PR #1126, opened 2016-01-26
Updating units direction is now configurable.

csnv:
Added update_direction as bitmask option in battle/client.conf, defaults
to 0 (no updates at all).

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -7258,6 +7258,7 @@ static const struct battle_data {
 	{ "min_body_style",                     &battle_config.min_body_style,                  0,      0,      SHRT_MAX,       },
 	{ "max_body_style",                     &battle_config.max_body_style,                  4,      0,      SHRT_MAX,       },
 	{ "save_body_style",                    &battle_config.save_body_style,                 0,      0,      1,              },
+	{ "update_direction",                   &battle_config.update_direction,               BL_NUL,  BL_NUL, BL_ALL,         },
 };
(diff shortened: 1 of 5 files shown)


MishimaHaruna:
Any reason why it defaults to 0? Is that what it happens on official servers? (considering it's different from our behavior before merging this)

src/map/battle.c
```diff
@@ -7258,6 +7258,7 @@ static const struct battle_data {
 	{ "min_body_style",                     &battle_config.min_body_style,                  0,      0,      SHRT_MAX,       },
 	{ "max_body_style",                     &battle_config.max_body_style,                  4,      0,      SHRT_MAX,       },
 	{ "save_body_style",                    &battle_config.save_body_style,                 0,      0,      1,              },
+	{ "update_direction",                   &battle_config.update_direction,               BL_NUL,  BL_NUL, BL_ALL,         },
```
dastgir: Default should be BC_ALL, isn't it? (Same applies to client.conf)

csnv:
No info from official server was considered when this feature was introduced, that's why it should default to 0.

If somebody can test it in an Aegis server would be great, but otherwise and due to lack of info, I would let it as it was before this feature was introduced.

Smokexyz:
This shouldn't be an optional thing. All unit's directions should be updated by default.

HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1126)

PR #1129, opened 2016-01-31
Move P2PTR Macro and Define Packet DB enums

hemagx:
Move P2PTR Macro to socket header and split the macro into two macros for WFIFO/RFIFO.
Change packet database enums to defines in mmo.h


MishimaHaruna:
There will be a separate pull request to change the RFIFO2PTR/RP2PTR macros to return a const pointer.

PR #1134, opened 2016-02-03
Enable use of specialeffect 74,SELF,<npc>;

vykimo:
Fix a forget of dev : when you want to display an effect from a NPC for only one player.


MishimaHaruna:
Merged via 2993842616623a4ad1e4d58f02d75f21d2700a58 (includes some extra cleanup), thank you

PR #1165, opened 2016-02-20
Const correctness (and typecast removal) for [RW]{FIFO,BUF}[PBWLQ] functions

MishimaHaruna:
This extends the work started in #1162, and ensures const correctness of several socket.h macros.


MishimaHaruna:
Issues addressed

MishimaHaruna:
Updated

PR #1181, opened 2016-03-04
Fixes #1114

### LINKED ISSUE #1114: Visual Effect Costumes
Despite having the setting of "cannot be reset by clearance", 
Visual Effect Costumes aka SC Costumes can be removed by Clearance
proof

Also tested with dispell, it did not remove the status.
Reference: https://github.com/HerculesWS/Hercules/blob/master/db/sc_config.txt#L476

Jedzkie:


dastgir:
Seems good to be merged
(No idea why that limitation was there)

MishimaHaruna:
Thank you. No idea either as to why it was limited... If this is how it works, I'm merging it

PR #1182, opened 2016-03-04
NPC Updates

Jedzkie:
Based on kRO (Baphomet Server)

### DIFF
--- a/npc/merchants/shops.txt
+++ b/npc/merchants/shops.txt
@@ -143,37 +143,6 @@ OnInit:
 	sellitem Cinquedea;
 }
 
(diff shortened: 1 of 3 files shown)


dastgir:
Is it done?

Jedzkie:
@dastgir Nope, still in-progress.

Ridley8819:
@Jedzkie do you plan to finish this?

Smokexyz:
@Jedzkie This would be good to merge, if you can squash the commits into 1 and describing all the changes.

Ridley8819:
Squashing is not the problem, since we can do this on merge. But the thing is, this is still in progress. Also i doubt we need any additional infos for the actual changes, since only thing changed are sold items.

PR #1184, opened 2016-03-06
Increase Character Slots from 9 to 12

Jedzkie:
- kRO Mains recently increased their character slots from 11 to 12.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1184)

### DIFF
--- a/src/common/mmo.h
+++ b/src/common/mmo.h
@@ -114,7 +114,15 @@
 
 #define MAX_INVENTORY 100
 //Max number of characters per account. Note that changing this setting alone is not enough if the client is not hexed to support more characters as well.
-#define MAX_CHARS 9
+#if PACKETVER >= 20100413
+#ifndef MAX_CHARS
+	#define MAX_CHARS 12
+#endif
+#else
+#ifndef MAX_CHARS
+	#define MAX_CHARS 9
+#endif
+#endif
 //Number of slots carded equipment can have. Never set to less than 4 as they are also used to keep the data of forged items/equipment. [Skotlex]
 //Note: The client seems unable to receive data for more than 4 slots due to all related packets having a fixed size.
 #define MAX_SLOTS 4


hemagx:
i remember old clients need to be hexed for this.
can someone check all client limits in old clients ? this one may need to be put under PACKETVER condition

Jedzkie:
@hemagx Please do check.

dastgir:
Nice +1

Travesso:
Nice +1

Jedzkie:
Any news on this @hemagx ?

src/common/mmo.h
```diff
@@ -114,7 +114,11 @@
 
 #define MAX_INVENTORY 100
 //Max number of characters per account. Note that changing this setting alone is not enough if the client is not hexed to support more characters as well.
-#define MAX_CHARS 9
+#if PACKETVER < 20090617
```
hemagx: sure it's < and not <= ?

EPuncker:
:+1:

Ridley8819:
<= 20100407

Ridley8819:
@4144 @MishimaHaruna this can be merged

4144:
Not sure is this correct for this version or not,
but while it changed, better add before line:
<code>#ifndef MAX_CHARS</code>
and after line:
<code>#endif</code>

Asheraf:
@4144

4144:
@Asheraf code looks fine, but i dont know is this change correct for this clients.
Need some one else to confirm this.

Asheraf:
Me and ridley have test it, clients before that date need to be diffed to support 12 chars also the diff for increasing char slots got removed after this date.

in 2010-04-07aRagexeRE:
https://github.com/MStr3am/roclientside/blob/master/Diff_Project/kRO/RagexeRE/2010/2010-04-07/2010-04-07aRagexeRE.diff#L1427-L1437

in 2010-04-14eRagexeRE:
https://github.com/MStr3am/roclientside/blob/master/Diff_Project/kRO/RagexeRE/2010/2010-04-14/2010-04-14eRagexeRE.diff

MishimaHaruna:
This packetver check should be kept consistent with the check in `char_mmo_send_characters()`, since it's closely related. That function says that for any clients `>= 20100413` we send the amount of available slots to the client (which matches what you observed in the client diffs, where they removed the slot modification, since starting with this date it's no longer hardcoded)

While I don't know if a 2010-04-13 client exists, and if it doesn't, this won't make any difference, I believe that if `char_mmo_send_characters()` uses 2010-04-13 as a threshold, then this patch also should. Otherwise both should be changed to use 2010-04-07.

Asheraf:
@MishimaHaruna done changed to check 20100413.

MishimaHaruna:
Thank you, merging

PR #1200, opened 2016-03-15
Hide area packets from hidden players/gms.

4144:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1200)

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -950,6 +950,7 @@ ACMD(hide) {
 		//bugreport:2266
 		map->foreachinmovearea(clif->insight, &sd->bl, AREA_SIZE, sd->bl.x, sd->bl.y, BL_ALL, &sd->bl);
 	} else {
+		clif->clearunit_area(&sd->bl, CLR_OUTSIGHT);
 		sd->sc.option |= OPTION_INVISIBLE;
 		sd->vd.class = INVISIBLE_CLASS;
 		clif->message(fd, msg_fd(fd,11)); // Invisible: On
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -419,8 +419,14 @@ bool clif_send(const void* buf, int len, struct block_list* bl, enum send_target
 
 	sd = BL_CAST(BL_PC, bl);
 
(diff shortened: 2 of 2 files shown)


ZiriusPH:
will this allow players occupy the tile an invisible GM is using? sometimes a spectating GM can be a problem for players. would be a great addition.

anacondaqq:
yes it is, thx a lot for fix, and that is very very very very strange, that noone fix it earlier.

EPuncker:
:+1:

AnisotropicDefixation:
If someone goes into @hiding on sight, this will leave the character's afterimage on the spot until refresh or map change.

Smokexyz:
@4144 @hemagx this looks good, why is it kept on hold?

Ridley8819:
'If someone goes into @hiding on sight, this will leave the character's afterimage on the spot until refresh or map change.'

Smokexyz:
He pushed on March 15, 2016, assuming that a fix was added.

Smokexyz:
Probable cause is that area packets are sent after `OPTION_INVISIBLE` is set.

EyesOfAHawk:
bump

EyesOfAHawk:
I found a simple, band-aid fix for this is to simply add `pc->setpos` to the hide() command.

At the very start of the command, save the characters current coordinates, then warp them to their savepoint (or somewhere random on the map).
```
ACMD(hide) {
+ unsigned short sd_map = sd->mapindex;
+ int x = sd->bl.x, y = sd->bl.y;
if (pc_isinvisible(sd)) {
```

Then after clif->changeoption
```
clif->changeoption(&sd->bl);
+ pc->setpos(sd, sd_map, x, y, CLR_TELEPORT);

return true;
```

It's a really bad fix for the after-image issue, but if you want to use this bit, it works well enough.

mekolat:
@EyesOfAHawk yes this is exactly what we had to do [in tmwAthena](https://github.com/themanaworld/tmwa/blob/master/src/map/pc.cpp#L5217).
It needs to behave as if the player logged out

PR #1210, opened 2016-03-20
Added -Wcast-qual compiler flag

MishimaHaruna:
This continues the work started on #1162 to add more diagnostic messages to the compiler output (and clean up any noncompliant code).

`-Wcast-qual` ensures that the `const` qualifier isn't (accidentally or deliberately) dropped with an explicit typecast.


MishimaHaruna:
Putting this on @4144 's code-review list (if you can, please)

Issue #840, opened 2015-10-30
when player die, Hp is 1 ???


AnnieRuru:
the exact same script

```
prontera,155,185,5  script  dsfjhsdkfj  1_F_MARIA,{
OnPCDieEvent:
    dispbottom Hp +" "+ readparam(Hp);
    end;
}
```

in rathena, 
when player alive, display `4138 4138`
when player dead, display `0 0`

in hercules
when player alive, display `1000000 0`
when player dead, display `1 0`

there is currently no way to determine the player is dead in hercules
unlike in the source, there is a pc_isdead() function, but no script command to do that
I like the way how rathena just use readparam(Hp) to check
.
.
if you need a screenshot, 
http://herc.ws/board/topic/10903-mvp-ladder-game-problem/#entry65045

Reilaen:
Does that also happen when you use the Hp variable? It shouldn't be 1 for whatever reason you use it when your Hp should be 0. I am just curios if that happens with the variable too.

AnnieRuru:
yes, both are broken in hercules
`dispbottom Hp +" "+ readparam(Hp);`I tested with both `hp` and `readparam(Hp)`

in hercules, with `Hp` variable
when alive, it shows your current hp <-- correct
when dead, it shows `1` <-- wrong

with `readparam(Hp)`
both alive and dead just shows 0

rathena got both of them right ... that's why ...

I got a feeling that with readparam script command,
the Hp already is your hp, becomes readparam(100000) <-- not confirm

MishimaHaruna:
You can't use `readparam(Hp)`, but you should instead use `Hp` directly. Documentation is incomplete and slightly misleading when it mentions "you can use a parameter name if it is defined in 'db/const.txt'". In fact, the following can't be used with readparam, but should be used directly (yes, it's just as you guessed, `Hp` will be replaced with its value - that's by design):

```
StatusPoint	9	1
BaseLevel	11	1
SkillPoint	12	1
Class	19	1
Upper	56	1
Zeny	20	1
Sex	21	1
Weight	24	1
MaxWeight	25	1
JobLevel	55	1
BaseExp	1	1
JobExp	2	1
Karma	3	1
Manner	4	1
NextBaseExp	22	1
NextJobExp	23	1
Hp	5	1
MaxHp	6	1
Sp	7	1
MaxSp	8	1
BaseJob	119	1
BaseClass	120	1
killerrid	121 1
killedrid	122 1
SlotChange	123	1
[...]

PR #1215, opened 2016-03-21
Fixes #889 & #840

### LINKED ISSUE #889: Hp displayed when dead
On iro/euro the client display 1 hp when you have died (on rathena -> 0 hp).

Image on left, with rathena ; offi on right

image

Credits to @Atemo 
https://github.com/rathena/rathena/issues/758

Jedzkie:
- Novice classes will never go below 50% hp when died.
- Other classes will show 1 HP rather than 0 when died.

---

This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1215)


hemagx:
What's the point of show fake HP, that what we actually have ?

Jedzkie:
Maybe its only a visual glitch, but its confirmed that when i respawn and if i'm a novice, my hp goes half.

dastgir:
Maybe a new battle config for novice to set respawn HP is needed rather than visual glitch...

hemagx:
then it's no sense to fix this way, it will be better to fix the actual HP and not send Fake HP.

Jedzkie:
We need the opinion of @MishimaHaruna

MishimaHaruna:
I believe this is just an AEGIS limitation, no reason to emulate it. I'd rather have 0 HP on dead characters, I don't see a point in displaying what the HP after respawning will be.

dastgir:
@MishimaHaruna 
The problem here is, novice should respawn with 50% HP and not 1Hp

MishimaHaruna:
That's a separate problem (I thought we already have that?) This PR only changes what's displayed *before* respawning, not the actual HP *after*. (when you are dead, you have 0 hp regardless of what this PR does, it just cheats by sending a different value to the client)

Playtester:
@MishimaHaruna 
Reason why Gravity does this is because the client HP display messes up when you send "0 HP" to it. The numbers will disappear in the display completely then.

MishimaHaruna:
@Playtester I wouldn't call that 'messes up', it looks more like they actually took their time to add special handling for the '0'. Or, that's what I've always thought when I saw it.

MishimaHaruna:
Edited so that it uses the same logic that's actually used at respawn.

malufett:
I believe we need to investigate further cause it seems like when after a character dies it actually revive but flagged as dead  so some functions work and some are not according to that flag..

MishimaHaruna:
I don't think we want dead characters to behave like alive ones (even if that happens on Aegis, I refuse to consider it anything other than a bug...)

This pull request only changes the HP display (which I consider really stupid by the way)

MishimaHaruna:
Since it's not clear what we want to do here, I'm putting it on hold. I'm fine with anything, as long as:
1. users can still revert to our old, sane, behavior
2. we don't emulate any aegis bugs, especially with regard to the value of HP that the scripts will see (through the `Hp` parameter)

Feel free to merge, reject or improve this, as long as the two above points are met.

Smokexyz:
@Jedzkie PR fulfill's Haru's requirements. There are conflicting files and the build is failing. Do you want to update this?

MishimaHaruna:
I extended this PR with a candidate fix for the issue discussed in IRC (characters respawning with 2 HP instead of 1).

@Jedzkie could you test that this still works as intended (with both the setting enabled and disabled), and we lost nothing during the rebase and the recent changes I added?

MrKeiKun:
this can be removed from project board. @MishimaHaruna

PR #1226, opened 2016-03-29
Unittalk script cmd shouldn't display the hidden name of npc

Angelmelody:
my test:

    prontera,156,153,5	script	afda#123	4_F_MAID,{
     unittalk getnpcid(0),"Hi I m a npc!";
     close;
    }

npctest

after fixes

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -16818,8 +16818,12 @@ BUILDIN(unittalk) {
 	bl = map->id2bl(unit_id);
 	if( bl != NULL ) {
 		struct StringBuf sbuf;
+		char blname[NAME_LENGTH];
 		StrBuf->Init(&sbuf);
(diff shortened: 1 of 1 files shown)


dastgir:
@HerculesWS/developers

MishimaHaruna:
Whoops :D Nice catch

Emistry:
lol, i always thought that was actually intended and then they created the *npctalk to overcome this issue haha.

MishimaHaruna:
I don't think it really was intended. The purpose for npctalk is that it's automatically attached to the current NPC (or can target a NPC by name), but the expected visible output should be the same. The `#hiddenpart` (just like the`::internalname`) of NPCs isn't something that players should care about, so it has no reason to be displayed in chat messages.

MishimaHaruna:
Merged, thank you

PR #1229, opened 2016-03-29
Update parse of mobs, now we can use 'monster ID constants'.

Cretino:
Update parse of mobs, now we can use 'monster ID constants'.
eg:
prontera,154,154,0,0	monster	Brown Snake	SIDE_WINDER,1,0,0,0

Old way sill works. :)

Signed-off-by: Cretino <yuri-lolz@hotmail.com>


MishimaHaruna:
Awesome! Wanted this for a while, too

Could you please edit the code to follow the [coding style guidelines](https://github.com/HerculesWS/Hercules/wiki/Coding-Style)? Thank you very much :)

Cretino:
Ok, I'll see how do make it.

PR #1232, opened 2016-03-29
Tarot Card/Gospel vs Chemical Protection

dastgir:
* The High Priestess Card from Tarot Card of Fate now removes Chemical Protection (all)
* Gospel now removes Chemical Protection (all) on the caster

Merge of https://github.com/rathena/rathena/commit/894d6f2fe64a401a62ceb57ec48f803e1e7f72be

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -7753,7 +7753,7 @@ int skill_castend_nodamage_id(struct block_list *src, struct block_list *bl, uin
 						sc_start(src,bl,SC_INCMATKRATE,100,-50,skill->get_time2(skill_id,skill_lv));
 						break;
 					case 2: // all buffs removed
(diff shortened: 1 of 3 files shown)


kyeme:
It can be configure to sc_config.txt

dastgir:
No @kyeme, because only some skill removes Chemical Protection, so we cannot make it into sc_config

Playtester:
You could add a new option in the config to define the status change as something else, though.

What I noticed while testing is that food buffs weren't removed on official servers by using Gospel and Tarot Card. I think on Herc food buffs are removed as well.

HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1232)

Issue #1201, opened 2016-03-16
VIT DEF is increasing damage taken in certain situations


AnisotropicDefixation:
Easiest way to test it:

Get high VIT (100+) and Revitalize or Prestige yourself, spawn a large amount of mobs and enjoy getting hit for massive amounts of damage.

This can also happen with lesser buffs (Angelus) or with nothing at all, randomly.

Seems like the more sources there are and the more hits are going through, the more calculations are getting lost and possibly misinterpreted.

WillTheDoggy:
Yes, VIT def is reduced for the number of mobs. 

Though I haven't checked if this value goes negative and adds more damage than the monsters are capable of.

AnisotropicDefixation:
I am aware, the value shouldn't go below 0 though, and the damage numbers are inconsistent as well. Let's say you're getting hit for 100 and suddenlly for 40k.

Playtester:
Yeah the multi-hit penalty is not implemented as on official servers yet.

On official servers (pre-renewal) the count is capped to 22 and the resulting DEF that is returned is:
DEF = DEF * (100 - 5 * (attacker_count - 2)) / 100;

Meaning the reduction is percentual between 0% and 100% in steps of 5%. Reduction below 0 is consequently impossible.

On rA/Herc is simply substracts a fixed "5" from def and def2 and does not even cap def to 0.

rAthena:
```	if( def1 == -400 ) /* being hit by a gazillion units, -400 creates a division by 0 and subsequently crashes */
		def1 = -399;```

Herc:
```			if( def1 < -399 ) // it stops at -399
				def1 = 399; // in aegis it set to 1 but in our case it may lead to exploitation so limit it to 399```

Funny, on Herc you actually get 399 DEF when your DEF goes below -399. How is that not more exploitable than returning 1 DEF?

Probably was supposed to be -399.

AnisotropicDefixation:
Eh, this is actually related to a block above that deals with VIT def reductions and pulls vit_penalty_ battle.conf values.

Adding a simple
`if (target_count > 22)
	target_count = 22;`
seems to solve the issue, as crude as it may be.

Playtester:
Yeah, or better, calculate it depending on the config settings.

```If (target_count > (100/penalty)+(mincount-1)) target_count = (100/penalty)+(mincount-1);```

I know it's related to the code above, just wanted to show that it's not limited to 0.

Playtester:
[...]

PR #1233, opened 2016-03-29
Fixed defense overflow exploit

dastgirp:
- The defense penalty for being attacked by many enemies at the same time can no longer make DEF go negative
- Fixed two potential overflow exploits

When being hit by a lot of monsters, your DEF will become negative and then eventually overflow, making you almost invincible.

On official servers the simultaneous attacker count is limited to 22. So at max, your def is reduced by ((22-2)*5%) = 100%. So it should neither be able to make your DEF negative nor cause an overflow.

Merge from https://github.com/rathena/rathena/commit/590f42cd15c58de78cff8be6053109852375bce1#diff-d96b6365b4bdad78139e676d6e7e3295R4589

---

This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1233)

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -1365,25 +1365,28 @@ static int64 battle_calc_defense(int attack_type, struct block_list *src, struct
 #endif
 			}
 
-			if( battle_config.vit_penalty_type && battle_config.vit_penalty_target&target->type ) {
-				unsigned char target_count; //256 max targets should be a sane max
-				target_count = unit->counttargeted(target);
-				if(target_count >= battle_config.vit_penalty_count) {
-					if(battle_config.vit_penalty_type == 1) {
-						if( !tsc || !tsc->data[SC_STEELBODY] )
-							def1 = (def1 * (100 - (target_count - (battle_config.vit_penalty_count - 1))*battle_config.vit_penalty_num))/100;
-						def2 = (def2 * (100 - (target_count - (battle_config.vit_penalty_count - 1))*battle_config.vit_penalty_num))/100;
-					} else { //Assume type 2
-						if( !tsc || !tsc->data[SC_STEELBODY] )
-							def1 -= (target_count - (battle_config.vit_penalty_count - 1))*battle_config.vit_penalty_num;
-						def2 -= (target_count - (battle_config.vit_penalty_count - 1))*battle_config.vit_penalty_num;
+			if (battle_config.vit_penalty_type != 0 && (battle_config.vit_penalty_target & target->type) != 0) {
+				int target_count = unit->counttargeted(target);
+				if (target_count >= battle_config.vit_penalty_count) {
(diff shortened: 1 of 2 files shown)


Playtester:
Might want to link to Herc Issue: https://github.com/HerculesWS/Hercules/issues/1201

dastgir:
@HerculesWS/developers

MishimaHaruna:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/map/battle.c, line 1370 \[r1\]](https://reviewable.io:443/reviews/herculesws/hercules/1233#-KLcsCJDH4wcmS8DLsA_:-KLcsCJDH4wcmS8DLsAa:1617978202) ([raw file](https://github.com/herculesws/hercules/blob/9d910a401461eeca0f9c9b0f0f930d840e50dfdf/src/map/battle.c#L1370)):*
> ```C
> 			if( battle_config.vit_penalty_type && battle_config.vit_penalty_target&target->type ) {
> 				unsigned char target_count; //256 max targets should be a sane max
> 				target_count = min(unit->counttargeted(target), (100 / battle_config.vit_penalty_num) + (battle_config.vit_penalty_count - 1));
> ```

This looks crashy were someone to set `vit_penalty_num` to 0.

Also, see `vit_penalty_type`. The line above assumes a value of 1, and doesn't handle the value 2.

```
// Type of penalty that is applied to both equipment and vit DEF when more than vit_penalty_count monsters are targetting player
// 0 = no penalty is applied
// 1 = vit_penalty_num is reduced from DEF as a %
// 2 = vit_penalty_num is reduced from DEF as an exact amount
vit_penalty_type: 1
```

I think it'd be more appropriate to calculate the def value, then cap it perhaps (in both the cases below separately)

---

*[src/map/unit.c, line 1797 \[r1\]](https://reviewable.io:443/reviews/herculesws/hercules/1233#-KLctIyD05JGuJHd-l6k:-KLctIyD05JGuJHd-l6l:928616489) ([raw file](https://github.com/herculesws/hercules/blob/9d910a401461eeca0f9c9b0f0f930d840e50dfdf/src/map/unit.c#L1797)):*
> ```C
> 		if (ud->target && (target = map->id2bl(ud->target)) != NULL && (ux = unit->bl2ud(target)) != NULL && ux->target_count > 0)
> 			--ux->target_count;
> 		if (target_id && (target = map->id2bl(target_id)) != NULL && (ux = unit->bl2ud(target)) != NULL  && ux->target_count < 255)
> ```

`UCHAR_MAX` instead of `255` (so when we'll replace it with a saner type, it'll be easier to spot than this magic value)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1233)*

Smokexyz:
@HerculesWS/developers

dastgirp:
@4144 @MishimaHaruna

PR #1235, opened 2016-03-29
Power Swing Update

dastgir:
* Power Swing should only cast while wearing a 1H Axe or 2H Axe.


AnisotropicDefixation:
You probably should also update the source so the axe boomerang autocast only works with axes like it's supposed to.

malufett:
in aegis it is usable to any weapon

PR #1237, opened 2016-03-30
Add missing checks into some files in map server

4144:


src/map/map.c
```diff
@@ -1403,6 +1421,7 @@ void map_clearflooritem(struct block_list *bl)
 
 	nullpo_retv(fitem);
 
+	nullpo_retv(fitem);
```
MishimaHaruna: Duplicate check? It's already 2 lines above

src/map/mob.c
```diff
@@ -2662,6 +2712,7 @@ int mob_dead(struct mob_data *md, struct block_list *src, int type) {
 
 void mob_revive(struct mob_data *md, unsigned int hp)
 {
+	nullpo_retv(md);
```
MishimaHaruna: Please swap with the next line (MSVC2012 compatibility)

4144:
fixed issues

PR #1252, opened 2016-04-14
Make packet_len usable in plugins

inhyositsu:
Original packet_len and all packet struct can't not be used in plugins,
so I changed it.


hemagx:
packet_db should not be interfaced it's meant to be private use for client interface only (clif.c).
however with start usage of private headers it will be directly usable by plugins, for now you can use the function clif->packet to get such info

PR #1253, opened 2016-04-14
Hpmhooking improvements

MishimaHaruna:
- Compile-time checks of the HPMHooking functions. This means that, if a plugin hooks into a function through HPMHooking, and the core function changes, the plugin will show a compile-time warning instead of silently compiling (and crashing at runtime or causing undesired effects).
- Added one level of indirection to pointer variables in pre/post hook functions (previously only non-pointer variables had that). This makes it possible to override const pointers from HPMHooking hooks.


EPuncker:
hey @MishimaHaruna can you please guide us on how to update our current plugins in order to work with those new changes since it broke every single one :P

dastgir:
Structures will have one more level of indirection in preHook
(struct xyz * will now be struct xyz **)
And all parameters in postHook will be same as the functions in source (no indirection)

Emistry:
just curious ...
i noticed some of it still declared like this
`(struct xyz ** will now be struct xyz *)`
or
`(struct xyz * will now be struct xyz **)`
or
`(struct xyz * will now be struct xyz *)`
or
`(struct xyz ** will now be struct xyz **)`

as you can see the amount of asterisk symbol`*` used are different sometime.

dastgir:
For preHook, 2nd one.
For postHook, 3rd one

MishimaHaruna:
I just posted a topic explaining the changes: http://herc.ws/board/topic/12583-2016-05-01-hpmhooking-improvements/

PR #1263, opened 2016-04-23
chatmessage refactoring

MishimaHaruna:
This is some refactoring in order to better support the 20151001 version of the chat packets (that apparently don't have a NUL terminator)

Highlights:
- Some non-clif code was moved from clif.c to pc.c
- We no longer require that the client sends a NUL terminator in all chat-related packets (public, party, guild, whisper, etc.)

While I'm unable to test on a 2015 client, I've tested it on old clients, and didn't spot any regressions.


dastgir:
Frost Joke still contains NUL Terminator.

hemagx:
@dastgir it's designed to not require the nul so should work fine if nul
there

MishimaHaruna:
Yeah, safestrncpy will take care of handling the string terminator (and stripping any extra terminators that the client might send)
```
safestrncpy(out_buf, packet->message, textlen+1); // [!] packet->message is not necessarily NUL terminated
message = out_buf + namelen + 3;
```

PR #1271, opened 2016-04-28
Add new gcc sanitize flag -fsanitize=bounds-strict

4144:


4144:
fix flag works in gcc 6

PR #1287, opened 2016-05-14
Fix compilation with lto enabled on gcc

4144:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1287)


Ridley8819:
@4144 is this still in progress?

4144:
This pr for now works only for gcc, and not clang.
Probably need add check for gcc compiler...

MishimaHaruna:
Tested on clang 3.7.1 (Gentoo Linux i686) and apple clang 8.0.0 (macOS 10.12 x86_64), and there are no regressions. Merging.

PR #1289, opened 2016-05-14
Remove swap testing code (leftover in cbasetypes)

mkbu95:
I'm not sure how I managed to squeeze this in the repo a while ago, but it did, and it just doesn't do anything.
I was just testing how MSVC optimized `swap`. Anyway, it should be removed :D

Thanks!


MishimaHaruna:
I was wondering about that code, and whether it was intended to be used at some point, thank you for clarifying :)

PR #1318, opened 2016-06-20
Added missed npc Scamp in renewal location of izlude from Shadow Rekember quest.

AtlantisRO:


dastgir:
Maybe they should be in izlude file? Instead of lighthalzen?

AtlantisRO:
We think it should be in quests_lighthalzen.txt because this npc is from a Lighthalzen Quest:

- http://irowiki.org/wiki/Shadow_Quest_(Rekenber_Job_Quest)

This npc is only missed in Renewal, he is available in pre-renewal in the file:

- npc\pre-re\quests\quests_lighthalzen.txt

Also, npc "::Scamp" is defined in npc\quests\quests_lighthalzen.txt

PR #1319, opened 2016-06-21
Added support to Para_Team_Mark_ item in Eden Group

AtlantisRO:
In 08/01/2012 Maintenance was added an improved version for Eden Group Mark:

"- Added exchange function for 'Lime Evenor' NPC where you can exchange  'Eden Group Mark' for a consumable item 'Eden Group Mark' with  butterflywing-like function that can be used infinite number of times."
Source: http://forums.irowiki.org/index.php?topic=95103.0

This commit gives support to this new Eden Group Mark. All dialogues are official and have been taken from iRO.

If you need any reference about dialogues, can provide a replay using this npc and receiving this item in iRO.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1319)


AnisotropicDefixation:
So, basically this is in an infinite butterfly wing? Might want to make this optional and add warnings then.

AtlantisRO:
The use of Eden Mark Group has a delay of 20 minutes in kRO (according to it's description), so, it's use is limited to a max of 3 per hour:

http://ro.gnjoy.com/guide/runemidgarts/itemview.asp?itemSeq=6&ID=22508

However, actually in iRO it has a delay of 30 minutes (even when in iRO's description of the item says 120 minutes).

This doesn't replace the use of a butterfly wing, but, it is useful when player forgets to get a Butterfly Wing.

AnisotropicDefixation:
Oh, I see now.

db/re/item_db.conf
```diff
@@ -145960,6 +145960,27 @@ item_db: (
...
+		nodrop: true
+		notrade: true
+		noselltonpc: true
+		nocart: true
+		nostorage: true
+		nogstorage: true
+		nomail: true
+		noauction: true
+	}
+	Script: <"
```
Ridley8819: please add space after comma, and if the script is only one line like it is here, do it like this:
Script: <" unitskilluseid getcharid(3), "AL_TELEPORT" ,3; ">

npc/instances/EndlessTower.txt
```diff
@@ -216,6 +216,9 @@ e_tower,81,105,0	script	Tower Protection Stone	2_MONEMUS,{
 	.@etower_timer = questprogress(60200,PLAYTIME); // 1 week
 	.@etower_timer2 = questprogress(60201,PLAYTIME); // 4 hours
 
+	if (has_instance2("1@tower") >= 0)
```
Ridley8819: why does this PR include changes to Endless Tower?

AtlantisRO: It was a commit non-related to this pull, I have delete it. Thanks for the help explaining how to do it!

PR #1320, opened 2016-06-22
Endless Tower Fix

### LINKED ISSUE #732: Endless Tower CD only applies to party leader
So the Endless Tower's 7 day cool down only applies to a party's leader.
Leaving the party after completion and assigning someone else as a party leader allows for infinite entry to Endless Tower.

I have no solution, honestly. Just quite concerned since this allows for some early level MVPs and Mini-MVPs to be farmed non-stop.

dastgirp:
Player cannot re-enter Endless Tower until the cooldown has finished
Fixes #732

---

This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1320)

### DIFF
--- a/npc/instances/EndlessTower.txt
+++ b/npc/instances/EndlessTower.txt
@@ -267,7 +267,7 @@ e_tower,81,105,0	script	Tower Protection Stone	2_MONEMUS,{
 		case 3:
 			end;
 		}
-	} else if (.@etower_timer == 1 && .@etower_timer2 == 2) {
+	} else if (.@etower_timer == 1 && (.@etower_timer2 == 2 || etower_partyid != getcharid(1))) {
 
 		.@dun_lim_time = etower_timer+604800; // 1 week
 		.@dun_lim_time2 = etower_timer+14400; // 4 hours
@@ -315,6 +315,7 @@ L_Enter:
 		mapannounce "e_tower", strcharinfo(0)+" of the party, " +getarg(3)+", is entering the dungeon, Endless Tower.",bc_map,"0x00ff99",FW_NORMAL,12;
 		if (getarg(1)) {
 			etower_timer = gettimetick(2);
+			etower_partyid = getcharid(1);
 			setquest 60200;
 			setquest 60201;
 		}


AtlantisRO:
Player still can enter to new instances created by other members of the party by changing party leader, because they conserves the same party id.

dastgir:
Ah, didn't thought that, the reproduce method given to me was change party

AtlantisRO:
I know and that was a great idea, but it still gives a chance to bypass. If you don't mind, we want to send our fix to this bug.

Emistry:
I think another way would be just assign the "Endless Tower Cooldown Quest" to every party member once they entered or during instance generated.

this way it would solve any trick they tried to re-enter the instance as long as the cooldown isn't completed

dastgir:
But emistry, they won't be able to re enter the tower with same party too

dastgir:
@AtlantisRO, sure you can show how you fixed it

AtlantisRO:
Sorry for the late. The trick is to make individual and unique each instance and check if the instance is a diferent one to the one they registered. This can be done using has_instance2 command and identifying the instance with gettimetick command, since it is the only way (we observe) that cannot be modifyed by player.

AtlantisRO/Hercules@a0c1fcdc8a98177277c6c10199b4c1a07ee1bf35

Ridley8819:
From what I know, you aren't supposed to re-enter during a run, the ashes are supposed to be used on the next run, not the same. So I see no issue in assigning the cooldown quest upon entering. I think this is wrong since years

dastgir:
Thats what I have mentioned in issue, that one cannot enter ET again in kRO/official servers

Ridley8819:
So what is the matter then?

Ridley8819:
After talking with Haru we decided to make it official, but add an variable option to decide to use that old behavior. Could you do this change please?

xBeret:
@dastgir @AtlantisRO @Ridley8819  Please go back in the discussion of this pr, many players abuse this in the servers.

dastgirp:
Closing in favor of #1594

PR #1321, opened 2016-06-23
Clif fixes

hemagx:


src/map/clif.c
```diff
@@ -267,19 +267,39 @@ uint32 clif_refresh_ip(void)
 
 #if PACKETVER >= 20071106
 static inline unsigned char clif_bl_type(struct block_list *bl) {
+	struct view_data *vd;
 	nullpo_retr(0x1, bl);
+
+	vd = status->get_viewdata(bl);
```
4144: I think need add nullpo_retr(CLUT_NPC, vd);
Because if vd is NULL this is error.

src/map/clif.c
```diff
@@ -267,19 +267,39 @@ uint32 clif_refresh_ip(void)
 
 #if PACKETVER >= 20071106
 static inline unsigned char clif_bl_type(struct block_list *bl) {
+	struct view_data *vd;
 	nullpo_retr(0x1, bl);
```
4144: Here need nullpo_retr(CLUT_NPC, bl);

src/map/clif.c
4144: And call to status->get_viewdata need do from case

hemagx: we can't do a nullpo check on vd as it's not always used and maybe can be null in cases it's not used, check rest of he code.

src/map/clif.c
hemagx: woah a leftover :D

PR #1325, opened 2016-06-24
Correct variable type to instance variable in Old Glast Heim script.

Angelmelody:
variable **.MyMobs** should be an instance variable not  npc variable

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1325)


MishimaHaruna:
This was merged via #1370. Thank you very much

PR #1351, opened 2016-07-10
Huld improvements (2)

MishimaHaruna:
- Added support for concatenated strings in the .po files, as described in the [gettext documentation](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html):

```
msgid ""
"Here is an example of how one might continue a very long string\n"
"for the common case the string represents multi-line output.\n"
```

- Added header to the generated .pot file (improves compatibility with online translation services such as transifex)

- Converted npc/airports/ and npc/battleground/ to a HULD-compliant format (simplifies translation, especially to languages that have different sentence order or are much more terse than English, as per HULD best practices)

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1351)

### DIFF
--- a/npc/airports/airships.txt
+++ b/npc/airports/airships.txt
@@ -9,7 +9,7 @@
 //= This file is part of Hercules.
 //= http://herc.ws - http://github.com/HerculesWS/Hercules
 //=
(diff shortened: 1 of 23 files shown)


EPuncker:
can't wait to test it

Issue #1340, opened 2016-07-04
Missing Items in the Pre-Renewal item_db.conf listed in skill_require_db.txt


Smokexyz:
There are items in the pre-re skill_require_db.txt that don't exist in item_db.conf -
```
2804
2805
2806
2808
2807
13294
6493
1549
2139
2800
2801
2802
```
this is gr8 m8 i r8 8/8. If in case the skill db libconfig takes a while to merge, the pre renewal servers might want to take note of these.

PR #1352, opened 2016-07-10
Some skilldb fixed after conversion

4144:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1352)

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -19295,9 +19295,9 @@ void skill_validate_damagetype(struct config_setting_t *conf, struct s_skill_db 
 				}
 			} else if (strcmpi(type, "SplashArea") == 0) {
 				if (on) {
-					sk->nk |= NK_SPLASH;
+					sk->nk |= NK_SPLASH_ONLY;
 				} else {
-					sk->nk &= ~NK_SPLASH;
+					sk->nk &= ~NK_SPLASH_ONLY;
 				}
 			} else if (strcmpi(type, "SplitDamage") == 0) {
 				if (on) {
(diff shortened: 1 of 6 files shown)


src/map/skill.c
```diff
@@ -20088,8 +20088,8 @@ bool skill_read_skilldb(const char *filename)
 			skill->level_set_value(tmp_db.num, 1); // Default 1
 
 		/* Interrupt Cast */
-		if (libconfig->setting_lookup_bool(conf, "InterruptCast", &tmp_db.castcancel) == 0)
-			tmp_db.castcancel = 1;
+		if (libconfig->setting_lookup_bool(conf, "InterruptCast", &tmp_db.castcancel) == CONFIG_FALSE)
+			tmp_db.castcancel = 0;
```
Smokexyz: Didn't we decide this defaults to true?
`InterruptCast: Cast Interruption              (bool, defaults to true)`

tools/skilldbconverter.php
```diff
@@ -654,7 +654,7 @@ function getnk($nk)
 {
 	$bitmask = array(
 		"NoDamage"         =>   intval(0x01), //- No damage skill
-		"SplashArea"       =>   intval(0x02)+intval(0x04), //- Has splash area (requires source modification)
```
Smokexyz: This in source is 
`NK_SPLASH         = 0x02|0x04, // 0x4 = splash & split`
therefore adding both options to the file as -
` 
	DamageType: {
		SplashArea: true
		SplitDamage: true
	}
`
is pointless.

hemagx: 1- it's not pointless, but yes it can be prettier (which will happen later when this fix finished)
2- this check you made caused skills to have been defined as splasharea by wrong, many skills is bugged.

src/map/skill.c
MishimaHaruna: Good point. Then the converter needs to be fixed.

tools/skilldbconverter.php
Smokexyz: Damn, yup, wrong check. Sorry about that.

MishimaHaruna:
Also fixes #1340

Issue #1240, opened 2016-04-04
Warp Facing Direction


kyeme:
When respawning from teleport/warp the player should always be facing north.. it looks that way on your screen but on others screen they will see you facing the last direction you were facing before you warped.
(Related to https://github.com/HerculesWS/Hercules/commit/c0c9af6d91be054390b1ab797bc292140f25287c)
Hercules Video:
https://youtu.be/3tIyqgcbHA4

MishimaHaruna:
@Jedzkie please don't close other repositories issues, or I'll have to revoke push rights.

Jedzkie:
Sorry @MishimaHaruna i diudnt noticed it. My bad.

PR #1353, opened 2016-07-11
Fixes warp facing direction. Issue #1240

KirieZ:
This fixes issue #1240 by changing facing direction when setting character position.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1353)


hemagx:
While it's probably bugged aegis behavior would not be better to update
direction on player side to fix the issue than force north facing?

MishimaHaruna:
Or it could be made configurable (via battle_config flag)

KirieZ:
I'll take a look, I thought it was forced by the client and as other commits were fixing it that way I just followed.

KirieZ:
I like the idea of making it configurable, as we add aegis behavior and at the same time one that make more sense. It's possible to change the direction after warping, but to keep it after logging out we need a new column, what should I do? where is the best place to put this on battle conf? player.conf? and what should I call it? Thanks

hemagx:
It's useless to log this, keep it reset on relog but only remembered when
warping

KirieZ:
Ok, done.

kyeme:
Why not set to default?

Emistry:
instead of keep the facing direction or not, why not just use the config to set the default direction after warped?

enable it to set for all 8 directions.
set the facing direction according to the config value.

conf/battle/player.conf
```diff
@@ -185,3 +185,7 @@ idletime_criteria: 0x1F
 costume_refine_def: yes
 shadow_refine_def: yes
 shadow_refine_atk: yes
+
+// Keep player facing direction after warping?
+// Default: no (on official servers players always faces north)
+player_warp_keep_direction: yes
```
hemagx: Please add a new line

hemagx:
@kyeme cause no sense Aegis behavior doesn't deserve to bet set as default :)
@Emistry It's another customize suggestion, but it's no sense to customize facing at all.

src/map/clif.c
```diff
@@ -9547,7 +9547,8 @@ void clif_parse_LoadEndAck(int fd, struct map_session_data *sd) {
 		clif->clearunit_area(&sd->bl, CLR_DEAD);
 	else {
 		skill->usave_trigger(sd);
-		sd->ud.dir = 0;/* enforce north-facing (not visually, virtually) */
+		if (battle_config.player_warp_keep_direction)
```
hemagx: kindly use == 1 here please

src/map/pc.c
```diff
@@ -5580,6 +5580,8 @@ int pc_setpos(struct map_session_data* sd, unsigned short map_index, int x, int
 		npc->script_event(sd, NPCE_LOGOUT);
 		//remove from map, THEN change x/y coordinates
 		unit->remove_map_pc(sd,clrtype);
+		if (!battle_config.player_warp_keep_direction)
+			sd->ud.dir = 0; // makes character face north
```
hemagx: same here, please use == 0

src/map/pc.c
```diff
@@ -5610,6 +5612,9 @@ int pc_setpos(struct map_session_data* sd, unsigned short map_index, int x, int
 		vending->close(sd);
 	}
 
+	if (!battle_config.player_warp_keep_direction)
```
hemagx: same as above

KirieZ:
Maybe it would be better to not use "Default" word on the comment? Just say that official servers does that? I think what @kyeme is saying is that we should be using the comment's default value. I think I didn't use it because I tested with yes and forgot to change back to no before commiting.

PR #1368, opened 2016-07-17
Adds official behavior when selling to NPCs. Fixes issue #1214

### LINKED ISSUE #1214: Official Trader Selling Behavior
If you reached server's max zeny and sell some items to the trader npc, a message will show 'The Deal has failed.' and the item will not deleted.

KirieZ:
This fixes #1214 . I chose not to avoid the deal when it's a script controlled shop.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1368)

### DIFF
--- a/src/map/npc.c
+++ b/src/map/npc.c
@@ -2173,6 +2173,9 @@ int npc_selllist(struct map_session_data *sd, struct itemlist *item_list)
 
 	z = 0;
 
+	if (sd->status.zeny == MAX_ZENY && !nd->master_nd)
+		return 1;
+
 	// verify the sell list
 	for (i = 0; i < VECTOR_LENGTH(*item_list); i++) {
 		struct itemlist_entry *entry = &VECTOR_INDEX(*item_list, i);


hemagx:
@Kiriez i believe it would be cool to have config for this one, cause what if i just wanan check the selling price?

KirieZ:
@hemagx I didn't understand. you mean for script controlled shops?

KirieZ:
@hemagx Ah, I think I understood now, you mean if you want to open the sell window to see the item sell price (not to actually sell it)? In this case, this change doesn't affect it as npc_selllist is called when the player confirms the sell, not when he asks to show the sell window.
You can still open the sell window and even put the items there, but when you press OK it shows the "deal failed" message

MishimaHaruna:
From what I see this only affects the case when the player already has `MAX_ZENY` zeny. Does anyone know how we should handle (or if we're already handling it correctly) the case when the player has `< MAX_ZENY` but would exceed it during the trade (i.e. `MAX_ZENY - 10`, and sell something for > 10 zeny)?

PR #1387, opened 2016-08-03
OctopusCave instance HULD support

Asheraf:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1387)

### DIFF
--- a/npc/re/instances/OctopusCave.txt
+++ b/npc/re/instances/OctopusCave.txt
@@ -261,7 +261,7 @@ OnEnable:
 	.@map$ = instance_mapname("1@cash");
 	switch (.@i) {
 	case 1:
-		areamonster(.@map$, 34, 96, 38, 100, _("Octopus's Henchman"0, MD_OCTOPUS, 3, .@label$);
+		areamonster(.@map$, 34, 96, 38, 100, _("Octopus's Henchman"), MD_OCTOPUS, 3, .@label$);
 		areamonster(.@map$, 39, 90, 41, 92, _("Octopus's Henchman"), MD_OCTOPUS, 2, .@label$);
 		areamonster(.@map$, 38, 76, 40, 78, _("Octopus's Henchman"), MD_OCTOPUS, 2, .@label$);
 		areamonster(.@map$, 39, 58, 41, 60, _("Octopus's Henchman"), MD_OCTOPUS, 2, .@label$);
--- a/npc/re/instances/OctopusCave.txt
+++ b/npc/re/instances/OctopusCave.txt
@@ -64,33 +64,33 @@ mal_dun01,151,235,5	script	Starfish	4_ASTER,{
 		switch (select("Ask what's going on.", "Ask to open the gate.", "Go to other location.")) {
 		case 1:
 			mes("[Starfish]");
-			mes("Lately, our Starfish lady is\r"
-				"suffering with some issues, ooh ooh.");
-			mes("Something bad happen in this\r"
-				"peaceful place, hehe!");
+			mes("Lately, our Starfish lady is");
+			mes("suffering with some issues, ooh ooh.");
+			mes("Something bad happen in this");
+			mes("peaceful place, hehe!");
 			next;
 			mes("[Starfish]");
-			mes("Weird looking limbs came out\r"
-				"from the hole there,\r"
-				"tried to kidnap our lady Starfish.");
+			mes("Weird looking limbs came out");
+			mes("from the hole there,");
+			mes("tried to kidnap our lady Starfish.");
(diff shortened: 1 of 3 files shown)


EPuncker:
why do you randomly breaks "mes" with \r and then just don't xD keep a fixed thing :P other than that, great job!

Asheraf:
well as i understand (not sure if correct ^^) if the mes end with `. or ! ...` i end it else if it end with `,` or nothing i use \r correct me if it's not the right way ^^

Ridley8819:
I would also make the mes to 1 liners instead of having several lines. Also you can rename the file (remove camel case) and adjust it in the scripts.conf :D

npc/re/instances/octopus_cave.txt
```diff
@@ -43,63 +43,44 @@ mal_dun01,151,235,5	script	Starfish	4_ASTER,{
 
 	if (!.@party_id) {
 		mes("[Starfish]");
-		mes("You alone is powerless, hehe!");
-		mes("Better get someone to help you out.");
-		mes("Make a party, and come back later.");
+		mes("You alone is powerless, hehe! Better get someone to help you out. Make a party, and come back later.");
```
MishimaHaruna: Are these intended to be in one line? Is it desirable that the line break is not where the exclamation mark or the period are?

npc/re/instances/OctopusCave.txt
```diff
@@ -64,33 +64,33 @@ mal_dun01,151,235,5	script	Starfish	4_ASTER,{
 		switch (select("Ask what's going on.", "Ask to open the gate.", "Go to other location.")) {
 		case 1:
 			mes("[Starfish]");
-			mes("Lately, our Starfish lady is\r"
-				"suffering with some issues, ooh ooh.");
-			mes("Something bad happen in this\r"
-				"peaceful place, hehe!");
+			mes("Lately, our Starfish lady is");
```
MishimaHaruna: This commit promised HULD support, but it removes `\r` (which was added for HULD support. Undo it?

Asheraf: i have removed the \r because they said it's messy and change it to 1 liner instead and i think the pr would be merged into 1 commit.

npc/re/instances/octopus_cave.txt
Asheraf: @MishimaHaruna then should i leave it normal and just break the lines every time ?

MishimaHaruna: I think there's a lot of confusion about the whole `\r` matter, for which I apologize.

Many Aegis scripts introduce hard wraps, in the places where they want the lines to be split. This, in private servers, is desirable for some, and undesirable for others (especially when translating, or when using a different font in the client). For example:

```
mes("[Parmy Gianino]");
mes("Well then, would");
mes("you like to apply");
mes("for registration?");
```

Client output:

> [Parmy Gianino]
> Well then, would
> you like to apply
> for registration?

This produces 4 separate lines in the client output, which is fine in the original language and with the original font. If you use a larger font, those lines might get additionally split in other places, or if you translate it, the translated sentence might be longer or shorter. Additionally, those sentences ("you like to apply", "for registration?") are quite meaningless to translate, without enough context. That's really undesirable.

For that reason, we introduced the `\r`, which is a special character that we can decide whether to strip out from the script when parsing it (we have a setting in the emulator), so that, those who want to preserve the exact line breaks as in official servers, don't have to do anything; those who want to let the client format it properly regardless of language and font, can toggle the setting and strip the `\r`.

```
mes("[Parmy Gianino]");
mes("Well then, would\r"
    "you like to apply\r"
    "for registration?");
```

Client output (default):

> [Parmy Gianino]
> Well then, would
> you like to apply
> for registration?

Client output (optional, for best HULD or custom font output):

> [Parmy Gianino]
> Well then, would you like to apply for registration?

This can produce either 4 separate lines, or just two, and always produces a meaningful string to translate in the generated_translations.pot ("Well then, would\ryou like to apply\rfor registration?"), which contains enough context so that people are able to translate it properly, and even allows for languages that require more or less than 3 lines.

Now, this only makes sense if the sentence is the same. If you have separate sentences, there's no point in using \r to join them. They're separate sentences, and each of them works on its own. Additionally, the line break to separate two different sentences, is likely desirable for everyone, regardless of font and language. That's why I'd keep them in separate `mes` commands, instead of merging them.

Of course, when you don't want to introduce a line break between two sentences, then it's fine to merge them into just one `mes`.

I hope this makes sense, otherwise please ask me for clarification. (I'll copy this content into wiki page later on, if it's of any help)

dastgir:
Branch has conflicts

PR #1393, opened 2016-08-08
Add support to *monster to save ids in array.

Asheraf:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1393)


dastgir:
Seems Good

hemagx:
So, what's the point of save same monster id repeated in array ?

hemagx:
You can do .@mobid = monster.
making an global array or global variable to hold it is bad idea, also a global array that full of same monster id is bad idea as well.

PR #1394, opened 2016-08-09
Fix login server null pointer access if packet handler in plugin init…

4144:
…iate server termination.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1394)


MishimaHaruna:
Merged, thank you

PR #1402, opened 2016-08-18
Update quest_db.conf to Episode 15.1

Ridley8819:
Episode 14.2
- Devil's Tower
- Faceworm's Nest
- Geffen Magic Tournament
- Ghost Palace

Episode 14.3
- End of Morocc
- Isle of Bios
- Morse' s Cave

Episode 15.1
- Airship Assault
- Charlston Crisis
- Nightmarish Jitterbug
- Sara and Fenrir

Double Space to some older quests to fit the rest of the layout.
Updated comment on old glast heim and sara's memory to fit with the rest.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1402)

### DIFF
--- a/db/quest_db.conf
+++ b/db/quest_db.conf
@@ -121,7 +121,7 @@ quest_db: (
 	Id: 1016
 	Name: "Gaining base levels"
 },
-// Solo Box Quest
+//  Solo Box Quest
 {
 	Id: 1100
 	Name: "Solo at Sphinx Dungeon!"
@@ -584,6 +584,24 @@ quest_db: (
 {
 	Id: 1228
 	Name: "Getting back Professor Worm's memory"
+},
+//  Episode 14.2 Eclage, Memorial Dungeon: Ghost Palace
+{
+	Id: 1260
+	Name: "Cursed Swordman"
+},
+{
+	Id: 1261
+	Name: "Cursed Swordman"
+	TimeLimit: 82800
+},
+{
+	Id: 1262
+	Name: "Cursed Swordman"
+},
+{
+	Id: 1263
+	Name: "Cursed Swordman"
 },
 {
 	Id: 2000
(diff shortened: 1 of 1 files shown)


hemagx:
There's no reason to add this entries in database unless the content assigned with it added as well.
also at some point it can cause issues in trace when content added or removed.

Emistry:
hmm, but why close the PR ? 
isn't that these quests are still required to merge in the future ? Herc still can make use of this and merge it in the future. With these quests added to the db, who know it may encourage other ppl to help complete the rest of the remaining contents?

Yes, it's true that sometime when they create small PR like this, it caused other dev who working on same updates having some extra work to keep track of it. I dont think it would create much issue, since the map-server will just throw a new line of `Duplicate quest ID: xxxxx` .
Isn't that there exist a checklist of what contents has been completed, what's not ? The TODO list.

I remember Herc once said 
> It's better than having nothing...

you cant really expect everyone to create a PR that have 100% of the contents, and ignore/removed all those PR that doesn't meet you expectation. o-o

I believe sometime these actions discourage other people from commit a PR.
> I cant do a huge/complete PR like others did, but I can help provide some small updates that Herc currently missing, but Herc dont want I commit such a small updates, they expecting more... I guess I'll just leave it to other dev who capable to commit a complete contents. let's just wait for it ...`

isnt that right? just saying.

MishimaHaruna:
It's true that something is always better than nothing, but I'd personally prefer to have this kind of incomplete things in a development branch, rather than in the master branch. These entries will likely need to be touched again when working on the script they're meant for (even if just to keep the names consistent with those in the script), and they can't be used at all on their own, so it'd make more sense that each of them was added (or merged to master) at the same time as its script

Ridley8819:
sure no problems, it makes sense to have the updates in branches and not in master.

PR #1408, opened 2016-08-25
Swap calls to send info about solt item to seller and actual item del…

4144:
…etion from cart.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1408)


MishimaHaruna:
Merged, thank you

PR #1410, opened 2016-08-26
Modified monster and areamonster command.

dastgirp:
Now all monster ID's are saved into $@mobid and number of monsters are
saved in $@mobid_count when spawned using above mentioned commands.

---

This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1410)

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -2018,10 +2018,8 @@ ACMD(monster)
 	count = 0;
 	range = (int)sqrt((float)number) +2; // calculation of an odd number (+ 4 area around)
 	for (i = 0; i < number; i++) {
-		int k;
 		map->search_freecell(&sd->bl, 0, &mx,  &my, range, range, 0);
-		k = mob->once_spawn(sd, sd->bl.m, mx, my, name, mob_id, 1, eventname, size, AI_NONE|(mob_id == MOBID_EMPELIUM?0x200:0x0));
-		count += (k != 0) ? 1 : 0;
+		count += VECTOR_LENGTH(mob->once_spawn(sd, sd->bl.m, mx, my, name, mob_id, 1, eventname, size, AI_NONE|(mob_id == MOBID_EMPELIUM?0x200:0x0)));
 	}
 
 	if (count != 0)
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -486,18 +486,22 @@ struct mob_data *mob_once_spawn_sub(struct block_list *bl, int16 m, int16 x, int
 /*==========================================
  * Spawn a single mob on the specified coordinates.
  *------------------------------------------*/
-int mob_once_spawn(struct map_session_data* sd, int16 m, int16 x, int16 y, const char* mobname, int class_, int amount, const char* event, unsigned int size, unsigned int ai) {
+struct mob_spawn_list mob_once_spawn(struct map_session_data* sd, int16 m, int16 x, int16 y, const char* mobname, int class_, int amount, const char* event, unsigned int size, unsigned int ai) {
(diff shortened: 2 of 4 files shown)


Emistry:
Although I would love to see it implemented, but ...
https://github.com/HerculesWS/Hercules/pull/1393

dastgir:
Didn't knew there was already one @Emistry .
However, Ema missed one point soo badly.
The variable store Mob GID and not MobID

hemagx:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/map/mob.c, line 489 \[r1\]](https://reviewable.io:443/reviews/herculesws/hercules/1410#-KQ77A6Div2ebUcmdoga:-KQ77A6Ew7atIOdYDDQi:b36kcwn) ([raw file](https://github.com/herculesws/hercules/blob/5e55be17127a1720bfcea036d9ace635db8a85f2/src/map/mob.c#L489)):*
> ```C
>  * Spawn a single mob on the specified coordinates.
>  *------------------------------------------*/
> struct mob_spawn_list mob_once_spawn(struct map_session_data* sd, int16 m, int16 x, int16 y, const char* mobname, int class_, int amount, const char* event, unsigned int size, unsigned int ai) {
> ```

Instead of returning struct please return error code and take struct pointer as argument for vector.

---

*[src/map/mob.c, line 546 \[r1\]](https://reviewable.io:443/reviews/herculesws/hercules/1410#-KQ77p1ag-AHOfvf5SQ6:-KQ77p1bZufMeZHoWwRZ:b36kcwn) ([raw file](https://github.com/herculesws/hercules/blob/5e55be17127a1720bfcea036d9ace635db8a85f2/src/map/mob.c#L546)):*
> ```C
>  * Spawn mobs in the specified area.
>  *------------------------------------------*/
> struct mob_spawn_list mob_once_spawn_area(struct map_session_data* sd, int16 m, int16 x0, int16 y0, int16 x1, int16 y1, const char* mobname, int class_, int amount, const char* event, unsigned int size, unsigned int ai)
> ```

Instead of returning struct please return error code and take struct pointer as argument for vector.

---

*[src/map/mob.h, line 411 \[r1\]](https://reviewable.io:443/reviews/herculesws/hercules/1410#-KQ78312aK78bZ6uHHLE:-KQ78312aK78bZ6uHHLF:b-nvue2g) ([raw file](https://github.com/herculesws/hercules/blob/5e55be17127a1720bfcea036d9ace635db8a85f2/src/map/mob.h#L411)):*
> ```objc
> };
> 
> VECTOR_STRUCT_DECL(mob_spawn_list, int);
> ```

please use int32, using compiler depended data types is not preferred and old code should be refactored at some point.

---

*[src/map/mob.h, line 431 \[r1\]](https://reviewable.io:443/reviews/herculesws/hercules/1410#-KQ78IbrM-ffo0Bogy2W:-KQ78IbrM-ffo0Bogy2X:b-1xowfy) ([raw file](https://github.com/herculesws/hercules/blob/5e55be17127a1720bfcea036d9ace635db8a85f2/src/map/mob.h#L431)):*
> ```objc
> 	int mora[5];
> 
> 	struct mob_spawn_list spawn_list;
> ```

global variables is not preferred unless it's private global, also depends on the other changes i suggested it may not be needed at all. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1410)*

hemagx:
@dastgir i would have appreciated if you actually checked previous PR code ;)

Asheraf:
it store the GID too but duplicated in the array ^^"

PR #1417, opened 2016-09-04
Cleanup CT & GH Nightmare

Ridley8819:
removed unnecessary lines (thx to Aleos)
added missing mapflags for c_tower2_ and c_tower3_

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1417)


npc/mapflag/nomemo.txt
```diff
@@ -33,7 +33,7 @@
 //=========================================================================
 //= Mapflag: Disable Warp Portal memory.
 //================= Current Version =======================================
-//= 1.8b
+//= 1.8.1
```
hemagx: 1.8b means it was 1.8.1 previously, please change it to 1.8.2

npc/re/quests/quests_aldebaran.txt
```diff
@@ -45,41 +46,18 @@ c_tower1,232,222,5	script	Belljamin Button#ct	4_M_RUSMAN2,{
...
-		case 3:
-			mes("[Belljamin Button]");
-			mes("I am here all day for the safety of adventurers. I always alert them about the danger. That's why I need some pay.");
-			emotion e_swt;
-			close;
-		}
 		break;
+	case 2:
 	case 3:
+		if (@menu == 2) {
```
hemagx: Okay, this horrible feature which added to switch code (save to character temporary variable).
I would prefer if you assign the switch return to a NPC temporary variable .@

PR #1422, opened 2016-09-08
Update some packet ids for newer clients and add navigation script command.

4144:
Changed functions:
clif_skill_nodamage
clif_heal
clif_hominfo
clif_vendingreport
clif_buyingstore_update_item
clif_bg_hp

New script function:
~~~
*navigateto("<map>"{,<x>,<y>,<flag>,<hide_window>,<monster_id>,<char_id>});
~~~

Information about packets get from 3CeAM and rAthena emulators code.
Based on this reports:
https://gitlab.com/evol/evol-tools/blob/f9e6037c7759139338f54cf3234f171e669da35f/servergreps/hercules/20150805/hercules_3ceam_outpackets.txt
https://gitlab.com/evol/evol-tools/blob/f9e6037c7759139338f54cf3234f171e669da35f/servergreps/hercules/20150805/hercules_rathena_outpackets.txt

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1422)


4144:
I cant run gravity client, because this need testing by other people.

Ridley8819:
Started with testing

Ridley8819:
clif_skill_nodamage
 - don't feel any difference
clif_heal
- works
clif_hominfo
- isn't supported by any client yet
clif_vending[report
- works, closes issue#1384
-- is the "tax" of 2% intended? i know there is one if autotrading, but i was ingame and instead of 100/1000/10000 zeny i got 98/980/9800
clif_buyingstore_update_item
- works
clif_bg_hp
- works
new commad navigateto
- works, has some long loading times in some cases (prontera to izlude will air ship enabled or similar) but it was always like this. navigation on the same map works instant and smooth

MishimaHaruna:
The vending tax isn't only when autotrading, it's on all vending transactions

About navigateto, I'd expect the processing time to be the same as when you manually set the destination from the in-game navigation settings window, is that not it?

Ridley8819:
About navigateto, I'd expect the processing time to be the same as when you manually set the destination from the in-game navigation settings window, is that not it?

Yes thats what i meant :)

Ridley8819:
that new commit fixed issue Fixes #1424

PR #1427, opened 2016-09-12
Update Intro (Renewal)

Ridley8819:
This is a kro playing conversion and a rough translation by myself.
- New starting iz_int 18 27 *attention* this requires[ the new iz_int map](http://herc.ws/board/topic/12600-starter-map-iz-int-both-versions)
- for the navigation to work, you're required to use the latest [navi luas whichi pr'd here](https://github.com/ROClientSide/Translation/pull/67/files)
- iteminfo.lub and questid2display is already [in the master branch](https://github.com/ROClientSide/Translation)
- iz_int01 - 04 and int_land01 - 04 are added and full functional, but not used by default (we use only iz_int and int_land while aegis splits the users to different maps)
- the rest of the script was updated to latest script standards

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1427)


conf/char/char-server.conf
```diff
@@ -106,8 +106,8 @@ char_configuration: {
 			// Start point (Renewal)
 			start_point_re: {
 				map: "iz_int"
```
dastgir: I thought the map too was changed

npc/re/warps/cities/izlude.txt
```diff
@@ -72,6 +73,82 @@ izlude_b,52,172,0	warp	iz04_b	2,2,izlude_in,74,161
...
+		sleep 8000;
+	}
+	end;
+}
+
+iz_int,47,30,3	script	noviship#room1-2	WARPNPC,2,2,{
+	end;
+
+OnTouch:
+	@display = 0;
```
dastgir: Why this?

npc/re/warps/cities/izlude.txt
```diff
@@ -72,6 +73,82 @@ izlude_b,52,172,0	warp	iz04_b	2,2,izlude_in,74,161
...
+iz_int,27,30,0	script	noviship#room1-1	WARPNPC,2,2,{
+	end;
+
+OnTouch:
+	warp(strnpcinfo(NPC_MAP), 51, 30);
+	end;
+OnInit:
+	while (true) {
+		emotion e_gasp, 0;
+		sleep 8000;
```
dastgir: I guess something bad here.
The npc shows emotion every 8 second? I thought it showed when player is near the npc.

conf/char/char-server.conf
Ridley8819: The map changed because it was replaced, it is still iz_int. Above I provided a download for both iz_int maps

npc/re/warps/cities/izlude.txt
Ridley8819: Nope it isn't, it's a time interval.

npc/re/warps/cities/izlude.txt
Ridley8819: To clear the variable of wounded swordsman in case player switches map (classchange command). I thought its cleared after the end but it wasn't the case. Works fine if player logs out in iz_int but if switched maps the npc sprite didn't change back.

npc/re/warps/cities/izlude.txt
MishimaHaruna: An infinite loop constantly running? Please avoid this no matter what. (besides, it wouldn't work unless you enable `freeloop`, but regardless, it's terrible and should be avoided at all costs)

Are you sure you don't need something like `showevent` here?

PR #1437, opened 2016-09-18
fix some items bonus

Asheraf:
bSkillAtk not increasing damage for some skills

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1437)

### DIFF
--- a/db/re/item_db.conf
+++ b/db/re/item_db.conf
@@ -22064,19 +22064,22 @@ item_db: (
 	EquipLv: 130
 	View: 13
 	Script: <"
-		bonus bInt,getrefine();
-		bonus bDex,getrefine();
-		if(getrefine()>4) {
-			bonus2 bSkillAtk,WM_REVERBERATION,10;
-			bonus2 bVariableCastrate,WM_REVERBERATION,-10;
+		bonus bInt, getrefine();
+		bonus bDex, getrefine();
+		if(getrefine() > 4) {
+			bonus2 bSkillAtk, WM_REVERBERATION_MELEE, 10;
+			bonus2 bSkillAtk, WM_REVERBERATION_MAGIC, 10;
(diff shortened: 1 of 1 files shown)


db/re/item_db.conf
```diff
@@ -49442,9 +49442,9 @@ item_db: (
 	Weight: 10
 	Loc: 2
 	Script: <"
-		bonus2 bSkillAtk,WM_REVERBERATION,50;
-		if (getequipweaponlv(EQI_HAND_R) == 4) bonus2 bSkillAtk,WM_REVERBERATION,50;
-		if (getrefine() >= 10) bonus2 bSkillAtk,WM_REVERBERATION,50;
+		bonus2 bSkillAtk,WM_REVERBERATION_MELEE,50;
```
dastgir: according to me,
WM_REVERBERATION_MAGIC too should be added
Let me know what you think about it

Asheraf: after testing it looks you're right it should increase the damage for the both hits the melee and the magic.

Asheraf:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[db/re/item_db.conf, line 49445 at r2](https://reviewable.io:443/reviews/herculesws/hercules/1437#-KRx_PjJ0j3RXIAg7b48-r2-49445:-KT16vLapATzOp_qebWi:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/eb131481c74389638f17aa89a6ad2b013a17aa60/db/re/item_db.conf#L49445)):*
<details><summary><i>Previously, Asheraf (Asheraf) wrote…</i></summary>
> after testing it looks you're right it should increase the damage for the both hits the melee and the magic.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1437)*

PR #1450, opened 2016-09-30
Add different fixes for gcc 7 warnings.

4144:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1450)

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -557,6 +557,7 @@ int skillnotok (uint16 skill_id, struct map_session_data *sd)
 				clif->skill_fail(sd,skill_id,USESKILL_FAIL_THERE_ARE_NPC_AROUND,0);
 				return 1;
 			}
+			FALLTHROUGH
 		case MC_IDENTIFY:
 			return 0; // always allowed
 		case WZ_ICEWALL:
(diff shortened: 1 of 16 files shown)


src/map/skill.c
```diff
@@ -7630,8 +7643,11 @@ int skill_castend_nodamage_id(struct block_list *src, struct block_list *bl, uin
 				sc_start4(src,bl,type,100,skill_lv,skill_id,src->id,skill->get_time(skill_id,skill_lv),1000));
 		#ifndef RENEWAL
 			if (sd) skill->blockpc_start (sd, skill_id, skill->get_time(skill_id, skill_lv)+3000);
-		#endif
+				break;
+			__attribute__ ((fallthrough));
```
dastgir: fallthrough after break?

4144: break here inside if
This mean code may ignore break

hemagx:
can you please include the compile log?

src/map/skill.c
dastgir:
```
if (sd) skill->blockpc_start (sd, skill_id, skill->get_time(skill_id, skill_lv)+3000);
```
if only includes skill->blockpc_start

PR #1451, opened 2016-09-30
use mediumint for equip_locations in item_db.sql & item_db_re.sql

gdesatrigraha:
Trying to fix this error when importing `item_db_re.sql`.

```
ERROR 1264 (22003) at line 9768: Out of range value for column 'equip_locations' at row 1
```
Seems we are trying to insert `65536` to an `unsigned smallint` which could only handle `65535`.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1451)

### DIFF
--- a/src/plugins/db2sql.c
+++ b/src/plugins/db2sql.c
@@ -457,7 +457,7 @@ void itemdb2sql_tableheader(void)
 			"  `equip_jobs` bigint(20) UNSIGNED DEFAULT NULL,\n"
 			"  `equip_upper` tinyint(8) UNSIGNED DEFAULT NULL,\n"
 			"  `equip_genders` tinyint(2) UNSIGNED DEFAULT NULL,\n"
-			"  `equip_locations` smallint(4) UNSIGNED DEFAULT NULL,\n"
+			"  `equip_locations` mediumint(8) UNSIGNED DEFAULT NULL,\n"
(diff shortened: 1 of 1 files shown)


dastgir:
Those are auto-generated files and should not be edited ever.
Instead, edit this file: https://github.com/HerculesWS/Hercules/blob/master/src/plugins/db2sql.c
Search for 
```
"  `equip_locations` smallint(4) UNSIGNED DEFAULT NULL,\n"
```
Change with
```
"  `equip_locations` mediumint(8) UNSIGNED DEFAULT NULL,\n"
```

gdesatrigraha:
@dastgir once I updated `db2sql.c`, do I need to run something to regenerate the sql files?

hemagx:
No, our bot will take care of it but first we need to check if this check
would work with our source if at all? Or the reason of this issue?

dastgir:
@hemagx There's locations in renewal which have value > 65535 (Shadow Equipments),
Also since item_db(_re).sql isn't used anywhere in hercules source, it would not affect other things.
@gdesatrigraha Just change the lines in db2sql, everything else will be autogenerated by the Hercules bot.

gdesatrigraha:
Updated guys.
Also checked in the wiki installation guide: `If your Control Panel software or website requires it, you may also import the item, monster and monster skill databases (not necessary for the correct operation of Hercules)`. So it seems that hercules doesn't need this data.

PR #1452, opened 2016-09-30
Add missing checks in common directory

4144:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1452)

### DIFF
--- a/src/common/HPM.c
+++ b/src/common/HPM.c
@@ -103,6 +103,7 @@ void hplugin_export_symbol(void *value, const char *name)
 void *hplugin_import_symbol(char *name, unsigned int pID)
 {
 	int i;
+	nullpo_retr(NULL, name);
 	ARR_FIND(0, VECTOR_LENGTH(HPM->symbols), i, strcmp(VECTOR_INDEX(HPM->symbols, i)->name, name) == 0);
 
 	if (i != VECTOR_LENGTH(HPM->symbols))
@@ -133,6 +134,7 @@ bool hplugin_iscompatible(char* version) {
 bool hplugin_exists(const char *filename)
 {
 	int i;
+	nullpo_retr(false, filename);
 	for (i = 0; i < VECTOR_LENGTH(HPM->plugins); i++) {
 		if (strcmpi(VECTOR_INDEX(HPM->plugins, i)->filename,filename) == 0)
 			return true;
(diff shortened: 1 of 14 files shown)


src/common/HPM.c
```diff
@@ -207,6 +209,7 @@ bool hplugin_data_store_validate(enum HPluginDataTypes type, struct hplugin_data
 	struct hplugin_data_store *store;
 	nullpo_retr(false, storeptr);
 	store = *storeptr;
+	nullpo_retr(false, store);
```
MishimaHaruna: `store` can be NULL here (it's checked below). The function is a bit convoluted (my fault), but all cases should be handled.

src/common/HPM.c
```diff
@@ -294,13 +298,13 @@ void *hplugins_getFromHPData(enum HPluginDataTypes type, uint32 pluginID, struct
 {
 	int i;
 
+	if (!store)
```
MishimaHaruna: it's okay for `store` to be a NULL pointer here (it gets initialized by `HPM->data_store_validate()`).

The check was correct after the call

src/common/HPM.c
```diff
@@ -322,13 +326,13 @@ void hplugins_removeFromHPData(enum HPluginDataTypes type, uint32 pluginID, stru
 	struct hplugin_data_entry *entry;
 	int i;
 
+	if (!store)
```
MishimaHaruna: Same as the above case, `store` can be NULL, and it gets initialized in the function (it's marked as [in,out] in the documentation)

src/common/core.c
```diff
@@ -292,9 +302,12 @@ const char *cmdline_arg_source(struct CmdlineArgData *arg) {
  * @param options   options associated to the command-line argument. @see enum cmdline_options.
  * @return the success status.
  */
-bool cmdline_arg_add(unsigned int pluginID, const char *name, char shortname, CmdlineExecFunc func, const char *help, unsigned int options) {
+bool cmdline_arg_add(unsigned int pluginID, const char *name, char shortname, CmdlineExecFunc func, const char *help, unsigned int options)
+{
 	struct CmdlineArgData *data = NULL;
 
+	nullpo_retr(false, name);
+	nullpo_retr(false, help);
```
MishimaHaruna: `help` can be NULL (if a command doesn't have any help message)

src/common/utils.c
```diff
@@ -108,6 +113,10 @@ static char* checkpath(char *path, const char *srcpath)
 {
 	// just make sure the char*path is not const
 	char *p = path;
+
+	nullpo_retr(NULL, path);
```
MishimaHaruna: According to the check below, `path` and `srcpath` seem like they can safely be NULL

src/common/utils.c
```diff
@@ -170,6 +179,9 @@ static char* checkpath(char *path, const char*srcpath)
 	// just make sure the char*path is not const
 	char *p=path;
 
+	nullpo_retr(NULL, path);
```
MishimaHaruna: Same as above

Issue #841, opened 2015-10-31
(default server misconfiguration) Skill range bypass: 14 cells skills usage


100percent-zz:
Just a followup to this issue:

http://herc.ws/board/tracker/issue-8302-potnd-bypass-range-skill-delay/

This bug is still existent in the present build of Hercules ( and rathena too! ) but not on any eathena servers.

How to reproduce:

1 - Record packet via WPE / RPE.
2 - Send the recorded packet.
3 - Try to send the packet outside the range of skills.

NOTE: If you followed the temporary fix on the link above...it would only fix the vertical and horizontal range of the skill..but if you try the diagonal way - the bug still persists! I have a screenshot of a guild abusing this, please see screenshot below:

http://i.imgur.com/TPmzNxt.jpg

As you can see, their guild programmer made a program that modify the client binaries to prevent the client from sending the move packet when using skills...that's why it turns out to be behaving like it's only sending packets without moving - and the bug shows up! Notice how he is casting dispell outside the correct range. Please fix this, a cheat that is on its way on becoming famous is abusing it ( xrag ) and its not healthy for the competitive WoE community and the whole RO community in general.

kyeme:
@MishimaHaruna @4144

anacondaqq:
i can confirm this issue since 2012. Still not fixed, and very dangerous.

100percent:
The range exploit is the most OP and must be fixed immediately because this is a GAME BREAKING bug! Imagine 7 cells Asura Strike and Sonic Blow!?

The sad thing here is that...the range hack doesn't work on eathena! What a time to be alive when old codebase are outperforming newer ones and it's left in the dust forever.

I already explained how to reproduce the bug, I hope some devs can try and see the bug for themselves..to see how alarming this stuff exists since forever...

anacondaqq:
oh, @100percent i've report about bugs since 2011, nobody listen to me, i will be surpriced, if someone will react of this. Usually people do not care about it, and do not understand a lot of child-problems is result of this issue. eAthena have a lot of mechanic-breaking things, and most of them fixed already at rAthena & Hercules, but not all, still a lot of them exist, and they are reported at their bugtrackers or at issues.

anacondaqq:
Btw, many people do not even hear about sit bug, when you can break any mechanic behavior because of fast sitting. A lot of software using that shit to make instant asura strike after body relocation. And this bug still not fixed:)

4144:
[...]

PR #1457, opened 2016-10-03
Fixes Range Exploit

Ridley8819:
issue #841
[Original Commit by Playtester on rAthena](https://github.com/rathena/rathena/commit/6f74f67da060a063183e1147dfc893c9454af853)
- Players will no longer get +2 range server-sided when using a target spell
- This will prevent players from modifying the client to get extra range

---

This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1457)


anacondaqq:
As result you will get this bug.
https://github.com/rathena/rathena/issues/1427

dastgir:
@anacondaqq I guess, that would be caused by this commit: https://github.com/rathena/rathena/commit/2c86ee435670168ff2710b1a135ac7bf001f81b2

dastgir:
@MishimaHaruna @4144

PR #1459, opened 2016-10-03
CRI should only be calculated if flag have SCB_CRI.

### LINKED ISSUE #1442: show_katar_crit_bonus:
i don't know if it supposed to behave like this:

1. set at battle config "show_katar_crit_bonus: true"
2. wear infiltrator
3. buff your self with enchant poison

findings:
- critical will increase by 2x each time you buff.

dastgir:
Fixes #1442

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1459)

### DIFF
--- a/src/map/status.c
+++ b/src/map/status.c
@@ -3806,13 +3806,15 @@ void status_calc_bl_main(struct block_list *bl, /*enum scb_flag*/int flag)
 	}
 
 	if(flag&SCB_CRI && bst->cri) {
-		if (st->luk == bst->luk)
+		if (st->luk == bst->luk) {
 			st->cri = status->calc_critical(bl, sc, bst->cri, true);
-		else
+		} else {
 			st->cri = status->calc_critical(bl, sc, bst->cri + 3*(st->luk - bst->luk), true);
+		}
+		if (battle_config.show_katar_crit_bonus && bl->type == BL_PC && BL_UCAST(BL_PC, bl)->status.weapon == W_KATAR) {
+			st->cri *= 2;
+		}
 	}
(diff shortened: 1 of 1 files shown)


Ridley8819:
fix works, confirming this

src/map/status.c
```diff
@@ -3806,13 +3806,15 @@ void status_calc_bl_main(struct block_list *bl, /*enum scb_flag*/int flag)
 	}
 
 	if(flag&SCB_CRI && bst->cri) {
-		if (st->luk == bst->luk)
+		if (st->luk == bst->luk) {
```
Ridley8819: why do you use those brackets here? those aren't needed at all

anacondaqq: i guess this is good style =)

dastgir: As mentioned above :)

hemagx: if there's conditions inside each others it's prefers to use { } even if one lined, it's easier to spot it this way.
Dastgir code style is the right here :)

src/map/status.c
```diff
@@ -3806,13 +3806,15 @@ void status_calc_bl_main(struct block_list *bl, /*enum scb_flag*/int flag)
...
 			st->cri = status->calc_critical(bl, sc, bst->cri, true);
-		else
+		} else {
 			st->cri = status->calc_critical(bl, sc, bst->cri + 3*(st->luk - bst->luk), true);
+		}
+		if (battle_config.show_katar_crit_bonus && bl->type == BL_PC && BL_UCAST(BL_PC, bl)->status.weapon == W_KATAR) {
+			st->cri <<= 1;
```
hemagx: please don't use >> and << instead of * 2 and / 2.
it's compiler job to optimize, and it's not bitwise field to do any bit operations on.

dastgir:
@Ridley8819

PR #1246, opened 2016-04-06
King's Grace cooldown

ajslashsd:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1246)

PR #1466, opened 2016-10-08
Added RG King's Grace cooldown

Ridley8819:
closes #1246
reference:
http://irowiki.org/wiki/King's_Grace

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1466)

### DIFF
--- a/db/re/skill_db.conf
+++ b/db/re/skill_db.conf
@@ -32971,6 +32971,13 @@ skill_db: (
 	SplashRange: 5
 	CastTime: 1000
 	SkillData1: 5000
+	CoolDown: {
+		Lv1: 100000
+		Lv2: 90000
+		Lv3: 80000
+		Lv4: 70000
+		Lv5: 60000
+	}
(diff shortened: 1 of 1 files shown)


dastgir:
As per IRC talk, we need those extra Level defined. you can just leave those as it is.
@hemagx or @MishimaHaruna would be able to better explain those LvX: lines

dastgir:
@Ridley8819 those lines were added in https://github.com/HerculesWS/Hercules/commit/61a240c20d46c00ee93b6861c1dfcfce5683ff15 from PR https://github.com/HerculesWS/Hercules/pull/1352 , so this one would revert the fix, so, better leave it,
We just need Cooldown thing from this PR

Ridley8819:
From what I talked with @4144 they aren't needed, that's why I removed them in the first place.

Ridley8819:
1st commit removed, this now only adds CD for King's Grace

Ridley8819:
@dastgir

PR #1472, opened 2016-10-09
Change *getcharid to use constants

Asheraf:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1472)

### DIFF
--- a/db/constants.conf
+++ b/db/constants.conf
@@ -3656,6 +3656,13 @@ constants_db: {
 	NPC_NAME_HIDDEN:  2
 	NPC_NAME_UNIQUE:  3
 	NPC_MAP:         4
+	
+	comment__: "getcharid"
+	CHAR_ID_CHAR:         0
+	CHAR_ID_PARTY:        1
+	CHAR_ID_GUILD:        2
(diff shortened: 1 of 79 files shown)


db/constants.conf
```diff
@@ -3656,6 +3656,13 @@ constants_db: {
 	NPC_NAME_HIDDEN:  2
 	NPC_NAME_UNIQUE:  3
 	NPC_MAP:         4
+	
+	comment__: "getcharid"
+	ID_CHAR:         0
+	ID_PARTY:        1
+	ID_GUILD:        2
+	ID_ACCOUNT:      3
+	ID_BATTLEGROUND: 4
```
Emistry: Maybe making the prefix more unique than just `ID_*` ?

Something like
```
CHARID_CHAR
CHARID_PARTY
CHARID_GUILD
CHARID_ACCOUNT
CHARID_BG
```
Who know we might end up create same constant in future? haha

hemagx: I would suggest to make it CHAR_PARTY_ID and so.
after this change PR will be merged.

Issue #1221, opened 2016-03-26
NPC allow changing of equipments while interacting


Emistry:
# issue Preview

When first talk to the NPC...
![GitHub Logo](http://i.imgur.com/1kvuwGO.png)

After the **Shop Window** opened.
![GitHub Logo](http://i.imgur.com/dMwNqN3.gif)
# How to Reproduce the Issue

```
// Enable all NPC to allow changing of equipments while interacting? (Note 1)
// Script commands 'enable_items/disable_items' will not be override. (see doc/script_commands.txt)
// 1 : yes(official)
// 0 : no
item_enabled_npc: 0
```
1. Change the setting in **conf/battle/item.conf**.
2. Create several items with same ID. _(Example: Katana [3])_
3. Place the items in the Hotkey bars.
4. Load the NPC script. https://pastebin.com/kZmUQx8D
5. Talk to the NPC and spam click the Hotkey to switch equipments.
# Result
- At first, your character unable to switch equipment no matter how you spam click the hotkey.
- After **Enter** pressed, the **Shop window** is opened.
  - first few click, might still success to block equipment switching.
  - **afterward, it will success to switch the equipment**.
- although success to switch equipment at the end, the selling progress might still failed, but the behaviour isn't correct. Players are suppose remain unable to exchange equipment.

josetaas:
Yeah, I remember this being abused before on servers with a costume preview NPC. It seems that ```end``` will lead to a ```free_state``` call which sets ```npc_id``` to 0. Hence, after a short moment, you can start equipping items again. The only way I can think of at the moment is to just flag shops with a script attached to them as an NPC interaction, as well.

MishimaHaruna:
Other than incorrect behavior (my opinion is that it's stupid to restrict actions without a reason, but oh well, so much for emulating Aegis limitations and their lack of user input validation), there really shouldn't be anything abusable here.

This kind of script seems pretty much an abuse of the scripting system. After an 'end', the script terminates, and none of its side effects is supposed to have effect anymore (correct behavior would be that the effect of disable_items ends immediately). Expecting it to still work sounds wrong (and if it does, I hope it wasn't intentional)

Emistry:
yes currently it doesn't seem like bringing any negative effect to server and it's also a trick/hack to use npc like this. :D

but isn't opening a shop would also consider as "interaction with NPC" ?

MishimaHaruna:
[...]

PR #1475, opened 2016-10-12
Fixes Equip Switch Bug for quest shop

SamuelHercules:
as stated at http://herc.ws/board/topic/13459-good-day-preview-bug-quest-shop-npc/

---

This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1475)

### DIFF
--- a/npc/custom/quests/quest_shop.txt
+++ b/npc/custom/quests/quest_shop.txt
@@ -16,6 +16,7 @@
 //= 1.5 Replaced categories with shop IDs.
 //= 1.6 Added support for purchasing stackables.
 //= 1.6a Added support for previewing costumes and robes.
+//= 1.6b Added 'disable_items' command.
 //============================================================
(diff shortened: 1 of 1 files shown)


Emistry:
hmm i think the issue would still exist, like this here.
https://github.com/HerculesWS/Hercules/issues/1221

Asheraf:
@emistry then spaming the equipment would break It?

SamuelHercules:
@Asheraf done edit

Ridley8819:
I'll doubt this change will make the script running as intended, it seems it has more flaws, but I'm not sure how we handle scripts from the custom folder?
@MishimaHaruna? We talked about this before

Ridley8819:
Since I'm not sure how are the future plans for support of the custom is handled, I'm closing this. But thank you for your intention to improve this!

PR #1486, opened 2016-10-23
Fixed airships.txt

MuzTank:
Fixed airships.txt if player and NPC get same number of dice but below than 8 then it was only showing NPC name

---

This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1486)


Asheraf:
@dastgir

PR #1495, opened 2016-10-27
Update Duel System Cooldown

Emistry:
- Able to set cooldown in `seconds` format.
- New `duel_time_interval` value will be applied to any existing cooldown
  if new value is reloaded in-game using `@reloadbattleconf`.
- Display the cooldown tick before can create new duel.
## Example:

image

---

This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1495)

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -8048,10 +8048,11 @@ ACMD(duel)
 		return false;
 	}
 
-	if (!duel->checktime(sd)) {
+	int64 diff = duel->difftime(sd);
+	if (diff > 0) {
 		char output[CHAT_SIZE_MAX];
-		// "Duel: You can take part in duel only one time per %d minutes."
-		sprintf(output, msg_fd(fd,356), battle_config.duel_time_interval);
+		// "Duel: You can take part in duel again after %d secconds."
+		sprintf(output, msg_fd(fd,356), (int)diff);
 		clif->message(fd, output);
 		return false;
 	}
@@ -8101,10 +8102,11 @@ ACMD(leave)
 
 ACMD(accept)
 {
-	if (!duel->checktime(sd)) {
+	int64 diff = duel->difftime(sd);
+	if (diff > 0) {
 		char output[CHAT_SIZE_MAX];
-		// "Duel: You can take part in duel only one time per %d minutes."
-		sprintf(output, msg_fd(fd,356), battle_config.duel_time_interval);
+		// "Duel: You can take part in duel again after %d seconds."
+		sprintf(output, msg_fd(fd,356), (int)diff);
 		clif->message(fd, output);
 		return false;
 	}
--- a/src/map/duel.c
+++ b/src/map/duel.c
@@ -41,27 +41,12 @@ struct duel_interface *duel;
  *------------------------------------------*/
 static void duel_savetime(struct map_session_data *sd)
 {
-	time_t clock;
-	struct tm *t;
-
-	time(&clock);
-	t = localtime(&clock);
-
-	pc_setglobalreg(sd, script->add_variable("PC_LAST_DUEL_TIME"), t->tm_mday*24*60 + t->tm_hour*60 + t->tm_min);
+	pc_setglobalreg(sd, script->add_variable("PC_LAST_DUEL_TIME"), (int)time(NULL));
 }
 
-static int duel_checktime(struct map_session_data *sd)
+static int64 duel_difftime(struct map_session_data *sd)
 {
-	int diff;
-	time_t clock;
-	struct tm *t;
-
-	time(&clock);
-	t = localtime(&clock);
-
-	diff = t->tm_mday*24*60 + t->tm_hour*60 + t->tm_min - pc_readglobalreg(sd, script->add_variable("PC_LAST_DUEL_TIME") );
-
-	return !(diff >= 0 && diff < battle_config.duel_time_interval);
+	return (pc_readglobalreg(sd, script->add_variable("PC_LAST_DUEL_TIME")) + battle_config.duel_time_interval - (int)time(NULL));
 }
 
 static int duel_showinfo_sub(struct map_session_data *sd, va_list va)
@@ -233,7 +218,7 @@ void duel_defaults(void)
 	duel->reject = duel_reject;
 	duel->leave = duel_leave;
 	duel->showinfo = duel_showinfo;
-	duel->checktime = duel_checktime;
+	duel->difftime = duel_difftime;
 
 	duel->init = do_init_duel;
 	duel->final = do_final_duel;
--- a/conf/map/battle/misc.conf
+++ b/conf/map/battle/misc.conf
@@ -101,7 +101,7 @@ duel_allow_teleport: false
 // Autoleave duel when die
 duel_autoleave_when_die: true
 
-// Delay between using @duel in minutes
+// Delay between using @duel in seconds
 duel_time_interval: 60
 
 // Restrict duel usage to same map
--- a/conf/messages.conf
+++ b/conf/messages.conf
@@ -368,7 +368,7 @@
 353: Duel: The Player is in the duel already.
 354: Duel: Invitation has been sent.
 355: Duel: You can't use @duel without @reject.
-356: Duel: You can take part in duel once per %d minutes.
+356: Duel: You can take part in duel again after %d seconds.
 357: Duel: Invalid value.
 358: Duel: You can't use @leave. You aren't a duelist.
 359: Duel: You've left the duel.
--- a/src/map/duel.h
+++ b/src/map/duel.h
@@ -52,7 +52,7 @@ struct duel_interface {
 	void (*reject) (const unsigned int did, struct map_session_data* sd);
 	void (*leave) (const unsigned int did, struct map_session_data* sd);
 	void (*showinfo) (const unsigned int did, struct map_session_data* sd);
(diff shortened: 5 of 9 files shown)


Emistry:
bump fore review? status?
@MishimaHaruna @dastgir @hemagx  etc...

src/map/duel.c
```diff
@@ -41,25 +41,19 @@ struct duel_interface *duel;
...
-	t = localtime(&clock);
 
-	pc_setglobalreg(sd, script->add_str("PC_LAST_DUEL_TIME"), t->tm_mday*24*60 + t->tm_hour*60 + t->tm_min);
+	pc_setglobalreg(sd, script->add_str("PC_LAST_DUEL_TIME"), (int)time(NULL));
 }
 
 int duel_checktime(struct map_session_data* sd) {
-	int diff;
+	int diff = 0;
 	time_t clock;
```
hemagx: use int64 instead

src/map/duel.c
```diff
@@ -41,25 +41,19 @@ struct duel_interface *duel;
...
-	struct tm *t;
-
 	time(&clock);
```
hemagx: you can just clock = time(NULL);

src/map/duel.c
```diff
@@ -41,25 +41,19 @@ struct duel_interface *duel;
...
-	t = localtime(&clock);
 
-	diff = t->tm_mday*24*60 + t->tm_hour*60 + t->tm_min - pc_readglobalreg(sd, script->add_str("PC_LAST_DUEL_TIME") );
+	diff = pc_readglobalreg(sd, script->add_str("PC_LAST_DUEL_TIME")) + battle_config.duel_time_interval - (int)time(NULL);
```
hemagx: why did you call time(NULL) again? tho you saved it in clock?

src/map/duel.c
```diff
@@ -41,25 +41,19 @@ struct duel_interface *duel;
...
 
-	return !(diff >= 0 && diff < battle_config.duel_time_interval);
+	return diff;
```
hemagx: you can just put the whole thing in the return and get rid of all the other variables and not-needed calls.

src/map/atcommand.c
```diff
@@ -7695,17 +7696,18 @@ ACMD(duel) {
...
+		// "Duel: You can take part in duel again after %d secconds."
+		sprintf(output, msg_fd(fd,356), diff);
 		clif->message(fd, output);
 		return false;
 	}
 
 	if (message[0]) {
 		if (sscanf(message, "%12u", &maxpl) >= 1) {
-			if (maxpl < 2 || maxpl > 65535) {
+			if (maxpl < 2 || maxpl > MAX_DUEL) {
```
hemagx: I don't know how duel command works, but you've replaced a 65535 with 1024 in this check..
so what is that exactly ?
and MAX_DUEL seem to be the possible active duels server wide.

Emistry: I thought this `MAX_DUEL` are created to set the limit of max player to duel ...
tho I never heard anyone would any that high amount of duel in server before, they would just pick a pvp map rather than create duel and invite so many peoples.

I will revert this changes.

src/map/duel.c
Emistry: you meant `int64 diff = 0;` or `int64 clock` ??

src/map/duel.c
Emistry: you meant change `time(&clock);` into `clock = time(NULL);` ??

src/map/duel.c
Emistry: will add this changes on next update.

Smokexyz:
@Emistry This is good but the check has failed. Please update the HPM hooking function for `duel_checktime()` as you have changed the return type from `int` to `int64`.

Emistry:
as told by @hemagx previously, I am not suppose to touch/add/edit/delete any codes in the HPM functions, he said there exist somekind of cronjob that will auto handle it.

Smokexyz:
@Emistry you can go ahead and edit the HPM function, it will not cause any issues.

Emistry:
I know... I did it in first commit, but @hemagx  told to remove it.

Smokexyz:
@Emistry I will review and approve, don't worry.

Ridley8819:
@Emistry, dastgir said the same thing recently (to add Hpm change directly)

src/map/atcommand.c
```diff
@@ -7746,10 +7748,11 @@ ACMD(leave) {
...
 
 ACMD(accept) {
-	if (!duel->checktime(sd)) {
+	int64 diff = duel->checktime(sd);
+	if (diff > 0) {
 		char output[CHAT_SIZE_MAX];
-		// "Duel: You can take part in duel only one time per %d minutes."
-		sprintf(output, msg_fd(fd,356), battle_config.duel_time_interval);
+		// "Duel: You can take part in duel again after %d seconds."
+		sprintf(output, msg_fd(fd,356), diff);
```
Smokexyz: Consider the use of function `txt_time` from atcommand.c. It could be added to the atcommand interface and used elsewhere.

Emistry: `duel->checktime()` current return int64 ... the `txt_time` use `int` ... 
beside I believe that nobody ever set the duel settings to higher than 1 minute, I am sure they set it as less as possible, so the minute or hour or even day display are kinda useless, and hence unnecessary processing.

Smokexyz: Good that you pointed that out, you can use `time_t` as a return type for a more architecture compliant integer return type.

Smokexyz: Good that you pointed that out, for a more architectural compliant data type you can use `time_t`. As for the duel setting limit, it's just better to leave it flexible.

src/map/atcommand.c
AnnieRuru: MAX_DUEL is the maximum number of DUEL group allow to create server wide
if somehow all the players (1027++ players ?) makes it own duel, which is highly unlikely
the `duel_create` function just return 0 without any error ...

I think better make another variable like MAX_DUEL_PLAYERS and save it as 65535 ?
since this 65535 actually has a meaning ...

Emistry: create another variable that aren't used in any other places look unnecessary.

anyway this is outdated diff...

EyesOfAHawk: hope you would consider finishing emistry. great work dont waste :)

src/map/duel.c
```diff
@@ -39,29 +39,12 @@ struct duel_interface *duel;
...
-static void duel_savetime(struct map_session_data *sd)
-{
-	time_t clock;
-	struct tm *t;
-
-	time(&clock);
-	t = localtime(&clock);
-
-	pc_setglobalreg(sd, script->add_variable("PC_LAST_DUEL_TIME"), t->tm_mday*24*60 + t->tm_hour*60 + t->tm_min);
+static void duel_savetime(struct map_session_data* sd) {
```
4144: at functions ``{`` should be on new line

src/map/duel.c
```diff
@@ -39,29 +39,12 @@ struct duel_interface *duel;
...
-	time_t clock;
-	struct tm *t;
-
-	time(&clock);
-	t = localtime(&clock);
-
-	diff = t->tm_mday*24*60 + t->tm_hour*60 + t->tm_min - pc_readglobalreg(sd, script->add_variable("PC_LAST_DUEL_TIME") );
-
-	return !(diff >= 0 && diff < battle_config.duel_time_interval);
+static int64 duel_checktime(struct map_session_data* sd) {
```
4144: ``{`` should be at new line

src/map/duel.c
```diff
@@ -39,29 +39,12 @@ struct duel_interface *duel;
...
-{
-	time_t clock;
-	struct tm *t;
-
-	time(&clock);
-	t = localtime(&clock);
-
-	pc_setglobalreg(sd, script->add_variable("PC_LAST_DUEL_TIME"), t->tm_mday*24*60 + t->tm_hour*60 + t->tm_min);
+static void duel_savetime(struct map_session_data* sd) {
+	pc_setglobalreg(sd, script->add_str("PC_LAST_DUEL_TIME"), (int)time(NULL));
```
MishimaHaruna: Is there any reason to use `script->add_str()` instead of the previous `script->add_variable()`

note: add_variable marks the string you're adding as a variable name, add_str does not - this means that the variable can't be accessed by a script (unless the script is loaded before this function is called, and the name then gets marked as variable by the script, but that's just a coincidence, and not the intended way for it to work)

src/map/duel.c
```diff
@@ -39,29 +39,12 @@ struct duel_interface *duel;
...
-	struct tm *t;
-
-	time(&clock);
-	t = localtime(&clock);
-
-	diff = t->tm_mday*24*60 + t->tm_hour*60 + t->tm_min - pc_readglobalreg(sd, script->add_variable("PC_LAST_DUEL_TIME") );
-
-	return !(diff >= 0 && diff < battle_config.duel_time_interval);
+static int64 duel_checktime(struct map_session_data* sd) {
+	return (pc_readglobalreg(sd, script->add_str("PC_LAST_DUEL_TIME")) + battle_config.duel_time_interval - (int)time(NULL));
```
MishimaHaruna: Same as above about the add_str bit

src/map/duel.c
Emistry: didn't aware about the different, I was probably copy from somewhere else, for examples.
fixed.

PR #1505, opened 2016-11-04
Fix reading addBattleConf config from includes.

Smokexyz:
This fixes the reading of battle_configuration from all conf/map/battle/*.conf files instead of just conf/map/battle.conf. 
(As it was intended I believe.)

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1505)


Smokexyz:
Wait there's a mishap while parsing the config keys 😛

hemagx:
It is intended the way it is now, to give ability to load configs from custom block in battle.conf or in battle.conf itself. (the main one not the on in conf/battle/battle.conf).

PR #1514, opened 2016-11-16
Itemdb AegisName fixes

### LINKED ISSUE #830: Unable to use AegisName as constant  if it contain full stop symbol

MishimaHaruna:
This is a general review of the item DB data, focusing mostly on the AegisName.

- There were many inconsistencies between the AegisName in the pre-renewal item DB and their renewal counterparts, for the same items
- From a check against other sources (i.e. Divine-Pride), many of our AegisNames seemed made up, especially for the items added in recent times.
- Some AegisNames even contained special symbols that couldn't be converted to constants. One especially funny (or sad) case was `Thanatos_Two-Handed_Staff`, which was not only inserted in the item DB, but was also distractedly used in a script, producing hilarious (or disastrous) results. (hint: `Thanatos_Two-Handed_Staff == (Thanatos_Two) - (Handed_Staff) == 0`.)
- While working on the item DB, I also realized that some items had the wrong sub-type: `Gymnastics_Ribbon` and `Floral_Mic_Of_Igu` were marked as musical instruments rather than whips. This has been fixed (but a deeper review will be part of another upcoming pull request.)
- This also contains the related pull request #956, rebased.

Closes #956
Fixes #830

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1514)


Lemongrass3110:
Having a dot inside a variable or constant name seems so unlike what you usually go for.
For a lot of other cases you were blocking things off, as far as I have seen, because you did not want to mimic stupid official behaviour and then you end up doing a pull request like this? I do not know what else to say about this, but to say that I am really surprised.
It is also somewhat confusing to see you doing all that hard work just for adding dot support to those names and replacing a lot of other symbols by underscores.
I do not mean to offend you in any way with this comment, but I rather want to make you aware of what you are going to do with these changes.

MishimaHaruna:
That's an interesting remark, thank you. Yes, I'm not fond of those names, just like how I hate the ones that have an apostrophe in the middle (and we already support).
I'd personally prefer to replace anything except `A-Za-z0-9_` with an underscore (or alternatively, just suppress the characters without replacing them with anything). That'd mean either `Tiger_s_Skin` or `Tigers_Skin`. If I'm not the only one to support this, then I'll gladly retire the commits 031487c and ccfd5c5.

On a side note, I do have medium/long-term plans about using the `.` symbol in the script engine, for a more object-oriented purpose.

Emistry:
prefer to remain with `A-Za-z0-9_` ... `Tigers_Skin` ...   
the support for using `dot` doesn't look like a good method/user friendly?
> Why should a "constant name" consist a dot inside it? Look weird ...

it would look more weird when the `.` symbol support in script are implemented..
it could create confusion between `constant name` and `object-oriented variable` ?

Lemongrass3110:
Personally I would just leave apostrophes out. By taking your example as a reference: Tiger's_Skin will be a lot more readable if it was Tigers_Skin as if it was Tiger_s_Skin, like Emistry already mentioned.

All others could be left out, if this does not cause a duplicated name. Otherwise it should be replaced by a underscore like you already did.

Having a dot, an exclamation mark or a hyphen inside your constant will possibly just cause a big confusion.
- Hyphen's are the worst example I have to admit, because it looks like a arithmetical operation between two variables or constants rather than a single variable or constant name.
- The exclamation mark is also a character you should think twice about supporting, because it could also be read as an operator.

I know this goes beyond the coverage of your pull request, since you already replaced hyphens and exclamation marks by underscores right now, but I just wanted to point out why it would not be a good idea to implement support for those too.

If you really want to make object-oriented approach available to Hercules' script engine I would rather suggest you to use something unused, not a dot which is already in use as a prefix for npc variables. I would rather use a `->` for field access for example.

MishimaHaruna:
Thanks for the feedback. I agree about the apostrophe removal. I'll edit this pull request.

About the `.`, don't assume we'll keep using that as prefix! One of the reasons for the object-oriented syntax is in order to get rid of those ridiculous prefixes (including the awful `'instancevars`). I'd rather move many of those prefixed variables to `npc.foo`, `pc.foo`, `instance.foo`, `server.foo` etc (or something similar). That'd free up the `$`-prefixed namespace for the most commonly used variables (`.@scopevars`) and the un-prefixed namespace would be left for commands, functions and constants (allowing us to do some validation and typo checking at parse time).

MishimaHaruna:
Updated, ready for review

PR #1519, opened 2016-11-22
Fix wrong command usage at the sign quest

### LINKED ISSUE #1518: The sign quest (dancing part)
The official "the sign quest" (https://github.com/HerculesWS/Hercules/blob/master/npc/quests/the_sign_quest.txt) does not work as it should. At the dancing part, it should be displaying the "EF_SUMMONSLAVE" effect to signal the player which way to "dance" but it does not. It doesn't show any sort of effect at all despite it being present in the script. Something's missing somewhere?

Asheraf:
disabled npc's do not show the effects
Fixes #1518 .

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1519)

### DIFF
--- a/npc/quests/the_sign_quest.txt
+++ b/npc/quests/the_sign_quest.txt
@@ -5165,8 +5165,8 @@ OnInit:
 	end;
 
 OnEnable:
-	specialeffect EF_SUMMONSLAVE;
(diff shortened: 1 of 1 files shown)


dastgir:
Merged , thank you

PR #1521, opened 2016-12-01
Fixed areawarp cell check

Lemongrass3110:
The script command areawarp has been checking the wrong target map for ages. It has been checking the mapindex rather than the mapid and therefore returned a wrong result.

Original commit: rathena/rathena@ac2ba09

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1521)

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -6391,6 +6391,9 @@ int buildin_areawarp_sub(struct block_list *bl, va_list ap)
 		pc->randomwarp(sd, CLR_TELEPORT);
 	} else if (x3 != 0 && y3 != 0) {
 		int max, tx, ty, j = 0;
+		int16 m;
+
+		m = map->mapindex2mapid(index);
 
 		// choose a suitable max number of attempts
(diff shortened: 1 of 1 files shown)


src/map/script.c
```diff
@@ -6391,6 +6391,9 @@ int buildin_areawarp_sub(struct block_list *bl, va_list ap)
 		pc->randomwarp(sd, CLR_TELEPORT);
 	} else if (x3 != 0 && y3 != 0) {
 		int max, tx, ty, j = 0;
+		int16 m;
+
+		m = map_mapindex2mapid(index);
```
dastgir: map->mapindex2mapid

Lemongrass3110: I knew I should have hit compile first... Sorry

AnisotropicDefixation:
This is sorta related, unitwarp also fails, at least when compiled with older GCC versions.

hemagx:
@AnisotropicDefixation maybe you should post information about this issue in a new issue.

PR #1522, opened 2016-12-02
Job Class vs Job MapID cleanup

MishimaHaruna:
This pull request clarifies some confusion between the Job Classes (client-compatible enumerated constants) and the Job MapIDs (bitmasks used by the map server) and the variables that use them.

See the individual commits messages for details on the actual issues that are fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1522)


Lemongrass3110:
Why not strictly go with jobclass or job_class instead of class to not reuse a (possible) keyword?
I know you do not plan to support C++ but it might be a better decision if you are going to rename it anyway.

PR #1525, opened 2016-12-07
Clear wedding_sign variable when divorce.

Emistry:
http://herc.ws/board/topic/9608-reset-variable-on-divorce/

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1525)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1300,7 +1300,7 @@ Linux* )
 	LIBS="$LIBS -ldl"
 	;;
 FreeBSD*)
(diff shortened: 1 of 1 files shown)


MishimaHaruna:
Umm, can't we just clear it from the script that sets it? It's only needed to keep track of the marriage applications, and never needed again. I never understood why we keep it set forever.  (if you marry two characters through atcommands, it doesn't even get set)

Emistry:
if the `wedding_sign` variable isn't some official things, then ya, it's better to just fix the script. 
I have no idea why a permanent variable was used tho...

PR #1530, opened 2016-12-20
Changed behavior when party leader leaves

### LINKED ISSUE #1107: Party dismissed when leader leave
Test Mode: Pre-Re
Hash: 1ac0a4831923f7c02338696d83a009371c28c6ad
Client: 2010-04-20aRagexeRE

If i remember correctly, when party leader leave the party (/leave)
the party will not dismissed, and the settings (exp and items) will be reseted,
But in Herc, your party will be dismissed.

guilherme-gm:
Now party leader is changed to the second user. Closes #1107

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1530)

### DIFF
--- a/src/char/int_party.c
+++ b/src/char/int_party.c
@@ -614,7 +614,6 @@ int mapif_parse_PartyLeave(int fd, int party_id, int account_id, int char_id)
 {
 	struct party_data *p;
 	int i,j;
-	unsigned int leader;
 
 	p = inter_party->fromsql(party_id);
 	if( p == NULL )
@@ -634,10 +633,7 @@ int mapif_parse_PartyLeave(int fd, int party_id, int account_id, int char_id)
 		return 0; //Member not found?
 
 	mapif->party_withdraw(party_id, account_id, char_id);
-
-	leader = p->party.member[i].leader; // member's leader state
 	
-	inter_party->tosql(&p->party,PS_DELMEMBER,i);
 	j = p->party.member[i].lv;
 	if (p->party.member[i].online > 0)
 		p->party.count--;
@@ -648,19 +644,10 @@ int mapif_parse_PartyLeave(int fd, int party_id, int account_id, int char_id)
 		inter_party->check_lv(p);
 	}
 
-	if (leader) {
-		// Member was party leader, pick a new leader
-		ARR_FIND(0, MAX_PARTY, i, p->party.member[i].account_id != 0);
-
-		if (i < MAX_PARTY) {
-			// Update party's leader
-			p->party.member[i].leader = 1;
-			inter_party->tosql(&p->party, PS_LEADER, i);
-		}
-	}
-
-	if (inter_party->check_empty(p) == 0)
+	if (inter_party->check_empty(p) == 0) {
+		inter_party->tosql(&p->party, PS_DELMEMBER, i);
 		mapif->party_info(-1, &p->party, 0);
+	}
 	return 0;
 }
 // When member goes to other map or levels up.
--- a/src/char/int_party.c
+++ b/src/char/int_party.c
@@ -639,23 +639,20 @@ int mapif_parse_PartyLeave(int fd, int party_id, int account_id, int char_id)
 	
 	inter_party->tosql(&p->party,PS_DELMEMBER,i);
 	j = p->party.member[i].lv;
-	if(p->party.member[i].online) p->party.count--;
+	if (p->party.member[i].online > 0) p->party.count--;
 	memset(&p->party.member[i], 0, sizeof(struct party_member));
 	p->size--;
-	if (j == p->min_lv || j == p->max_lv || p->family)
-	{
+	if (j == p->min_lv || j == p->max_lv || p->family) {
 		if(p->family) p->family = 0; //Family state broken.
 		inter_party->check_lv(p);
 	}
 
-	if( leader ) {
+	if (leader) {
 		// Member was party leader, pick a new leader
 		i = 0;
-		while (i < MAX_PARTY && p->party.member[i].account_id == 0) {
-			i++;
-		}
+		while (i < MAX_PARTY && p->party.member[i].account_id == 0) i++;
 
-		if( i < MAX_PARTY ) {
+		if (i < MAX_PARTY) {
 			// Update party's leader
 			p->party.member[i].leader = 1;
 			inter_party->tosql(&p->party, PS_LEADER, i);
--- a/src/char/int_party.c
+++ b/src/char/int_party.c
@@ -614,6 +614,7 @@ int mapif_parse_PartyLeave(int fd, int party_id, int account_id, int char_id)
 {
 	struct party_data *p;
 	int i,j;
+	unsigned int leader;
 
 	p = inter_party->fromsql(party_id);
 	if( p == NULL )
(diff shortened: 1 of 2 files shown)


src/char/int_party.c
```diff
@@ -614,6 +614,7 @@ int mapif_parse_PartyLeave(int fd, int party_id, int account_id, int char_id)
 {
 	struct party_data *p;
 	int i,j;
+	unsigned int leader;
```
hemagx: the usage of unsigned integers is not preferred unless explicitly needed, please change it to int

src/map/party.c
```diff
@@ -300,6 +300,18 @@ int party_recv_info(const struct party *sp, int char_id)
 			if (i == MAX_PARTY)
 				added[added_count++] = member_id;
 		}
+		
+		ARR_FIND(0, MAX_PARTY, j, p->party.member[j].leader == 1);
+		if( j == MAX_PARTY ) {
+			// Leader has changed
+			int i;
```
hemagx: i is already defined before, please use another name like j k v

src/char/int_party.c
```diff
@@ -634,25 +635,30 @@ int mapif_parse_PartyLeave(int fd, int party_id, int account_id, int char_id)
...
-		for (j = 0; j < MAX_PARTY; j++) {
-			if (!p->party.member[j].account_id)
-				continue;
-			mapif->party_withdraw(party_id, p->party.member[j].account_id, p->party.member[j].char_id);
-			p->party.member[j].account_id = 0;
+	leader = p->party.member[i].leader; // member's leader state
+	
+	inter_party->tosql(&p->party,PS_DELMEMBER,i);
+	j = p->party.member[i].lv;
+	if(p->party.member[i].online) p->party.count--;
```
hemagx: please change this check to > 0

src/char/int_party.c
guilherme-gm: I've used unsigned integer because party structure uses it (at: p->party.member[i].leader;), it's safe to change to int as its values are only 0 or 1, but it will be different from the structure, so should I change?

hemagx: ah.. no signed/unsigned mix is even worse, keep it unsigned till we change it later.

guilherme-gm:
@hemagx I did the changes, please review it when you have some time

src/map/party.c
```diff
@@ -300,6 +300,18 @@ int party_recv_info(const struct party *sp, int char_id)
 			if (i == MAX_PARTY)
 				added[added_count++] = member_id;
 		}
+		
+		ARR_FIND(0, MAX_PARTY, j, p->party.member[j].leader == 1);
+		if (j == MAX_PARTY) {
+			// Leader has changed
+			int k;
+			ARR_FIND(0, MAX_PARTY, k, sp->member[k].leader == 1);
+			if (i < MAX_PARTY) {
```
hemagx: forgot to change this i ?

src/char/int_party.c
```diff
@@ -634,25 +635,27 @@ int mapif_parse_PartyLeave(int fd, int party_id, int account_id, int char_id)
...
-		p->size--;
-		if (j == p->min_lv || j == p->max_lv || p->family)
-		{
-			if(p->family) p->family = 0; //Family state broken.
-			inter_party->check_lv(p);
+	leader = p->party.member[i].leader; // member's leader state
+	
+	inter_party->tosql(&p->party,PS_DELMEMBER,i);
+	j = p->party.member[i].lv;
+	if (p->party.member[i].online > 0) p->party.count--;
```
hemagx: new line here please :)

src/char/int_party.c
```diff
@@ -634,25 +635,27 @@ int mapif_parse_PartyLeave(int fd, int party_id, int account_id, int char_id)
...
+	p->size--;
+	if (j == p->min_lv || j == p->max_lv || p->family) {
+		if(p->family) p->family = 0; //Family state broken.
+		inter_party->check_lv(p);
+	}
+
+	if (leader) {
+		// Member was party leader, pick a new leader
+		i = 0;
+		while (i < MAX_PARTY && p->party.member[i].account_id == 0) i++;
```
hemagx: use ARR_FIND please

guilherme-gm:
@hemagx done

hemagx:
@MishimaHaruna can you give a final review please :)

src/char/int_party.c
```diff
@@ -634,25 +635,27 @@ int mapif_parse_PartyLeave(int fd, int party_id, int account_id, int char_id)
...
+		p->party.count--;
+	memset(&p->party.member[i], 0, sizeof(struct party_member));
+	p->size--;
+	if (j == p->min_lv || j == p->max_lv || p->family) {
+		if(p->family) p->family = 0; //Family state broken.
+		inter_party->check_lv(p);
+	}
+
+	if (leader) {
+		// Member was party leader, pick a new leader
```
MishimaHaruna: When picking a new leader, you might want to give priority to the online members. It would be really annoying (read: would make this patch almost pointless) if an offline member were to be elected leader here, and everyone else was forced to leave in order to recreate a party they can manage.

src/map/party.c
```diff
@@ -300,6 +300,18 @@ int party_recv_info(const struct party *sp, int char_id)
...
 				added[added_count++] = member_id;
 		}
+		
+		ARR_FIND(0, MAX_PARTY, j, p->party.member[j].leader == 1);
+		if (j == MAX_PARTY) {
+			// Leader has changed
+			int k;
+			ARR_FIND(0, MAX_PARTY, k, sp->member[k].leader == 1);
+			if (k < MAX_PARTY) {
+				clif->PartyLeaderChanged(map->id2sd(sp->member[k].account_id), 0, sp->member[k].account_id);
```
MishimaHaruna: Is there any reason to lookup `map->id2sd(sp->member[k].account_id)`, when the sd pointer is already cached in `p->data[k].sd`?
But either way, `map->id2sd()` may return a `NULL` pointer if the player is offline. The correct behavior here would be to find the first available sd in the party, and then use that one to pass to `clif->PartyLeaderChanged()`. The second argument should be the old account ID, which is still available at the beginning of this function, so you could retrieve it.
If you move this to the `while (removed_count > 0)` block below, and check for the removed member's `.leader`, you can even save an `ARR_FIND()`, and only do this when the leader is actually getting removed.

src/map/party.c
```diff
@@ -300,6 +300,18 @@ int party_recv_info(const struct party *sp, int char_id)
...
+			} else {
+				party->broken(p->party.party_id); // Should not happen, Party is leaderless, disband
```
MishimaHaruna: I know that this is just a sanity check and it should never happen, but you can't call this here. Everything will blow up below, when it tries to edit a party that was just deleted (and its memory released).

src/map/party.c
guilherme-gm: I think when I wrote it I thought `sp` and `p` could be completely different (so indexes wouldn't match), but this doesn't make much sense when I look at it now.
If I prioritize online members to change leader to, can I be sure that if leader is offline the entire party is offline?

MishimaHaruna: It's likely that there are some cases where that isn't true, since operations between the char and map server are asynchronous (I'm thinking of cases where the leader leaves, and another player logs in at the same time)

guilherme-gm:
@MishimaHaruna I just noticed that in `party_recv_info` `while (removed_count > 0)` is always 0, even in clean updated hercules with any party member leaving. Party members seems to be removed on `party_member_withdraw` and this `while` is just to make sure that party is not out of sync.
So by the time `party_recv_info` is called, the old leader is not in `p` anymore so I can't get his `account_id`.
I moved the party leader changing to `party_member_withdraw` so it's possible to have current member state and old party leader data.

guilherme-gm:
@MishimaHaruna can you take a look when you've some time? thanks

MishimaHaruna:
Good point about the `party_recv_info` bit, I didn't realize that.

This looks good to me, I'm going to merge it.

MishimaHaruna:
Merged, thank you very much for your contribution

PR #1542, opened 2017-01-04
Fixed announces from Koschei the Immortal (Moscovia Quests)

landergate:
Some ancient typo, which results in `script_rid2sd: fatal error ! player not attached!` and never showing those messages.

This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1542)


dastgir:
Thanks for Contribution +1

PR #1544, opened 2017-01-13
getstatus() Update

dastgir:
Infinite Duration status would result in error since there was no timers
attached to it.
Added a condition to check if status is having infinite duration and
return value according to it.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1544)

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -11591,14 +11591,16 @@ BUILDIN(getstatus)
 		case 3: script_pushint(st, sd->sc.data[id]->val3); break;
 		case 4: script_pushint(st, sd->sc.data[id]->val4); break;
 		case 5:
-		{
-			const struct TimerData *td = timer->get(sd->sc.data[id]->timer);
+			if (sd->sc.data[id]->infinite_duration) {
+				script_pushint(st, INFINITE_DURATION);
+			} else {
+				const struct TimerData *td = timer->get(sd->sc.data[id]->timer);
 
-			if (td != NULL) {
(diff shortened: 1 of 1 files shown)


src/map/script.c
```diff
@@ -11591,14 +11591,16 @@ BUILDIN(getstatus)
 		case 3: script_pushint(st, sd->sc.data[id]->val3); break;
 		case 4: script_pushint(st, sd->sc.data[id]->val4); break;
 		case 5:
-		{
-			const struct TimerData *td = timer->get(sd->sc.data[id]->timer);
+			if (sd->sc.data[id].infinite_duration) {
```
Lemongrass3110: You need to use -> here. I don't know if you noticed that your build failed.
But actually the assertion in timer_get in timer.c is still wrong.
The only thing you would have to add here is to return infinite duration in the else case(td == NULL).
The error @Ragn0 got is because no return value is pushed to the script engine in this case.

dastgir: Ah, I didn't noticed, was in hurry and with no build environment :x
Yup, I know error was due to no return value, but infinite timers would trigger that ShowError part. And it was sensible to return -1 instead

dastgir: I guess I need to check if some of the status timer expire or I can add else directly.
But it's best to check infinite_duration variable, since it's made for that.

Smokexyz:
@dastgir This looks okay, is it still in progress?

PR #1552, opened 2017-01-21
New map cache system

hemagx:
* new mapcache system!
* Now each map is in separated file so now it will be easier to know
  which map got updated or delete on updates
* Now there's md5 checksum check for each map
* Now the map cache is platform safe, the old format was not packed
  which may result in undefined behavior
* The map cache tool got converted into hercules plugin

-- Before merge, we need forum topic ready for upgrade instructions

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1552)


src/map/map.h
```diff
@@ -1160,7 +1155,7 @@ END_ZEROED_BLOCK;
 	struct map_data *list;
 	/* [Ind/Hercules] */
 	struct eri *iterator_ers;
-	char *cache_buffer; // Has the uncompressed gat data of all maps, so just one allocation has to be made
+	//char *cache_buffer; // Has the uncompressed gat data of all maps, so just one allocation has to be made
```
4144: probably better remove it and not left commented line?

src/map/map.c
```diff
@@ -3240,97 +3239,107 @@ int map_eraseipport(unsigned short map_index, uint32 ip, uint16 port) {
...
-		info = (struct map_cache_map_info *)p;
+int map_readfromcache_v1(FILE *fp, struct map_data *m, unsigned int file_size)
+{
+	struct map_cache_header mheader = {0};
+	uint8 md5buf[16] = {0};
+	int map_size;
+	nullpo_retr(false, fp);
+	nullpo_retr(false, m);
+
+	if (file_size < sizeof(mheader)) {
```
4144: Here should be check of <= sizeof(mheader) because at least one cell should be in map

src/map/map.c
```diff
@@ -3240,97 +3239,107 @@ int map_eraseipport(unsigned short map_index, uint32 ip, uint16 port) {
...
-		if( strcmp(m->name, info->name) == 0 )
-			break; // Map found
+	if (fread(&mheader, sizeof(mheader), 1, fp) < 1) {
+		ShowError("map_readfromcache: faild to read cache header for map '%s'\n", m->name);
+		return 0;
+	}
 
-		// Jump to next entry..
-		p += sizeof(struct map_cache_map_info) + info->len;
+	if (file_size < sizeof(mheader) + mheader.len) {
```
4144: probably also need check mheader.len for correct values

hemagx: resolved

src/map/map.c
hemagx: resolved

src/map/map.h
hemagx: resolved

Ragn0:
Suggestion: Could be possible that maps that are identical in both pre-renewal and renewal, be shared in common folder?

Also, I noticed that some map files that I think should be equal in both pre-renewal and renewal are differents, like "1@cash.mcache" or even "geffen.mcache". Why does these be different in both folders?

MishimaHaruna:
There are indeed some differences between pre-re and re maps. It's possible that the maps were different when those mapcache entries were generated (possibly the pre-renewal one being more outdated?). Keep in mind that the .mcache files added here are generated from the map_cache.dat we had in the repository until now, so there won't be any change in behavior.
It's definitely worth analyzing the differences though (I'll be probably writing some tool to analyze the contents of a .mcache file, or diff two of them)

Ragn0:
It can't be denied the useful this would be with actual results. It hasn't been implemented yet and we can see the differences in maps between pre-re and re. Awesome.

AnisotropicDefixation:
I honestly don't get how is this a good idea at all and not a complete mess, make it optional at least.

MishimaHaruna:
The point here is to make it easier to replace a single map's .mcache data file when it gets updated clientside (rather than having to rebuild the entire mapcache, and end up with garbage in it, like it always happens). It's true that this produces many more files, but they're much smaller (and generate much smaller commits when updated individually), and their total size is the same as the current data file.
The idea is also to eventually move certain data files out of the repository, but there are no near-future plans for this yet.

This change also makes it pretty easy to make corrections to the .mcache files or inspect them, may one wish to do so. See the screenshot for an example of what you can do with a simple plugin (I'll publish that in the Staff Plugins repository after this PR will be completed)

screen shot 2017-02-14 at 04 14 02

iRagno:
@AnisotropicDefixation, in my opinion, the main usage I would bring is to easily make duplicated maps. I'm not used into map duplications (my topic is scripting :P) and being able to duplicate a map by just copy paste one single file would allow me to do more things. Not even talking about sharing map files, it would make easier to implement them having cache precompiled.

I like ideas like this that make easier things for emulation.

Jedzkie:
@MishimaHaruna will this PR be merged in the future?

MishimaHaruna:
@Jedzkie yes, but it still needs a bit more work to add more flexibility to the mapcache plugin.

4144:
is this pr still active?

MishimaHaruna:
Yes, I'm currently testing it in a production environment, with good results so far. It's still missing some tweaks to make it suitable for merging, but I'll take care of those

MishimaHaruna:
This is ready from my side.

I fixed a few errors and loose ends, documented the plugin's command line arguments.

I added an unrelated commit (appveyor configuration file), to prevent appveyor build failures.

MishimaHaruna:
For those interested that might be subscribed to this PR, I pushed the plugin showcased above to the StaffPlugins repository:

Folder: https://github.com/HerculesWS/StaffPlugins/tree/master/Haru/mapcache2txt
Commit: HerculesWS/StaffPlugins@97b696bc1ab744c065a93f5d7cfdeb9dc5b74f18

PR #1562, opened 2017-01-27
remove hardcoded chat area size

mekolat:
This patch removes the hardcoded chat area size and instead reads it from battle.conf

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1562)

### DIFF
--- a/conf/map/battle/client.conf
+++ b/conf/map/battle/client.conf
@@ -69,6 +69,9 @@ pet_hair_style: 100
 // Visible area size (how many squares away from a player can they see)
 area_size: 14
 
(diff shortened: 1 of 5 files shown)


hemagx:
Pretty nice, we always prefer to get rid of hardcoded settings, however as suggestion, you may want to make minimum 1, and maximum (500*500), to ensure it's always in valid range.

mekolat:
`area_size` has a range of `0` to `INT_MAX` so I thought I could use the same settings for `chat_area_size`, but if you want me to I'll change it

PR #1567, opened 2017-02-06
Fixes named items having their names changed when near other players

### LINKED ISSUE #1206: The name from the named item is not permanent
Bug: When you and the owner of the 'Named Item' are on the same map, the 'Named Item' becomes 'Nameless'. But, if the owner of the 'Named Item' is not on your map, the name of the 'Named Item' is as is, nothings change.

ss1

~ I can't reproduced this bug on old revision (october 2015).

guilherme-gm:
Fixes named items having their names changed when near other players/name's owner. Fixes #1206 when merged.

There was 2 issues there:
- first: item name disappeared, because the client needs the character name when it's near
- second: name changing to another player, that's because when the player wich name is on the item walked near the one who had the item, unit_walking was sending the GID of the one who had the item instead of the one walking near, thus changing the name.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1567)

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -1077,9 +1077,11 @@ void clif_set_unit_idle(struct block_list* bl, struct map_session_data *tsd, enu
 #endif
 #if PACKETVER >= 20150513
 	p.body = vd->body_style;
+#endif
+/* Might be earlier, this is when the named item bug began */
+#if PACKETVER >= 20131223
 	safestrncpy(p.name, clif->get_bl_name(bl), NAME_LENGTH);
 #endif
-
 	clif->send(&p,sizeof(p),tsd?&tsd->bl:bl,target);
 
 	if (clif->isdisguised(bl)) {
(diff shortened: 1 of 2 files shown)


4144:
For packet version 20131223 in clif_set_unit_idle, hercules using packet id 0x09dd, this id was introduced in client version 2013-10-23aRagexe and 2013-10-23aRagexeRE. May be this change related to this date?
Without testing exact client versions i cant be sure. Because packet len is -1.

https://gitlab.com/4144/servergreps/blob/master/client_re/packets/2013-10-23aRagexeRE/packets.h#L37
https://gitlab.com/4144/servergreps/blob/master/client_pre/packets/2013-10-23aRagexe/packets.h#L37

Same for other packets with this change.

guilherme-gm:
Maybe, but isn't 0x9dd only used for 2013-12-23 and above?
https://github.com/HerculesWS/Hercules/blob/master/src/map/clif.c#L955
https://github.com/HerculesWS/Hercules/blob/master/src/map/packets_struct.h#L97

Ridley8819:
Is this ready to be merged?

guilherme-gm:
I wasn't able to find a client between 20131023 and before 20131223 so I can't be sure of the PACKETVER

Ridley8819:
thank you for this, I am merging this. Even we can't be sure about the exact date, I believe it is better to have this fix. Blame me if something bad happen

PR #1570, opened 2017-02-08
Added extra parameters to *getgroupid and *npctalk script commands

RandonsRO:
*getgroupid({<char_name>|<account_id>|<char_id>})
    if no parameter is given will return group id for invoking player.
*npctalk(<"message">{,<"npc name">{,<name_display>}})
   display values:
      0: false (default) - shows npc name
      1: true - hides npc name

- Requested by Ridley for official scripts

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1570)


Ridley8819:
you're my man <3

Emistry:
i suggest the `display` rename it to something more meaningful and appropriate. like `hide_npcname` ...
if you want to use the `display` perhaps should default the value to `1 = show npc name`, it doesn't sound right if you want the display but you set the value to `0` or vice versa.

Asheraf:
I believe the name hide edit should be added to unittalk as well

PR #1579, opened 2017-02-19
New Enchants

Ridley8819:
new commit of  #1462

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1579)


db/re/item_db.conf
```diff
@@ -147047,6 +147358,20 @@ item_db: (
...
 },
 {
+	Id: 22826
+	AegisName: "Costume_Enchantment_Stone_Box"
+	Name: "Costume Enchantment Stone Box"
+	Buy: 10
+	Type: 2
+	Weight: 10
+	Script: <"
+		getitem callfunc("F_Rand", Stone_ASPD_Garment, Attack_Stone_Upper, Magic_Stone_Upper, Health_Stone_Upper,
```
dastgir: why F_Rand?
why not item_group or item package?
same with other boxes

db/re/item_db.conf
```diff
@@ -151568,4 +151929,664 @@ item_db: (
...
+	Loc: 1023
+	Script: <"
+		bonus bHealPower, 30;
+		bonus bMdef, 18;
+	">
+},
+{
+	Id: 29117
+	AegisName: "Mercy7Lv"
+	Name: "Mercy7Lv"
```
dastgir: Name it something like "Mercy Lv.7"?

Ridley8819: got to disagree here. Why would want a dot in item names? Haru even did commits to remove special characters from it. Also those are the names i found on rms and dp

db/re/item_db.conf
```diff
@@ -147073,6 +147393,24 @@ item_db: (
...
 	Script: <" getrandgroupitem 22838,1; ">
 },
+{
+	Id: 22868
+	AegisName: "Costume_Enchantment_Stone_Box_5"
+	Name: "Costume Enchantment Stone Box 5"
+	Buy: 10
+	Type: 2
+	Weight: 10
+	Script: <" getrandgroupitem 22868, 1; ">
```
Asheraf: same here

db/re/item_db.conf
```diff
@@ -147073,6 +147393,24 @@ item_db: (
...
+},
+{
+	Id: 22905
+	AegisName: "Costume_Enchantment_Stone_Box_6"
+	Name: "Costume Enchantment Stone Box 6"
+	Buy: 10
+	Type: 2
+	Weight: 10
+	Script: <" getrandgroupitem 22905, 1; ">
```
Asheraf: same here

db/re/item_db.conf
```diff
@@ -147047,6 +147358,15 @@ item_db: (
 	">
 },
 {
+	Id: 22826
+	AegisName: "Costume_Enchantment_Stone_Box"
+	Name: "Costume Enchantment Stone Box"
+	Buy: 10
+	Type: 2
+	Weight: 10
+	Script: <" getrandgroupitem 22826, 1; ">
```
Asheraf: Use aegis name?

db/re/item_db.conf
Emistry: i think the special characters removal only for `AegisName`.

dastgir: ^ Exactly, special characters removal was for AegisName.
"Name" Field is for use at like getitemname command. Having Mercy7Lv would in my opinion create confusion.

Smokexyz:
This is good, why the hold-up?

Ridley8819:
I no time yet to change the item names, will do tomorrow

Ridley8819:
Thank you @Asheraf

db/re/item_db.conf
MishimaHaruna: Name field is the user-visible name. It should match the official name used in the client where possible.

AegisName should be an identifier usable in the scripting language (so only letters, underscores, numbers).

PR #1586, opened 2017-02-24
getd enhancement

mekolat:
It feels awkward to have to use `attachrid` just to get a variable of another player.. with this improved `getd` you can now do it without ever changing the rid. It can be used for both players and npcs, so it's basically a `getvariableofnpc` on steroids.

## new syntax
`getd("<var>"{, <GID>{, <default value>}})`

## example
`getd("@var", getcharid(CHAR_ID_ACCOUNT, "player"), -1)`  

The above will return a reference to variable `@var` from player `player`, or else return `-1` if the player could not be found.

<br>
<br>

---
In the future I might make a `getparam(Param, <GID>)` which would return a reference to a param of another player (ie `Zeny`)

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1586)

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -2861,32 +2861,44 @@ struct script_data *get_val(struct script_state* st, struct script_data* data) {
 		const char *str = NULL;
 
 		switch (prefix) {
-			case '@':
+		case '@':
+			if (data->ref) {
+				str = script->get_val_ref_str(st, data->ref, data);
+			} else {
 				str = pc->readregstr(sd, data->u.num);
-				break;
-			case '$':
-				str = mapreg->readregstr(data->u.num);
-				break;
-			case '#':
-				if (name[1] == '#')
-					str = pc_readaccountreg2str(sd, data->u.num);// global
-				else
-					str = pc_readaccountregstr(sd, data->u.num);// local
-				break;
-			case '.':
-				if (data->ref)
-					str = script->get_val_ref_str(st, data->ref, data);
-				else if (name[1] == '@')
-					str = script->get_val_scope_str(st, &st->stack->scope, data);
-				else
-					str = script->get_val_npc_str(st, &st->script->local, data);
-				break;
-			case '\'':
-				str = script->get_val_instance_str(st, name, data);
-				break;
-			default:
+			}
+			break;
+		case '$':
+			str = mapreg->readregstr(data->u.num);
+			break;
+		case '#':
+			if (data->ref) {
+				str = script->get_val_ref_str(st, data->ref, data);
+			} else if (name[1] == '#') {
+				str = pc_readaccountreg2str(sd, data->u.num);// global
+			} else {
+				str = pc_readaccountregstr(sd, data->u.num);// local
+			}
+			break;
+		case '.':
+			if (data->ref) {
+				str = script->get_val_ref_str(st, data->ref, data);
+			} else if (name[1] == '@') {
+				str = script->get_val_scope_str(st, &st->stack->scope, data);
+			} else {
+				str = script->get_val_npc_str(st, &st->script->local, data);
+			}
+			break;
+		case '\'':
+			str = script->get_val_instance_str(st, name, data);
+			break;
+		default:
+			if (data->ref) {
+				str = script->get_val_ref_str(st, data->ref, data);
+			} else {
 				str = pc_readglobalreg_str(sd, data->u.num);
-				break;
+			}
+			break;
 		}
 
 		if (str == NULL || str[0] == '\0') {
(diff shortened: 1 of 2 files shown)


Ridley8819:
´getvariableofnpc´ on steroids. Approved

mekolat:
@Ridley8819 done

Ridley8819:
regarding coding style: please do the brackets { always one the same line as the condition (except for functions)

mekolat:
@Ridley8819 I modified my code to match the coding style, but it seems Hercules code in general doesn't respect its own coding style:
- there's other places where there's a blank line before the `{` in non-functions
- the switch statements are double-indented
- there's lines longer than 120 characters
- there's spaces at some place inside parenthesis, ie `if ( foo == bar )`
- there's tons of places where there is no space after the words `if, switch, for, do, while`

Ridley8819:
@mekolat yes, these are most of the time old codes which weren't modified In ages. For new commits we take care these guidelines are followed :)

4144:
@mekolat feel free to fix any style issues in existing code

mekolat:
@4144 you already know how lazy I am haha, not gonna waste time on that stuff

src/map/script.c
```diff
@@ -2862,13 +2862,18 @@ struct script_data *get_val(struct script_state* st, struct script_data* data) {
 
 		switch (prefix) {
 			case '@':
```
Ridley8819: the cases are on the wrong indentation, they should be on same one was the switch

src/map/script.c
```diff
@@ -2909,13 +2917,18 @@ struct script_data *get_val(struct script_state* st, struct script_data* data) {
 		} else
```
Ridley8819: since the command is rewritten heavily, please apply coding style here too. If you do an if / else where one is using brackets, you do it for the other condition as well even if not needed
`} else {

src/map/script.c
```diff
@@ -2909,13 +2917,18 @@ struct script_data *get_val(struct script_state* st, struct script_data* data) {
...
 			switch( prefix ) {
 				case '@':
```
Ridley8819: cases on wrong indentation again

src/map/script.c
```diff
@@ -16470,8 +16489,62 @@ BUILDIN(getd) {
...
+		st->state = END;
+		return false;
+	}
+
+	if (script_hasdata(st, 3)) {
+		bl = map->id2bl(script_getnum(st, 3));
+
+		if (!bl) {
+			// being not found, push default value
+			if(script_hasdata(st, 4))
```
Ridley8819: conditions inside other conditions always use brackets

src/map/script.c
```diff
@@ -16470,8 +16489,62 @@ BUILDIN(getd) {
...
+		script->reportdata(script_getdata(st, 2));
+		script_pushnil(st);
+		st->state = END;
+		return false;
+	}
+
+	if (script_hasdata(st, 3)) {
+		bl = map->id2bl(script_getnum(st, 3));
+
+		if (!bl) {
```
Ridley8819: please use `bl != NULL` instead

mekolat: more like `bl == NULL`

Ridley8819: *facepalm yes xD

src/map/script.c
mekolat: this is the already existing indentation, should I change it?

src/map/script.c
mekolat: this is the existing code, not mine

mekolat:
@Ridley8819 applied your requested changes

dastgir:
:+1:

MishimaHaruna:
Can you please refrain from merging pull requests without even giving me a chance to read them, especially if they touch the source part?

mekolat:
@MishimaHaruna sorry, didn't expect it to get merged, I can open a new PR if you want me to change something

src/map/script.c
```diff
@@ -2861,32 +2861,44 @@ struct script_data *get_val(struct script_state* st, struct script_data* data) {
 		const char *str = NULL;
 
 		switch (prefix) {
-			case '@':
+		case '@':
+			if (data->ref) {
+				str = script->get_val_ref_str(st, data->ref, data);
```
MishimaHaruna: Never ever re-indent a block AND change some lines in it in the same commit. It makes it really hard to review the changes and/or merge them for users.

mekolat: sorry about that, didn't get a chance to see your comments before it got merged

MishimaHaruna: Yeah, not your fault. The error was to merge this so suddenly without proper review.

MishimaHaruna:
Side note: I'm against this. The use of `getd()` and `setd()` should be DISCOURAGED, while this seems to encourage it when you want to grab another character's variable. I would have gone for a `getvariableofpc()` command (compare with `getvariableofnpc()`), which, if someone really wants to use `getd()`, can be paired in some ugly construct such as `getvariableofpc(getd(.@foo$), .@target)`. But now it's too late, meh.

src/map/script.c
```diff
@@ -16493,8 +16552,63 @@ BUILDIN(getd) {
 	if (sscanf(buffer, "%99[^[][%d]", varname, &elem) < 2)
 		elem = 0;
 
+	if (strlen(varname) < 1) {
```
MishimaHaruna: In proper C, this is `if (varname[0] == '\0') {`

mekolat:
@MishimaHaruna want me to make a revert commit?

PR #1584, opened 2017-02-22
Implementation of unit controlling script commands.

Smokexyz:
```Setunitdata <GUID>, <DataType>, <Val1> {, <Val2>, <Val3>}```
```Getunitdata <GUID>, <DataType> {, <Variable>}```
```Getunitname <GUID>```
```Setunitname <GUID>, <name>```

Performs alteration and retrieval of real-time unit data for game objects of the following types -
NPCs, Pets, Monsters, Homunuculus', Mercenaries, Elementals.

Applicable data types (available as script constants) -
```
UDT_TYPE:       Unit Type
UDT_SIZE:       Unit Size
UDT_LEVEL:      Level
UDT_HP:         Current HP
UDT_MAXHP:      Max HP
UDT_SP:         SP
UDT_MAXSP:      MAX SP
UDT_MASTERAID:  Master Account ID (for Summons)
UDT_MASTERCID:  Master Char ID (for Summons)
UDT_MAPIDXY:    Warp a Unit to a map.
UDT_WALKTOXY:   Make a unit walk to certain co-ordinates.
UDT_SPEED:      Unit Speed
UDT_MODE:       Mode (Mobs)
UDT_AI:         Unit AI Type
UDT_SCOPTION:   Status Options.
UDT_SEX:        mm Sex.
UDT_CLASS:      Class of the unit.
UDT_HAIRSTYLE:  Hair Style of the unit.
UDT_HAIRCOLOR:  Hair Color.
UDT_HEADBOTTOM: Headgear Bottom Sprite.
UDT_HEADMIDDLE: Headgear Middle Sprite.
UDT_HEADTOP:    Headegar Top Sprite.
UDT_CLOTHCOLOR: Cloth Color.
UDT_SHIELD:     Shield Sprite.
UDT_WEAPON:     Weapon Sprite.
UDT_LOOKDIR:    Face direction.
UDT_CANMOVETICK:Stop a unit from move for n seconds.
UDT_STR:        Unit STR.
UDT_AGI:        Unit AGI.
UDT_VIT:        Unit VIT.
UDT_INT:        Unit INT.
UDT_DEX:        Unit DEX.
UDT_LUK:        Unit LUK.
UDT_ATKRANGE:   Attack range of a unit.
UDT_ATKMIN:     Min Atk of a unit.
UDT_ATKMAX:     Max Atk of a unit.
UDT_MATKMIN:    Min MATK of a unit.
UDT_MATKMAX:    Max MATK of a unit.
[...]

PR #1571, opened 2017-02-09
Script commands Update

RandonsRO:
## Updated _getgroupid_ script command

- [x] Added an extra optional parameter which lets you specify a player to get his/her groupid you can use the account_id, char_id or character name.

## Updated _npctalk_ and _unittalk_ script commands

- [x] Added an extra optional parameter to show/hide npc/unit name. By default unit/npc names will be shown.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1571)

PR #1589, opened 2017-02-25
allow strcharinfo() and strnpcinfo() to take a GID

mekolat:
`strcharinfo()` and `strnpcinfo()` can now work without an attached rid, and can get details from another GID than the attached pc/npc. It can also return a default value if the attached or target gid is not found.

* `strcharinfo(0)` will return the name of the attached player, or an empty string if there is no attached player
* `strcharinfo(0, "foobar")` will return the name of the attached player, or `foobar` if there is no attached player
* `strcharinfo(0, "foobar", id)` will return the name of the player with account id `id`, or `foobar` if the player could not be found

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1589)

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -8685,39 +8685,48 @@ BUILDIN(getguildmember)
  *------------------------------------------*/
 BUILDIN(strcharinfo)
 {
-	int num;
 	struct guild* g;
 	struct party_data* p;
-	struct map_session_data *sd = script->rid2sd(st);
-	if (sd == NULL) //Avoid crashing....
-		return true;
+	struct map_session_data *sd;
 
-	num=script_getnum(st,2);
-	switch(num) {
-		case 0:
-			script_pushstrcopy(st,sd->status.name);
-			break;
-		case 1:
-			if( ( p = party->search(sd->status.party_id) ) != NULL ) {
-				script_pushstrcopy(st,p->party.name);
-			} else {
-				script_pushconststr(st,"");
-			}
-			break;
-		case 2:
-			if( ( g = sd->guild ) != NULL ) {
-				script_pushstrcopy(st,g->name);
-			} else {
-				script_pushconststr(st,"");
-			}
-			break;
-		case 3:
-			script_pushconststr(st,map->list[sd->bl.m].name);
-			break;
-		default:
-			ShowWarning("buildin_strcharinfo: unknown parameter.\n");
-			script_pushconststr(st,"");
-			break;
+	if (script_hasdata(st, 4))
+		sd = map->id2sd(script_getnum(st, 4));
+	else
+		sd = script->rid2sd(st);
+
+	if (sd == NULL) {
+		if(script_hasdata(st, 3)) {
+			script_pushcopy(st, 3);
+		} else {
+			script_pushconststr(st, "");
+		}
+		return true;
+	}
+
+	switch (script_getnum(st, 2)) {
+	case 0:
+		script_pushstrcopy(st, sd->status.name);
+		break;
+	case 1:
+		if ((p = party->search(sd->status.party_id)) != NULL) {
+			script_pushstrcopy(st, p->party.name);
+		} else {
+			script_pushconststr(st, "");
+		}
+		break;
+	case 2:
+		if ((g = sd->guild) != NULL) {
+			script_pushstrcopy(st, g->name);
+		} else {
+			script_pushconststr(st, "");
+		}
+		break;
+	case 3:
+		script_pushconststr(st, map->list[sd->bl.m].name);
+		break;
+	default:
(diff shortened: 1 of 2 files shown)


src/map/script.c
```diff
@@ -8685,39 +8685,45 @@ BUILDIN(getguildmember)
...
-	if (sd == NULL) //Avoid crashing....
+	struct map_session_data *sd;
+
+	if (script_hasdata(st, 3))
+		sd = map->id2sd(script_getnum(st, 3));
+	else
+		sd = script->rid2sd(st);
+
+	if (sd == NULL) {
+		if(script_hasdata(st, 4))
```
Ridley8819: for conditions inside other conditions we always use brackets, regardless if they needed or not, also missing space after `if`

src/map/script.c
```diff
@@ -8685,39 +8685,45 @@ BUILDIN(getguildmember)
...
+			script_pushcopy(st, 4);
+		else
```
Ridley8819: same as above, for conditions inside conditions always use brackets

src/map/script.c
```diff
@@ -8734,41 +8740,48 @@ BUILDIN(strcharinfo)
...
+	struct npc_data *nd;
+
+	if (script_hasdata(st, 3))
+		nd = map->id2nd(script_getnum(st, 3));
+	else
+		nd = map->id2nd(st->oid);
+
 	if (nd == NULL) {
-		script_pushconststr(st, "");
+		if(script_hasdata(st, 4))
```
Ridley8819: missing space after `if` and brackets for conditions.

src/map/script.c
```diff
@@ -8734,41 +8740,48 @@ BUILDIN(strcharinfo)
...
+		break;
+	case 2: // # fragment
+		if ((buf = strchr(nd->name,'#')) != NULL)
+			name = aStrdup(buf+1);
+		break;
+	case 3: // unique name
+		name = aStrdup(nd->exname);
+		break;
+	case 4: // map name
+		if (nd->bl.m >= 0) // Only valid map indexes allowed (issue:8034)
```
Ridley8819: which issue you're mentioning here?

mekolat: I kept the existing comment, it's not mine

Ridley8819: ah okay I didn't pay attention to this x)

Ridley8819:
@dastgir any additional comments?

dastgir:
@MishimaHaruna 
Tagging Haru, because I think I have seen haru talked about not liking this thing

Ridley8819:
Right, there was something in #1571

MishimaHaruna:
Looks good to merge. While I'm not especially fond of having this kind of thing implemented in so many commands (it'd be nicer if there was a global facility at the language level to support this), I discussed this with @mekolat, who showed me an interesting use case.

mekolat:
@MishimaHaruna in [TMWA](https://github.com/themanaworld/tmwa) we had commands that allowed to get/set pretty much anything from anything.
- `get(BL_ID, "player name")`
- `set Hp, value, "player name";`

and also the params were all exposed so you were able to do things like
- `if (Hp >= (MaxHp / 2))`
- `Sex = 3;`

it seems #1584 does something similar, and it could easily be extended to return strings too instead of just integers

src/map/script.c
```diff
@@ -8685,39 +8685,48 @@ BUILDIN(getguildmember)
...
+			script_pushstrcopy(st, g->name);
+		} else {
+			script_pushconststr(st, "");
+		}
+		break;
+	case 3:
+		script_pushconststr(st, map->list[sd->bl.m].name);
+		break;
+	default:
+		ShowWarning("script:strcharinfo: unknown parameter.\n");
```
Emistry: should be `ShowWarning("buildin_strcharinfo: unknown parameter.\n");` ??

mekolat: @Emistry seems this should be added to the code style file then. There's a shit ton of buildins using the `script:function: message` syntax so I didn't know which one to use

Emistry: I believe it should be `buildin_xxxxxxx: message`.

MishimaHaruna: We have a mix of both. While the formally correct one would be `buildin_xxxxx` (that's the actual function name), I don't see a reason to enforce it for two reasons:

- those function names are never spelled out, but generated by a macro (so we can't argue that it makes it easy to find the error source, since you can't search the file for that)
- the end user or scripter the message is targeted to, doesn't need to know that it's from a function with that name. They are more interested in knowing that it comes from a script, and from the xxxx command.

I'd accept either one, possibly favoring the `script:xxxxx` format used here.

MishimaHaruna:
Merged, thank you for your contribution!

PR #1593, opened 2017-02-27
Added time limit parameter to setquest

guilherme-gm:
This adds an extra optional parameter to set quest that allows user to give an UNIX epoch time to this quest time limit, that will ignore the one in quest_db.
I used 0 as to use quest_db value to avoid an extra parameter on add_quest and I doubt someone will want to set a quest that ends in the past. Also unsigned int because it's the same as in the quest structure.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1593)


iRagno:
As a reference, this is not only an improvement or suggestion. It has been seen cases where in official server sets time limit with "dinamic" duration (time limit finishes until specific hour on server, and not giving a time limit with fixed duration).

I noticed that behaviour on iRO's 13th Anniversary Quest Event.

src/map/quest.c
```diff
@@ -103,22 +103,23 @@ int quest_pc_login(struct map_session_data *sd)
...
  */
-int quest_add(struct map_session_data *sd, int quest_id)
+int quest_add(struct map_session_data *sd, int quest_id, unsigned int time_limit)
 {
 	int n;
 	struct quest_db *qi = quest->db(quest_id);
 
 	nullpo_retr(-1, sd);
-	if( qi == &quest->dummy ) {
+	if(qi == &quest->dummy) {
```
Ridley8819: if (qi == &quest->dummy) {
you just forgot space after most of your ifs :)

src/map/quest.c
```diff
@@ -103,22 +103,23 @@ int quest_pc_login(struct map_session_data *sd)
...
 		ShowError("quest_add: quest %d not found in DB.\n", quest_id);
 		return -1;
 	}
 
-	if( quest->check(sd, quest_id, HAVEQUEST) >= 0 ) {
+	if(quest->check(sd, quest_id, HAVEQUEST) >= 0) {
```
Ridley8819: if (

src/map/quest.c
```diff
@@ -130,15 +131,17 @@ int quest_add(struct map_session_data *sd, int quest_id)
 	sd->avail_quests++;
 	RECREATE(sd->quest_log, struct quest, sd->num_quests);
 
-	if( sd->avail_quests != sd->num_quests ) {
+	if(sd->avail_quests != sd->num_quests) {
```
Ridley8819: if (

src/map/quest.c
```diff
@@ -130,15 +131,17 @@ int quest_add(struct map_session_data *sd, int quest_id)
...
 		memmove(&sd->quest_log[n+1], &sd->quest_log[n], sizeof(struct quest)*(sd->num_quests-sd->avail_quests));
 	}
 
 	memset(&sd->quest_log[n], 0, sizeof(struct quest));
 
 	sd->quest_log[n].quest_id = qi->id;
-	if( qi->time )
+	if (time_limit > 0)
+		sd->quest_log[n].time = time_limit;
+	else if(qi->time)
```
Ridley8819: else if (

src/map/quest.c
```diff
@@ -147,8 +150,8 @@ int quest_add(struct map_session_data *sd, int quest_id)
 	clif->quest_add(sd, &sd->quest_log[n]);
 	clif->quest_update_objective(sd, &sd->quest_log[n]);
 
-	if( map->save_settings&64 )
-		chrif->save(sd,0);
+	if(map->save_settings&64)
```
Ridley8819: if (

dastgir:
What if someone provided past unix time? it would just accept it .

src/map/quest.c
guilherme-gm: Oops tried to fix everything and forgot them hehe thanks

doc/script_commands.txt
```diff
@@ -1,4 +1,4 @@
-//===== Hercules Documentation ===============================
+﻿//===== Hercules Documentation ===============================
```
dastgir: file ending changed?

guilherme-gm:
@dastgir yes, it would accept and make a quest that is already expired. When I made it I chose to use unix time because I though it would be easier to calculate expire times, but now I can't think in an example that doesn't need `gettimetick(2)`.
I can change to instead of a unix time use just how many seconds from now it will expire, like `quest_db` does, but in this case, `time limit = 0` is a completely valid value, so I need another way to tell if I want or not to set a custom time, maybe `UINT32_MAX`?

Smokexyz:
What is the use of this?

> (time limit finishes until specific hour on server, and not giving a time limit with fixed duration).

makes absolutely no sense because you're contradicting yourself in the second half of the sentence.

Why is it not possible to add a quest with PLAYTIME for this particular purpose?

iRagno:
@Smokexyz: It's (new) official behaviour. Let's try to bring an example.

Let's say there is a repeatable quest with variable time expiration delay that completes on 01:00 hours on official server and there is Player A that sets quest on 08:00 hrs and Player B that sets quest on 21:00 hrs.

Player A will have a quest timer to wait 17 hours and Player B will have a quest timer to wait 4 hours, because that quest specific ends on 01:00 am server time to all players.

Actually in emulator this can't be done. Both players must wait the same time in all cases. If quest has 23 hours of delay on quest_db, players must wait 23 hours without consider the hour they set their quest.

Asheraf:
The idea of smoke about using `TimeLimit` is pretty good as it keep the timers in the database file, but beside that its also good to be able to set the timer dynamically for custom content.

Smokexyz:
@iRagno imho, players waiting for the same amount of time is more just. 
**BUT**, the norm for this kind of behaviour, for official servers in most MMORPGs is to define a global setting for instance cool-down times. For example cool-down settings in a config file -
```
InstanceCoolDownDayOfTheWeek: 0-6
InstanceCooldownTime: 0100
```
This would reset the cool-down of all instances on the day of the week. This is not the case in ALL instances though, so I there should be a better design to address all cool-down situations of the instance system.

iRagno:
Just to be clear, this is not for custom content, the idea for this is to have a more accurate emulation of oficial (and actual) behaviour.

I see your point @Smokexyz and I'm agree with you. It would be better to include new config on quest_db.conf that makes server calculate delay for player instead than setting it on setquest command, because it would require on script to have mechanism to maintain updated the time limit.

However, please keep in mind that this is not a custom feature. It has been seen cases on official server (more exactly iRO), where time limit expires at fixed hours, instead of fixed time to wait.

Asheraf:
@iRagno 
> (more exactly iRO)
this is used in kRO for all the new content.
ex: The new eden group quests, daily quests for illusion of (moonlight, vampire, frozen) all of them have a cooldown that reset at 04:00 am korean time.

Smokexyz:
Then designing this dynamic system of adding time via script command is the same as adding the time limit option in a database. To emulate the same reset system as official servers, you must re-design.

guilherme-gm:
@Smokexyz Although I personally prefer to change the command as it also allows you to do other things if you want to, I can redesign it, in that case, should I use these settings you presented earlier?

> InstanceResetDayOfTheWeek: 0-6
> InstanceResetTimeOfTheDay: 0100

Asheraf:
This system should work in general with quests. I'd prefere avoiding use of
instance as name.

Smokexyz:
@guilherme-gm My example wouldn't work for all cases. For example, how would the reset cool-down of 3 days work? granted that the time of the day would be fixed, though day of the week only works for weekly cool-downs.
Requesting @MishimaHaruna 's input on the matter.

MishimaHaruna:
Is there a problem with using unix times? The idea of keeping the time limits in the quest database makes sense, but if we want to design something flexible enough using just those, we need to redesign a lot of things in it. It'd need to accept monthly, weekly, daily, hourly expiration settings at the very least.

By using an implementation like the one provided in this pull request, you get much more flexibility for free, and it even allows you to design scripts that have different expiration times based on other factors (character's level, class or whatever), whether it is currently needed for official scripts or not.

An improvement I feel like suggesting is a function that can return the unix timestamp of an arbitrary future relative date (i.e. next friday at 7:52). Something like

```
// getcalendartime(<day of month>, <day of week>, <hour>, <minute>);
getcalendartime(7, -1, 4, 27); // Returns the unix timestamp of the next 7 of the month, at 4:27
getcalendartime(-1, SATURDAY, 18, 30); // The next saturday at 18:30
getcalendartime(-1, -1, 0, 0); // The next midnight
```

Something like that would make it much easier to use unix timestamps in functions such as this one.

Smokexyz:
If the intention of the PR was to design something flexible for quests alone, this is a good idea, along with the `getcalendartime` suggestion.

guilherme-gm:
flexibility was indeed one of my main concerns, I like the `getcalendartime` suggestion and I can work on it if you want :)

@Smokexyz Well, instances cooldown are basically quests, so this brings flexibility to everything, right? Sure, in the case of instances if you wanted to customize it you would have to actually open instance scripts and change it there, but would it be worth all the work to think of all cases to put in the db file just because of that? also with `getcalendartime` it wouldn't be that hard to change. just my opinion :)

MishimaHaruna:
@guilherme-gm Sure, please go ahead if you may!

guilherme-gm:
@MishimaHaruna not sure if it's the best way to do it, if there's a better one let me now and I'll update

src/map/script.c
```diff
@@ -20723,6 +20724,82 @@ BUILDIN(mergeitem)
...
+BUILDIN(getcalendartime)
+{
+	struct tm info = { 0 };
+	int day_of_month = script_hasdata(st, 4) ? script_getnum(st, 4) : -1,
+		day_of_week = script_hasdata(st, 5) ? script_getnum(st, 5) : -1,
+		year = date_get_year(),
+		month = date_get_month(),
+		day = date_get_day(),
+		hour = script_getnum(st, 2),
+		minute = script_getnum(st, 3);
```
MishimaHaruna: Would you mind splitting these into separate declarations (add `int` to each line)? I find it cleaner than using commas and indentation, when multiple lines are needed

src/map/script.c
```diff
@@ -20723,6 +20724,82 @@ BUILDIN(mergeitem)
...
+		ShowError("script:getcalendartime: Day of Month in invalid range. Must be between 1 and 31.\n");
+		script_pushint(st, -1);
+		return false;
+	}
+	if (day_of_week > -1 && (day_of_week < 0 || day_of_week > 6)) {
+		ShowError("script:getcalendartime: Day of Week in invalid range. Must be between 0 and 6.\n");
+		script_pushint(st, -1);
+		return false;
+	}
+	if (hour == -1 || minute == -1) {
```
MishimaHaruna: Since you check the range for the other variables, then you should check it for these two as well

src/map/script.c
```diff
@@ -20723,6 +20724,82 @@ BUILDIN(mergeitem)
...
+		} else if (day_of_week == cur_wday) { // Today
+			if (hour < date_get_hour() ||
+				(hour == date_get_hour() && minute < date_get_min())) {
+				info.tm_mday += 7; // Next week
+			}
+		} else if (day_of_week < cur_wday) { // Next week
+			info.tm_mday += (7 - cur_wday + day_of_week);
+		}
+	} else if (day_of_week == -1 && day_of_month == -1) { // Next occurence of hour/min
+		if (hour < date_get_hour() ||
```
MishimaHaruna: I'd avoid so many calls to these date functions, you could just call them once and cache the result

src/map/script.c
```diff
@@ -20723,6 +20724,82 @@ BUILDIN(mergeitem)
...
+		script_pushint(st, -1);
+		return false;
+	}
+
+	if (day_of_month > -1) {
+		if (day_of_month < day) { // Next Month
+			info.tm_mon++;
+		} else if (day_of_month == day) { // Today
+			if (hour < date_get_hour() ||
+				(hour == date_get_hour() && minute < date_get_min())) { // But past time, next month
```
MishimaHaruna: I think the minute checks should be done with `<=` (so it never returns the current timestamp but always the next occurrence of the requested time)

src/map/script.c
```diff
@@ -20723,6 +20724,82 @@ BUILDIN(mergeitem)
...
+				info.tm_mon++;
+			}
+		}
+		info.tm_mday = day_of_month;
```
MishimaHaruna: Something that could be improved is the case you request the timestamp of 31st, during a shorter month. This implementation will likely just return the 1st (or 2nd or 3rd depending on the case) of the next month, which I'm not sure is optimal.

If you want to find the 31st of the next month that actually has 31 days, you could try something like this, to replace this ilne (untested code):

```
do {
    time_t t;
    struct tm *lt;
    info.tm_mday = day_of_month;
    t = mktime(&info);
    lt = localtime(&t);
    info = *lt;
} while (info.tm_mday != day_of_month);

guilherme-gm:
@MishimaHaruna I've just pushed the fixes, sorry for the delay

MishimaHaruna:
Merged, thank you!

Issue #1582, opened 2017-02-20
Pre-Re instances cooldown check


Ridley8819:
Ok so I had a longer discussion with @iRagno regarding instances, and the way we do check for CD in athena, how its done on iRO, and how its done on kRO. So we thought to have a new battleflag in features.conf to offer the following behaviour for instances.

--> athena behaviour = check CD on party leader, allow leader switch (bug?) 
--> iro behaviour = CD quest got applied to every party member, but it doesn't get checked until the instance is completed/destroyed. This allows party member who disconnect, for whatever reason, to rejoin.
--> kro behaviour (will not be implemented immediately since we're missing dialogues for it) = instant apply CD on all party member, no CD applied

This would fix the leader switch bug for all pre re instances, as well as allowing disconnecting party member to reenter in time.

When talking with @iRagno, he was excited to take this task.

Mysteries:
I agree. I think since we're all use to it... we should add it as a setting in the features.conf =/

iRagno:
About new behaviour, that is, disable re-entering the instance when player comes out from instance map, I think we can use actual messages for when cool down has been seted, until having official dialogues that requires ingame test.

In that way, we can bring an option asap according to the needs.

AnisotropicDefixation:
Some of the new instances instead add a quest entry that allows you to reenter freely for the first 5 or so minutes.

Asheraf:
he is talking about `Pre-Re instances`

iRagno:
Added pull request with fix to entrance exploit in Endless Tower, Nydhogur Nest and Orcs Memorial instance dungeons.

Sealed Shrine has disabled reentering instance once was created and needs more investigation if that should be default behaviour. I assume it could be fixed with this, but messages are kind of weird, needs to confirm behaviour with officials.

PR #1594, opened 2017-02-27
Pre-Renewal instances cooldown mode

### LINKED ISSUE #732: Endless Tower CD only applies to party leader

iRagno:
Fixes entrance exploits and changes behavior of instances cooldown with this values:
0 - Classic behaviour with exploit entrance (not recommended)
1 - Classic behaviour without exploit entrance (default)
2 - Updated behaviour to disable re-entrance on same instance
Fixes #732 #1582

### DIFF
--- a/conf/map/battle/feature.conf
+++ b/conf/map/battle/feature.conf
@@ -55,4 +55,11 @@ features: {
 	// Requires: 2014-10-22bRagexe or later
 	// Disabled by default while test version is out; enable at your own risk -- the mean dev.
 	roulette: false
+
+	// Pre-Renewal instances cooldown mode
+	// Fixes entrance exploits and changes behavior of instances cooldown with this values:
+	// 0 - Classic behaviour with exploit entrance (not recommended)
+	// 1 - Classic behaviour without exploit entrance (default)
+	// 2 - Updated behaviour to disable re-entrance on same instance
+	pre_re_instance_mode: 1
 }
--- a/npc/instances/EndlessTower.txt
+++ b/npc/instances/EndlessTower.txt
@@ -312,6 +312,33 @@ L_Enter:
 		mes "The party leader did not generate the dungeon yet.";
 		close;
 	} else {
+		switch (getbattleflag("features/pre_re_instance_mode")) {
+		case 0:
+			break;
+		case 1:
+			.@instance = has_instance2("1@tower");
+			instance_attach(.@instance);
+			if (!'ins_etower_id) {
+				$@ins_etower_id[.@instance] = gettimetick(2);
+				'ins_etower_id = $@ins_etower_id[.@instance];
+			}
+			if (!questprogress(60200,PLAYTIME))
+				ins_etower_id = 'ins_etower_id;
+			if (ins_etower_id == 'ins_etower_id)
+				break;
+			// fall through
+		case 2:
+			.@dun_ent_t = ((etower_timer+604800) - gettimetick(2));
+			.@dun_h = (.@dun_ent_t / 3600);
+			.@dun_m = (.@dun_ent_t - (.@dun_h * 3600)) / 60;
+			.@dun_s = .@dun_ent_t - ((.@dun_h * 3600) + (.@dun_m * 60));
+			mesf("Due to the tower's aftereffects, you cannot enter the dungeon right now, %d hours %d minutes %d seconds left to enter the next dungeon.", .@dun_h, .@dun_m, .@dun_s);
+			next;
+			mes "It is dangerous here. Let me move you to Alberta.";
+			close2;
+			warp "alberta",223,36;
+			end;
+		}
 		mapannounce "e_tower", strcharinfo(PC_NAME)+" of the party, " +getarg(3)+", is entering the dungeon, Endless Tower.",bc_map,"0x00ff99",FW_NORMAL,12;
 		if (getarg(1)) {
 			etower_timer = gettimetick(2);
--- a/npc/instances/NydhoggsNest.txt
+++ b/npc/instances/NydhoggsNest.txt
@@ -224,6 +224,26 @@ L_Enter:
 		mes "You did not request for entrance. Please let your leader request entrance.";
 		close;
 	} else {
+		switch (getbattleflag("features/pre_re_instance_mode")) {
+		case 0:
+			break;
+		case 1:
+			.@instance = has_instance2("1@nyd");
+			instance_attach(.@instance);
+			if (!'ins_nyd_id) {
+				$@ins_nyd_id[.@instance] = gettimetick(2);
+				'ins_nyd_id = $@ins_nyd_id[.@instance];
+			}
+			if (!questprogress(3135,PLAYTIME))
+				ins_nyd_id = 'ins_nyd_id;
+			if (ins_nyd_id == 'ins_nyd_id)
+				break;
+			// fall through
+		case 2:
+			mes "[Yggdrasil Gatekeeper]";
+			mes "The time limit to enter the dungeon has expired. You must wait for the World Tree to stabilize its power before trying to re-enter.";
+			close;
+		}
 		mapannounce "nyd_dun02", getpartyname(getcharid(CHAR_ID_PARTY))+"'s party member "+strcharinfo(PC_NAME)+" has entered Nidhoggur's Nest.",bc_map,"0x00ff99";
 		if (!questprogress(3135)) setquest 3135;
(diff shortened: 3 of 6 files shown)


conf/map/battle/feature.conf
```diff
@@ -55,4 +55,11 @@ features: {
 	// Requires: 2014-10-22bRagexe or later
 	// Disabled by default while test version is out; enable at your own risk -- the mean dev.
 	roulette: false
+
+	// Pre-Renewal instances cooldown mode
+	// Fixes entrance exploits and changes behavior of instances cooldown with this values:
+	// 0 - Classic behaviour with exploit entrance (not recommended)
```
Ridley8819: i think it makes more sense to have 0 - fixed behavior. After all, this is supposed to be a bugfix, but with an option to go back to old behavior. Thats why i think it should be:
0 - Classic behavior without exploit entrance (default)
1 - Athena behavior with exploit entrance (not recommended)
2 - Renewal behavior to disable re-entrance on same instance

conf/map/battle/feature.conf
```diff
@@ -55,4 +55,11 @@ features: {
...
+	// 1 - Classic behaviour without exploit entrance (default)
+	// 2 - Updated behaviour to disable re-entrance on same instance
+	pre_re_instance_mode: 1
```
Ridley8819: if following my suggestion above, this can be disabled by default

npc/instances/EndlessTower.txt
```diff
@@ -312,6 +312,33 @@ L_Enter:
 		mes "The party leader did not generate the dungeon yet.";
 		close;
 	} else {
+		switch (getbattleflag("features/pre_re_instance_mode")) {
+		case 0:
```
Ridley8819: This would need some adjustments then, I'd also like to see `default:` rather than `case 0:`

npc/instances/EndlessTower.txt
```diff
@@ -312,6 +312,33 @@ L_Enter:
...
+		case 2:
+			.@dun_ent_t = ((etower_timer+604800) - gettimetick(2));
+			.@dun_h = (.@dun_ent_t / 3600);
+			.@dun_m = (.@dun_ent_t - (.@dun_h * 3600)) / 60;
+			.@dun_s = .@dun_ent_t - ((.@dun_h * 3600) + (.@dun_m * 60));
+			mesf("Due to the tower's aftereffects, you cannot enter the dungeon right now, %d hours %d minutes %d seconds left to enter the next dungeon.", .@dun_h, .@dun_m, .@dun_s);
+			next;
+			mes "It is dangerous here. Let me move you to Alberta.";
+			close2;
+			warp "alberta",223,36;
```
Ridley8819: missing spaces

npc/instances/EndlessTower.txt
```diff
@@ -312,6 +312,33 @@ L_Enter:
...
+		case 0:
+			break;
+		case 1:
+			.@instance = has_instance2("1@tower");
+			instance_attach(.@instance);
+			if (!'ins_etower_id) {
+				$@ins_etower_id[.@instance] = gettimetick(2);
+				'ins_etower_id = $@ins_etower_id[.@instance];
+			}
+			if (!questprogress(60200,PLAYTIME))
```
Ridley8819: missing spaces

npc/instances/NydhoggsNest.txt
```diff
@@ -224,6 +224,26 @@ L_Enter:
 		mes "You did not request for entrance. Please let your leader request entrance.";
 		close;
 	} else {
+		switch (getbattleflag("features/pre_re_instance_mode")) {
+		case 0:
```
Ridley8819: `default:` > `case 0:` here (at least thats my impression for a setting :D)

npc/instances/OrcsMemory.txt
```diff
@@ -107,6 +107,26 @@ gef_fild10,242,202,0	script	Dimensional Gorge Piece	2_MONEMUS,{
...
+	case 0:
+		break;
+	case 1:
+		.@instance = has_instance2("1@orcs");
+		instance_attach(.@instance);
+		if (!'ins_orc_id) {
+			$@ins_orc_id[.@instance] = gettimetick(2);
+			'ins_orc_id = $@ins_orc_id[.@instance];
+		}
+		if (!questprogress(12059,PLAYTIME))
```
Ridley8819: missing space

src/map/battle.c
```diff
@@ -7321,6 +7321,7 @@ static const struct battle_data {
 	{ "player_warp_keep_direction",         &battle_config.player_warp_keep_direction,      0,      0,      1,              },
 	{ "atcommand_levelup_events",	        &battle_config.atcommand_levelup_events,	    0,      0,      1,				},
 	{ "max_summoner_parameter",             &battle_config.max_summoner_parameter,          120,    10,     10000,          },
+	{ "features/pre_re_instance_mode",      &battle_config.pre_re_instance_mode,            1,      0,      2,              },
```
Ridley8819: defaulting to 0 then, too

4144:
Why need create configuration option for one instance?
Probably better put variable at top of script and add comment how it works?

Ridley8819:
It's not only one instance, it's for all pre re ones but the pr is still in progress

AnisotropicDefixation:
Wouldn't it be less revolting to just add quests with timers to check for the cooldown?

iRagno:
@4144, thisfixes several instances that has entrance exploit, that is why I suggest to have it as a common configuration, and suggest to have it as feature because this would allow to switch between different modes, from having issue fixed, to having issue without fix and change beheviour to updated one.

I make the suggestion and Ridley ask a week ago in #1582, of course this can be done through var scopes on scripts, I would be ok to do like that, if it would be better.

@AnisotropicDefixation, I remember a comment from @MishimaHaruna telling it is not a problem to have more variables when I suggest to return to use a variable with bit check (the one for bioethics), because have an extra variable would be easier to read than the bit check, so I don't see the need to avoid usage of new vars in this case.

By the other side, creating new quest entryes to set timmer checks doesn't fix the problem. Endless Tower has already two quest entryes, one for delay to 1 week and another to delay with 4 hours (to complete the instance), and that delay can be bypasseable having only those checks, that is why it's needed the new vars.

However, it's true that can empty the var when cd quest are erased, but I make it in that way to leave the less differences possibles with original script.

iRagno:
Since there was an oposition to battleflag introduced before I think that may not be user friendly so I remove it to introduce proper switch in scripts changed and also fix spaces.

@Ridley8819, I cannot use the default value to fix, only if default value is before updated behaviour, something like this:

```
switch (.@instance_mode) {
default:
	// fixed behaviour
	// fall through
case 1:
	// updated behaviour
	// break;
case 2:
	// old bugged behaviour
	// break;
}
```
And that because fixed behaviour uses message from updated behaviour. Otherwise, it would requiere to copy the code from inside updated behaviour to fixed behaviour and I try to keep it as simple as possible (even make it to add only just a block of code instead of having several lines modified).

Please tell me if it is needed to change it to use default value and I will change it.

iRagno:
Changed the temporary global var to be the same for all instances since instances can't repeat id while running and this would avoid values 0 on unused indexes.

@Ridley8819, @Asheraf @Jedzkie: Please tell me after review changes to squash commits.

iRagno:
@Mysteries: Done, thanks for the observation.

iRagno:
Squashed commits and updated pull request with new template.

MishimaHaruna:
Is this ready?

Ridley8819:
@MishimaHaruna I still think it's better to have a battleflag rather than a variable inside the related scripts. While at the moment this only affects 3 pre re instances, same can be done with renewal ones.

MishimaHaruna:
@Ridley8819 it's true that a more centralized place to handle this would be better, to avoid having to edit several scripts and end up forgetting some, but this feels like a misuse of a battle configuration flag. The alternatives aren't really appealing though (using a global server variable like many scripts do is even more wrong), so we might as well go with the battle configuration flag.

The other viable option would be to use an approach similar to how `F_GM_NPC()` works. Up to you which approach to take, whichever feels more convenient.

PR #1598, opened 2017-03-02
Implementation of Item Options System.

Smokexyz:
Allows the infusing of equipments with up to 5 bonus item options.
Discussion Topic - http://herc.ws/board/topic/14021-random-item-option-system/
- This feature is constrained to clients of packet versions greater than or equal to `20150226`.
- Item Options and their effects are defined server-side in `db/item_options.conf` and client side in `data/luafiles514/lua files/datainfo/addrandomoptionnametable.lub`
- The ID of the option must tally with the correct index of the description provided in the client side lua file.
- A basic custom script to `npc/custom/item_options.txt` to make use of this feature and demonstrate examples of usage.
- `IT_OPT_*` keys and `MAX_ITEM_OPTIONS` macro are also exported from the source as constants.
- An additional flag `disable_options` has been added to sql, and as `DisableOptions: true/false    (boolean, defaults to false !!for equipments only!!)` to db/<pre-re/re>/item_db.conf files.
- Script commands documentation is included.
- SQL File updates are included.

Credits: [smokexyz](https://github.com/Smokexyz)/Hercules 
Code Style Formatting by [Asheraf](https://github.com/Asheraf)/Hercules
Initial design idea: [secretdataz](https://github.com/secretdataz)/rAthena

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1598)


db/item_options.conf
```diff
@@ -0,0 +1,979 @@
...
+//=      | | | |                   | |
+//=      | |_| | ___ _ __ ___ _   _| | ___  ___
+//=      |  _  |/ _ \ '__/ __| | | | |/ _ \/ __|
+//=      | | | |  __/ | | (__| |_| | |  __/\__ \
+//=      \_| |_/\___|_|  \___|\__,_|_|\___||___/
+//================= License ===============================================
+//= This file is part of Hercules.
+//= http://herc.ws - http://github.com/HerculesWS/Hercules
+//=
+//= Copyright (C) 2014-2016  Hercules Dev Team
```
dastgir: 2017

npc/custom/item_options.txt
```diff
@@ -0,0 +1,234 @@
...
+		mes(.@name$);
+		mes("^009900Praise Jesus^000000");
+		mes("I have added an option to your item.");
+		mes("My skills are flawless!");
+	}
+	next;
+	mes(.@name$);
+	mes("See you around!");
+	close;
+}
```
dastgir: Newline missing

src/map/clif.c
```diff
@@ -2445,10 +2457,18 @@ void clif_additem(struct map_session_data *sd, int n, int amount, int fail) {
...
-			p.option_data[i].param = 0;
+		for (i = 0; i < MAX_ITEM_OPTIONS; i++) {
+			if (sd->status.inventory[n].option[i].index) {
+				p.option_data[j].index = sd->status.inventory[n].option[i].index;
+				p.option_data[j].value = sd->status.inventory[n].option[i].value;
+				p.option_data[j].param = sd->status.inventory[n].option[i].param;
+				j++;
+			}
+		}
+		for (; j < MAX_ITEM_OPTIONS || j < 5; j++) {
```
dastgir: MAX_ITEM_OPTIONS is there, why < 5 check?

Smokexyz: This is in the case that MAX_ITEM_OPTIONS is < 5. The remaining buffer is appended with `0`. If MAX_ITEM_OPTIONS >= 5, it is made to revert to 5 in mmo.h.

secretdataz:
:)

src/common/mmo.h
```diff
@@ -256,6 +256,12 @@
 #define MAX_ELESKILLTREE 3
 #endif
 
+#define MAX_ITEM_OPTIONS 5 // Maximum item options [Smokexyz]
```
4144: Please add #ifndef MAX_ITEM_OPTIONS before 259 line
and #endif after 263 line
This will allow override this settings from build script without changing code.

src/map/clif.c
```diff
@@ -2386,14 +2386,26 @@ void clif_addcards2(unsigned short *cards, struct item* item) {
  * @param buf[in,out] The buffer to write to. The pointer must be valid and initialized.
  * @param item[in]    The source item.
  */
-void clif_add_random_options(unsigned char* buf, struct item* item)
+void clif_add_random_options(struct RndOptions *buf, const struct item it)
```
4144: Why not use pointer here?
const struct item it -> const struct item *it
And why rename it from item to it?

src/map/clif.c
```diff
@@ -2445,10 +2457,18 @@ void clif_additem(struct map_session_data *sd, int n, int amount, int fail) {
 		p.bindOnEquipType = sd->status.inventory[n].bound && !itemdb->isstackable2(sd->inventory_data[n]) ? 2 : sd->inventory_data[n]->flag.bindonequip ? 1 : 0;
 #endif
 #if PACKETVER >= 20150226
-		for (i=0; i<5; i++){
-			p.option_data[i].index = 0;
-			p.option_data[i].value = 0;
-			p.option_data[i].param = 0;
+		for (i = 0; i < MAX_ITEM_OPTIONS; i++) {
```
4144: Block with adding options to pointer probably better move to separate function.
This will allow for plugins to change options what will be send to client

src/map/clif.c
```diff
@@ -2543,39 +2564,43 @@ void clif_item_equip(short idx, struct EQUIPITEM_INFO *p, struct item *i, struct
...
 	p->Flag.SpareBits    = 0;
 #endif
 
 #if PACKETVER >= 20150226
 	p->option_count = 0;
-	for (j=0; j<5; j++){
-		p->option_data[j].index = 0;
-		p->option_data[j].value = 0;
-		p->option_data[j].param = 0;
+	for (i=0, j=0; i < MAX_ITEM_OPTIONS; i++) {
```
4144: Can be moved to same function like previous options block from clif_additem

src/map/itemdb.c
```diff
@@ -1299,6 +1309,69 @@ void itemdb_read_packages(void) {
 	ShowStatus("Done reading '"CL_WHITE"%d"CL_RESET"' entries in '"CL_WHITE"%s"CL_RESET"'.\n", count, config_filename);
 }
 
+void itemdb_read_options(void) {
```
4144: Here { should be on new line

src/map/script.c
```diff
@@ -9019,6 +9019,32 @@ BUILDIN(getequipisenableref)
...
+ * 
+ * @param equipment_index as the inventory index of the equipment.
+ * @return 1 on enabled 0 on disabled.
+ */
+BUILDIN(getequipisenableopt)
+{
+	int i = -1, index = script_getnum(st,2);
+	struct map_session_data *sd = script->rid2sd(st);
+	
+	if (sd == NULL)
```
4144: Probably here need add script_pushint(st, -1) or like this for prevent script from termination

db/item_options.conf
```diff
@@ -0,0 +1,979 @@
...
+
+{
+	Id: 1
+	Name: "VAR_MAXHPAMOUNT"
+	Script: <" bonus bMaxHP,getequippedoptioninfo(IT_OPT_VALUE); ">
+},
+{
+	Id: 2
+	Name: "VAR_MAXSPAMOUNT"
+	Script: <" bonus bMaxSP,getequippedoptioninfo(IT_OPT_VALUE); ">
```
Jedzkie: Same as above.

db/item_options.conf
```diff
@@ -0,0 +1,979 @@
...
+},
+{
+	Id: 3
+	Name: "VAR_STRAMOUNT"
+	Script: <" bonus bStr,getequippedoptioninfo(IT_OPT_VALUE); ">
```
Jedzkie: Same.

db/item_options.conf
```diff
@@ -0,0 +1,979 @@
...
+},
+{
+	Id: 4
+	Name: "VAR_AGIAMOUNT"
+	Script: <" bonus bAgi,getequippedoptioninfo(IT_OPT_VALUE); ">
```
Jedzkie: Same.

src/map/script.c
```diff
@@ -13546,6 +13572,184 @@ BUILDIN(getiteminfo)
...
+		ShowError("buildin_setequipoption: Invalid equipment index %d provided.\n", equip_index);
+		return false;
+	} else if (slot < 0 || slot >= MAX_ITEM_OPTIONS) {
+		script_pushint(st, 0);
+		ShowError("buildin_setequipoption: Invalid option index %d (Min: 1, Max: %d) provided.\n", slot, MAX_ITEM_OPTIONS-1);
+		return false;
+	} else if (equip_index > 0 && equip_index <= ARRAYLENGTH(script->equip))
+		i = pc->checkequip(sd,script->equip[equip_index-1]);
+	
+	if (&sd->status.inventory[i] != NULL) {
```
4144: This condition always true. May be you want check something else?

src/map/script.c
```diff
@@ -13546,6 +13572,184 @@ BUILDIN(getiteminfo)
...
+		script_pushint(st, 0);
+		ShowError("buildin_setequipoption: Invalid equipment index %d provided.\n", equip_index);
+		return false;
+	} else if (slot < 0 || slot >= MAX_ITEM_OPTIONS) {
+		script_pushint(st, 0);
+		ShowError("buildin_setequipoption: Invalid option index %d (Min: 1, Max: %d) provided.\n", slot, MAX_ITEM_OPTIONS-1);
+		return false;
+	} else if (equip_index > 0 && equip_index <= ARRAYLENGTH(script->equip))
+		i = pc->checkequip(sd,script->equip[equip_index-1]);
+
```
4144: Here "i" can have -1, and in next lines it will be buffer underflow
like sd->status.inventory[-1]. This is error

Mysteries:
Nice 👍

Smokexyz:
*[src/map/clif.c, line 2389 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZu-r1-2389:-KeEkqfsCLHGUote2LKR:bv23h85) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/src/map/clif.c#L2389)):*
<details><summary><i>Previously, 4144 (Andrei Karas) wrote…</i></summary><blockquote>

Why not use pointer here?
const struct item it -> const struct item *it
And why rename it from item to it?
</blockquote></details>

Why should there be a pointer here when the function isn't writing any data to the pointer? 
and because struct item item is is like like weird weird.. :smile: :smile: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1598)*

4144:
@Smokexyz because passing structs by value is bad idea.
It copy whole struct to stack. Pointers created for avoid this issues.
You can add const to pointers too. Even in two places.

Example: const struct item *const it

First const here mean data in struct item is constant.
Second const mean variable it is constant. Mean you cant assign it new value.

Smokexyz:
Review status: 0 of 29 files reviewed at latest revision, 15 unresolved discussions.

---

*[db/item_options.conf, line 12 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZi-r1-12:-KeFP27DW08r9YT3YCMM:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/db/item_options.conf#L12)):*
<details><summary><i>Previously, dastgir (Dastgir) wrote…</i></summary><blockquote>

2017
</blockquote></details>

Done.

---

*[db/item_options.conf, line 50 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZi-r1-50:-KeFP3CR3-mXFuRBYYB7:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/db/item_options.conf#L50)):*
<details><summary><i>Previously, Jedzkie (Jedzkie) wrote…</i></summary><blockquote>

Space between comma

`bonus bMaxHP, getequippedoptioninfo(IT_OPT_VALUE);`
</blockquote></details>

Done.

---

*[db/item_options.conf, line 55 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZi-r1-55:-KeFP3qoukEYDAasyjqQ:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/db/item_options.conf#L55)):*
<details><summary><i>Previously, Jedzkie (Jedzkie) wrote…</i></summary><blockquote>

Same as above.
</blockquote></details>

Done.

---

*[db/item_options.conf, line 60 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZi-r1-60:-KeFP48mUkz_uH5iU-wN:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/db/item_options.conf#L60)):*
<details><summary><i>Previously, Jedzkie (Jedzkie) wrote…</i></summary><blockquote>

Same.
</blockquote></details>

Done.

---

*[db/item_options.conf, line 65 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZi-r1-65:-KeFP4S24AmXCeo4pRKw:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/db/item_options.conf#L65)):*
<details><summary><i>Previously, Jedzkie (Jedzkie) wrote…</i></summary><blockquote>

Same until the last Id.
</blockquote></details>

Done.

---

*[npc/custom/item_options.txt, line 234 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZm-r1-234:-KeFP5p_jpdMzYNBlE2u:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/npc/custom/item_options.txt#L234)):*
<details><summary><i>Previously, dastgir (Dastgir) wrote…</i></summary><blockquote>

Newline missing
</blockquote></details>

Done.

---

*[src/common/mmo.h, line 259 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZt-r1-259:-KeFP6hspzzhzEp6HOim:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/src/common/mmo.h#L259)):*
<details><summary><i>Previously, 4144 (Andrei Karas) wrote…</i></summary><blockquote>

Please add #ifndef MAX_ITEM_OPTIONS before 259 line
and #endif after 263 line
This will allow override this settings from build script without changing code.
</blockquote></details>

Done.

---

*[src/map/clif.c, line 2389 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZu-r1-2389:-KeFP7CEAQO_hmIf50Kt:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/src/map/clif.c#L2389)):*
<details><summary><i>Previously, Smokexyz (smokexyz) wrote…</i></summary><blockquote>

Why should there be a pointer here when the function isn't writing any data to the pointer? 
and because struct item item is is like like weird weird.. :smile: :smile: 

</blockquote></details>

Done.

---

*[src/map/clif.c, line 2460 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZu-r1-2460:-KeFP8LDte5DEl-vLLYZ:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/src/map/clif.c#L2460)):*
<details><summary><i>Previously, 4144 (Andrei Karas) wrote…</i></summary><blockquote>

Block with adding options to pointer probably better move to separate function.
This will allow for plugins to change options what will be send to client
</blockquote></details>

Done.

---

*[src/map/clif.c, line 2468 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZu-r1-2468:-KeFP8rASSVcxcWhPrLX:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/src/map/clif.c#L2468)):*
<details><summary><i>Previously, Smokexyz (smokexyz) wrote…</i></summary><blockquote>

This is in the case that MAX_ITEM_OPTIONS is < 5. The remaining buffer is appended with `0`. If MAX_ITEM_OPTIONS > 5, it is made to revert to 5 in mmo.h.
</blockquote></details>

Done.

---

*[src/map/clif.c, line 2596 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZu-r1-2596:-KeFP9D97nQpH6rUIzJe:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/src/map/clif.c#L2596)):*
<details><summary><i>Previously, 4144 (Andrei Karas) wrote…</i></summary><blockquote>

Can be moved to same function like previous options block from clif_additem
</blockquote></details>

Done.

---

*[src/map/itemdb.c, line 1312 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieHZw-r1-1312:-KeFP9d1ewMXgyka4Zwx:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/src/map/itemdb.c#L1312)):*
<details><summary><i>Previously, 4144 (Andrei Karas) wrote…</i></summary><blockquote>

Here { should be on new line
</blockquote></details>

Done.

---

*[src/map/script.c, line 9034 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieH_--r1-9034:-KeFPAWsLIO8seVgiHti:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/src/map/script.c#L9034)):*
<details><summary><i>Previously, 4144 (Andrei Karas) wrote…</i></summary><blockquote>

Probably here need add script_pushint(st, -1) or like this for prevent script from termination
</blockquote></details>

Done.

---

*[src/map/script.c, line 13704 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieH_--r1-13704:-KeFPApSsC3Efdm1egZl:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/src/map/script.c#L13704)):*
<details><summary><i>Previously, 4144 (Andrei Karas) wrote…</i></summary><blockquote>

Here "i" can have -1, and in next lines it will be buffer underflow
like sd->status.inventory[-1]. This is error
</blockquote></details>

Done.

---

*[src/map/script.c, line 13705 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1598#-KeDmFq6xPSfjO1ieH_--r1-13705:-KeFPBLfHAI7OQkszLKe:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/85928252069614c2fdedd55134595cc81f697ecb/src/map/script.c#L13705)):*
<details><summary><i>Previously, 4144 (Andrei Karas) wrote…</i></summary><blockquote>

This condition always true. May be you want check something else?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1598)*

Smokexyz:
Reviewed 29 of 29 files at r1.
Review status: 18 of 29 files reviewed at latest revision, 15 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1598)*

src/map/script.c
Smokexyz: Checked for in next line as `if ( i >= 0 && ...)`

db/item_options.conf
```diff
@@ -0,0 +1,979 @@
...
+//=      | | | |                   | |
+//=      | |_| | ___ _ __ ___ _   _| | ___  ___
+//=      |  _  |/ _ \ '__/ __| | | | |/ _ \/ __|
+//=      | | | |  __/ | | (__| |_| | |  __/\__ \
+//=      \_| |_/\___|_|  \___|\__,_|_|\___||___/
+//================= License ===============================================
+//= This file is part of Hercules.
+//= http://herc.ws - http://github.com/HerculesWS/Hercules
+//=
+//= Copyright (C) 2014-2017  Hercules Dev Team
```
Asheraf: `Copyright (C) 2017  Hercules Dev Team`

db/item_options.conf
```diff
@@ -0,0 +1,979 @@
...
+//= but WITHOUT ANY WARRANTY; without even the implied warranty of
+//= MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+//= GNU General Public License for more details.
+//=
+//= You should have received a copy of the GNU General Public License
+//= along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//=========================================================================
+//= Items Options Database
+//============================= Credits ===================================
+// Smokexyz
```
Asheraf: move this up

npc/custom/item_options.txt
```diff
@@ -0,0 +1,234 @@
...
+//= along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//=========================================================================
+//= Option Master (Item Options NPC @see db/item_options.conf)
+//================= Description ===========================================
+//= Select from a list of your equipments, one to enhance with up to 5 options.
+//= A list of the first 10 options are provided in this npc.
+//================= Current Version =======================================
+//= 1.0
+//==================== Credits ============================================
+//= Smokexyz
```
Asheraf: credit already added up

npc/custom/item_options.txt
```diff
@@ -0,0 +1,234 @@
...
+	 */
+	.minimum_bonus_amount = -100;
+	/* Maximum amount of the bonus value. 
+	 * Maximum possible value is INT16_MAX */
+	.maximum_bonus_amount = 100;
+	/* Disable selection of bonus value (true/false) */
+	.enable_random_bonus = false;
+	
+	/* Item Option Descriptions */
+	setarray .options$[0],"Max HP", "Max SP", "STR", "AGI", "VIT", "INT", "DEX", "LUK";
```
Asheraf: missed a space `.options$[0], "Max HP",`

npc/custom/item_options.txt
```diff
@@ -0,0 +1,234 @@
...
+	mes(.@name$);
+	mes("Would you like me to add extra stats to your equipment?");
+	next;
+	if (select("Of course!", "Meh, not right now.") == 2) {
+		mes(.@name$);
+		mes("Alright then, catch you later.");
+		close;
+	}
+	// Build the Menu.
+	setarray .@position$[1],"Head","Body","Left Hand","Right Hand","Robe","Shoes","Accessory 1","Accessory 2","Head 2","Head 3";
```
Asheraf: spaces here

npc/custom/item_options.txt
```diff
@@ -0,0 +1,234 @@
...
+	.@menu$ = "";
+	for (.@i = 1; .@i<=10; ++.@i)
```
Asheraf: `.@i <= 10`

npc/custom/item_options.txt
```diff
@@ -0,0 +1,234 @@
...
+		.@menu$ += ((getequipisequiped(.@i))?getequipname(.@i):.@position$[.@i]+"-[Not equipped]")+":";
```
Asheraf: spaces

npc/custom/item_options.txt
```diff
@@ -0,0 +1,234 @@
...
+	.@menu$ = "";
+	for (.@i = 0; .@i < getarraysize(.options$); .@i++)
+		.@menu$ += (.@i+1)+") "+.options$[.@i]+":";
+		
+	do {
+		// Select the options!
+		.@option_variable = select(.@menu$);
+		next;
+		mes(.@name$);
+		mes("You chose ^009900"+.options$[.@option_variable-1]+"^000000!");
```
Asheraf: use mesf ?

src/char/char.c
```diff
@@ -770,8 +773,12 @@ int char_memitemdata_to_sql(const struct item items[], int max, int id, int tabl
 	SQL->StmtBindColumn(stmt, 9, SQLDT_UINT64,    &item.unique_id,   0, NULL, NULL);
 	for (j = 0; j < MAX_SLOTS; ++j)
 		SQL->StmtBindColumn(stmt, 10+j, SQLDT_SHORT, &item.card[j], 0, NULL, NULL);
+	for (j = 0; j < MAX_ITEM_OPTIONS; ++j) {
+		SQL->StmtBindColumn(stmt, 10+MAX_SLOTS+j*2, SQLDT_INT16, &item.option[j].index, 0, NULL, NULL);
```
Asheraf: `10 + MAX_SLOTS + j * 2`

npc/custom/item_options.txt
```diff
@@ -0,0 +1,234 @@
+//================= Hercules Script =======================================
```
Ridley8819: Actually not sure about the whole custom script. We talked before to drop support for custom folder entirely. Maybe better to remove it here and offer it in forums?
you also forgot parantheses supposed to be included in all commands, this means also in next(); or close();
exceptions here are break and end

Emistry:
## Database ##
by the way, I think previously Herc was also planning to do this Item Bonus system by storing these item_bonus_id / item_bonus_value in a separate SQL tables, the idea was dropped?

## Script Command ##
do we support script command that allow direct creation of item with item bonus?
like 
`getitem3( ...................<item_bonus_id1>,<item_bonus_value1>,....,<item_bonus_id5>,<item_bonus_value5>);`
or 
`getitem3( .............. <item_bonus_id_array>,<item_bonus_value_array>);`

about the script command to get the current info of item bonus from selected equipment slot ...
why not just do something like .. `getinventorylist` ?
```
getequipoptioninfo( <equipment_slot>{, <option_number>} );
equipment_slot - equipment index
option_number
 - which index of item bonus in the equipment (value 1-MAX_ITEM_OPTIONS lines)
 - 0 = list all index

return an array that consist of info of these options in the equipment.
- @item_option_id[] 
- @item_option_value[]
```

I am not familiar with the item bonus option, but this script command
```*getequipoptioninfo(<equipment_slot>,<type>,<option_slot>);```
the `type` parameter seem like troublesome to use.
Example:
when you try to retrieve the item bonus info from equipments, you gonna have to loop through every single `<type>` unless you trying to hardcode the parameters value during the usage of this script command

just my opinion...

File: npc/custom/item_options.txt [hunk 25]
```diff
@@ -0,0 +1,234 @@
...
+//================= Description ===========================================
+//= Select from a list of your equipments, one to enhance with up to 5 options.
+//= A list of the first 10 options are provided in this npc.
+//================= Current Version =======================================
+//= 1.0
+//==================== Credits ============================================
+//= Smokexyz
+//=========================================================================
+prontera,153,180,4	script	Option Master	4_DOG01,{
+	callsub(StartTalking);
```
Emistry: this is unnecessary, remove this line and move the whole `StartTalking` label part up here .... ?

npc/custom/item_options.txt
Emistry: are not this script suppose to be added into the `doc/sample/` folder?

npc/custom/item_options.txt
Emistry: personally, its not really a good practice to initialize the array at index 1 ... the index 0 is wasted ... 
perhaps you should consider to adjust it to index 0 ?

npc/custom/item_options.txt
Emistry: perhaps the value `.@i <= 10` could change to `.@i <= getarraysize(.@array)` ? since it actually depend on the array values.
and same with above, the `.@i` should start from 0 since its actually refer to an index of an array.

npc/custom/item_options.txt
```diff
@@ -0,0 +1,234 @@
...
+	
+	// Select a random bonus.
+	if (.enable_random_bonus) {
+		.@value = rand(.maximum_bonus_amount);
+	} else {
+		do {
+			mes(.@name$);
+			mes("Please input the bonus amount of ^009900"+.options$[.@option_variable-1]+"^000000 you want to add!");
+			mes("(Min: "+.minimum_bonus_amount+", Max: "+.maximum_bonus_amount+")");
+			input .@value;
```
Emistry: consider to use this since there are `min` and `max` value.
`input(.@value, .minimum_bonus_amount, .maximum_bonus_amount);`

npc/custom/item_options.txt
```diff
@@ -0,0 +1,234 @@
...
+			next;
+		} while (.@value < .minimum_bonus_amount || .@value == 0 || .@value > .maximum_bonus_amount);
```
Emistry: extra condition checking can be avoided if you applied the `input(<variable>,<min>,<max);`

npc/custom/item_options.txt
Smokexyz: @MishimaHaruna

Smokexyz:
@Emistry 

> by the way, I think previously Herc was also planning to do this Item Bonus system by storing these item_bonus_id / item_bonus_value in a separate SQL tables, the idea was dropped?

Not a very good idea for this implementation.

> do we support script command that allow direct creation of item with item bonus?

not yet, no. But something like this is required for crimson weapons?

> return an array that consist of info of these options in the equipment.

Can be done, it's a good suggestion.

> you gonna have to loop through every single `<type>`

only IT_OPT_INDEX and IT_OPT_VALUE. (0 or 1)

> and by the way is this based on rA implementation? o-o

Well, since the code structure of the methods involved are almost the same, there aren't many ways to implement such a system without it looking similar. Consider this a system similar to item cards, but with added fields. The item_options.conf and script commands allows potential to add custom bonus options (as long as the client side is edited.). The param field was dropped since we don't know it's use. There are many improvements in both code and design over rA's implementation. It is for many reasons not based on theirs.

Smokexyz:
Considering the number of cosmetic changes requested in this PR, I feel that the "reviewers" should be urged to add these minor changes as a PR to the original author's branch. This would allow faster processing of PRs as well as more team work and co-ordination in the team,  so to speak.

npc/custom/item_options.txt
Smokexyz: Copied from the hd refiner file. Although, this is done by taking into consideration the index that is passed to the following `getequipisequiped()`, `getequipisenableopt()`, `getequipisidentify()` checks. It is, in source, checked for `if (num > 0 && num <= ARRAYLENGTH(script->equip))`. It would fail if the index is 0. And if we began the array at [0] we'd have to add a +1 to all the following places we require the check. @MishimaHaruna @4144 Shouldn't this be changed in source? maybe in a commit changing all scripts using such commands.

npc/custom/item_options.txt
Smokexyz: They array here is not dynamic in size, therefore there is no need to call on a size check.

npc/custom/item_options.txt
Smokexyz: Thanks, I wasn't aware.

npc/custom/item_options.txt
Emistry: nope, i mean this ...
```setarray .@position$[1]....```
it should be 
```setarray .@position$[0]....```
same goes for the other script... hd refiner..in case this is copied from there.

should avoid initialize the array and left the value remain unused.

File: npc/custom/item_options.txt (hunk 20 above)
Comment (Emistry) [inline npc/custom/item_options.txt]:
```
EQI_HEAD_TOP (1)          - Upper head gear
EQI_ARMOR (2)             - Armor (Where you keep your Jackets and Robes)
EQI_HAND_L (3)            - What is in your Left hand.
EQI_HAND_R (4)            - What is in your Right hand.
EQI_GARMENT (5)           - The garment slot (Mufflers, Hoods, Manteaus)
EQI_SHOES (6)             - What foot gear the player has on.
EQI_ACC_L (7)             - Accessory 1.
EQI_ACC_R (8)             - Accessory 2.
EQI_HEAD_MID (9)          - Middle Headgear (masks and glasses)
EQI_HEAD_LOW (10)         - Lower Headgear (beards, some masks)
```
btw, if you're referring to this list... to use the number 1 ~ 10 ... it's better to replace with the `EQI_` constant.
it could help avoid potential script broken if these EQI values changed in future..... just how what happened in rA when they decided to update the values.

File: npc/custom/item_options.txt (hunk 19 above)
Comment (Smokexyz) [inline npc/custom/item_options.txt]: > the index that is passed to the following getequipisequiped(), getequipisenableopt(), getequipisidentify() checks. It is, in source, checked for if (num > 0 && num <= ARRAYLENGTH(script->equip)). It would fail if the index is 0.

File: npc/custom/item_options.txt (hunk 25 above)
Comment (Smokexyz) [inline npc/custom/item_options.txt]: This was done to add the OnInit configuration at the beginning of the file.

File: npc/custom/item_options.txt (hunk 17 above)
Comment (Smokexyz) [inline npc/custom/item_options.txt]: woops

File: npc/custom/item_options.txt (hunk 20 above)
Comment (Smokexyz) [inline npc/custom/item_options.txt]: I'm don't think we're on the same page, you can hit me up on the IRC to clarify.

File: npc/custom/item_options.txt (hunk 19 above)
Comment (Emistry) [inline npc/custom/item_options.txt]:
```
// Build the Menu.
setarray .@position$,"Head","Body","Left Hand","Right Hand","Robe","Shoes","Accessory 1","Accessory 2","Head 2","Head 3";

for (.@i = 0; .@i <= 10; .@i++) {
	.@menu$ += (getequipisequiped(.@i + 1) ? getequipname(.@i + 1) : .@position$[.@i] + "-[Not equipped]") + ":";
}
.@equip_index = select(.@menu$);
```
or better
```
// Build the Menu.
setarray .@eqi_list,EQI_HEAD_TOP,EQI_ARMOR,EQI_HAND_L,EQI_HAND_R,EQI_GARMENT,EQI_SHOES,EQI_ACC_L,EQI_ACC_R,EQI_HEAD_MID,EQI_HEAD_LOW;
setarray .@position$,"Head","Body","Left Hand","Right Hand","Robe","Shoes","Accessory 1","Accessory 2","Head 2","Head 3";

for (.@i = 0; .@i <= 10; .@i++) {
	.@menu$ += (getequipisequiped(.@eqi_list[.@i]) ? getequipname(.@eqi_list[.@i]) : .@position$[.@i] + "-[Not equipped]") + ":";
}
.@equip_index = .@eqi_list[select(.@menu$) - 1];
```
I believe these approach are better than having the array creating unused index 0.

File: npc/custom/item_options.txt (hunk 25 above)
Comment (Emistry) [inline npc/custom/item_options.txt]: `OnInit` label can be located at any place in the script ...  personally, I often place the `OnInit` label at the end of the script ...

File: npc/custom/item_options.txt (hunk 19 above)
Comment (Smokexyz) [inline npc/custom/item_options.txt]: Your first case would work, in the second case, you're creating a whole new array[10] to make up for one unused index.

Comment (Emistry) [inline npc/custom/item_options.txt]: Because it's more logical. It's more elegant, it makes more sense ... to have array start with index 0.

File: npc/custom/item_options.txt (hunk 20 above)
Comment (Emistry) [inline npc/custom/item_options.txt]: I am just suggesting the usage of dynamic size because users are normally confused with this kind of settings if they are not familiar ..
Example:

> User: I added a new item to the array `.@position$` but it doesn't show up in the menu, why ?

tho its not really something that really need to focus on ...

File: npc/custom/item_options.txt (hunk 19 above)
Comment (Smokexyz) [inline npc/custom/item_options.txt]: I agree, but this scenario involves passing an incremented the value to functions that I believe don't handle indexes properly. And so I'm requesting the input of @MishimaHaruna and @4144

File: npc/custom/item_options.txt (hunk 22 above)
Comment (Smokexyz) [inline npc/custom/item_options.txt]: Life saving function.

**Comment by Smokexyz** [comment]
2 script commands have been changed to -
```

*getequipoptioninfo(<equip_index>,<slot>,<type>);

Gets the option information of an equipment.

<equipment_index> For a list of equipment indexes see getequipid().
<option_slot> can range from 1 to MAX_ITEM_OPTIONS
<type> can be IT_OPT_INDEX (the ID of the option bonus, @see "Id" or "Name" in db/item_options.conf)
       or IT_OPT_VALUE (the value of the bonus script of the equipment, @see "Script" in db_item_options.conf).

returns the value of the slot if exists or -1 for invalid slot, type or slots.


*setequipoption(<equip_index>,<slot>,<opt_index>,<value>);

Set an equipment's option index or value for the specified option slot.

<equipment_index> For a list of equipment indexes see getequipid().
<option_slot> can range from 1 to MAX_ITEM_OPTIONS
<type> can be IT_OPT_INDEX (the ID of the option bonus, @see "Id" or "Name" in db/item_options.conf)
<value> The value of the type to be set.

returns 0 if value couldn't be set, 1 on success.
```

db2sql update will be added before a merge.

**Comment by Atemo** [comment]
>Well, since the code structure of the methods involved are almost the same between both emulators, there aren't many ways to implement such a system without it looking similar. Consider this a system like item cards, but with added fields. The item_options.conf and script commands allows potential to add custom bonus options (as long as the client side is edited.). The ID field allows specifying the index of the client side equivalent. The param field was dropped since we don't know it's use. There are many improvements in both code and design over rA's implementation. The packet structures were added by Dastgir last year. It is for many reasons not based on theirs.

Logical issue, please review.

**Comment by Smokexyz** [comment]
@Atemo what?

File: npc/custom/item_options.txt (hunk 24 above)
Comment (MishimaHaruna) [inline npc/custom/item_options.txt]: Good point about the sample folder, I always forget it exists.

My opinion is:

- If the script is meant to be an example for users that want to implement their own custom script, it should go into doc/sample.
- If the script is meant to be usable as is, it's fine to keep it in npc/custom - keeping in mind that it'll eventually be split to another repository and won't be maintained officially.

File: sql-files/upgrades/2017-03-02--11-40.sql [hunk 28]
```diff
@@ -0,0 +1,105 @@
...
+	ADD COLUMN `opt_idx1` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_val0`,
+	ADD COLUMN `opt_val1` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_idx1`,
+	ADD COLUMN `opt_idx2` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_val1`,
+	ADD COLUMN `opt_val2` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_idx2`,
+	ADD COLUMN `opt_idx3` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_val2`,
+	ADD COLUMN `opt_val3` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_idx3`,
+	ADD COLUMN `opt_idx4` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_val3`,
+	ADD COLUMN `opt_val4` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_idx4`;
+
+ALTER TABLE `picklog`
```
MishimaHaruna: When a migration affects the main database and the logs database, it should always be split into two, so that people can easily run the two scripts separately when they have separate databases.

sql-files/main.sql
```diff
@@ -79,6 +79,16 @@ CREATE TABLE IF NOT EXISTS `auction` (
   `card1` SMALLINT(11) NOT NULL DEFAULT '0',
   `card2` SMALLINT(11) NOT NULL DEFAULT '0',
   `card3` SMALLINT(11) NOT NULL DEFAULT '0',
+  `opt_idx0` smallint(5) unsigned NOT NULL default '0',
```
MishimaHaruna: Missing the sql_updates addition

src/char/char.c
```diff
@@ -1188,8 +1204,10 @@ int char_mmo_char_fromsql(int char_id, struct mmo_charstatus* p, bool load_every
 	//`inventory` (`id`,`char_id`, `nameid`, `amount`, `equip`, `identify`, `refine`, `attribute`, `card0`, `card1`, `card2`, `card3`, `expire_time`, `favorite`, `bound`, `unique_id`)
 	StrBuf->Init(&buf);
 	StrBuf->AppendStr(&buf, "SELECT `id`, `nameid`, `amount`, `equip`, `identify`, `refine`, `attribute`, `expire_time`, `favorite`, `bound`, `unique_id`");
-	for( i = 0; i < MAX_SLOTS; ++i )
+	for (i = 0; i < MAX_SLOTS; ++i)
 		StrBuf->Printf(&buf, ", `card%d`", i);
+	for (i = 0; i < MAX_SLOTS; ++i)
```
MishimaHaruna: `MAX_SLOTS`? Typo?

src/common/mmo.h
```diff
@@ -256,6 +256,14 @@
 #define MAX_ELESKILLTREE 3
 #endif
 
+// Maximum item options [Smokexyz]
+#ifndef MAX_ITEM_OPTIONS
+#define MAX_ITEM_OPTIONS 5
+#if MAX_ITEM_OPTIONS > 5 // (constrained to 5 through client and database !!Do not exceed!!)  [Smokexyz]
```
MishimaHaruna: `STATIC_ASSERT(MAX_ITEM_OPTIONS <= 5, "This value is limited by the client and database layout and should only be increased if you know the consequences.")`

src/common/mmo.h
```diff
@@ -63,7 +63,7 @@
 // 20120307 - 2012-03-07aRagexeRE+ - 0x970
 
 #ifndef PACKETVER
-	#define PACKETVER 20141022
+	#define PACKETVER 20150513
```
MishimaHaruna: Is this intended? Do we fully support that packetver? I think it would be worth doing this in a separate commit in case.

src/map/clif.c
```diff
@@ -2378,23 +2378,37 @@ void clif_addcards2(unsigned short *cards, struct item* item) {
...
  */
-void clif_add_random_options(unsigned char* buf, struct item* item)
-{
-	int i;
-	nullpo_retv(buf);
-	for (i = 0; i < 5; i++){
-		WBUFW(buf,i*5+0) = 0; // OptIndex
-		WBUFW(buf,i*5+2) = 0; // Value
-		WBUFB(buf,i*5+4) = 0; // Param1
+unsigned int clif_add_random_options(struct ItemOptions *buf, struct item *it)
```
MishimaHaruna: From the look of it, `it` should be const. And the return value `int`

src/map/itemdb.c
```diff
@@ -1299,6 +1309,70 @@ void itemdb_read_packages(void) {
...
+			ShowError("itemdb_read_options: Script code not found for entry %s (Id: %d) in '%s', skipping...\n", str, t_opt.index, filepath);
+			continue;
+		}
+		
+		/* Set Script */
+		t_opt.script = *str ? script->parse(str, filepath, t_opt.index, SCRIPT_IGNORE_EXTERNAL_BRACKETS, NULL) : NULL;
+		
+		/* Allocate memory and copy contents */
+		CREATE(s_opt, struct item_option, 1);
+		memcpy(s_opt, &t_opt, sizeof(struct item_option));
```
MishimaHaruna: `*s_opt = t_opt;` ?

src/map/script.c
```diff
@@ -13546,6 +13575,176 @@ BUILDIN(getiteminfo)
...
+	struct map_session_data *sd = NULL;
+	
+	if ((sd = script->rid2sd(st)) == NULL || status->current_equip_item_index == -1
+		|| !sd->status.inventory[status->current_equip_item_index].option[status->current_equip_option_index].index) {
+		script_pushint(st, -1);
+		return false;
+	}
+	
+	switch (type) {
+		case IT_OPT_INDEX:
```
MishimaHaruna: Too much indentation (elsewhere as well, whenever there's a switch), please fix your editor settings.

src/common/mmo.h
Smokexyz: this was picked up in a rebase somehow, sorry.

src/map/clif.c
Smokexyz: The problem here is that the HPM generators do not append the second const, I have not checked the scripts but `struct item *const it` would be generated as `struct item *constit`.

MishimaHaruna: But what about `const struct item *it`? (so that it's a pointer to constant data, not the other way around)

The HPM in general doesn't handle constant pointers to non-constant data, but only pointers to constant data. This is because I'm not aware of any case where passing a `struct foo * const foo` to a function has any relevance and makes any difference from passing `struct foo *foo` - the caller doesn't need to care about what happens to the pointer itself inside the function, only cares that a constant buffer doesn't get modified.

src/map/itemdb.c
```diff
@@ -1299,6 +1309,71 @@ void itemdb_read_packages(void) {
...
+		script->set_constant(str, t_opt.index, false, false);
+		
+		if (!libconfig->setting_lookup_string(conf, "Script", &str)) {
+			ShowError("itemdb_read_options: Script code not found for entry %s (Id: %d) in '%s', skipping...\n", str, t_opt.index, filepath);
+			continue;
+		}
+		
+		/* Set Script */
+		t_opt.script = *str ? script->parse(str, filepath, t_opt.index, SCRIPT_IGNORE_EXTERNAL_BRACKETS, NULL) : NULL;
+
```
4144: Please add here call to function for read additional fields. For example like in mob.c mob->read_db_additional_fields.
This can be usefull for plugins too

Smokexyz: Done.

src/map/clif.c
Smokexyz: Corrected and updated.

sql-files/main.sql
```diff
@@ -848,6 +898,16 @@ CREATE TABLE IF NOT EXISTS `storage` (
...
+  `opt_idx0` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0',
+  `opt_val0` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0',
+  `opt_idx1` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0',
+  `opt_val1` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0',
+  `opt_idx2` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0',
+  `opt_val2` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0',
+  `opt_idx3` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0',
+  `opt_val3` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0',
+  `opt_idx4` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0',
+  `opt_val4` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0',
```
MishimaHaruna: Still missing the `INSERT INTO sql_updates` lines

src/common/mmo.h
```diff
@@ -256,6 +256,12 @@
 #define MAX_ELESKILLTREE 3
 #endif
 
+// Maximum item options [Smokexyz]
+#ifndef MAX_ITEM_OPTIONS
+#define MAX_ITEM_OPTIONS 5
+STATIC_ASSERT(MAX_ITEM_OPTIONS <= 5, "This value is limited by the client and database layout and should only be increased if you know the consequences.");
```
MishimaHaruna: Can you move this line to after the `#endif`? (so that if it's defined elsewhere it still asserts)

src/map/clif.c
MishimaHaruna: Yes, it's a bug in the parser. The use of const pointers (just like const scalar variables) was never considered, since it doesn't affect the function's contract with the caller. It should be fixed in the parser, but it's not a very high priority task

sql-files/main.sql
Smokexyz: Done

src/common/mmo.h
Smokexyz: done

src/map/clif.h
```diff
@@ -1351,7 +1351,7 @@ struct clif_interface {
 	void (*pNPCMarketClosed) (int fd, struct map_session_data *sd);
 	void (*pNPCMarketPurchase) (int fd, struct map_session_data *sd);
 	/* */
-	void (*add_random_options) (unsigned char* buf, struct item* item);
+	int (*add_random_options) (struct ItemOptions *buf, const struct item *it);
```
MishimaHaruna: since we're calling them Item Options rather than Random Options, I think this method should also be renamed accordingly

src/map/itemdb.c
```diff
@@ -1299,6 +1309,90 @@ void itemdb_read_packages(void) {
...
+			continue;
+		}
+		
+		if (!libconfig->setting_lookup_string(conf, "Name", &str)) {
+			ShowError("itemdb_read_options: Invalid Option variable name provided for Option Id %d in '%s', skipping...\n", t_opt.index, filepath);
+			continue;
+		}
+		
+		/* Set Constant */
+		script->set_constant(str, t_opt.index, false, false);
```
MishimaHaruna: What happens when you reload the item DB? You might want to use `script->set_constant2()` instead, that allows redefinition as long as the value doesn't change.

src/map/itemdb.c
```diff
@@ -1299,6 +1309,90 @@ void itemdb_read_packages(void) {
...
+		libconfig->destroy(&item_options_db);
+		return;
+	}
+	
+	while ((conf = libconfig->setting_get_elem(ito, index++))) {
+		struct item_option t_opt = { 0 }, *s_opt = NULL;
+		const char *str = NULL;
+		
+		if (!libconfig->setting_lookup_int16(conf, "Id", &t_opt.index) || t_opt.index <= 0) {
+			ShowError("itemdb_read_options: Invalid Option Id provided for entry %d in '%s', skipping...\n", t_opt.index, filepath);
```
MishimaHaruna: What if it's a duplicated ID?

src/map/itemdb.c
```diff
@@ -1299,6 +1309,90 @@ void itemdb_read_packages(void) {
...
+			continue;
+		}
+		
+		if (!libconfig->setting_lookup_string(conf, "Name", &str)) {
+			ShowError("itemdb_read_options: Invalid Option variable name provided for Option Id %d in '%s', skipping...\n", t_opt.index, filepath);
```
MishimaHaruna: What if it's a duplicated name? And what if it contains invalid characters? This should be handled similarly to the AegisName

src/map/itemdb.c
```diff
@@ -2373,13 +2486,15 @@ void do_final_itemdb(void) {
...
 	itemdb->other->destroy(itemdb->other, itemdb->final_sub);
+	itemdb->options->destroy(itemdb->options, itemdb->options_final_sub);
 	itemdb->destroy_item_data(&itemdb->dummy, 0);
 	db_destroy(itemdb->names);
 }
 
 void do_init_itemdb(bool minimal) {
 	memset(itemdb->array, 0, sizeof(itemdb->array));
 	itemdb->other = idb_alloc(DB_OPT_BASE);
+	itemdb->options = idb_alloc(DB_OPT_RELEASE_BOTH);
```
MishimaHaruna: You probably want `DB_OPT_RELEASE_DATA` here (the key is numeric, there's nothing to release). It's not safe to use options that include `DB_OPT_RELEASE_KEY` in DBMaps that have integers as keys.

src/map/script.c
```diff
@@ -13546,6 +13575,176 @@ BUILDIN(getiteminfo)
...
+ *        IT_OPT_INDEX     ID of the item option.
+ *        IT_OPT_VALUE     Amount of the bonus to be added.
+ * @return value of the type or -1.
+ */
+BUILDIN(getequippedoptioninfo)
+{
+	int val = 0, type = script_getnum(st, 2);
+	struct map_session_data *sd = NULL;
+	
+	if ((sd = script->rid2sd(st)) == NULL || status->current_equip_item_index == -1
```
MishimaHaruna: Shouldn't you also check the value of `status->current_equip_option_info`?

src/map/script.c
```diff
@@ -13546,6 +13575,176 @@ BUILDIN(getiteminfo)
...
+ */
+BUILDIN(getequipoption)
+{
+	int val = 0, equip_index = script_getnum(st, 2);
+	int slot = script_getnum(st, 3);
+	int opt_type = script_getnum(st, 4);
+	int i = -1;
+	struct map_session_data *sd = script->rid2sd(st);
+	
+	if (sd == NULL) {
```
MishimaHaruna: Can I ask to break this elseif chain? The last check has especially nothing to do with the rest...

src/map/script.c
```diff
@@ -13546,6 +13575,176 @@ BUILDIN(getiteminfo)
...
+		return false;
+	} else if (slot <= 0 || slot > MAX_ITEM_OPTIONS) {
+		script_pushint(st, -1);
+		ShowError("buildin_getequipoptioninfo: Invalid option slot %d (Min: 1, Max: %d) provided.\n", slot, MAX_ITEM_OPTIONS);
+		return false;
+	} else if (equip_index > 0 && equip_index <= ARRAYLENGTH(script->equip)) {
+		i = pc->checkequip(sd, script->equip[equip_index - 1]);
+	}
+	
+	if (i >= 0 && sd->status.inventory[i].nameid) {
```
MishimaHaruna: And else? The documentation above says that I should expect -1 on failure

src/map/status.c
```diff
@@ -2646,7 +2646,42 @@ int status_calc_pc_(struct map_session_data* sd, enum e_status_calc_opt opt)
...
+		if (i == EQI_HAND_R && sd->equip_index[EQI_HAND_L] == index)
+			continue;
+		else if (i == EQI_HEAD_MID && sd->equip_index[EQI_HEAD_LOW] == index)
+			continue;
+		else if (i == EQI_HEAD_TOP && (sd->equip_index[EQI_HEAD_MID] == index || sd->equip_index[EQI_HEAD_LOW] == index))
+			continue;
+		
+		if (index >= 0 && sd->inventory_data[index]) {
+			int j;
+			struct item_option *ito = NULL;
```
MishimaHaruna: Move this to inside the for loop?

src/map/status.c
```diff
@@ -2646,7 +2646,42 @@ int status_calc_pc_(struct map_session_data* sd, enum e_status_calc_opt opt)
...
+			for (j = 0; j < MAX_ITEM_OPTIONS; j++) {
+				short option_index = sd->status.inventory[index].option[j].index;
```
MishimaHaruna: Any reason to use short here instead of int?

src/map/status.c
```diff
@@ -2646,7 +2646,42 @@ int status_calc_pc_(struct map_session_data* sd, enum e_status_calc_opt opt)
...
+				
+				if (option_index == 0)
+					continue;
+				
+				status->current_equip_option_index = j;
+				
+				if ((ito = itemdb->option_exists(option_index)) == NULL || ito->script == NULL)
+					continue;
+				else
```
MishimaHaruna: This else looks very redundant after the continue

src/map/status.c
```diff
@@ -2646,7 +2646,42 @@ int status_calc_pc_(struct map_session_data* sd, enum e_status_calc_opt opt)
...
+					script->run(ito->script, 0, sd->bl.id, 0);
+				
+				if (calculating == 0) //Abort, script->run his function. [Skotlex]
+					return 1;
+			}
+		}
+		status->current_equip_option_index = -1;
```
MishimaHaruna: Redundant. It's already cleared at the beginning of the loop. If you want to clean it up at the end, move this two lines below, after the end of the loop.
Also, does `status->current_equip_index` not need to be cleared?

sql-files/upgrades/2017-03-05--08-09.sql
```diff
@@ -0,0 +1,15 @@
...
+	ADD COLUMN `opt_idx1` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_val0`,
+	ADD COLUMN `opt_val1` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_idx1`,
+	ADD COLUMN `opt_idx2` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_val1`,
+	ADD COLUMN `opt_val2` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_idx2`,
+	ADD COLUMN `opt_idx3` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_val2`,
+	ADD COLUMN `opt_val3` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_idx3`,
+	ADD COLUMN `opt_idx4` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_val3`,
+	ADD COLUMN `opt_val4` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `opt_idx4`; 
+	
+INSERT INTO `sql_updates` (`timestamp`, `ignored`) VALUES (1488744559 , 'No');
```
Smokexyz: Done

src/map/clif.h
Smokexyz: Done

src/map/itemdb.c
Smokexyz: Done

src/map/itemdb.c
Smokexyz: Done

Smokexyz: Fixed, thanks.

src/char/char.c
```diff
@@ -790,9 +797,12 @@ int char_memitemdata_to_sql(const struct item items[], int max, int id, int tabl
 			 && items[i].card[2] == item.card[2]
 			 && items[i].card[3] == item.card[3]
 			) {
+				int k = 0;
 				// They are the same item.
 				ARR_FIND(0, MAX_SLOTS, j, items[i].card[j] != item.card[j]);
-				if (j == MAX_SLOTS
+				ARR_FIND(0, MAX_ITEM_OPTIONS, k, items[i].option[k].index != item.option[k].index);
```
Smokexyz: Fixed, thanks.

src/map/script.c
```diff
@@ -13546,6 +13575,200 @@ BUILDIN(getiteminfo)
...
+	int i = -1;
+	struct map_session_data *sd = script->rid2sd(st);
+	
+	if (sd == NULL) {
+		script_pushint(st, -1);
+		ShowError("buildin_getequipoptioninfo: Player not attached!\n");
+		return false;
+	}
+
+	if (equip_index < 0) {
```
MishimaHaruna: This check happens twice (see line 13653). This one should be removed

src/map/script.c
```diff
@@ -13546,6 +13575,200 @@ BUILDIN(getiteminfo)
...
+	struct map_session_data *sd = script->rid2sd(st);
+	struct item_option *ito = NULL;
+	
+	if (sd == NULL) {
+		script_pushint(st, 0);
+		ShowError("buildin_setequipoption: Player not attached!\n");
+		return false;
+	}
+
+	if (equip_index < 0) {
```
MishimaHaruna: Duplicated check here as well (line 13726)

src/map/script.c
```diff
@@ -13546,6 +13575,200 @@ BUILDIN(getiteminfo)
...
+			script_pushint(st, 0);
+			return false;
+		}
+	} else {
+		ShowError("buildin_setequipoptioninfo: Invalid equipment index %d provided.\n", equip_index);
+		script_pushint(st, 0);
+		return false;
+	}
+	
+	if (sd->status.inventory[i].nameid) {
```
MishimaHaruna: Please consider using an explicit `!= 0` when checking non-boolean numeric variables

src/map/script.c
```diff
@@ -13546,6 +13575,200 @@ BUILDIN(getiteminfo)
...
+		sd->status.inventory[i].option[slot-1].index = ito->index;
+		/* Add Option Value */
+		sd->status.inventory[i].option[slot-1].value = value;
+		
+		/* Unequip and simulate deletion of the item. */
+		pc->unequipitem(sd, i, PCUNEQUIPITEM_FORCE); // status calc will happen in pc->equipitem() below
+		clif->refine(sd->fd, 0, i, sd->status.inventory[i].refine); // notify client of a refine.
+		clif->delitem(sd, i, 1, DELITEM_MATERIALCHANGE); // notify client to simulate item deletion.
+		/* Log the Item */
+		logs->pick_pc(sd, LOG_TYPE_SCRIPT, 1, &sd->status.inventory[i], sd->inventory_data[i]);
```
MishimaHaruna: Logs should happen in pairs, there should be a matching entry with `-1` before (see `BUILDIN(successrefitem)` for reference)

src/map/status.c
```diff
@@ -2646,6 +2646,39 @@ int status_calc_pc_(struct map_session_data* sd, enum e_status_calc_opt opt)
...
+			continue;
+		else if (i == EQI_HEAD_MID && sd->equip_index[EQI_HEAD_LOW] == index)
+			continue;
+		else if (i == EQI_HEAD_TOP && (sd->equip_index[EQI_HEAD_MID] == index || sd->equip_index[EQI_HEAD_LOW] == index))
+			continue;
+		
+		if (index >= 0 && sd->inventory_data[index]) {
+			int j = 0;
+			for (j = 0; j < MAX_ITEM_OPTIONS; j++) {
+				uint16 option_index = sd->status.inventory[index].option[j].index;
```
MishimaHaruna: `sd->status.inventory[index].option[j].index` is `int16`, not `uint16`

src/map/storage.c
```diff
@@ -133,9 +133,12 @@ int compare_item(struct item *a, struct item *b)
 		a->bound == b->bound &&
 		a->unique_id == b->unique_id)
 	{
-		int i;
-		for (i = 0; i < MAX_SLOTS && (a->card[i] == b->card[i]); i++);
-		return (i == MAX_SLOTS);
+		int i = 0, k = 0;
+		ARR_FIND(0, MAX_SLOTS, i, a->card[i] == b->card[i]);
```
MishimaHaruna: Did you mean `!=`?

Smokexyz: Oops. Yes, fixed.

src/map/status.c
Smokexyz: Fixed.

src/map/script.c
Smokexyz: Added.

src/map/script.c
Smokexyz: Removed.

src/map/script.c
Smokexyz: Removed.

src/map/script.c
Smokexyz: Fixed.

MishimaHaruna:
Thank you, all clear from my side. Looks good to merge once rebased, if @4144 approves

src/map/script.c
```diff
@@ -13706,7 +13919,7 @@ BUILDIN(petloot)
 BUILDIN(getinventorylist)
 {
 	struct map_session_data *sd = script->rid2sd(st);
-	char card_var[NAME_LENGTH];
+	char card_var[26];
```
Emistry: using a new define constant? or something like `char card_var[NAME_LENGTH + 10];` ?

src/map/script.c
```diff
@@ -13739,7 +13960,7 @@ BUILDIN(getinventorylist)
 BUILDIN(getcartinventorylist)
 {
 	struct map_session_data *sd = script->rid2sd(st);
-	char card_var[26];
+	char card_var[30];
```
Emistry: using a new define constant? or something like `char card_var[NAME_LENGTH + 10];` ?

MishimaHaruna: it's not really `NAME_LENGTH` -- it should rather be `SCRIPT_VARNAME_LENGTH`, since it's a variable name

src/map/itemdb.c
```diff
@@ -1299,6 +1309,125 @@ void itemdb_read_packages(void) {
...
+		t_opt.script = *str ? script->parse(str, filepath, t_opt.index, SCRIPT_IGNORE_EXTERNAL_BRACKETS, NULL) : NULL;
+
+		/* Allocate memory and copy contents */
+		CREATE(s_opt, struct item_option, 1);
+
+		*s_opt = t_opt;
+		
+		/* Store ptr in the database */
+		idb_put(itemdb->options, t_opt.index, s_opt);
+
```
4144: Here need call additional function for read custom plugin option flags. Same like in itemdb.

Smokexyz: Woops, forgot the call the function 🤣

src/map/script.c
Smokexyz: Fixed, it is of `SCRIPT_VAR_LENGTH`.

sql-files/main.sql
```diff
@@ -79,6 +79,16 @@ CREATE TABLE IF NOT EXISTS `auction` (
   `card1` SMALLINT(11) NOT NULL DEFAULT '0',
   `card2` SMALLINT(11) NOT NULL DEFAULT '0',
   `card3` SMALLINT(11) NOT NULL DEFAULT '0',
+  `opt_idx0` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0',
+  `opt_val0` SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0',
```
dastgir: I guess val shouldn't be unsigned?

Smokexyz: Yes, fixed.

4144:
I think from my side clear too

MishimaHaruna:
Ok, let's merge this then, thank you!

PR #1601, opened 2017-03-03
Update Global Function - F_Sex

Emistry:
- fix invalid value returned.
- fix invalid example.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1601)

### DIFF
--- a/npc/other/Global_Functions.txt
+++ b/npc/other/Global_Functions.txt
@@ -159,11 +159,8 @@ function	script	F_Rand	{
 //== Function F_Sex ========================================
 // Returns 1st argument if female, 2nd argument otherwise
 // Example: mes callfunc("F_Sex","What a beautiful lady!","What a handsome man!");
-function	script	F_SexMes	{
-	if (Sex == SEX_MALE)
-		return getarg(0);
-	else
-		return getarg(1);
+function	script	F_Sex	{
+	return getarg(Sex == SEX_MALE);
 }
 
 //== Function F_Hi =========================================


Asheraf:
I think the function should be changed to F_Sex instead, it could be used with other commands beside mes.

Ridley8819:
Can you give an example where to use this function? I mean you can simply do `if (sex)  do this` for male and continue normal for females

Emistry:
@Ridley8819 
```mes callfunc("F_Sex","What a beautiful lady!","What a handsome man!");```
Anyway, this function are not using in any scripts right now, probably has been removed in the past.
I am just fixing it since it still exist in the repo.

**Comment by Ridley8819** [comment]
I just find it not useful, and rather have it removed entirely since you could either do 
`If (sex) 
do this 
else
do that `
Or
`mes "You are a"+ ((sex)?"handsome man":"beautiful lady");`

File: npc/other/Global_Functions.txt [hunk 1]
```diff
@@ -159,11 +159,8 @@ function	script	F_Rand	{
 //== Function F_Sex ========================================
 // Returns 1st argument if female, 2nd argument otherwise
 // Example: mes callfunc("F_Sex","What a beautiful lady!","What a handsome man!");
-function	script	F_SexMes	{
-	if (Sex == SEX_MALE)
-		return getarg(0);
-	else
-		return getarg(1);
+function	script	F_Sex	{
+	return getarg(Sex == SEX_MALE);
```
Smokexyz: `if (Sex == SEX_MALE)` would return 0, but in this case, getarg(1) would be returned.

Smokexyz: Since there is a constant for Sex, I don't see the need for this function at all. Current functions using this function could be replaced with the constant in a new PR.

Smokexyz:
Since there is a constant for Sex, I don't see the need for this function at all. Current functions using this function could be replaced with the constant in a new PR.

npc/other/Global_Functions.txt
Emistry: ```callfunc("F_Sex","What a beautiful lady!","What a handsome man!");```
`getarg(0)` - for female , result should be `What a beautiful lady`
`getarg(1)` - for male , result should be `What a handsome man!`

`SEX_FEMALE = 0`
`SEX_MALE = 1`

when `(Sex == SEX_MALE)`,
if your char's Sex = Female , which is value `0` .... the condition become `if ( 0 == 1 )`, which is `false`, and become `getarg(0)`
if your char's Sex = Male , which is value `1` .... the condition become `if ( 1 == 1 )`, which is `true`, and become `getarg(1)`

it create the correct output. I dont see how it cause issues to any of current scripts.

Smokexyz: Basically, if (Sex == SEX_MALE) would return 1 because the clause is true.

Smokexyz: you are correct, sorry.

Ridley8819:
The function isn't used anywhere, not to mention i still think we don't need it at all. We can remove it with a new pr

Issue #1524, opened 2016-12-05
Chance in Shadow refiner.


Atemo:
Githash e1453434af18b30119114077dfc52d27ad196cef

The __double chance__ in the Shadow refiner should only be available for Enrich ore.
https://github.com/HerculesWS/Hercules/blob/master/npc/re/merchants/shadow_refiner.txt#L161

hemagx:
Correct, actually according to the official script it doesn't support enriched at all.
@dastgir @Asheraf @Emistry wanna help?

Asheraf:
I would take it.

PR #1602, opened 2017-03-03
Fix issue #1524

Emistry:
- double chance should only available for Enrich ores.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1602)


Smokexyz:
This is good, unnecessary to keep it waiting.

PR #1607, opened 2017-03-04
Update Global Function - Remove F_SexMes

Emistry:
- not using in any script, could be easily replace by simply if-else
  condition.

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1607)


Ridley8819:
Thank you Emistry :)

PR #1610, opened 2017-03-05
allow the PC timer buildins to run on another player

mekolat:
### Pull Request Prelude

### Changes Proposed
- add a new parameter to `addtimer`, `addtimercount`, and `deltimer` to make them usable on another player
- ~~~add new buildin: `areatimer("map", x1, y1, x2, y2, tick, "npc::event")`~~~

**Issues addressed:** (none)

### Known Issues
(none)

### TODO
- [x] update documentation

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1610)

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -6475,15 +6475,16 @@ Size is 0 = normal 1 = small 2 = big.
 //=====================================
 ---------------------------------------
 
-*addtimer(<ticks>, "NPC::OnLabel")
-*deltimer("NPC::OnLabel")
-*addtimercount("NPC::OnLabel", <ticks>)
+*addtimer(<ticks>, "NPC::OnLabel"{, <account id>})
+*deltimer("NPC::OnLabel"{, <account id>})
+*addtimercount("NPC::OnLabel", <ticks>{, <account id>})
 
 These commands will create, destroy, and delay a countdown timer - 
 addtimer() to create, deltimer() to destroy and addtimercount() to delay
 it by the specified number of ticks. For all three cases, the event label 
 given is the identifier of that timer. The timer runs on the character 
-object that is attached to the script, and can have multiple instances. 
+object that is attached to the script, and can have multiple instances.
+If <acccount id> is passed, this player will be used instead.
 When the label is run, it is run as if the player that the timer runs on 
 has clicked the NPC.
 
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -10790,7 +10790,11 @@ BUILDIN(addtimer)
 	struct map_session_data *sd;
 
 	script->check_event(st, event);
-	sd = script->rid2sd(st);
+
+	if (script_hasdata(st, 4))
+		sd = map->id2sd(script_getnum(st, 4));
+	else
+		sd = script->rid2sd(st);
 
 	if (sd == NULL) {
 		script_pushint(st, 0);
@@ -21092,7 +21096,7 @@ void script_parse_builtin(void) {
 		BUILDIN_DEF(clone,"siisi????"),
 		BUILDIN_DEF(doevent,"s"),
 		BUILDIN_DEF(donpcevent,"s"),
-		BUILDIN_DEF(addtimer,"is"),
+		BUILDIN_DEF(addtimer,"is?"),
 		BUILDIN_DEF(deltimer,"s"),
 		BUILDIN_DEF(addtimercount,"si"),
 		BUILDIN_DEF(initnpctimer,"??"),
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -10841,7 +10841,11 @@ BUILDIN(addtimercount)
 
 	event = script_getstr(st, 2);
 	tick = script_getnum(st, 3);
-	sd = script->rid2sd(st);
+
+	if (script_hasdata(st, 4))
+		sd = map->id2sd(script_getnum(st, 4));
+	else
+		sd = script->rid2sd(st);
 
 	if (sd == NULL)
 		return true;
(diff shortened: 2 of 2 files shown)


Emistry:
`areatimer("map", x1, y1, x2, y2, tick, "npc::event")`
if you added this, perhaps you may consider to add for `delareatimer` to remove the timer also?

normally the old timer will be removed before the new timer is added to player.

mekolat:
@Emistry why would you want to check if the player has a timer before deleting a timer? deltimer doesn't care if you already have it or not.

what I could do is remove `areatimer` and add a `getareaunits("<map>", <x1>, <y1>, <x2>, <y2>, <type>, <array>)` so you'd be able to do the following:

```c
    npctalk("I am the most beautiful being the Universe ever begat!");
    sleep(500);
    npctalk("...and you shall all love me!");

    getareaunits("map", 45, 67, 109, 111, UNITTYPE_PC, .@gid);

    for (.@i = 0; .@i < getarraysize(.@gid); ++.@i)
    {
        deltimer(strnpcinfo(0) + "::OnEvent", .@gid[.@i]); // remove old timer, if any
        addtimer(500, strnpcinfo(0) + "::OnEvent", .@gid[.@i]);
    }
    end;

OnEvent:
    emotion(e_heart); // force everyone in the area to do a heart emote
    end;
```

@MishimaHaruna what would you prefer?

Emistry:
definitely downvote for that `getareaunits` thing.... @AnnieRuru show a better approach for `getmemberaid` previously.

the reason to add the `deltimerarea`, because you added the `areatimer` thing ... 
almost all the time, when you assign the timer in any scripts , you would deltimer any existing timers before you assign it. You commit lack of this script command.

mekolat:
@Emistry what do you mean about `getmemberaid`?

get on IRC please?

Emistry:
hmm
> getunits(AREA, UNITTYPE_PC, 50, .@account_id, "map", x1, y1, x2, y2) which would fill .@account_id with up to 50 players (UNITTYPE_PC) in the area
why need to limit the size? if compare with the `getmemberaid` did by @AnnieRuru  the advantages of `getunits` are just the array assignment + unit type to filter, which are nice to have.
if insist to add the parameter for `size`, perhaps it should be add as the last optional parameter.

anyway, if you insist to add the `getunits` please at least make it meaningful ? something like `getunitsaid` or `getunitsrid` or `getunitsid` ? at least we know the script command return `account_id` or `gid` from the unit by looking at it.

for the `delareatimer` implementation...
> areatimer("map", x1, y1, x2, y2, tick, "npc::event")
> if you added this, perhaps you may consider to add for `delareatimer` to remove the timer also?
> normally the old timer should be removed before the new timer is added to player to avoid duplicate 
> timer running.

You are trying to attach the timer to everyone in that area ... but you didnt have a way to delete their existing timer before you attach a new timer to them, which end up potentially create multiple timers that execute the same event label.
unless you trying to tell me you gonna do it this way...
```
.@size = getunits( AREA, UNITTYPE_PC, .@aid );
for ( .@i = 0; .@i < .@size; .@i++ ) {
	deltimer "npc::event",.@aid[.@i];
}
areatimer("map", x1, y1, x2, y2, tick, "npc::event");
```
and not something like this..
```
deltimer( "npc::event","map",x1,y1,x2,y2 );
areatimer("map", x1, y1, x2, y2, tick, "npc::event");
```
or 
```
.@size = getunits( AREA, UNITTYPE_PC, .@aid );
for ( .@i = 0; .@i < .@size; .@i++ ) {
	deltimer "npc::event",.@aid[.@i];
	addtimer 10000, "npc::event",.@aid[.@i];
}
```
which end up `areatimer()` are kinda useless in this case.

beside, all these new script commands are just extra bonus use in custom scripts, you rarely find any official scripts are using these timer.

mekolat:
@Emistry 

> why need to limit the size? aren't Herc already support array up to 2 Billions index?

>> `<_41444d494e_>` if i not wrong now getunits() get all in area and put into array
`<_41444d494e_>` but it can be slow if here 1k or 10k units
(...)
`<meko>` yeah let's say you just want to execute it on 10 nearby players, it would pointless to fill the whole array if you just want 10
`<meko>` I could make so that if you put 0 as limit then there's no cap besides MAX_ARRAY_SIZE
`<smoke_>` makes sense
`<smoke_>` you could alternatively make the limit an optional entry
`<meko>` nah I want to make that explicit
`<meko>` if you don't want a limit pass false (0), else pass a number
`<smoke_>` Yeah it's not a problem, but I don't see the point of limiting in that case, when you could just reduce the area size.
`<smoke_>` But since 0 is unlimited, it's okay I guess.
`<meko>` you can reduce the area size indeed but while extremely unlikely it's not impossible to have 1k players standing on one single tile
`<smoke_>` true :P

<br><br>

> `getunits` I dont know what it return by looking at it, return id? units struct? aid? gid? cid?

it returns unit id (aka BL_ID, aka GID), which is independent of unit type, so adding the `aid` or `rid` prefix here makes no sense since those are PC-specific and this buildin is type-agnostic
<br><br>

> You are trying to attach the timer to everyone in that area ... but you didnt have a way to delete their existing timer before you attach a new timer to them, which end up potentially create multiple timers that execute the same event label.

didn't you read my earlier comment? or even look at the commits in this PR? this PR is mostly to add the `<account id>` parameter to the existing timer buildins; getunits() is just icing on the cake, and GEEZ, please just forget about `areatimer`, the `getunits` makes it obsolete so I'm not adding it (nor a `delareatimer`)

```c
    for (.@i = 0; .@i < getarraysize(.@gid); ++.@i)
    {
        deltimer(strnpcinfo(0) + "::OnEvent", .@gid[.@i]); // remove old timer, if any
        addtimer(500, strnpcinfo(0) + "::OnEvent", .@gid[.@i]);
    }
```

Emistry:
I do look at the commits, and I still can see the `areatimer`  tho. 
In my point of view `getunits` simply meaningless it should at least `getunitsid` or whatever to represent what it returned from the result in case it get implemented.

anyway, please fix the doc/script_commands.txt why are there almost a empty space modification at every single line.

mekolat:
@Emistry the reason you still see `areatimer` is because I didn't yet remove it, nor start working on `getunits()`, because I'm waiting till we can finalize the buildin name and its arguments, to not end up doing the same work a dozen time, hence the `WIP` in the title. and yes, I'll fix the doc once this PR is done.

mekolat:
@Emistry @MishimaHaruna I'm removing the getunits/areatimer from this PR so we can just merge the timer buildin modifications; gonna make another PR for getunits later on

MishimaHaruna:
Merged, thank you!

Issue #1533, opened 2016-12-27
Peuz's Greaves - Wrong bonus script


ItsWolfeh:
When refine is below 7, it gives -1 agility, and when above, it gives +1 and only +1. Item description states no negative agility and that you get more than +1 Agi (it's for every refine above 7.)

Script:
bonus bAgi,-7+min(getequiprefinerycnt(EQI_SHOES),8);
Suggested Fix:
if(getequiprefinerycnt(EQI_SHOES)>7) bonus bAgi,getequiprefinerycnt(EQI_SHOES)-7;

Mysteries:
@iRagno

iRagno:
This is description from kRO:

> MSP + 40.
> 8제련 시 부터 제련도 1당 AGI + 1 증가.

And this is description from jRO:

> MaxSP + 40 
> ――――――――――――― 
> [ペオースグリーブ]の 
> 精錬値が8以上の時、 
> 精錬値が1上がる度に追加で 
> Agi + 1 

From translation on Google and Bing, I can say it increases 1 Agi per every refine over 8 (refine 9: +1 agi, refine 10: +2 agi).

bRO description even say it more clearly:

> SP Máx. +40. 
> AGI +1 por refino acima do +8. 

Actual script comes from iRO description:

> MaxSP + 40.
> Reduces AGI by 7. Refine rate reduces the AGI reduction by 1% for each level up to +8.

Since kRO and jRO are coincident, I would say maybe iRO description has a translation error or custom effect.

Script could be changed to:

`bonus bAgi, (getequiprefinerycnt(EQI_SHOES) > 8 ? getequiprefinerycnt(EQI_SHOES) - 8, 0);`

iRagno:
I just get one Peuz Greaves without refine on iRO, and keeps intact Agi when equipped:

![peuz](http://i65.tinypic.com/3022vyb.jpg)

I can say bug is confirmed.

iRagno:
Since I don't know korean and use translators, I have doubt about required level and search for items with similar description. I found this:

13061 - Black Wing [1] :
> Korean: 6제련 시 부터, 제련도 당 해당 스킬의 공격력 2% 추가 증가 , MATK + 3씩 증가.
> English: 2% ATK and +3 MATK increase with every **upgrade above level 6**.

> Hercules Effect:
```
		if(getrefine() >= 6) {
			bonus2 bSkillAtk,SC_FATALMENACE,(getrefine()-5)*2;
[...]

PR #1620, opened 2017-03-11
Item DB Fixes.

iRagno:

### Pull Request Prelude

### Changes Proposed

- Fixes Peuz_Greave effect with refine level.

[//]: # (Master? Slave?)

**Issues addressed:**

- #1533 Peuz's Greaves - Wrong bonus script

None

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1620)

### DIFF
--- a/db/re/item_db.conf
+++ b/db/re/item_db.conf
@@ -31571,7 +31571,8 @@ item_db: (
 	EquipLv: 100
 	Script: <"
 		bonus bMaxSP,40;
-		bonus bAgi,-7+min(getequiprefinerycnt(EQI_SHOES),8);
+		if (getrefine() >= 8)
+			bonus bAgi, getrefine() - 7;
 	">
 },
 {


db/re/item_db.conf
```diff
@@ -61939,6 +61940,11 @@ item_db: (
 	Slots: 1
 	Loc: 256
 	View: 525
+	Script: <"
+		bonus bDex, 2;
+		skill TF_DOUBLE, 5;
+		bonus bDoubleRate, 25;
```
Asheraf: I cant find this effect in kRO description. from where did you get it?

db/re/item_db.conf
```diff
@@ -61939,6 +61940,11 @@ item_db: (
 	Slots: 1
 	Loc: 256
 	View: 525
+	Script: <"
+		bonus bDex, 2;
+		skill TF_DOUBLE, 5;
```
Asheraf: according to kRO description this should be
`skill TF_DOUBLE, max(5, getskilllv(TF_DOUBLE));`

iRagno: When description says "activate it at that level" it doesn't refer to enable skill to a higher level, but to the chance to do a double attack, and that refer to bonus bDoubleRate, wich uses the maximum skill level learned for that. It's like when having two buffs of diferent level but the same kind, only is applied the higher one.

db/re/item_db.conf
iRagno: It comes implicit with the "activate it at that level" part and is present in other items such as sidewinder card and chick hat, both of them doesn't have that effect on aegis script, but it's also applied to work. It is exactly the same, if this effect is wrong, then all the items with the same effect also are.

db/re/item_db.conf
Ridley8819: @iRagno you are wrong here. The Hat is supposed to give you Level 5 Double Attack (+ your own skill level in it). Sidewinder Card just gives a chance to cast it, while the hat actually gives you the skill. This is something entirely different.
Asheraf is right here.

iRagno: Well, let's check description again. This is kRO description (in korean) :

> 머리 위에서 잠이 든 귀여운 아기 용.
> 이대로 무얼하든 깨어나지 않을 것 같다.
> DEX+2
> 5레벨 더블어택 사용 가능.
> 더 높은 레벨을 배우고 있을 시, 해당 레벨로 발동.
> 계열 : 투구 방어 : 1
> 위치 : 상단 무게 : 10
> 요구 레벨 : 1
> 장착 : 전 직업

This is translation according to Google:

> For a cute baby sleeping on the head.
> No matter what I do, I will not wake up.
> DEX + 2
> Level 5 Double Attack available.
> If you are learning a higher level, activate it at that level.
> Category: Helm Defense: 1
> Location: Top Weight: 10
> Requires Level: 1
> Mounting: former occupation

And this is translation to propose to english, according to *athena format:

> A baby dragon who has been trained to help its master.
> DEX + 2
> Enables Level 5 Double Attack (When you learn a higher level, activate the corresponding level).
> Class: Headgear
> Defense: 1
> Location: Upper
> Weight: 10
> Level Requirement: 1
> Jobs: All classes

This headgear enables usage of Level 5 Double Attack with equipped weapon, but, if player has learned double attack at higher level, the chance is influenced by the level double attack is learned.

What you both propose is to enable Double Attack at level 5 or enable it at a higher level if player has already learned it at higher level. That is what that proposal script does and that doesn't have any sense, why would enable usage of a skill already learned?

There are several items who already brings Double Attack usage on actual database and are these:

```
1130	Nagan		skill TF_DOUBLE,5;	bonus bDoubleRate,25;
2820	Thief Manual	skill TF_DOUBLE,3;	bonus bDoubleRate,15;
4117	Sidewinder Card	skill TF_DOUBLE,1;	bonus bDoubleRate,5;
5283	Chick Hat	skill TF_DOUBLE,2;	bonus bDoubleRate,10;
5388	Snake Head	skill TF_DOUBLE,5;	bonus bDoubleRate,25;
13407	Refined Nagan	skill TF_DOUBLE,5;	bonus bDoubleRate,25;
```
When I say it's implicit on script I mean this `bonus bDoubleRate` activates it at learned level because effect gets influenced by the level of Double Attack learned, coming from the enabled skill by the gear or coming from naturally learned this skill.

I can't agree with you :( This has been the behaviour that this kind of gears have had since ever and I cannot change it.

Emistry: Does it mean, if a thief has level 10 Double Attack  equipped the item which give him another Level 5 Double Attack, hence result in a Level 15 Double Attack ?

iRagno: No, it will have level 10 Double Attack, but will be able to do a Double Attack with 50% chance with every weapon he equips, such as Swords, Bow and Axe, and when being an Assassin, will be able to do Double Attack with katars, just because have equipped that kind of gear, otherwise they couldn't do the Double Attack (even if learns the skill through gear), because Double Attack only works with daggers.

iRagno: I even remember there were a time when these items wasn't enabling skill usage, and somebody reported it as an issue and then was added the enable skill script to those items.

I just don't remember where was reported, but if needed I can search for it.

Ridley8819: The most important thing you miss is: This is your interpretation of it. The item is stating clear how it is supposed to work, it does not matter if it makes sense or not, and it's a huge difference to enable to cast it or to give the skill. 
I actually wanted to test this on kRO but I couldn't find the item. After talking with some korean players, they told me they've never seen the item on kRO so a live test doesn't seem possible atm. 
Therefore (imo) we need to go with the translation and infos we have, rather than interpreting and changing it.

iRagno: I don't know what to say Ridley. Sincerelly I don't think that have `skill TF_DOUBLE, max(5, getskilllv(TF_DOUBLE));` is the right way to understand `더 높은 레벨을 배우고 있을 시, 해당 레벨로 발동` since that text is also present (exactly) in Chick_Hat and it doesn't require to evaluate any skill learned, because it is implicit.

Even Sidewinder Card has that clausule on it's description:

> [더블어택] Lv 1
> 더블어택 스킬을 익힌 경우, 익힌 더블어택 스킬 레벨의 발동확률에 따른다.
> [Double Attack] Lv 1
> If you learn Double Attack skill, it will depend on the probability of playing Double Attack skill level.

And it does not make any check on player's skill level, just like Chick_Hat.

If after what I have explained you still can't share opinion, I can't, I won't and I will not push on this fix. I will take myself apart to allow other to send proper fix.

Emistry: I agree with @iRagno , it should not be giving the skill ... 
should be the `bonus bDoubleRate,x;`

since kRO has no way to test it, what about the AEGIS ? does it exist in there? someone take a look there?

iRagno:
I'm sorry guys, I sincerelly not agree to changes required to Baby Dragon Hat, and since I don't want to push on it or be an obstacle to solve the issue, I removed the effect proposed to Baby Dragon Hat:

```
	Script: <"
		bonus bDex, 2;
		skill TF_DOUBLE, 5;
		bonus bDoubleRate, 25;
 	">
```

I'm really sorry for the bothering I'm causing, this is not what I want and I understand there is things I just can't fix. I would be grateful with anyone who promoves a better fix to that issue. Again, sorry for the inconveniences.

db/re/item_db.conf
iRagno: That is the way it works once, some years ago, but somebody reported it as a bug, an also add the skill enabling into skill tree and not only give bDoubleRate bonus.

You can check this commits about it: e1cd46aacaab8a0e398d809cfda67aebc3cc77f2, 137096757561eb466f2bc7cbb89d65ca70ba22a3

Issue 5935 it's pretty informative: http://herc.ws/board/tracker/issue-5935-sidewinder-card-effect The other one I couldn't find it. Anyway, maybe those people are wrong too.

iRagno: If you want to check it, @Emistry, you can check this commits: e1cd46aacaab8a0e398d809cfda67aebc3cc77f2, 137096757561eb466f2bc7cbb89d65ca70ba22a3

Issue 5935: http://herc.ws/board/tracker/issue-5935-sidewinder-card-effect The other one I couldn't find it.

Ridley8819:
For now I merge the change to Peutz's, about the Baby Dragon Hat we can continue discussion in IRC

PR #1636, opened 2017-03-15
add gettimer() buildin

mekolat:
### Pull Request Prelude

### Changes Proposed
As requested by @Emistry, here's a buildin to check if the player has timers
* new buildin: `gettimer(<mode>{, <account id>{, "<event>"}})`

**Examples:**
* `gettimer(0)`
    * will return the total number of timers attached to the player
* `gettimer(0, getcharid(CHAR_ID_ACCOUNT), "NPC::EVENT")`
    * will return the number of timers attached to the player for event `NPC::EVENT`
* `gettimer(1, getcharid(CHAR_ID_ACCOUNT), "NPC::EVENT")`
    * will return the number of ticks remaining before the next call to `NPC::EVENT`
* `gettimer(2, getcharid(CHAR_ID_ACCOUNT), "NPC::EVENT")`
    * will return the number of ticks remaining before the last call to `NPC::EVENT`

**Affected Branches:** 
* `master`

**Issues addressed:**
<sub>(none)</sub>

### TODO
- [ ] documentation

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1636)

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -6479,17 +6479,13 @@ Size is 0 = normal 1 = small 2 = big.
 ---------------------------------------
 
 *addtimer(<ticks>, "NPC::OnLabel"{, <account id>})
-*deltimer("NPC::OnLabel"{, <account id>})
-*addtimercount("NPC::OnLabel", <ticks>{, <account id>})
 
-These commands will create, destroy, and delay a countdown timer - 
-addtimer() to create, deltimer() to destroy and addtimercount() to delay
-it by the specified number of ticks. For all three cases, the event label 
-given is the identifier of that timer. The timer runs on the character 
-object that is attached to the script, and can have multiple instances.
-If <acccount id> is passed, this player will be used instead.
-When the label is run, it is run as if the player that the timer runs on 
-has clicked the NPC.
+This command will create a countdown timer.
+The event label given is the identifier of that timer.
+The timer runs on the character object that is attached to the script,
+and can have multiple instances. If <acccount id> is passed, this player
+will be used instead. When the label is run, it is run as if the player
+that the timer runs on has clicked the NPC.
 
 When this timer runs out, a new execution thread will start in the 
 specified NPC object at the specified label.
@@ -6510,6 +6506,39 @@ On5secs:
 	dispbottom("5 seconds have passed!");
 	end;
 }
+
+---------------------------------------
+
+*deltimer("NPC::OnLabel"{, <account id>})
+
+Deletes timers created by addtimer() that matches the given event
+label. Refer to addtimer() for additional information.
+
+---------------------------------------
+
+*addtimercount("NPC::OnLabel", <ticks>{, <account id>})
+
+Delays a timer that was created with addtimer() by <ticks> ticks
+if it matches the given event label. Refer to addtimer() for additional
+information.
+
+---------------------------------------
+
+*gettimer(<type>{, <account id>{, "<event>"}})
+
+Returns informations on timers that were created by addtimer().
+
+valid <type> for gettimer() are:
+
+(0) TIMER_COUNT
+	Will return the total number of timers for the specified or
+	attached player. Can be filtered by <event>.
+(1) TIMER_TICK_NEXT
+	Will return the number of ticks until the next timer runs
+	for the specified or attached player. Can be filtered by <event>.
+(2) TIMER_TICK_LAST
+	Will return the number of ticks until the last timer runs
+	for the specified or attached player. Can be filtered by <event>.
 
 ---------------------------------------
 
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -10861,6 +10861,100 @@ BUILDIN(addtimercount)
 	return true;
 }
 
+enum gettimer_mode {
+	GETTIMER_COUNT = 0,
+	GETTIMER_TICK_NEXT = 1,
+	GETTIMER_TICK_LAST = 2,
+};
+
+BUILDIN(gettimer)
+{
+	struct map_session_data *sd;
+	const struct TimerData *td;
+	int i;
+	int tick;
+	const char *event = NULL;
+	int val = 0;
+	bool first = true;
+	short mode = script_getnum(st, 2);
+
+	if (script_hasdata(st, 3))
+		sd = map->id2sd(script_getnum(st, 3));
+	else
+		sd = script->rid2sd(st);
+
+	if (script_hasdata(st, 4)) {
+		event = script_getstr(st, 4);
+		script->check_event(st, event);
+	}
+
+	if (sd == NULL) {
+		script_pushint(st, -1);
+		return true;
+	}
+
+	switch (mode) {
+	case GETTIMER_COUNT:
+		// get number of timers
+		for (i = 0; i < MAX_EVENTTIMER; i++) {
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				if (event != NULL) {
+					td = timer->get(sd->eventtimer[i]);
+					Assert_retr(false, td != NULL);
+
+					if (strcmp((char *)(td->data), event) == 0) {
+						val++;
+					}
+				} else {
+					val++;
+				}
+			}
+		}
+		break;
+	case GETTIMER_TICK_NEXT:
+		// get the number of tick before the next timer runs
+		for (i = 0; i < MAX_EVENTTIMER; i++) {
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				td = timer->get(sd->eventtimer[i]);
+				Assert_retr(false, td != NULL);
+				tick = max(0, DIFF_TICK32(td->tick, timer->gettick()));
+
+				if (event != NULL) {
+					if ((first == true || tick < val) && strcmp((char *)(td->data), event) == 0) {
+						val = tick;
+						first = false;
+					}
+				} else if (first == true || tick < val) {
+					val = tick;
+					first = false;
+				}
+			}
+		}
+		break;
+	case GETTIMER_TICK_LAST:
+		// get the number of ticks before the last timer runs
+		for (i = MAX_EVENTTIMER - 1; i >= 0; i--) {
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				td = timer->get(sd->eventtimer[i]);
+				Assert_retr(false, td != NULL);
+				tick = max(0, DIFF_TICK32(td->tick, timer->gettick()));
+
+				if (event != NULL) {
+					if (strcmp((char *)(td->data), event) == 0) {
+						val = max(val, tick);
+					}
+				} else {
+					val = max(val, tick);
+				}
+			}
+		}
+		break;
+	}
+
+	script_pushint(st, val);
+	return true;
+}
+
 /*==========================================
  *------------------------------------------*/
 BUILDIN(initnpctimer)
@@ -21115,6 +21209,7 @@ void script_parse_builtin(void) {
 		BUILDIN_DEF(addtimer,"is?"),
 		BUILDIN_DEF(deltimer,"s?"),
 		BUILDIN_DEF(addtimercount,"si?"),
+		BUILDIN_DEF(gettimer,"i??"),
 		BUILDIN_DEF(initnpctimer,"??"),
 		BUILDIN_DEF(stopnpctimer,"??"),
 		BUILDIN_DEF(startnpctimer,"??"),
--- a/db/constants.conf
+++ b/db/constants.conf
@@ -3664,14 +3664,14 @@ constants_db: {
 	PC_PARTY: 1
 	PC_GUILD: 2
 	PC_MAP:   3
-	
+
 	comment__: "strnpcinfo"
 	NPC_NAME:        0
 	NPC_NAME_VISIBLE: 1
 	NPC_NAME_HIDDEN:  2
 	NPC_NAME_UNIQUE:  3
 	NPC_MAP:         4
-	
+
 	comment__: "getcharid"
 	CHAR_ID_CHAR:         0
 	CHAR_ID_PARTY:        1
@@ -3708,6 +3708,11 @@ constants_db: {
 	GETTIME_MONTH:      6
 	GETTIME_YEAR:       7
 	GETTIME_DAYOFYEAR:  8
+
+	comment__: "gettimer"
+	TIMER_COUNT:        0
+	TIMER_TICK_NEXT:    1
+	TIMER_TICK_LAST:    2
 
 	comment__: "unit types"
 	UNITTYPE_PC:   0


Emistry:
missing doc/script_commands.txt

how about add constant to replace the value on the first parameter ?

src/map/script.c
```diff
@@ -10861,6 +10861,91 @@ BUILDIN(addtimercount)
 	return true;
 }
 
+BUILDIN(gettimer)
+{
+	const char *event;
```
Emistry: perhaps provide a default value? according to Travis CI build, it generate warning since the variable might not be initialized if the script command doesn't provide the parameters

src/map/script.c
```diff
@@ -10861,6 +10861,91 @@ BUILDIN(addtimercount)
...
+		script_pushint(st, -1);
+		return true;
+	}
+
+	switch (mode) {
+	case 0:
+		// get number of timers
+		for (i = 0; i < MAX_EVENTTIMER; i++) {
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				if (script_hasdata(st, 4)) {
```
Emistry: could be replaced with `if (event != NULL)`

src/map/script.c
```diff
@@ -10861,6 +10861,91 @@ BUILDIN(addtimercount)
...
+		for (i = 0; i < MAX_EVENTTIMER; i++) {
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				td = timer->get(sd->eventtimer[i]);
+				tick = max(0, DIFF_TICK32(td->tick, timer->gettick()));
+				if (first == true && tick > 0) {
+					val = tick;
+					first = false;
+				}
+
+				if (script_hasdata(st, 4)) {
```
Emistry: could be replaced with `if (event != NULL)`

src/map/script.c
```diff
@@ -10861,6 +10861,91 @@ BUILDIN(addtimercount)
...
+		}
+		break;
+	case 2:
+		// get the number of ticks before the last timer runs
+		for (i = MAX_EVENTTIMER - 1; i >= 0; i--) {
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				td = timer->get(sd->eventtimer[i]);
+				tick = max(0, DIFF_TICK32(td->tick, timer->gettick()));
+
+				if (script_hasdata(st, 4)) {
```
Emistry: could be replaced with `if (event != NULL)`

src/map/script.c
```diff
@@ -10861,6 +10861,91 @@ BUILDIN(addtimercount)
...
+		return true;
+	}
+
+	switch (mode) {
+	case 0:
+		// get number of timers
+		for (i = 0; i < MAX_EVENTTIMER; i++) {
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				if (script_hasdata(st, 4)) {
+					if (strcmp((char *)(timer->get(sd->eventtimer[i])->data), event) == 0) {
```
Emistry: `(char *)` should be `(char*)` ?

src/map/script.c
```diff
@@ -10861,6 +10861,91 @@ BUILDIN(addtimercount)
...
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				td = timer->get(sd->eventtimer[i]);
+				tick = max(0, DIFF_TICK32(td->tick, timer->gettick()));
+				if (first == true && tick > 0) {
+					val = tick;
+					first = false;
+				}
+
+				if (script_hasdata(st, 4)) {
+					if (tick < val && strcmp((char *)(td->data), event) == 0) {
```
Emistry: `(char *)` should be `(char*)` ?

src/map/script.c
```diff
@@ -10861,6 +10861,91 @@ BUILDIN(addtimercount)
...
+		break;
+	case 2:
+		// get the number of ticks before the last timer runs
+		for (i = MAX_EVENTTIMER - 1; i >= 0; i--) {
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				td = timer->get(sd->eventtimer[i]);
+				tick = max(0, DIFF_TICK32(td->tick, timer->gettick()));
+
+				if (script_hasdata(st, 4)) {
+					if (strcmp((char *)(td->data), event) == 0) {
```
Emistry: `(char *)` should be `(char*)` ?

mekolat:
@Emistry all the other PC timer buildins don't have `pc` in the name:
* `addtimer`
* `deltimer`
* `addtimercount`

so it makes sense to also not include it in `gettimer

src/map/script.c
mekolat: looks better as `(char *)` imho

src/map/script.c
```diff
@@ -10861,6 +10861,91 @@ BUILDIN(addtimercount)
...
+		return true;
+	}
+
+	switch (mode) {
+	case 0:
+		// get number of timers
+		for (i = 0; i < MAX_EVENTTIMER; i++) {
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				if (event != NULL) {
+					if (strcmp((char *)(timer->get(sd->eventtimer[i])->data), event) == 0) {
```
4144: timer->get(sd->eventtimer[i])->data) can return NULL
And here will be crash or memory corruption.
Before check for invalid time not mean what timer still registered. For some reason actual timer can be removed. Most time this is bug, but server must not crash because this.

src/map/script.c
```diff
@@ -10861,6 +10861,91 @@ BUILDIN(addtimercount)
...
+				}
+			}
+		}
+		break;
+	case 1:
+		// get the number of tick before the next timer runs
+		for (i = 0; i < MAX_EVENTTIMER; i++) {
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				td = timer->get(sd->eventtimer[i]);
+				tick = max(0, DIFF_TICK32(td->tick, timer->gettick()));
```
4144: td can be NULL

src/map/script.c
```diff
@@ -10861,6 +10861,91 @@ BUILDIN(addtimercount)
...
+				if (first == true && tick > 0) {
+					val = tick;
+					first = false;
+				}
+
+				if (event != NULL) {
+					if (tick < val && strcmp((char *)(td->data), event) == 0) {
```
4144: And here null pointer access in td

src/map/script.c
```diff
@@ -10861,6 +10861,91 @@ BUILDIN(addtimercount)
...
+				}
+			}
+		}
+		break;
+	case 2:
+		// get the number of ticks before the last timer runs
+		for (i = MAX_EVENTTIMER - 1; i >= 0; i--) {
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				td = timer->get(sd->eventtimer[i]);
+				tick = max(0, DIFF_TICK32(td->tick, timer->gettick()));
```
4144: Here possible null pointer in td too

src/map/script.c
```diff
@@ -10861,6 +10861,91 @@ BUILDIN(addtimercount)
...
+
+				if (event != NULL) {
+					if (strcmp((char *)(td->data), event) == 0) {
```
4144: and here

src/map/script.c
mekolat: @4144 I added asserts for the null pointers, please review when you have the time

mekolat:
@MishimaHaruna anything you want me to change here?

src/map/script.c
```diff
@@ -10861,6 +10861,97 @@ BUILDIN(addtimercount)
...
+	case 1:
+		// get the number of tick before the next timer runs
+		for (i = 0; i < MAX_EVENTTIMER; i++) {
+			if (sd->eventtimer[i] != INVALID_TIMER) {
+				td = timer->get(sd->eventtimer[i]);
+				Assert_retr(false, td != NULL);
+				tick = max(0, DIFF_TICK32(td->tick, timer->gettick()));
+
+				if (first == true && tick > 0) {
+					val = tick;
```
MishimaHaruna: Shouldn't this also be conditional to the value of `event`, when it's set?

mekolat:
@MishimaHaruna you're totally right. Fixed that.

MishimaHaruna:
@mekolat Ok, looks good to me now. Please go ahead with documentation and constants

mekolat:
@MishimaHaruna done

doc/script_commands.txt
```diff
@@ -6481,15 +6481,28 @@ Size is 0 = normal 1 = small 2 = big.
...
+and can have multiple instances. If <acccount id> is passed, this player
+will be used instead. When the label is run, it is run as if the player
+that the timer runs on has clicked the NPC.
+
+valid <type> for gettimer() are:
+
+(0) TIMER_COUNT
+	Will return the total number of timers for the specified or
+	attached player. Can be filtered by <event>.
+(1) TIMER_FIRST_TICK
```
Emistry: how about named as `TIMER_NEXT_TICK` ? since it return the next timer tick and not the "first" thing. Described better as "next" and not "first" tho timer only trigger once for it.

doc/script_commands.txt
```diff
@@ -6481,15 +6481,28 @@ Size is 0 = normal 1 = small 2 = big.
...
-When the label is run, it is run as if the player that the timer runs on 
-has clicked the NPC.
+addtimer() to create, deltimer() to destroy, addtimercount() to delay
+it by the specified number of ticks, and gettimer() to get misc info.
+For all three cases, the event label given is the identifier of that timer.
+The timer runs on the character object that is attached to the script,
+and can have multiple instances. If <acccount id> is passed, this player
+will be used instead. When the label is run, it is run as if the player
+that the timer runs on has clicked the NPC.
+
```
Emistry: although I have no idea who is the first to write the documentation for the timer script commands, combining them together look suck, each of them perform different tasks.

Maybe consider to split the documentations for these 4 script commands?

doc/script_commands.txt
```diff
@@ -6481,15 +6481,28 @@ Size is 0 = normal 1 = small 2 = big.
...
+valid <type> for gettimer() are:
+
+(0) TIMER_COUNT
+	Will return the total number of timers for the specified or
+	attached player. Can be filtered by <event>.
+(1) TIMER_FIRST_TICK
+	Will return the number of ticks until the next timer runs
+	for the specified or attached player. Can be filtered by <event>.
+(2) TIMER_LAST_TICK
```
Emistry: how about named as `TIMER_REMAINING_TICK` ? since it return the remaining timer tick and not the "last" thing. Described better as "remaining" and not "last" tho timer only trigger once for it.

Emistry:
as an addition, if we could make the naming group together, it can help to provide additional functionality to third party tools. like autocomplete feature in notepad++.

we can named it something like these..
`TIMER_TICK_NEXT`
`TIMER_TICK_LAST`
with this we could easily enable the autocomplete to show both constants when we typed only `TIMER_TICK_`

however, if follow the current naming ... we cant make this work. we could only make it show all possible constant by typed `TIMER_` but who know that it might confused with other constant that begin with `TIMER_`

doc/script_commands.txt
mekolat: well it isn't the next timer in the queue but rather the one that will execute the first, regardless of order so given this:
```c++
addtimer(700, "event");
addtimer(200, "event");
addtimer(900, "event");
```
the queue would end up being `[700, 200, 900]` so the next in queue would be the `700` but it's not the first to be called, which would be `200`

doc/script_commands.txt
mekolat: no this is the amount of ticks remaining until the last event runs so given this:
```c++
addtimer(700, "event");
addtimer(900, "event");
addtimer(200, "event");
```
the amount of ticks until `"event"` gets called for the last time would be `900` (while `REMAINING_TICK` would imply the amount of ticks until it is called next, which would be `200`)

- - - 
in short, `FIRST_TICK` would be `200` and `LAST_TICK` would be `900`, which correspond to the first and last times the event is invoked

mekolat:
@Emistry I feel sorry for you who is stuck with n++ while there's [Atom](https://atom.io/) and Sublime out there... but sure, I can use the `TIMER_TICK` prefix if it makes you happier

doc/script_commands.txt
mekolat: sure, I felt it was weird to see them all combined too, but went with it and just added the new one... splitting makes lots of sense

Emistry:
Please don't make it sound like personal ... using any kind of editor would have just give the same results.
Atom or Sublime wouldn't help you categorize the autocomplete list if your constant naming aren't consistent in a way.
I am not ordering your change it to the way I wanted it to be, I am just suggesting a way that could benefits others.

doc/script_commands.txt
Emistry: the doc sound confuse to people.
if its executing the next first timer in the queue, perhaps should mention it in the doc too then.
something like `return the next tick for the next timer timer in the current timer queue` ... something similar.

mekolat:
@Emistry I'm not taking it personal, just kidding you 😄 I will change the constant names don't worry

mekolat:
@Emistry @MishimaHaruna is that better?

src/map/script.c
```diff
@@ -10861,6 +10861,94 @@ BUILDIN(addtimercount)
...
+		script->check_event(st, event);
+	}
+
+	if (sd == NULL) {
+		script_pushint(st, -1);
+		return true;
+	}
+
+	switch (mode) {
+	case 0:
```
Emistry: how about making it into an `enum`?  with this, we can directly assign the enum value as constant in script too.

src/map/script.c
```diff
@@ -10861,6 +10861,94 @@ BUILDIN(addtimercount)
...
+					if (strcmp((char *)(td->data), event) == 0) {
+						val++;
+					}
+				} else {
+					val++;
+				}
+			}
+		}
+		break;
+	case 1:
```
Emistry: how about making it into an `enum`?  with this, we can directly assign the enum value as constant in script too.

src/map/script.c
```diff
@@ -10861,6 +10861,94 @@ BUILDIN(addtimercount)
...
+						first = false;
+					}
+				} else if (first == true || tick < val) {
+					val = tick;
+					first = false;
+				}
+			}
+		}
+		break;
+	case 2:
```
Emistry: how about making it into an `enum`?  with this, we can directly assign the enum value as constant in script too.

src/map/script.c
mekolat: pointless enum; it would only ever be used in that function (`buildin_gettimer`) and nowhere else so it's not worth it, and adds more code for nothing really

MishimaHaruna:
I believe this is in a reasonable state to be merged. Any objections, @Emistry?

As for the enums, I'd certainly like to have a way to avoid duplication (and risk of things getting outdated when a possible value is added or removed). If we don't want to add it now to avoid cluttering script.h with yet another enum, we can add it at a later point, along with many others (when we introduce a private script.p.h header)

Emistry:
nop, dont have any objections.

MishimaHaruna:
Yes, I agree with you about the enums (and jot just for readability, they're also handy in debug). If @mekolat doesn't mind adding it, it would be nice, else I'll still merge this (and handle the hardcoded constants issue at a later time, when I work on the private interfaces, as I mentioned)

mekolat:
ok so that's done, feel free to manually edit yourself if you need further changes

MishimaHaruna:
Thank you! Merging this now

PR #1637, opened 2017-03-15
Fix wrong order of dialogue at Crusader Jobchange Quest.

skyleo:
### Current Behavior
If a player has Patriotism_Marks and Sacred_Marks before starting the quest, he'll skip the introductionary dialogue of the Crusader automatically.
(This actually irritated a player of mine heavily, making him believe the quest is bugged.)

### Expected Behavior
He does his introduction and when you're willing to become Crusader, you'll be able to use your 2 Items to skip the Item Gathering part.

## Changes Proposed
I moved one of the alternative first quest parts to the part where the player already has accepted the Item Gathering part. This is exactly at the location where it is in the original aegis script.
This also makes the 2nd very same code-part not obsolete anymore.

**Affected Branches: Master** 

**Issues addressed: None**

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1637)

### DIFF
--- a/npc/jobs/2-2/crusader.txt
+++ b/npc/jobs/2-2/crusader.txt
@@ -210,6 +210,7 @@ prt_castle,45,169,5	script	Senior Crusader	4_M_CRU_OLD,{
 		mes "Have you proven your determination with the task I have given you, or do you possess the items proving that you have received your calling?";
 		next;
 		if (countitem(Patriotism_Marks) >= 1 && countitem(Sacred_Marks) >= 1) {
+			mes "[Michael Halig]";
 			mes "Ah...";
 			mes "I see that you have been called to become a Crusader. We are assured of your will, but now we must test your capabilities.";
 			next;
--- a/npc/jobs/2-2/crusader.txt
+++ b/npc/jobs/2-2/crusader.txt
@@ -78,20 +78,6 @@ prt_castle,45,169,5	script	Senior Crusader	4_M_CRU_OLD,{
 		mes "As it happened one thousand years ago, evil forces will one day attack in droves in an attempt to take over the world once again.";
 		close;
 	}
-	else if(CRUS_Q <= 3 && countitem(Patriotism_Marks) && countitem(Sacred_Marks)) {
-		mes "Ah...";
-		mes "I see that you have been called to become a Crusader. We are assured of your will, but now we must test your capabilities.";
-		next;
-		mes "[Michael Halig]";
-		mes "Meet with Moorenak Miyol who is training in the underground dungeon of the Prontera Castle. Go, and speak with him first.";
-		next;
-		delitem 1004, 1;
-		delitem 1009, 1;
-		CRUS_Q = 4;
-		mes "[Michael Halig]";
-		mes "Moorenak and others like him will test the limits of your capabilities and help you find your path. Return to me after you have completed their tests...";
-		close;
-	}
 	else if(CRUS_Q == 0) {
 		mes "We are Crusaders, warriors preparing for the Holy War.";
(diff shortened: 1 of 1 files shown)


iRagno:
I have a question. In actual version, when talking to the npc and the character has the Patriotism_Marks and Sacred_Marks, when clicking on the npc he would read this:
```
 	mes "[Michael Halig]";
	mes "Ah...";
	mes "I see that you have been called to become a Crusader. We are assured of your will, but now we must test your capabilities.";
	next;
```
Now, in same case he will read this:
```
 	mes "[Michael Halig]";
	mes "Have you proven your determination with the task I have given you, or do you possess the items proving that you have received your calling?";
	next;
	mes "Ah...";
	mes "I see that you have been called to become a Crusader. We are assured of your will, but now we must test your capabilities.";
	next;
```
Notice the missing of the [Michael Halig] header in second message. Perhaps check for those items may be before of the first message. What do you think?

npc/jobs/2-2/crusader.txt
```diff
@@ -223,6 +209,27 @@ prt_castle,45,169,5	script	Senior Crusader	4_M_CRU_OLD,{
...
+			next;
+			delitem Patriotism_Marks, 1;
+			delitem Sacred_Marks, 1;
+			CRUS_Q = 4;
+			if (questprogress(3006)) {
+				changequest 3006,3009;
+			} else if (questprogress(3007)) {
+				changequest 3007,3009;
+			} else {
+				changequest 3008,3009;
```
Ridley8819: I am no fan of changequest in such cases, simply because the quest progress is supposed to be 2 (completed) afterwards, but with changequest it got resetted (deleted) and a future check wouldn't be possible. 
With proper usage of questprogress, comoletequest and erasequest we could avoid so many additional variables.

skyleo: I'm just following the behavior the script is already using at line 259 (Master).
I could change that if actually needed, although I didn't quite get your reasoning for now.
While it seems to be possible to shorten down lines by using erasequest(3005, 3010) or completequest(3005,3010) (I hope I read the documentations correctly, it says *between*, not if it's excluding or including.).

Aegis script uses changequest at this part btw. (Implying our changequest is an exact imitation of theirs)

Ridley8819: That's now what I mean, my point is in general about the usage of changequest rather than completequest. 
Changequest erases the questprogress back to 0 while completequest removes it from quest window and sets it to 2.
So with changequest you have no future chance to check if quest was done or not. This is not required here but in other usages it is. Changequest actually makes only sense when using for some kind of cooldown quest

skyleo:
@iRagno I chose to position it that way because it's that way in the Aegis Script I compared it with.
And it does make sense to me dialogue-wise.
Thanks for noting out the missing mes "[Michael Halig]";

npc/jobs/2-2/crusader.txt
skyleo: I see what you mean now. I find it indeed weird, why all scripts use this "removal behavior" instead of preserving questlog, which is kind of the point of a log.

But I'd suggest that we won't change this for this PR, as it'd require me to change it everywhere in that script. I'd like to separate that in another PR if I were to do that, since I am not quite sure about the stance of Hercules staff on this.

(I was actually quite close of standardizing the code-style and syntax of the whole script.
But decided against it, as it'd slow down the PR process for the actual fix I want.)

Mysteries: I sorta agree with skyleo. If we're going to change it, we change a bunch of scripts in another PR rather than in this one.

PR #1652, opened 2017-03-26
Fixes some incorrect packet ver dates.

Jedzkie:
### Pull Request Prelude

### Changes Proposed

Upon testing 2015-09-23fRagexe, the chat messages are also broken, but in 2015-09-16aRagexe, it does not.

Also in 2015-10-21aRagexe, this client isn't yet supports new character creation.

**Affected Branches:** master

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1652)

### DIFF
--- a/src/char/char.c
+++ b/src/char/char.c
@@ -4633,7 +4633,10 @@ void char_parse_char_create_new_char(int fd, struct char_session_data* sd)
 		//turn character creation on/off [Kevin]
 		result = -2;
 	} else {
+#ifdef ENABLE_PACKETVER_RE
 #if PACKETVER >= 20151001
+#endif
+#elif PACKETVER >= 20151029
 		uint8 sex = RFIFOB(fd, 35);
 
 		switch (sex) {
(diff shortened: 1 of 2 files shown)


Jedzkie:
@MishimaHaruna @Smokexyz

4144:
@Jedzkie while you updating versions for chat end null pointer, can you also check for your clients input script command? I think it related to this versions. Where client add or not add null char at end of string.
Example:
```
input @var$;
mes @var$;
```

Jedzkie:
@4144 i tried changing the client check in `clif_parse_NpcStringInput` from 20151029 to 20151021 and got this result.

input .@test;
mesf("[%d]", .@test);

result is [01], etc. depends on what number you've input, it always adds 0 before the input number.

So the current packet ver check for `clif_parse_NpcStringInput` is correct.

4144:
@Jedzkie it works only for strings. Ints always works.
In short old clients add always \x0 char at end.
And new clients not adding \x0 char at end.

Jedzkie:
let me try again.

4144:
See check in clif_parse_NpcStringInput

Jedzkie:
I tested on 2 clients,

2015-10-21aRagexe = no null char at end
2015-11-04aRagexe = no null char at end

I use this script

`input @var$;
mes @var$;`

PR #1657, opened 2017-03-29
add getunits() buildin

mekolat:

### Pull Request Prelude

### Changes Proposed
See the comments in #1610. This PR adds a new buildin:
```
getunits(<type>, <var>[<index>], <limit>, "<map>"{, <x1>, <y1>, <x2>, <y2>})
```

### Example code
```c++
    npctalk("I am the most beautiful being the Universe ever begat!");
    sleep(500);
    npctalk("...and you shall all love me!");

    .@count = getunits(UNITTYPE_PC, .@gid[0], 5, "map", 45, 67, 109, 111);

    for (.@i = 0; .@i < .@count; ++.@i)
    {
        emotion(e_heart, true, strcharinfo(0, .@gid[.@i]));
    }
```

**Issues addressed:** (none)

### TO-DO:
- [X] ~~~`BL_` constants for bl types~~~
   - used the already existing `UNITTYPE_` constants instead
- [X] documentation

<br><br>

---
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1657)

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -3229,6 +3229,40 @@ Notice that NPC objects disabled with disablenpc() will still be located.
 
 ---------------------------------------
 
+*getunits(<type>, <variable>, <limit>, "<map>"{, <x1>, <y1>, <x2>, <y2>})
+
+This function searches a whole map or area for units and adds their GID to
+the provided <variable> array. It filters units by <type> and stops searching
+after <limit> units have been found. Set <limit> to false (0) if you wish to
+disable the limit altogether.
+
+Type is the type of unit to search for:
+
+	BL_PC    - Character object
+	BL_MOB   - Monster object
+	BL_PET   - Pet object
+	BL_HOM   - Homunculus object
+	BL_MER   - Mercenary object
+	BL_IEM   - Item object (item drops)
+	BL_SKILL - Skill object (skill fx & sfx)
+	BL_NPC   - NPC object
+	BL_CHAT  - Chat object
+	BL_ELEM  - Elemental object
+	BL_CHAR  - Shorthand for (BL_PC|BL_MOB|BL_HOM|BL_MER|BL_ELEM)
+	BL_ALL   - Any kind of object
+
+** Do NOT use UNITTYPE_ constants here, they have different values.
+
+Example:
+
+	.@count = getunits((BL_PC | BL_NPC), .@units, false, "prontera");
+
+The above example would search the map "prontera" for all PC and NPC units and
+add them to the .@units array, while setting .@count to the amount of units
+added to the array (useful in for() loops).
+
+---------------------------------------
+
 *getgmlevel()
 
 This function will return the (GM) level of player group the account to 
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -10955,6 +10955,91 @@ BUILDIN(gettimer)
 	return true;
 }
 
+int buildin_getunits_sub(struct block_list *bl, va_list ap)
+{
+	struct script_state *st = va_arg(ap, struct script_state *);
+	struct map_session_data *sd = va_arg(ap, struct map_session_data *);
+	int32 id = va_arg(ap, int32);
+	uint32 start = va_arg(ap, uint32);
+	uint32 *count = va_arg(ap, uint32 *);
+	uint32 limit = va_arg(ap, uint32);
+	const char *name = va_arg(ap, const char *);
+	struct reg_db *ref = va_arg(ap, struct reg_db *);
+	uint32 index = start + *count;
+
+	if (index >= SCRIPT_MAX_ARRAYSIZE || *count > limit) {
+		return 1;
+	}
+
+	script->set_reg(st, sd, reference_uid(id, index), name,
+		(const void *)h64BPTRSIZE(bl->id), ref);
+
+	(*count)++;
+	return 0;
+}
+
+BUILDIN(getunits)
+{
+	const char *mapname, *name;
+	int16 m, x1, y1, x2, y2;
+	int32 id;
+	uint32 start;
+	struct reg_db *ref;
+	enum bl_type type = script_getnum(st, 2);
+	struct script_data *data = script_getdata(st, 3);
+	uint32 count = 0, limit = script_getnum(st, 4);
+	struct map_session_data *sd = NULL;
+
+	if (!data_isreference(data) || reference_toconstant(data)) {
+		ShowError("script:getunits: second argument must be a variable\n");
+		script->reportdata(data);
+		st->state = END;
+		return false;
+	}
+
+	id = reference_getid(data);
+	start = reference_getindex(data);
+	name = reference_getname(data);
+	ref = reference_getref(data);
+
+	if (not_server_variable(*name)) {
+		sd = script->rid2sd(st);
+		if (sd == NULL) {
+			return true; // player variable but no player attached
+		}
+	}
+
+	if (is_string_variable(name)) {
+		ShowError("script:getunits: second argument must be an integer variable\n");
+		script->reportdata(data);
+		st->state = END;
+		return false;
+	}
+
+	if (limit < 1 || limit > SCRIPT_MAX_ARRAYSIZE) {
+		limit = SCRIPT_MAX_ARRAYSIZE;
+	}
+
+	mapname = script_getstr(st, 5);
+	m = map->mapname2mapid(mapname);
+
+	if (script_hasdata(st, 9)) {
+		x1 = script_getnum(st, 6);
+		y1 = script_getnum(st, 7);
+		x2 = script_getnum(st, 8);
+		y2 = script_getnum(st, 9);
+
+		map->foreachinarea(buildin_getunits_sub, m, x1, y1, x2, y2, type,
+			st, sd, id, start, &count, limit, name, ref);
+	} else {
+		map->foreachinmap(buildin_getunits_sub, m, type,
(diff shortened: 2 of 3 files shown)


mekolat:
@Emistry

src/map/script.c
```diff
@@ -10955,6 +10955,91 @@ BUILDIN(gettimer)
...
+
+	if (not_server_variable(*name)) {
+		sd = script->rid2sd(st);
+		if (sd == NULL) {
+			return true; // player variable but no player attached
+		}
+	}
+
+	if (is_string_variable(name)) {
+		ShowError("script:getunits: must be an integer variable\n");
```
Emistry: maybe pinpoint the exact parameter here? like
```
ShowError("script:getunits: array must be an integer variable\n");
```
I find the message vague when it only tell "script command must be integer variable". Its not helpful enough to tackle the issue.

src/map/script.c
```diff
@@ -10955,6 +10955,91 @@ BUILDIN(gettimer)
...
+	int32 id;
+	uint32 start;
+	short area = script_getnum(st, 2);
+	short bl_type = script_getnum(st, 3);
+	int count = 0, limit = script_getnum(st, 4);
+	struct script_data *data = script_getdata(st, 5);
+	struct map_session_data *sd = NULL;
+
+	if (!data_isreference(data) || reference_toconstant(data)) {
+		ShowError("script:getunits: not a variable\n");
```
Emistry: maybe pinpoint the exact parameter here? like
```
ShowError("script:getunits: parameter must be an integer variable\n");
```
I find the message vague when it only tell "script command must be integer variable". Its not helpful enough to tackle the issue.

src/map/script.c
```diff
@@ -10955,6 +10955,91 @@ BUILDIN(gettimer)
...
+	}
+
+	if (is_string_variable(name)) {
+		ShowError("script:getunits: must be an integer variable\n");
+		script->reportdata(data);
+		st->state = END;
+		return false;
+	}
+
+	switch (area) {
```
Emistry: missing  `case default:` ?

mekolat: nope that's on purpose: do nothing with the array when the `<area>` is not (yet) supported, or if no unit is found

src/map/script.c
mekolat: well then, have a look at `buildin_setarray`

mekolat:
@Emistry 
> I think you should consider try at least include a documentation every time you made a new PR.

I find it super annoying to have to re-write the documentation every time the parameters change. Changing the parameters just in the code is easy for me but when modifying the documentation I have to write in English (which is not my native language) and formulate sentences that make sense and are easy to read... takes much more time than just waiting till the end and then writing the documentation.  

> what are the `BL_` that are supported?

Any kind of unit (PC, NPC, Mob, Homunculus, Mercenary, Pet, etc). It's basically meant to return the GID of virtually any unit, so it can then be manipulated by unit commands (which will soon be extended in #1584).

> doesn't really have to depend on the area type.

Yes it does. What if we want to add support for other "send targets" like `BG` at some point? What if we want to deprecate `getpartymember()` and add support for `PARTY`?

> does it included both PC and Mobs GID if there were a `BL_` type for all?

Yes, `<type>` is a bit flag, which makes this command very powerful. You could make combinations like `BL_NPC | BL_MOB` if you wish so.

> doesn't really have to depend on the area type.

It does if we want to expand the command in the future, else we end up with a `getunits2()` and such. What if we want to support additional "send targets" in the future, like `BG`? What if at some point we want to deprecate `getpartymember()` and `getguildmember()` and add support for `GUILD` and `PARTY`? If you want `getunits` to be confined to the map/area forever then sure I can remove the `<area>` parameter

> it seem differences between current `getunits()` script command with the `getareausers()` are just returning the GID of the units regardless of player/mob, and allow to save the GID into an array specified.

Yes, exactly, a more flexible version of it, with no hardcoded stuff

Emistry:
the `AREA` parameter are not really needed.
you can determine it check for `area` when you provided `mapname` and provided `4 additional coordinates` at the end.
you can determine it check for `map` when you provided only `mapname`.
you can determine it check for `ALL` when you **didnt** provided any parameters (in case you want it that way).

get units within an area
```getunits(<type>, <array>, <map>, {<x1>,<y1>,<x2>,<y2>});```
get units within a map
```getunits(<type>, <array>, <map>);```
get units global
```getunits(<type>, <array>);```

**Comment by mekolat** [comment]
seems you didn't read all that I wrote:
> What if we want to support additional "send targets" in the future, like `BG`? What if at some point we want to deprecate `getpartymember()` and `getguildmember()` and add support for `GUILD` and `PARTY`?

those are all `<area>`s and if you remove that parameter there's no way to figure out what are the arguments that you pass: is it a map? is it a guild id? is it a party id? is it a battleground id?



Also why would you remove `<limit>` ? This parameter is important.

> why need to limit the size? aren't Herc already support array up to 2 Billions index?

>> `<_41444d494e_>` if i not wrong now getunits() get all in area and put into array
`<_41444d494e_>` but it can be slow if here 1k or 10k units
(...)
`<meko>` yeah let's say you just want to execute it on 10 nearby players, it would pointless to fill the whole array if you just want 10
`<meko>` I could make so that if you put 0 as limit then there's no cap besides MAX_ARRAY_SIZE
`<smoke_>` makes sense
`<smoke_>` you could alternatively make the limit an optional entry
`<meko>` nah I want to make that explicit
`<meko>` if you don't want a limit pass false (0), else pass a number
`<smoke_>` Yeah it's not a problem, but I don't see the point of limiting in that case, when you could just reduce the area size.
`<smoke_>` But since 0 is unlimited, it's okay I guess.
`<meko>` you can reduce the area size indeed but while extremely unlikely it's not impossible to have 1k players standing on one single tile
`<smoke_>` true \:P

**Comment by mekolat** [comment]
@Emistry I applied your requested changes, here's a [shim](https://en.wikipedia.org/wiki/Shim_(computing)) of `areatimer()` as an example:

```c++
// areatimer("<map>", <x1>, <y1>, <x2>, <y2>, <tick>, "<npc::event>")
function	script	areatimer	{
	.@count = getunits(UNITTYPE_PC, .@units[0], false, getarg(0), getarg(1), getarg(2), getarg(3), getarg(4));

		deltimer(getarg(6), .@units[.@i]);
		addtimer(getarg(5), getarg(6), .@units[.@i]);
	}

	return .@count;
}
```

Emistry:
Just curious, 
```
getunits(UNITTYPE_PC, .@units[0], false, getarg(0), getarg(1), getarg(2), getarg(3), getarg(4));
```
I noticed the way you write it as `.@units[0]`, nothing wrong but does it also meant the script command support to append the new values into existing array? lets say `.@units[100]` where all GID will be appended/replaced in the array from index 100 and onward?

mekolat:
@Emistry yup, it does support this, like every single command that take a variable/array. Just [grep](https://linux.die.net/man/1/grep) or [ack](https://linux.die.net/man/1/ack) for `reference_getindex` or `reference_uid` if curious

src/map/script.c
```diff
@@ -10955,6 +10955,89 @@ BUILDIN(gettimer)
...
+	m = map->mapname2mapid(mapname);
+
+	if (script_hasdata(st, 9)) {
+		x1 = script_getnum(st, 6);
+		y1 = script_getnum(st, 7);
+		x2 = script_getnum(st, 8);
+		y2 = script_getnum(st, 9);
+
+		map->foreachinarea(buildin_getunits_sub, m, x1, y1, x2, y2, type,
+			st, sd, id, start, &count, limit, name, reference_getref(data));
```
MishimaHaruna: In order to avoid hiding very subtle errors if the return type of `reference_getref()` ever changes, can you assign it to a variable before passing it to the variadic function `map->foreachinarea()`?

```
struct reg_db *ref = referene_getref(data);
/* ... */
map->foreachinarea(buildin_getunits_sub, m, x1, y1, x2, y2, type,
	st, sd, id, start, &count, limit, name, ref);
```

doc/script_commands.txt
```diff
@@ -3229,6 +3229,33 @@ Notice that NPC objects disabled with disablenpc() will still be located.
...
+	UNITTYPE_NPC  - NPC object
+	UNITTYPE_PET  - Pet object
+	UNITTYPE_MOB  - Monster object
+	UNITTYPE_HOM  - Homunculus object
+	UNITTYPE_MER  - Mercenary object
+	UNITTYPE_ELEM - Elemental object
+
+Example:
+
+	.@count = getunits((UNITTYPE_PC | UNITTYPE_NPC), .@units, false, "prontera");
```
mekolat: oh, I see, gonna expose the `BL_` constants then, like I originally intended (as hardcoded constants in `script.c`)

src/map/script.c
mekolat: sure

src/map/script.c
```diff
@@ -10955,6 +10955,89 @@ BUILDIN(gettimer)
...
+	return 0;
+}
+
+BUILDIN(getunits)
+{
+	const char *mapname, *name;
+	int16 m, x1, y1, x2, y2;
+	int32 id;
+	uint32 start;
+	enum bl_type type = script_getnum(st, 2);
```
mekolat: yeah at first I was using the `BL_` constants but saw we already had `UNITTYPE_` constants, so I naively assumed it had the same values and didn't check

MishimaHaruna:
Looks good now, I'm merging this. Thank you very much!

PR #1666, opened 2017-04-08
Implementation of the official Achievement System.

Smokexyz:
### Pull Request Prelude

### Changes Proposed
Source: http://ro.gnjoy.com/news/update/View.asp?seq=163&curpage=1

Script Commands -
achievement_progress(<`ach_id`>,<`obj_idx`>,<`progress`>,<`incremental`>{,<`char_id`>});
`aid`  - achievement ID
`obj_idx` - achievement objective index. (Client side equivalent)
`progress` - objective progress towards goal.
`incremental` - (boolean) true to add the progress towards the goal,
            	false to use the progress only as a comparand.
`char_id` - (optional) Character ID to perform on.
@return 1 on success, 0 on failure.

Achievement Types -
  [Type]                             [Description] (Trigger)
`ACH_QUEST`                          Specific achievement objective update (Script)
`ACH_KILL_PC_TOTAL`                  (Accumulative) Total kill count. (Player kill)
`ACH_KILL_PC_JOB`                    Kill a player of the specified job. (Player Kill)
`ACH_KILL_PC_JOBTYPE`                Kill a player of the specified job type. (Player Kill)
`ACH_KILL_MOB_CLASS`                 Kill a particular mob class. (Mob Kill)
`ACH_DAMAGE_PC_MAX`                  Maximum damage caused on a player. (Player Damage)
`ACH_DAMAGE_PC_TOTAL`                (Accumulative) Damage on players. (Player Damage)
`ACH_DAMAGE_PC_REC_MAX`              Maximum damage received by a player. (Receive Player Damage)
`ACH_DAMAGE_PC_REC_TOTAL`            (Accumulative) Damage received by players. (Receive Player Damage)
`ACH_DAMAGE_MOB_MAX`                 Maximum damage caused on a monster. (Monster Damage)
`ACH_DAMAGE_MOB_TOTAL`               (Accumulative) Damage caused on monsters. (Monster Damage)
`ACH_DAMAGE_MOB_REC_MAX`             Maximum damage received by a monster. (Receive Monster Damage)
`ACH_DAMAGE_MOB_REC_TOTAL`           (Accumulative) Damage received by monsters. (Receive Monster Damage)
`ACH_JOB_CHANGE`                     Change to a specific job. (Job Change)
`ACH_STATUS`                         Acquire a specific amount of a particular status type. (Stat Change)
`ACH_STATUS_BY_JOB`                  Acquire a specific amount of a status type as a job class. (Stat Change)
`ACH_STATUS_BY_JOBTYPE`              Acquire a specific amount of a status type as a job type. (Stat Change)
`ACH_CHATROOM_CREATE_DEAD`           (Accumulative) Create a chatroom when dead. (Chatroom Creation)
`ACH_CHATROOM_CREATE`                (Accumulative) Create a chatroom. (Chatroom Creation)
`ACH_CHATROOM_MEMBERS`               Gather 'n' members in a chatroom. (Chatroom Join)
`ACH_FRIEND_ADD`                     Add a specific number of friends. (Friend Addition)
`ACH_PARTY_CREATE`                   (Accumulative) Create a specific number of parties. (Party Creation)
`ACH_PARTY_JOIN`                     (Accumulative) Join a specific number of parties. (Party Join)
`ACH_MARRY`                          (Accumulative) Marry a specified number of times. (Marriage)
`ACH_ADOPT_BABY`                     (Accumulative) Get Adopted. (Adoption)
`ACH_ADOPT_PARENT`                   (Accumulative) Adopt a Baby. (Adoption)
`ACH_ZENY_HOLD`                      Hold a specific amount of zeny in your inventory. (Gain Zeny)
`ACH_ZENY_GET_ONCE`                  Gain a specific amount of zeny in one transaction. (Gain Zeny)
`ACH_ZENY_GET_TOTAL`                 (Accumulative) Gain a specific amount of zeny in total. (Gain Zeny)
`ACH_ZENY_SPEND_ONCE`                Spend a specific amount of zeny in one transaction. (Pay Zeny)
`ACH_ZENY_SPEND_TOTAL`               (Accumulative) Spend a specific amount of zeny in total. (Pay Zeny)
`ACH_EQUIP_REFINE_SUCCESS`           Refine an item to +N. (Successful Refine)
`ACH_EQUIP_REFINE_FAILURE`           Fail to refine an item of +N refine. (Failed Refine)
`ACH_EQUIP_REFINE_SUCCESS_TOTAL`     (Accumulative) Refine an item successfully N times. (Success Refine)
`ACH_EQUIP_REFINE_FAILURE_TOTAL`     (Accumulative) Fail to refine an item N times. (Failed Refine)
`ACH_EQUIP_REFINE_SUCCESS_WLV`       Refine a Weapon of a particular Level to +N. (Success Refine)
`ACH_EQUIP_REFINE_FAILURE_WLV`       Fail to refine a Weapon of a particular level from +N. (Failed Refine)
`ACH_EQUIP_REFINE_SUCCESS_ID`        Refine a particular Item successfully to +N. (Success Refine)
`ACH_EQUIP_REFINE_FAILURE_ID`        Fail to refine a particular item successfully from +N. (Failed Refine)
`ACH_ITEM_GET_COUNT`                 Acquire N amount of an item of a particular ID. (Acquire Item)
`ACH_ITEM_GET_COUNT_ITEMTYPE`        Acquire N amount of items of a particular type mask. (Acquire Item)
`ACH_ITEM_GET_WORTH`                 Acquire an item of buy value N. (Acquire Item)
`ACH_ITEM_SELL_WORTH`                Sell an item of sell value N. (NPC Sell Item)
`ACH_PET_CREATE`                     Successfully tame a pet of a particular mob class. (Successful Pet Tame)
`ACH_ACHIEVE`                        Achieve an Achievement. (Achievement Completion)
`ACH_ACHIEVEMENT_RANK`               Achieve an Achievement Rank. (Achievement Rank Increase)

Includes an achievement_db.conf generator that reads from the item_db, mob_db (server side) and achievement_list.lub files to determine valid achievement entries based on item/monster availability. Achievements containing unsupported entries are commented out. Missing monsters, items and disabled achievements are listed at the bottom of the conf file.

This feature, although renewal-only in official servers, is capable of being used in pre-renewal mode on Hercules.

Does not include the title system yet.

**Affected Branches:** Master

- [x] Include new files in vcproj-11 and  vcproj-12
- [x] Fix achievement rank type achievement validation.
- [ ] Title System (Possible inclusion)
- [ ] NPCs and Scripts for Script/NPC based achievements. Achievement types in particular - `SEE`, `HEAR`, `EAT`, `ADVENTURE` and `CHATTING`. Thank you kRO for being confused af.

Achievement completion black bar announcement does not work properly for clients before `2016-11-*`.
Achievement window overview totals display `68` for clients after `2016-03-16`.

A big thanks to -
@MishimaHaruna for reviewing to perfection.
@4144 for all the support.
@Asheraf for a lot of official information.


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1666)

src/map/achievement.c
```diff
@@ -0,0 +1,1399 @@
...
+ * @param dstsd    pointer to target player's session data.
+ * @param damage   amount of damage dealt / received.
+ */
+void achievement_validate_pc_damage(struct map_session_data *sd, struct map_session_data *dstsd, unsigned int damage)
+{
+	struct achievement_objective criteria = { 0 };
+
+	nullpo_retv(sd);
+	nullpo_retv(dstsd);
+
```
Smokexyz: I don't understand what you mean here.

Smokexyz:
Waiting for the reviews to pile-up before making any changes on this, so I tackle them all at once.

src/map/achievement.h
```diff
@@ -0,0 +1,287 @@
...
+	int id;
+	char name[ACHIEVEMENT_NAME_LENGTH];
+	enum achievement_types type;
+	int points;
+	VECTOR_DECL(struct achievement_objective) objective;
+	struct achievement_rewards rewards;
+};
+
+// Achievements that require filtering by criteria
+#define achievement_has_criteria(t) ( \
```
Smokexyz: plugins cause so much extra work 😠

src/char/int_achievement.c
```diff
@@ -0,0 +1,314 @@
...
+	/* Ensure data exists */
+	cp = idb_ensure(inter_achievement->char_achievements, char_id, inter_achievement->ensure_char_achievements);
+
+	/* Load storage for char-server. */
+	inter_achievement->fromsql(char_id, cp);
+
+	/* Determine payload size */
+	data_size = sizeof(struct achievement) * VECTOR_LENGTH(*cp);
+
+	/* Send to the map server. */
```
Smokexyz: Done.

src/map/achievement.c
```diff
@@ -0,0 +1,1399 @@
...
+		idb_put(achievement->db, p_ad->id, p_ad);
+
+		/* Put achievement key in categories. */
+		VECTOR_ENSURE(achievement_category[p_ad->type], 1, 1);
+		VECTOR_PUSH(achievement_category[p_ad->type], p_ad->id);
+
+		/* Qualify for duplicate Id checks */
+		VECTOR_ENSURE(duplicate, 1, 1);
+		VECTOR_PUSH(duplicate, t_ad.id);
+
```
Smokexyz: Done

src/map/clif.c
```diff
@@ -19096,6 +19099,216 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+			p.ach[count].completed_at = (uint32) a->completed_at;
+			p.ach[count].reward = a->rewarded_at ? 1 : 0;
+			p.total_points += ad->points;
+		}
+
+		/* Acknowledgement of it's existence. */
+		count++;
+	}
+
+	/* Packet ID */
```
Smokexyz: Removed some of the pointless ones.

src/map/clif.c
```diff
@@ -19096,6 +19099,216 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+ * Send all of a session's achievement information to client.
+ * Called only once on login/char-loading. (PACKET_ZC_ALL_ACH_LIST)
+ * @packet [out] 0x0A23 <ID>.W <Length>.W <ach_count>.L <total_points>.L <rank>.W <current_rank_points>.L <next_rank_points>.L <struct ach_list_info *[]>.P
+ * @param fd   socket descriptor
+ * @param sd   pointer to map_session_data
+ */
+void clif_achievement_send_list(int fd, struct map_session_data *sd)
+{
+#if PACKETVER >= 20141016
+	int i = 0, count = 0, curr_exp_tmp = 0;
```
Smokexyz: Added.

src/map/achievement.c
```diff
@@ -0,0 +1,1399 @@
...
+			ShowError("achievement_readdb: Type field not provided for Achievement %d! Skipping...\n", t_ad.id);
+			continue;
+		} else if (!script->get_constant(string, &val)) {
+			ShowError("achievement_readdb: Invalid constant %s provided as type for Achievement %d! Skipping...\n", string, t_ad.id);
+			continue;
+		}
+
+		t_ad.type = (enum achievement_types) val;
+
+		/* Objectives */
```
Smokexyz: Done.

src/map/achievement.c
```diff
@@ -0,0 +1,1399 @@
...
+* Checks if the given criteria satisfies the achievement's objective.
+* @param objective pointer to the achievement's objectives data.
+* @param criteria  pointer to the current session's criteria as a comparand.
+* @return true if all criteria are satisfied, else false.
+*/
+bool achievement_check_criteria(const struct achievement_objective *objective, const struct achievement_objective *criteria)
+{
+	bool equal = true;
+	int i = 0, j = 0;
+
```
Smokexyz: Added.

src/char/int_achievement.c
```diff
@@ -0,0 +1,314 @@
...
+ * @param  char_id     character identifier.
+ * @param  cp          pointer to loaded achievements.
+ * @param  p           pointer to map-sent character achievements.
+ * @return number of achievements saved.
+ */
+int inter_achievement_tosql(int char_id, struct char_achievements *cp, struct char_achievements *p)
+{
+	StringBuf buf;
+	int i = 0, rows = 0;
+
```
Smokexyz: Added.

src/map/achievement.c
```diff
@@ -0,0 +1,1693 @@
...
+	nullpo_retr(false, objective);
+	nullpo_retr(false, criteria);
+
+	/* Monster Id */
+	equal = (objective->mobid && objective->mobid == criteria->mobid);
```
MishimaHaruna: Each check overwrites the previous value of `equal`. You should simply return false after each check if it failed (and you don't even need to cache the `equal` value in that case)

src/map/achievement.c
```diff
@@ -0,0 +1,1693 @@
...
+}
+
+/**
+ * Validates any achievement's specific objective index.
+ * @param sd       pointer to the session data.
+ * @param aid      ID of the achievement.
+ * @param index    index of the objective.
+ * @param progress progress to be added towards the goal.
+ */
+bool achievement_validate(struct map_session_data *sd, const int aid, const unsigned int index, const int progress, bool additive)
```
MishimaHaruna: Do the ints here have to be const? Is this intended to limit the HPM's ability to overwrite them?

src/map/achievement.c
```diff
@@ -0,0 +1,1693 @@
...
+	struct achievement_objective criteria = { 0 };
+	struct item_data *id = itemdb->exists(sd->status.inventory[idx].nameid);
+
+	nullpo_retv(sd);
+	Assert_retv(idx < MAX_INVENTORY);
+	Assert_retv(id != NULL);
+
+	/* Universal */
+	achievement->validate_type(sd,
+							   success ? ACH_EQUIP_REFINE_SUCCESS : ACH_EQUIP_REFINE_FAILURE,
```
MishimaHaruna: Not sure what happened here. If you wanted to align this, please use spaces (else just indent the continuation lines with the standard two tabs, to make better use of the precious horizontal space)

src/map/achievement.c
```diff
@@ -0,0 +1,1693 @@
...
+	equal = (objective->unique.itemid && objective->unique.itemid == criteria->unique.itemid);
+
+	/* Weapon Level */
+	equal = (objective->unique.weapon_lv && objective->unique.weapon_lv == criteria->unique.weapon_lv);
+
+	/* Achievement Id */
+	equal = (objective->unique.achieve_id && objective->unique.achieve_id == criteria->unique.achieve_id);
+
+	/* Status Types */
+	equal = (objective->unique.status_type && objective->unique.status_type == criteria->unique.status_type);
```
MishimaHaruna: `unique` is an union. You need to know which field of the union has been used rather than probing them (reading the other fields will return garbage!)

src/map/intif.c
```diff
@@ -1495,11 +1497,94 @@ void intif_parse_DeleteHomunculusOk(int fd) {
...
+		struct achievement t_ach = { 0 };
+
+		memcpy(&t_ach, RFIFOP(fd, 8 + i * sizeof(struct achievement)), sizeof(struct achievement));
+
+		if (achievement->get(t_ach.id) == NULL) {
+			ShowError("intif_parse_achievements_load: Invalid Achievement %d received from character %d. Ignoring...\n", t_ach.id, char_id);
+			continue;
+		}
+
+		VECTOR_PUSH(sd->achievement, t_ach);
```
MishimaHaruna: What if this is called twice? (in case the char server sends it twice for any reason)

src/map/intif.c
```diff
@@ -1495,11 +1497,94 @@ void intif_parse_DeleteHomunculusOk(int fd) {
...
+		return;
+	/* Return if no data. */
+	if (!(payload_size = VECTOR_LENGTH(sd->achievement) * sizeof(struct achievement)))
+		return;
+
+	packet_len = payload_size + 8;
+
+	WFIFOHEAD(inter_fd, packet_len);
+	WFIFOW(inter_fd, 0) = 0x3099;
+	WFIFOL(inter_fd, 2) = packet_len;
```
Smokexyz: Woops. Reminder to change to WFIFOW.

src/map/achievement.c
```diff
@@ -0,0 +1,1693 @@
...
+	struct achievement *ach = NULL;
+
+	nullpo_retv(sd);
+	nullpo_retv(ad);
+
+	Assert_retv(progress > 0);
+	Assert_retv(obj_idx < VECTOR_LENGTH(ad->objective));
+
+	// find existing or create a new entry.
+	ach = achievement->find_or_create(sd, ad->id, true);
```
Smokexyz: Hmm, I don't think there is any reason for `ad->id` to be invalid though. In the very unlikely event, it is also checked on line 72. `ach` here too cannot be null as we're creating an index if none was found, and so a null pointer is not possible.

src/map/achievement.c
```diff
@@ -0,0 +1,1693 @@
...
+	struct achievement *ach = NULL;
+
+	nullpo_retv(sd);
+	nullpo_retv(ad);
+
+	Assert_retv(progress > 0);
+	Assert_retv(obj_idx < VECTOR_LENGTH(ad->objective));
+
+	if (progress >= VECTOR_INDEX(ad->objective, obj_idx).goal) {
+		if ((ach = achievement->find_or_create(sd, ad->id, true)) && ach->completed_at)
```
Smokexyz: I don't think I follow, for the same reasons stated above.

src/map/achievement.c
Smokexyz: In the case where one objective has more than one criteria for example - mobid and itemid, your suggestion would return false every time without checking the second criteria.

src/map/achievement.c
Smokexyz: Not really, I actually did not think of HPM. Simply because they weren't meant to be changed within the function but allowing so through plugins makes sense.

src/char/int_achievement.c
```diff
@@ -0,0 +1,341 @@
...
+	if (SQL_ERROR == SQL->QueryStr(inter->sql_handle, StrBuf->Value(&buf))) {
+		Sql_ShowDebug(inter->sql_handle);
+		StrBuf->Destroy(&buf);
+		return false;
+	}
+
+	if ((num_rows = (int) SQL->NumRows(inter->sql_handle)) != 0) {
+		int j = 0;
+
+		VECTOR_INIT(*cp);
```
Smokexyz: This is actually initialised in db ensuring function so I'll remove it from here as it is unnecessary.

src/map/achievement.c
Smokexyz: Thanks, done as suggested with two tabs.

src/map/intif.c
Smokexyz: I think adding a check to see if vector length is `> 0` would be just as good as adding a flag to check if the data is received.

src/map/achievement.c
Smokexyz: Fixed using an enumerator to keep track of the union type in use. Thank you.

src/char/int_achievement.c
```diff
@@ -0,0 +1,341 @@
...
+			struct achievement t_ach = { 0 };
+			SQL->GetData(inter->sql_handle, 0, &data, NULL); t_ach.id = atoi(data);
+			SQL->GetData(inter->sql_handle, 1, &data, NULL); t_ach.completed_at = atoi(data);
+			SQL->GetData(inter->sql_handle, 2, &data, NULL); t_ach.rewarded_at = atoi(data);
+			/* Objectives */
+			for (j = 0; j < MAX_ACHIEVEMENT_OBJECTIVES; j++) {
+				SQL->GetData(inter->sql_handle, j + 3, &data, NULL);
+				t_ach.objective[j] = atoi(data);
+			}
+			VECTOR_ENSURE(*cp, 1, 1);
```
Smokexyz: Done.

src/char/int_achievement.c
```diff
@@ -0,0 +1,341 @@
...
+ * @param char_id  character Id.
+ */
+void mapif_achievement_load(int fd, int char_id)
+{
+	int i = 0;
+	int data_size = 0;
+	struct char_achievements *cp = NULL;
+
+	/* Ensure data exists */
+	cp = idb_ensure(inter_achievement->char_achievements, char_id, inter_achievement->ensure_char_achievements);
```
Smokexyz: Added the removal of the db entry in `char_set_char_offline()` as suggested.

src/map/achievement.c
```diff
@@ -0,0 +1,1693 @@
...
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#define HERCULES_CORE
+
+#include "map/achievement.h"
+
+#include "map/script.h"
```
Smokexyz: Done, thanks, good catch.

src/map/achievement.c
```diff
@@ -0,0 +1,1693 @@
...
+			if (completed != NULL)
+				(*completed)++;
+		}
+	}
+
+	if (total_points != NULL && *total_points > 0
+		&& (rank != NULL || curr_rank_points != NULL)) {
+		if (curr_rank_points != NULL )
+			*curr_rank_points = *total_points;
+		for (i = 0; i < MAX_ACHIEVEMENT_RANKS && *curr_rank_points >= VECTOR_INDEX(achievement->rank_exp, i) && i < VECTOR_LENGTH(achievement->rank_exp); i++) {
```
Smokexyz: Re-coded the function to use temporary buffers and assign the values on non-null pointers  at the end of the function.

src/map/achievement.c
MishimaHaruna: But it is exactly line 72 that I'm speaking about. If the id is invalid, line 72 causes this function to return NULL

src/map/achievement.c
MishimaHaruna: Which is exactly what the function's documentation says (`@return true if all criteria are satisfied, else false.` == "returns false if at least one of all the criteria is not satisfied")

src/map/achievement.c
MishimaHaruna: But using const for a scalar is entirely pointless, and only limits the HPM. Const only makes sense (to the caller of the function) when it's a pointer, since the caller knows that the pointer it owns won't be changed by the called function. If the function wants to change its local copy of variables passed by value, why would the function signature want to care at all? The public signature mustn't care about internal implementations of the function.

Smokexyz: Well, from the perspective of the caller, why should it not want the same for values (as it does for the pointers) that are not meant to be changed.

src/map/achievement.c
Smokexyz: Right, I totally misread that as return true if one criteria was satisfied. Sorry.

src/map/achievement.c
Smokexyz: Right, so then the `nullpo_retv` makes sense.

src/map/achievement.c
MishimaHaruna: A function signature is a contract with the caller. If it says const, it means that the caller can safely call it, and expect the values not to be changed after it returns. and this only applies to variables passed by reference (pointers).

With variables passed by value (scalars), the caller doesn't care, and doesn't want to care whether the function internally needs to change the values while using them, since they're passed by value, and changing them inside the function won't cause them to change in the caller function as well.

The caller doesn't want to have to care about implementation details of **local** variables inside a function (because that's what a variable passed by value is). And having them as const prevents HPMHooking from working the way it is designed. A pre-hook may have a valid reason to want to change a value before passing it to this function, and you're preventing it.

I insist, please remove const from scalars.

Smokexyz: > A function signature is a contract with the caller. If it says const, it means that the caller can safely call it, and expect the values not to be changed after it returns. and this only applies to variables passed by reference (pointers).
Well, correct but there are no pass-by references to this function. So if you're passing a const,  it would prevent any future statements from changing the value of the variable because we absolutely don't need it to. 

> And having them as const prevents HPMHooking from working the way it is designed.
Even in the case of plugins, I can't see why the achievement and objective Id would need to be changed in a pre-hook. Since changing the first two arguments **in this function** makes as much sense as wanting to change the pointer to a session data in a function that uses it as a key argument (example: pc_delitem - where in its pre/post-hook we allow the changing of the pointer since it's a non-const).

> I insist, please remove const from scalars.
If your design idea for HPM incorporates the changing of anything, anywhere, then I'll remove the consts. I just wanted a explanation that made sense because it is not wrong to pass scalars as const unless they require a change within the scope of the function.

MishimaHaruna: > Well, correct but there are no pass-by references to this function. So if you're passing a const, it would prevent any future statements from changing the value of the variable because we absolutely don't need it to. (Including plugin hooks)

In general (at least in Hercules) the only meaning of the `const` keyword is as a guarantee to the caller. And the caller doesn't care about what the called function does with the data (i.e. if it wishes to, it can make a copy of it, or three, and change them to whatever other value). A function signature mustn't limit, or even care aboout, what a function does with its own private copy of the data.

> Even in the case of plugins, I can't see why the achievement and objective Id would need to be changed in a pre-hook.

A plugin author may see why, and we aren't here to limit their freedom of expression.

> Since changing the first two arguments in this function makes as much sense as wanting to change the pointer to a session data in a function that uses it as a key argument (example: pc_delitem - where in its pre/post-hook we allow the changing of the pointer since it's a non-const).

Which is a perfectly valid thing for a plugin to do. If a plugin wants to remap IDs differently for any reason, it should be able to. Just like a plugin can prevent this function from being called at all by stopping the hook before the function is called (as long as it respect its contract with the caller). If a plugin wants to pass a different `sd` to `pc_delitem` why would you want to stop it? The plugin might be mapping a player's inventory to another player's for whatever reason the author wishes, and might still want to call `pc_delitem` with the swapped pointer -- and this won't affect the caller in any way, since the change only happens in the pre-hook and inside the called function.

> If your design idea for HPM incorporates the changing of anything, anywhere, then I'll remove the consts. I just wanted an explanation that made sense because it is not wrong to pass scalars as const unless they are allowed to be changed within the scope of the function.

Not "anything, anywhere". Using const is still a very good thing to do for pointers (and with this I mean pointers to const data, not const pointers to mutable data -- those would be as useless as const scalars). Const only has a reason to exist (in Hercules at least) when it guarantees to the caller that the called function doesn't change the value of the passed variables. I'll make an example on the only meaningful reason I'd ever want to use `const` for, in a function signature:

```

/*
 * These functions may be defined elsewhere and
 * we don't even want to care about their internal
 * implementation details
 */
int fun(const char *foo);
int notsofun(char *foo);
int notfunatall(char * const foo);

int caller(void)
{
    char buf[256] = "foo";

    /* ... */

    // I created or received 'buf' and I validated it, ensuring it contains a meaningful value of any kind

    fun(buf);
    // I still know that the content of buf hasn't changed, and I don't need to re-ensure anything

    notsofun(buf);
    // Now I have no guarantees about anything, because the function I passed this to
    // takes a pointer to mutable data and might have changed *my* data (the caller's data)

    notfunatall(buf);
    // Once again, no guarantee. The called function is telling us that in their implementation
    // they won't change the address their local copy of the pointer points to, but really, who
    // cares? They don't guarantee, though, that they won't mess with the pointed data (which
    // was my data to begin with, not theirs)
}
```

But with scalars this all becomes pointless:

```
/*
 * These functions may be defined elsewhere and
 * we don't even want to care about their internal
 * implementation details
 */
int constfun(const int foo);
int notconstfun(int foo);
int anotherfun(int foo);

/* over-simplified HPMHooking example */
int anotherfun_pre(int *foo)
{
    if (*foo == 1)
        *foo = 42; // And so what? I want the called function to process 42 instead, can't I?
}
int anotherfun_post(int ret, int *foo)
{
    if (*foo == 42 && ret == 1)
        return 0; // Why not
}
int anotherfun_with_hpmhooking(int foo)
{
    int ret_ = anotherfun_pre(&foo);
    /* ... */
    ret_ = anotherfun(foo);
    /* ... */
    return anotherfun_post(ret_, &foo);
}

int caller(void)
{
    int foo = 1;
    const int bar = 1;

    /* ... */
    // I validated or created my data and as such I know the value of both foo and bar.

    constfun(foo);
    constfun(bar);
    notconstfun(foo);
    notconstfun(bar);

    // No difference at all for me to call either function with either const or not const data.
    // Regardless of what I call (by value), I don't care about what they did with the data. But even
    // if their local copy of my data was not const, I know they weren't able to change my copy
    // and that's all I care about.


    anotherfun_with_hpmhooking(foo);
    anotherfun_with_hpmhooking(bar);
    // And even if a plugin needs to play around with the data, I still don't care or want to care.
    // It's working on a copy of the data, so why should I limit it? The plugin knows `anotherfun()`'s
    // internals better than I do, and knows what its doing. It's a plugin that aims to modify
    // modifies that function's behavior after all.
}
```

matheus00123:
Sorry my english because I'm using the translator
At the time of giving @zeny 99999999 to make some test achievements generates this error in mapserver

http://i.imgur.com/KC9hpDV.jpg

src/map/achievement.c
Smokexyz: @MishimaHaruna Thanks, I will remove the scalar consts. I did not take HPM into consideration before I added them. Maybe you might find it of interest to create a "Hercules Developer's Handbook" or something of the like where you could add these bits so future and current devs could be briefed before committing. It could potentially reduce PR processing time and keep you from having to repeat things over in similar situations if ever.

Smokexyz:
@matheus00123 Thanks for testing. This should be fixed in the latest commit.
@MishimaHaruna requesting a review of the latest commit.

src/map/achievement.c
4144: this one never crash inside func1 even if ptr is NULL
```
void func1(int *ptr)
{
  func2(ptr);
}
```

this one will crash if ptr is NULL
```
void func3(int *ptr)
{
  *ptr = 10;
}
```

This mean in func1 check ptr for NULL is useless. It only slow down execution.
But in func3 must be check for NULL, or it may crash.

I saw in many places where was useless nullpo checks in previous code.

Useless nullpo checks is not critical, but for performance better not add it.

src/map/achievement.c
```diff
@@ -0,0 +1,1759 @@
...
+ * @param amount     [in] (goal) amount of the item collected.
+ */
+void achievement_validate_item_get(struct map_session_data *sd, int nameid, int amount)
+{
+	struct item_data *it = itemdb->exists(nameid);
+	struct achievement_objective criteria = { 0 };
+
+	nullpo_retv(sd);
+	Assert_retv(amount > 0);
+	Assert_retv(it != NULL);
```
4144: for NULL check here better use nullpo_retv(it)

src/map/achievement.c
```diff
@@ -0,0 +1,1759 @@
...
+ * @param sd        [in] pointer to session data.
+ * @param class     [in] (criteria) class of the monster tamed.
+ */
+void achievement_validate_taming(struct map_session_data *sd, int class)
+{
+	struct achievement_objective criteria = { 0 };
+
+	nullpo_retv(sd);
+	Assert_retv(class > 0);
+	Assert_retv(mob->db(class) != NULL);
```
4144: for NULL check you can use nullpo

src/map/achievement.c
```diff
@@ -0,0 +1,1759 @@
...
+	nullpo_retr(false, conf);
+	nullpo_retr(false, entry);
+
+	if ((t = libconfig->setting_get_member(conf, "Objectives")) && config_setting_is_group(t)) {
+		int i = 0;
+		// Initialize the buffer objective vector.
+		VECTOR_INIT(entry->objective);
+
+		for (i = 0; i < libconfig->setting_length(t) && i < MAX_ACHIEVEMENT_OBJECTIVES; i++) {
+			struct config_setting_t *tt = NULL;
```
4144: Reading separate objective should be in separate function (body of this for in separate function)

src/map/achievement.c
```diff
@@ -0,0 +1,1759 @@
...
+	struct config_setting_t *tt = NULL;
+	int val = 0, i = 0;
+
+	nullpo_retv(t);
+	nullpo_retv(entry);
+
+	if ((tt = libconfig->setting_get_member(t, "Items")) && config_setting_is_group(t)) {
+
+		for (i = 0; i < libconfig->setting_length(tt) && i < MAX_ACHIEVEMENT_ITEM_REWARDS; i++) {
+			struct config_setting_t *it = NULL;
```
4144: body of for should be in separate function

src/map/intif.c
```diff
@@ -1495,11 +1497,99 @@ void intif_parse_DeleteHomunculusOk(int fd) {
...
+ * @param fd socket descriptor.
+ */
+void intif_parse_achievements_load(int fd)
+{
+	int size = RFIFOW(fd, 2);
+	int char_id = RFIFOL(fd, 4);
+	int payload_count = (size - 8) / sizeof(struct achievement);
+	struct map_session_data *sd = map->charid2sd(char_id);
+	int i = 0;
+
```
4144: i think here need add nullpo check for sd

kvri4e:
failed assertion triggered by ACH_ITEM_GET_WORTH(possibly).

vBrenth:
@Smokexyz

Smokexyz:
@4144 I've addressed your change requests, let me know if I've missed something. Thanks.

MishimaHaruna:
Reviewed 3 of 3 files at r5, 3 of 3 files at r6, 17 of 20 files at r7, 9 of 29 files at r8.
Review status: 32 of 55 files reviewed at latest revision, 52 unresolved discussions.

---

*[src/char/char.c, line 301 at r8](https://reviewable.io:443/reviews/herculesws/hercules/1666#-KliMUv-whUQhDdb2jth:-KliMUv-whUQhDdb2jti:b-2i4x7z) ([raw file](https://github.com/herculesws/hercules/blob/2a00779449573a7c19a4b69c7736a44fd0825dcc/src/char/char.c#L301)):*
> ```C
> 			idb_remove(chr->char_db_,char_id);
> 		if (c_ach != NULL)
> 			idb_remove(inter_achievement->char_achievements, char_id);
> ```

This would leak the allocated vector (a call to `VECTOR_CLEAR()` is missing)

---

*[src/char/int_achievement.c, line 51 at r8](https://reviewable.io:443/reviews/herculesws/hercules/1666#-KliFqu-K8WJk1_rbA-S:-KliFqu-K8WJk1_rbA-T:b-3lqi2x) ([raw file](https://github.com/herculesws/hercules/blob/2a00779449573a7c19a4b69c7736a44fd0825dcc/src/char/int_achievement.c#L51)):*
> ```C
>  * @return number of achievements saved.
>  */
> int inter_achievement_tosql(int char_id, struct char_achievements *cp, struct char_achievements *p)
> ```

Could you make `p` const? This would make the ownership more obvious (it's the caller's), and prevent this function from using `VECTOR_CLEAR()` on it.

---

*[src/char/int_achievement.c, line 90 at r8](https://reviewable.io:443/reviews/herculesws/hercules/1666#-KliHKelqWnkFF-BXl0g:-KliHKelqWnkFF-BXl0h:bc3if2d) ([raw file](https://github.com/herculesws/hercules/blob/2a00779449573a7c19a4b69c7736a44fd0825dcc/src/char/int_achievement.c#L90)):*
> ```C
> 		Sql_ShowDebug(inter->sql_handle);
> 		StrBuf->Destroy(&buf); // Destroy the buffer.
> 		VECTOR_CLEAR(*p); /* Free buffer data in the case of an error. */
> ```

This is done by the caller from what I see below

---

*[src/char/int_achievement.c, line 111 at r8](https://reviewable.io:443/reviews/herculesws/hercules/1666#-KliISg6HEH-9uzP_-w4:-KliISg6HEH-9uzP_-w5:bc4zvm1) ([raw file](https://github.com/herculesws/hercules/blob/2a00779449573a7c19a4b69c7736a44fd0825dcc/src/char/int_achievement.c#L111)):*
> ```C
>  * Retrieves all achievements of a character.
>  * @param char_id  character identifier.
>  * @param ca       pointer to character achievements structure.
> ```

Typo: `ca` instead of `cp`

Suggestion: this is an output argument, you can document that behavior with the following Doxygen syntax:

```
@param[out] cp   Pointer to the character achievements structure.
```

---

*[src/char/int_achievement.c, line 139 at r8](https://reviewable.io:443/reviews/herculesws/hercules/1666#-KliIzaNzIQTCUM2yelb:-KliIzaNzIQTCUM2yelc:bn4scr5) ([raw file](https://github.com/herculesws/hercules/blob/2a00779449573a7c19a4b69c7736a44fd0825dcc/src/char/int_achievement.c#L139)):*
> ```C
> 		int j = 0;
> 
> 		VECTOR_CLEAR(*cp);
> ```

I believe the vector should be cleared even if the query results have 0 rows (just above the `if ()`?)

---

*[src/char/int_achievement.c, line 209 at r8](https://reviewable.io:443/reviews/herculesws/hercules/1666#-KliJYzKTpctEJpRcM9s:-KliJYzKTpctEJpRcM9t:bj2fjrm) ([raw file](https://github.com/herculesws/hercules/blob/2a00779449573a7c19a4b69c7736a44fd0825dcc/src/char/int_achievement.c#L209)):*
> ```C
>  * @param  cp      [in]  Pointer to character's achievement data vector.
>  */
> void mapif_send_achievements_to_map(int fd, int char_id, struct char_achievements *cp)
> ```

`cp` can be const here

---

*[src/char/int_achievement.c, line 213 at r8](https://reviewable.io:443/reviews/herculesws/hercules/1666#-KliKzSbnHJi_yjvGwm0:-KliKzSbnHJi_yjvGwm1:bqgixu9) ([raw file](https://github.com/herculesws/hercules/blob/2a00779449573a7c19a4b69c7736a44fd0825dcc/src/char/int_achievement.c#L213)):*
> ```C
> 	int i = 0;
> 	int data_size = sizeof(struct achievement) * VECTOR_LENGTH(*cp);
> ```

Since we send everything in one packet (and right now it's a good idea, since it uses less than 1/3 of the maximum packet size), it's a good idea to add a compile-time sanity check here:

```
STATIC_ASSERT(sizeof(struct achievement) * MAX_ACHIEVEMENT_DB + 8 <= UINT16_MAX, "The achievements data can potentially be larger than the maximum packet size. This may cause errors at run-time.");
```

---

*[src/common/mmo.h, line 201 at r8](https://reviewable.io:443/reviews/herculesws/hercules/1666#-KliN4SqITnar9TonVVf:-KliN4SqITnar9TonVVg:b-t43nyx) ([raw file](https://github.com/herculesws/hercules/blob/2a00779449573a7c19a4b69c7736a44fd0825dcc/src/common/mmo.h#L201)):*
> ```objc
> #ifndef MAX_ACHIEVEMENT_DB
> #define MAX_ACHIEVEMENT_DB 360          // Maximum number of achievements
> #define MAX_ACHIEVEMENT_OBJECTIVES 10	// Maximum number of achievement objectives
> ```

There's an extra tab in this line (it should be aligned with spaces)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1666)*

vBrenth:
@Smokexyz <3

marky291:
`	CC	int_achievement.c
In file included from /usr/include/features.h:367:0,
                 from /usr/include/inttypes.h:25,
                 from ../common/cbasetypes.h:124,
                 from ../common/hercules.h:24,
                 from int_achievement.h:24,
                 from int_achievement.c:23:
int_achievement.c: In function 'mapif_send_achievements_to_map':
../common/cbasetypes.h:461:32: error: redundant redeclaration of '__Static_assert_function' [-Werror=redundant-decls]
 #define STATIC_ASSERT(ex, msg) _Static_assert(ex, msg)
                                ^
int_achievement.c:215:2: note: in expansion of macro 'STATIC_ASSERT'
  STATIC_ASSERT((sizeof(struct achievement) * MAX_ACHIEVEMENT_DB + 8 <= UINT16_MAX),
  ^
../common/cbasetypes.h:461:32: note: previous declaration of '__Static_assert_function' was here
 #define STATIC_ASSERT(ex, msg) _Static_assert(ex, msg)
                                ^
../common/mmo.h:291:1: note: in expansion of macro 'STATIC_ASSERT'
 STATIC_ASSERT(MAX_ITEM_OPTIONS <= 5, "This value is limited by the client and database layout and should only be increased if you know the consequences.");`

Smokexyz:
@marky291 weird because it passed all checks in CI.

marky291:
Thanks for your work on this, keep up the good job.  🥇

vBrenth:
@Smokexyz How this thing going on? I am so excited with this one!

MishimaHaruna:
Reviewed 1 of 55 files at r4, 1 of 20 files at r7, 16 of 29 files at r8, 11 of 11 files at r9.
Review status: all files reviewed at latest revision, 43 unresolved discussions.

---

*[src/char/int_achievement.c, line 231 at r9](https://reviewable.io:443/reviews/herculesws/hercules/1666#-Kog7a780SlL-BeM6Bis:-Kog7a780SlL-BeM6Bit:b-w4u6xl) ([raw file](https://github.com/herculesws/hercules/blob/178c047ff1c854822255157f39ce07c6a09b66eb/src/char/int_achievement.c#L231)):*
> ```C
>  * Handles achievement request and saves data from map server.
>  * @packet[in] 0x3099 <packet_size>.W <char_id>.L <char_achievement>.P
>  * @param[fd]  session socket descriptor.
> ```

Did you mean `@param[in] fd` here?

---

*[src/map/achievement.c, line 934 at r9](https://reviewable.io:443/reviews/herculesws/hercules/1666#-KogFxKOkUj8vkUFzJ3l:-KogFxKOkUj8vkUFzJ3m:bq5bbcx) ([raw file](https://github.com/herculesws/hercules/blob/178c047ff1c854822255157f39ce07c6a09b66eb/src/map/achievement.c#L934)):*
> ```C
> 	if (!entry) {
> 		ShowError("achievement_readdb_ranks: No ranks provided in '%s'!\n", filename);
> 		return;
> ```

Leaking the `ar_conf`. You should move the `libconfig->destroy()` call to before this `if`.

---

*[src/map/achievement.h, line 234 at r9](https://reviewable.io:443/reviews/herculesws/hercules/1666#-KogE_EE5x6aCv_jWpsa:-KogE_EE5x6aCv_jWpsb:b-he40wk) ([raw file](https://github.com/herculesws/hercules/blob/178c047ff1c854822255157f39ce07c6a09b66eb/src/map/achievement.h#L234)):*
> ```objc
> 
> // Checks if a (unique) union member from the objective structure is used.
> #define achievement_objective_unique(u) ( \
> ```

Can you remove this macro? it's not used, and it's not safe to use (checking the value of union fields that were never set)

---

*[src/map/pc.c, line 1522 at r9](https://reviewable.io:443/reviews/herculesws/hercules/1666#-KogH1V9JhYzo-Cye3f5:-KogH1V9JhYzo-Cye3f6:bc0jd0b) ([raw file](https://github.com/herculesws/hercules/blob/178c047ff1c854822255157f39ce07c6a09b66eb/src/map/pc.c#L1522)):*
> ```C
> 
> 	// Achievements [Smokexyz/Hercules]
> 	intif->achievements_request(sd);
> ```

What if an achievement is processes before the data is received from the char server? I haven't seen a `received` flag anywhere (if I missed it, please disregard this). We should block any achievements until the data is loaded and received from the char server.

---

*[src/map/script.c, line 20401 at r9](https://reviewable.io:443/reviews/herculesws/hercules/1666#-Kog8zvs-2NK6PeFtIQJ:-Kog8zvs-2NK6PeFtIQK:bd948zx) ([raw file](https://github.com/herculesws/hercules/blob/178c047ff1c854822255157f39ce07c6a09b66eb/src/map/script.c#L20401)):*
> ```C
> 	const struct achievement_data *ad = NULL;
> 
> 	nullpo_retr(false, sd);
> ```

Generating a backtrace here if the script was run without a RID attached isn't ideal (if anything, it should show an [Error] message, or just report the source by returning false, see other commands that have an sd check)

My understanding, though, is that the command should work without a RID attached, if the `char_id` argument is passed and valid. (why don't we use an account ID instead, for consistency with other commands)

---

*[src/map/script.c, line 20434 at r9](https://reviewable.io:443/reviews/herculesws/hercules/1666#-Kog9gISXIwchovDytAy:-Kog9gISXIwchovDytAz:b-rwb2fx) ([raw file](https://github.com/herculesws/hercules/blob/178c047ff1c854822255157f39ce07c6a09b66eb/src/map/script.c#L20434)):*
> ```C
> 			script_pushint(st, 0);
> 			return false;
> 		} else if (sd != map->charid2sd(char_id)) {
> ```

Why do you check against the currently attached RID? I don't fully understand the point of passing a char_id, if that's the case

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1666)*

vBrenth:
@Smokexyz Please progress this, Thank you.

Smokexyz:
This should rather wait until #1686 is merged, to complete the system by sending reward items through the mail. Also the title system.

vBrenth:
Okay, then gotta wait Haru to push that PR :'(

Cherminator38:
Hi, Very thank's for your job ! 
I have install diff, all is good but when i achieve for example "Organize a party " success message is: "<"
and in console i have this error: 

[Error]: --- failed assertion --------------------------------------------
[Error]: achievement.c:849: 'rank > 0 && rank <= ( (achievement->rank_exp)._len_ )' in function `achievement_validate_achievement_rank'
[Error]: ./map-server(assert_report+0x81) [0x64c0b1]
[Error]: ./map-server(achievement_validate_achievement_rank+0x55) [0x43b215]
[Error]: ./map-server(clif_achievement_send_update+0x134) [0x495514]
[Error]: ./map-server(achievement_validate_type+0x149) [0x43e229]
[Error]: ./map-server(achievement_validate_party_create+0x47) [0x43a937]
[Error]: ./map-server(party_create+0x114) [0x542b04]
[Error]: ./map-server(clif_parse_CreateParty2+0xaf) [0x4901ff]
[Error]: ./map-server(clif_parse+0x1a2) [0x4957d2]
[Error]: ./map-server(do_sockets+0x28e) [0x659f5e]
[Error]: ./map-server(main+0x182) [0x43a2a2]
[Error]: --- end failed assertion ----------------------------------------

So how can fix it ?

xEasycore:
Any news on this?
Since #1686 is already implemented.

tlacson7:
i hope to see this merged very soon.

arkangel2k4:
Really hope to see this get some attention soon :)

xBeret:
Something about when the system will be implemented? @Smokexyz ?

Smokexyz:
This is somewhat "complete" from my side. The current RoDEX I believe doesn't support sending of mail through non-player entities. Title system can be completed in another PR.
@MishimaHaruna @4144

Asheraf:
@Smokexyz it should support sending via npcs which is how its done in official see `enum MAIL_TYPE`

4144:
Yes rodex should support mail from npc. But here some limits on client versions. Old clients may not support this

4144:
@Smokexyz squash temporary commits, i will try review again after this

src/map/achievement.c
```diff
@@ -0,0 +1,1858 @@
...
+ * @return false on failure, true on success.
+ */
+bool achievement_readdb_validate_reward_item_sub(const struct config_setting_t *t, int index, struct achievement_data *entry)
+{
+	struct config_setting_t *it = NULL;
+	struct achievement_reward_item item = { 0 };
+	const char *name = NULL;
+	int amount = 0;
+	int val = 0;
+
```
4144: missing checks for t and entry

sql-files/main.sql
```diff
@@ -883,8 +905,10 @@ INSERT IGNORE INTO `sql_updates` (`timestamp`) VALUES (1475526420); -- 2016-10-0
 INSERT IGNORE INTO `sql_updates` (`timestamp`) VALUES (1477434595); -- 2016-10-26--10-29.sql
 INSERT IGNORE INTO `sql_updates` (`timestamp`) VALUES (1488454834); -- 2017-03-02--11-40.sql
 INSERT IGNORE INTO `sql_updates` (`timestamp`) VALUES (1488744559); -- 2017-03-05--08-09.sql
+INSERT IGNORE INTO `sql_updates` (`timestamp`) VALUES (1489389953); -- 2017-03-13--07-25.sql
```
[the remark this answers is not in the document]
Smokexyz: I don't think it will affect anything, the server will still complain of a missing upgrade.

Smokexyz:
@4144 I've addressed the issues and remerged. If anything else do let me know.

EyesOfAHawk:
sending my good wishes to the merging of this to herc main branch :D

xBeret:
@4144 @Smokexyz @MishimaHaruna Please look at this pr, for years that exists in kro, hercules users want to use this

4144:
Need rebase and sqush for followup commit, but other i think ok.

lucaslsb:
Awesome!
This works only in PACKETVER >= 20141016?

vBrenth:
@Smokexyz , @MishimaHaruna :D Ping!
Please give this feature for Hercules users! We are all excited for years :D

MishimaHaruna:
Reviewed 4 of 59 files at r16, 1 of 1 files at r17, 3 of 3 files at r18, 16 of 20 files at r19, 23 of 32 files at r20, 12 of 12 files at r21.
Review status: all files reviewed at latest revision, 44 unresolved discussions, some commit checks failed.

---

*[src/map/achievement.c, line 1278 at r21](https://reviewable.io:443/reviews/herculesws/hercules/1666#-L5enOHwbuadLLhVjtsB:-L5enOHwbuadLLhVjtsC:byhloic) ([raw file](https://github.com/herculesws/hercules/blob/27e83a6c4ce575eff0a1df68e84ebad8615db1e8/src/map/achievement.c#L1278)):*
> ```C
> 		while (j < libconfig->setting_length(tt)) {
> 			if ((val = libconfig->setting_get_int_elem(tt, j))) {
> 				if (val < IT_HEALING || val >= IT_MAX) {
> ```

This function is mixing bitmask constants (`enum objective_item_types`) with non-bitmask constants (`enum item_types`) and using them as a bitmask. This is likely not intended, can you please check?

---

*[src/map/achievement.h, line 240 at r21](https://reviewable.io:443/reviews/herculesws/hercules/1666#-L5eX7295FyBcss9yIEl:-L5eX7295FyBcss9yIEm:b-ptf6yd) ([raw file](https://github.com/herculesws/hercules/blob/27e83a6c4ce575eff0a1df68e84ebad8615db1e8/src/map/achievement.h#L240)):*
> ```objc
> 	void (*final) (void);
> 	/* */
> 	int (*db_finalise) (union DBKey key, struct DBData *data, va_list args);
> ```

Can you change this to the AmE spelling we use everywhere in Hercules? (`db_finalize`, or `db_clear` are both fine)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1666)*

vBrenth:
@Smokexyz

bryanveloso:
Was there any reason for this being randomly closed? I was looking at transferring from an rAthena instance already using the achievement system and this says to me that it won't be worked on any longer.

4144:
Author closed all his pull requests because dont want work on them anymore.
But all code still exists in his repository, if you know how do rebase, you can use it.
See line ``Smokexyz:achievement-system``. This is repository name + branch name

Issue #1669, opened 2017-04-09
Size of struct mmo_charstatus


MishimaHaruna:
### Issue Prelude

-  [x] I have not modified the source prior to reproducing this issue.
-  [x] I am using the latest version of Hercules.
-  [x] I am aware that this report will be closed or deleted if it becomes obvious that I am stating the false.

### Description

The size of `struct mmo_charstatus` has grown a lot recently, especially after merging #1598 and this poses a problem for future expansion, as well as customization possibilities.

In case someone isn't aware, this is a problem because the struct needs to be smaller than 64kB, since we send between char and map server as a single packet.

The struct size before merging #1598 was `39,896` bytes, and now is `65,496` bytes (very close to the maximum limit). Because of this, even simple modifications to it (i.e. increasing the size of the friend list) are pretty much impossible right now in Hercules.

The main offender is the storage (`struct storage_data` is now `43,208` bytes). I believe we should split that from the `struct mmo_charstatus`, and handle it separately (like we do with the char regs).

**Branch(es):**
- [x] master

**Hercules rev. hash/commit:** 

Git revision src: 7cc094df792340752c9637fc28b3a1108dcd53d7

```
[Info]: Hercules 64-bit for Mac OS X
[Info]: Git revision (src): '7cc094df792340752c9637fc28b3a1108dcd53d7'
[Info]: Git revision (scripts): '7cc094df792340752c9637fc28b3a1108dcd53d7'
[Info]: OS version: 'Mac OS X 10.12.4 16E195 [x86_64]'
[Info]: CPU: 'Intel Core i7 (2.8 GHz) [4]'
[Info]: Compiled with GCC v6.3.0
```

PR #1679, opened 2017-04-13
Dismemberment of storage_data from mmo_charstatus. 

Smokexyz:
### Pull Request Prelude

### Changes Proposed

Dismemberment of `storage_data` from `mmo_charstatus`. 
Add `storage_data` reception, parsing and sending from the map-server.
Remove loading and saving of `storage_data` through char.c
Re-declaration of structure storage_data as a vector.
Re-code of portions in the map-server using `storage_data`.
Unchanged: `storage_data` (now vector data) is still saved through `char_memitemdata_to_sql`.

**Affected Branches:** Master

[//]: # (Master? Slave?)

**Issues addressed:** #1669 

- [x] Search for unchanged places where storage_data was used and (if found) update to comply with the new vector format.

### DIFF
--- a/doc/constants.md
+++ b/doc/constants.md
@@ -3799,6 +3799,16 @@
 - `EQP_SHADOW_ACC_R`: 1048576
 - `EQP_SHADOW_ACC_L`: 2097152
 
+### Item Option Types
+
+- `IT_OPT_INDEX`: 0
+- `IT_OPT_VALUE`: 1
+- `IT_OPT_PARAM`: 2
+
+### Maximum Item Options
+
+- `MAX_ITEM_OPTIONS`: 5
+
 ### Navigation constants, use with *navigateto*
 
 - `NAV_NONE`: 0


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1679)

src/map/storage.c
```diff
@@ -329,22 +348,26 @@ int storage_storageaddfromcart(struct map_session_data* sd, int index, int amoun
...
 		return 0;
 
-	if( (flag = pc->cart_additem(sd,&sd->status.storage.items[index],amount,LOG_TYPE_STORAGE)) == 0 )
-		storage->delitem(sd,index,amount);
+	if ((flag = pc->cart_additem(sd, it, amount, LOG_TYPE_STORAGE)) == 0)
+		storage->delitem(sd, index, amount);
 	else {
 		clif->dropitem(sd, index,0);
-		clif->cart_additem_ack(sd,flag == 1?0x0:0x1);
+		clif->cart_additem_ack(sd, flag == 1?0x0:0x1);
```
Jedzkie: `clif->cart_additem_ack(sd, flag == 1 ? 0x0 : 0x1);`

src/map/storage.c
```diff
@@ -329,22 +348,26 @@ int storage_storageaddfromcart(struct map_session_data* sd, int index, int amoun
...
-	if( amount < 1 || amount > sd->status.storage.items[index].amount )
+	if (amount < 1 || amount > it->amount)
 		return 0;
 
-	if( (flag = pc->cart_additem(sd,&sd->status.storage.items[index],amount,LOG_TYPE_STORAGE)) == 0 )
-		storage->delitem(sd,index,amount);
+	if ((flag = pc->cart_additem(sd, it, amount, LOG_TYPE_STORAGE)) == 0)
+		storage->delitem(sd, index, amount);
 	else {
 		clif->dropitem(sd, index,0);
```
Jedzkie: `clif->dropitem(sd, index, 0);`

src/map/storage.c
```diff
@@ -180,29 +182,35 @@ int storage_additem(struct map_session_data* sd, struct item* item_data, int amo
 	{//Stackable
 		for( i = 0; i < MAX_STORAGE; i++ )
 		{
-			if( compare_item(&stor->items[i], item_data) )
+			it = &VECTOR_INDEX(sd->storage, i);
+			if( compare_item(it, item_data) )
```
Jedzkie: `if (compare_item(it, item_data))`

src/map/storage.c
```diff
@@ -180,29 +182,35 @@ int storage_additem(struct map_session_data* sd, struct item* item_data, int amo
...
 			{// existing items found, stack them
-				if( amount > MAX_AMOUNT - stor->items[i].amount || ( data->stack.storage && amount > data->stack.amount - stor->items[i].amount ) )
+				if( amount > MAX_AMOUNT - it->amount || ( data->stack.storage && amount > data->stack.amount - it->amount ) )
```
Jedzkie: `if (amount > MAX_AMOUNT - it->amount || (data->stack.storage && amount > data->stack.amount - it->amount))`

src/map/storage.c
Smokexyz: Code-styling will be handled in a separate commit, once the main commit has been reviewed. But thanks for your findings.

src/map/storage.c
Smokexyz: This line isn't even changed though.

src/char/int_storage.c
```diff
@@ -74,36 +75,41 @@ int inter_storage_fromsql(int account_id, struct storage_data* p)
...
-			item->card[j] = atoi(data);
-		}
-		/* Item Options */
-		for (j = 0; j < MAX_ITEM_OPTIONS; ++j) {
-			SQL->GetData(inter->sql_handle, 10 + MAX_SLOTS + j * 2, &data, NULL);
-			item->option[j].index = atoi(data);
-			SQL->GetData(inter->sql_handle, 11 + MAX_SLOTS + j * 2, &data, NULL);
-			item->option[j].value = atoi(data);
+	if ((num_rows = (int)SQL->NumRows(inter->sql_handle)) != 0) {
+		VECTOR_ENSURE(*p, num_rows, 1);
```
4144: Probably here better use bigger number for allocation step (replace 1 to other value). May be 4 or 8?
Bigger number will allocate bit more memory, but from other side it reduce number of allocations.

Smokexyz: Not really, step 1 would allocate 1 time on load. Only 
```
while ((_n) + VECTOR_LENGTH(_vec) > _newcapacity_) \
			_newcapacity_ += (_step); \
```
would be executed `(_n + length)/_step` times.

src/char/int_storage.c
```diff
@@ -41,24 +41,25 @@ struct inter_storage_interface inter_storage_s;
 struct inter_storage_interface *inter_storage;
 
 /// Save storage data to sql
-int inter_storage_tosql(int account_id, struct storage_data* p)
+int inter_storage_tosql(int account_id, const struct item *p)
```
MishimaHaruna: I think it's better for this function to stay as it was (and just add const). Similar things hurt us in the past, and I don't want to repeat it (it's the reason why I started using VECTOR wherever we were using pointers to the first element of an array)

`struct storage_data *` has the benefit over `struct item *` of containing an array of known size. If you really want to pass a pointer to the first entry, then you have to always pass the size along with it.


The following are both equally acceptable:
```
// This passes the knowledge that the item array is part of a storage, and as such the function knows that, by definition, it'll have MAX_STORAGE elements
int inter_storage_tosql(int account_id, const struct storage_data* p)

// This pushes the responsibility on the caller, which knows that the item array is part of a storage, of letting the function know the array size
int inter_storage_tosql(int account_id, const struct item *p, int length)
```

I don't want functions that take a pointer to the first element of an array, and arbitrarily decide the size of said array.

MishimaHaruna:
I started reviewing this PR, but at some point I got a bit lost. If this uses a VECTOR now, why does it always assume a fixed array length? (in all the `index >= 0 && index < MAX_STORAGE` checks)

src/char/int_storage.c
Smokexyz: Ah correct, this should use and send the length of the vector through to `char_ memitemdata_to_sql`

Smokexyz:
@MishimaHaruna If you could review the files please.

src/char/int_storage.c
```diff
@@ -41,24 +41,27 @@ struct inter_storage_interface inter_storage_s;
...
 	char* data;
 	int i;
 	int j;
+	int num_rows = 0;
 
 	nullpo_ret(p);
-	memset(p, 0, sizeof(struct storage_data)); //clean up memory
-	p->storage_amount = 0;
+
+	VECTOR_INIT(*p);
```
MishimaHaruna: `VECTOR_INIT()` should be on the caller's side (so that you don't have to pass an uninitialized variable to a function), and here you should have `VECTOR_CLEAR()` (just as a safety measure). This change would make the  caller the owner of this struct, which is what we want here so that allocations and deallocations are always in matching pair where the owner resides

src/char/int_storage.c
```diff
@@ -240,13 +248,100 @@ int mapif_save_guild_storage_ack(int fd, int account_id, int guild_id, int fail)
...
+ */
+int mapif_account_storage_load(int fd, int account_id)
+{
+	struct storage_data p = { 0 };
+	int count = 0, i = 0, len = 0;
+
+	Assert_ret(account_id > 0);
+
+	if ((count = inter_storage->fromsql(account_id, &p)) == 0)
+		return 0;
```
MishimaHaruna: I don't think an empty storage should be considered a failed load. It should still send the info ("the storage is empty") to the mapserver

src/char/int_storage.c
```diff
@@ -240,13 +248,100 @@ int mapif_save_guild_storage_ack(int fd, int account_id, int guild_id, int fail)
...
+ * @return 1 on success, 0 on failure.
+ */
+int mapif_parse_AccountStorageSave(int fd)
+{
+	int payload_size = RFIFOW(fd, 2) - 8, account_id = RFIFOL(fd, 4);
+	int i = 0, count = 0;
+	struct storage_data p_buf;
+
+	Assert_ret(fd > 0);
+	Assert_ret(account_id < 0);
```
MishimaHaruna: Do you mean > 0 here?

src/char/int_storage.c
```diff
@@ -240,13 +248,100 @@ int mapif_save_guild_storage_ack(int fd, int account_id, int guild_id, int fail)
...
+
+	if (payload_size > 0 || (count = payload_size/sizeof(struct item)) <= 0)
```
MishimaHaruna: `payload_size < 0`?

Also, `count == 0` is a reasonable value to accept. If a storage gets emptied, we still want to save it as empty (and delete old existing entries).

src/map/atcommand.c
```diff
@@ -5160,17 +5160,17 @@ ACMD(storeall)
...
 
 	if (sd->state.storage_flag == STORAGE_FLAG_NORMAL) {
 		clif->message(fd, msg_fd(fd,250));
 		return false;
 	}
 
-	j = sd->status.storage.storage_amount;
-	for (i = 0; i < j; ++i) {
-		storage->delitem(sd, i, sd->status.storage.items[i].amount);
+	for (i = 0; i < VECTOR_LENGTH(sd->storage); ++i) {
```
MishimaHaruna: This won't work properly: it's iterating on the same set it's deleting (the indexes as well as VECTOR_LENGTH will change during the loop).

If you really want to do this, and you really want to use vectors, then you have to loop in reverse:

```
while (VECTOR_LENGTH(sd->storage) > 0) {
    storage->delitem(sd, VECTOR_LENGTH(sd->storage) - 1, VECTOR_LAST(sd->storage).amount);
}
```

This will also be more efficient (deleting the last element of a vector doesn't cause it to have to move every other element multiple times)

src/map/chrif.c
```diff
@@ -324,6 +324,9 @@ bool chrif_save(struct map_session_data *sd, int flag) {
 	if( sd->save_quest )
 		intif->quest_save(sd);
 
+	if (VECTOR_LENGTH(sd->storage))
```
MishimaHaruna: You likely want to save the storage even if it's empty (to handle the case where the storage was full, then items get deleted from it)

src/map/intif.c
```diff
@@ -439,6 +439,97 @@ int intif_request_registry(struct map_session_data *sd, int flag)
...
+void intif_parse_account_storage(int fd)
+{
+	int account_id = 0, payload_size = 0, storage_count = 0;
+	int i = 0;
+	struct map_session_data *sd = NULL;
+
+	Assert_retv(fd > 0);
+
+	if ((payload_size = RFIFOW(fd, 2) - 8) == 0)
+		return; // no data, nothing to do.
```
MishimaHaruna: If there's no data, we likely still want to know whose it was, and mark the storage data as received for the character

src/map/intif.c
```diff
@@ -439,6 +439,97 @@ int intif_request_registry(struct map_session_data *sd, int flag)
...
+{
+	int len = 0, i = 0;
+
+	if (intif->CheckForCharServer())
+		return;
+
+	len = 8 + VECTOR_LENGTH(sd->storage) * sizeof(struct item);
+
+	if (len == 8)
+		return; // nothing to save.
```
MishimaHaruna: If there's nothing to save, we still want to save it

src/map/pc.c
```diff
@@ -10163,21 +10167,22 @@ int pc_checkitem(struct map_session_data *sd)
...
+			id = it->nameid;
 
 			if (!id)
 				continue;
 
 			if( id && !itemdb_available(id) ) {
-				ShowWarning("Removed invalid/disabled item id %d from storage (amount=%d, char_id=%d).\n", id, sd->status.storage.items[i].amount, sd->status.char_id);
-				storage->delitem(sd, i, sd->status.storage.items[i].amount);
+				ShowWarning("Removed invalid/disabled item id %d from storage (amount=%d, char_id=%d).\n", id, it->amount, sd->status.char_id);
+				storage->delitem(sd, i, it->amount);
```
MishimaHaruna: See the note in atcommand clearstorage on using `storage->delitem()` inside a loop on `VECTOR_LENGTH(sd->storage)`. The same issue occurs here

src/map/storage.c
```diff
@@ -212,21 +220,28 @@ int storage_additem(struct map_session_data* sd, struct item* item_data, int amo
...
-	if( sd->status.storage.items[n].amount == 0 )
-	{
-		memset(&sd->status.storage.items[n],0,sizeof(sd->status.storage.items[0]));
-		sd->status.storage.storage_amount--;
-		if( sd->state.storage_flag == STORAGE_FLAG_NORMAL )
-			clif->updatestorageamount(sd, sd->status.storage.storage_amount, MAX_STORAGE);
+	it->amount -= amount;
+
+	if (it->amount == 0) {
+		VECTOR_ERASE(sd->storage, n);
```
MishimaHaruna: This will cause the storage to be re-sorted, which the client doesn't expect. Any data in an open storage window will likely become invalid, and the client will start using wrong indexes for subsequent operations. This needs to be tested, but I don't expect it to work properly (I'm not sure how to let the client know that we changed the indexes from what we originally sent)

src/map/intif.c
Smokexyz: Hmm, why here though? I understand sending "no data" from char to map to flag it as received but why would the char server need this.

src/map/atcommand.c
Smokexyz: As discussed, vector indexes of storage items will be `memset` to 0 while the storage is open. When closed, zeroed entries will be erased from the vector.

src/map/chrif.c
Smokexyz: I don't understand what you mean here, if you could explain a little more in detail please.

MishimaHaruna: Saving an empty storage is a case that should be handled normally. A storage can currently have 0 items in it, and we still want to save it.
This happens when a character has items in its storage, logs in, then removes all the items from its storage. In that case we still want to save an empty storage, or the items will be duplicated.

Smokexyz: Ah right, silly me.

Smokexyz: Should be fixed now.

src/map/intif.c
Smokexyz: Done.

src/char/int_storage.c
```diff
@@ -74,36 +77,41 @@ int inter_storage_fromsql(int account_id, struct storage_data* p)
...
 		}
 	}
-	p->storage_amount = i;
+
 	SQL->FreeResult(inter->sql_handle);
 
-	ShowInfo("storage load complete from DB - id: %d (total: %d)\n", account_id, p->storage_amount);
-	return 1;
+	ShowInfo("storage load complete from DB - id: %d (total: %d)\n", account_id, VECTOR_LENGTH(*p));
+	return num_rows;
```
MishimaHaruna: Instead of `num_rows`, this should return `VECTOR_LENGTH(*p)` (they can be different in case there are more items in the SQL database than the `MAX_STORAGE` for example.  You can, in fact, get rid of the `num_rows` variable entirely.

src/map/storage.c
```diff
@@ -178,31 +180,37 @@ int storage_additem(struct map_session_data* sd, struct item* item_data, int amo
...
+
 	// add item to slot
-	memcpy(&stor->items[i],item_data,sizeof(stor->items[0]));
-	stor->storage_amount++;
-	stor->items[i].amount = amount;
-	clif->storageitemadded(sd,&stor->items[i],i,amount);
-	clif->updatestorageamount(sd, stor->storage_amount, MAX_STORAGE);
+	VECTOR_PUSH(sd->storage, *item_data);
+
+	it = &VECTOR_INDEX(sd->storage, VECTOR_LENGTH(sd->storage)-1);
```
MishimaHaruna: A shortand for this is:

```
it = &VECTOR_LAST(sd->storage);
```

src/map/storage.c
```diff
@@ -212,21 +220,29 @@ int storage_additem(struct map_session_data* sd, struct item* item_data, int amo
...
 {
+	struct item *it = NULL;
+
 	nullpo_retr(1, sd);
-	Assert_retr(1, n >= 0 && n < MAX_STORAGE);
-	if( sd->status.storage.items[n].nameid == 0 || sd->status.storage.items[n].amount < amount )
+
+	Assert_retr(1, n >= 0 && n < VECTOR_LENGTH(sd->storage));
+
+	if (sd->state.storage_flag != STORAGE_FLAG_NORMAL)
```
MishimaHaruna: Can you explain the reason for this line? Do we want delitem to silently fail (and return success) if the storage isn't open? I believe we have some functions that don't change the sd's storage_flag, but still want to delete items from the storage.

src/map/storage.c
```diff
@@ -212,21 +220,29 @@ int storage_additem(struct map_session_data* sd, struct item* item_data, int amo
...
-	{
-		memset(&sd->status.storage.items[n],0,sizeof(sd->status.storage.items[0]));
-		sd->status.storage.storage_amount--;
-		if( sd->state.storage_flag == STORAGE_FLAG_NORMAL )
-			clif->updatestorageamount(sd, sd->status.storage.storage_amount, MAX_STORAGE);
+	it->amount -= amount;
+
+	if (it->amount == 0) {
+		memset(it, 0, sizeof(struct item));
+		clif->updatestorageamount(sd, VECTOR_LENGTH(sd->storage), MAX_STORAGE);
```
MishimaHaruna: This isn't reporting the correct amount of items to the player (because by deleting an item in the middle, you can't rely on VECTOR_LENGTH() anymore. There needs to be a flag (in the sd) that counts the deleted items until the storage is closed and the vector is compacted...

src/map/storage.c
```diff
@@ -356,12 +380,24 @@ int storage_storagegettocart(struct map_session_data* sd, int index, int amount)
...
 
-	if( map->save_settings&4 )
-		chrif->save(sd,0); //Invokes the storage saving as well.
+	if (map->save_settings & 4)
+		chrif->save(sd, 0); //Invokes the storage saving as well.
+
+	/* Erase deleted account storage items from memory
+	 * and resize the vector. */
+	for (i = 0; i < VECTOR_LENGTH(sd->storage); i++) {
+		if (VECTOR_INDEX(sd->storage, i).nameid == 0) {
```
MishimaHaruna: This would skip adjacent empty slots. You need to `i--` whenever you erase an entry (or if you prefer, use a while loop and only do `i++` when you haven't deleted anything)

PR #1686, opened 2017-04-15
Implementation of RoDEX

guilherme-gm:

### Pull Request Prelude

### Changes Proposed

Implementation of RoDEX system (new mail system). Works with clients dated 2015-05-13 and above.
Adds `rodex_sendmail` and `rodex_sendmail2` script commands (see doc for information)

**Affected Branches:**  Master

[//]: # (Master? Slave?)

**Issues addressed:**
~~

~~

A big thanks to:
@hemagx - Base code
@Asheraf  - some official info
@dastgir - for giving me some directions when I was lost
Brought by Ragnarok United (http://board.ragnarokunited.com)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1686)

src/map/clif.c
```diff
@@ -19097,6 +19098,381 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+	sPacket->PacketType = rodexgetitem;
+	sPacket->MailID = rPacket->MailID;
+	sPacket->opentype = rPacket->opentype;
+	sPacket->result = result;
+	WFIFOSET(fd, sizeof(*sPacket));
+}
+
+void clif_rodex_icon(int fd, bool show)
+{
+	WFIFOHEAD(fd, 3);
```
4144: server must send packets only if compiled with supported version. Need add #ifdef

src/map/clif.c
```diff
@@ -19097,6 +19098,381 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+}
+
+void clif_parse_rodex_add_item(int fd, struct map_session_data *sd)
+{
+	struct PACKET_ZC_ADD_ITEM_TO_MAIL *packet;
+	const struct PACKET_CZ_ADD_ITEM_TO_MAIL *rPacket = RFIFOP(fd, 0);
+	int16 idx = rPacket->index - 2;
+	int j;
+
+	WFIFOHEAD(fd, sizeof(*packet));
```
4144: each packet must be sent from own function. This mean extract into other function sending packet

src/map/clif.c
```diff
@@ -19097,6 +19098,381 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+void clif_parse_rodex_remove_item(int fd, struct map_session_data *sd)
+{
+	const struct PACKET_CZ_REQ_REMOVE_ITEM_MAIL *rPacket = RFIFOP(fd, 0);
+	struct PACKET_ZC_ACK_REMOVE_ITEM_MAIL *packet;
+	int result;
+	int16 idx = rPacket->index - 2;
+
+	result = rodex->remove_item(sd, idx, (int16)rPacket->cnt);
+
+	WFIFOHEAD(fd, sizeof(*packet));
```
4144: each packet must be sent from own function

src/map/clif.c
```diff
@@ -19097,6 +19098,381 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+	struct mail_item *item;
+	int i, body_len, size;
+
+	if (msg == NULL)
+		return;
+
+	body_len = (int)strlen(msg->body) + 1;
+	size = sizeof(*sPacket);
+
+	WFIFOHEAD(fd, sizeof(*sPacket) + body_len + (sizeof(*item) * RODEX_MAX_ITEM));
```
4144: packets must be sent from own functions. extact into new function

src/map/clif.c
```diff
@@ -19097,6 +19098,381 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
 
 /*==========================================
+ * RoDEX
+ *------------------------------------------*/
+
+void clif_parse_rodex_open_write_mail(int fd, struct map_session_data *sd)
+{
+	const struct PACKET_CZ_REQ_OPEN_WRITE_MAIL *rPacket = RFIFOP(fd, 0);
+	struct PACKET_ZC_ACK_OPEN_WRITE_MAIL *sPacket = NULL;
+	WFIFOHEAD(fd, sizeof(*sPacket));
```
4144: packets must be send from own functions and not from parse. extract into new function

4144: server must send packets only if compiled with supported version. Need add #ifdef

src/map/clif.c
4144: server must send packets only if compiled with supported version. Need add #ifdef

src/map/clif.c
4144: server must send packets only if compiled with supported version. Need add #ifdef

src/map/clif.c
```diff
@@ -19097,6 +19098,381 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+
+void clif_rodex_checkname_result(struct map_session_data *sd, int char_id, short class_, int base_level, char name[NAME_LENGTH])
+{
+	struct PACKET_ZC_CHECKNAME *sPacket;
+	int fd;
+
+	nullpo_retv(sd);
+
+	fd = sd->fd;
+	WFIFOHEAD(fd, sizeof(*sPacket));
```
4144: server must send packets only if compiled with supported version. Need add #ifdef

src/map/clif.c
```diff
@@ -19097,6 +19098,381 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+	if (result != RODEX_SEND_MAIL_SUCCESS)
+		clif->rodex_send_mail_result(fd, sd, result);
+	rodex->clean(sd, 1);
+}
+
+void clif_rodex_send_mail_result(int fd, struct map_session_data *sd, int8 result)
+{
+	struct PACKET_ZC_WRITE_MAIL_RESULT *sPacket;
+
+	WFIFOHEAD(fd, sizeof(*sPacket));
```
4144: server must send packets only if compiled with supported version. Need add #ifdef

src/map/clif.c
```diff
@@ -19097,6 +19098,381 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+	WFIFOSET(fd, sizeof(*sPacket));
+}
+
+void clif_rodex_send_maillist(int fd, struct map_session_data *sd, int8 open_type, int64 page_start)
+{
+	struct PACKET_ZC_MAIL_LIST *packet;
+	struct maillistinfo *inner;
+	int16 size = sizeof(*packet);
+	int8 count = 0;
+	WFIFOHEAD(fd, sizeof(*packet) + (sizeof(*inner) + RODEX_TITLE_LENGTH) * 7);
```
4144: server must send packets only if compiled with supported version. Need add #ifdef

src/map/clif.c
```diff
@@ -19097,6 +19098,381 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+	WFIFOSET(fd, size);
+}
+
+void clif_rodex_send_refresh(int fd, struct map_session_data *sd, int8 open_type, int count)
+{
+	struct PACKET_ZC_MAIL_LIST *packet;
+	struct maillistinfo *inner;
+	int16 size = sizeof(*packet);
+	int i, j;
+	WFIFOHEAD(fd, sizeof(*packet) + (sizeof(*inner) + RODEX_TITLE_LENGTH) * 7);
```
4144: server must send packets only if compiled with supported version. Need add #ifdef

src/map/clif.c
4144: server must send packets only if compiled with supported version. Need add #ifdef

src/map/clif.c
```diff
@@ -19097,6 +19098,381 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+
+void clif_rodex_delete_mail(int fd, struct map_session_data *sd)
+{
+	const struct PACKET_CZ_REQ_DELETE_MAIL *rPacket = RFIFOP(fd, 0);
+	struct PACKET_ZC_ACK_DELETE_MAIL *sPacket;
+
+	if (rodex->delete_mail(sd, rPacket->MailID) == false)
+		return;
+
+	WFIFOHEAD(fd, sizeof(*sPacket));
```
4144: server must send packets only if compiled with supported version. Need add #ifdef

src/map/clif.c
```diff
@@ -19097,6 +19098,381 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+
+void clif_rodex_request_zeny(int fd, struct map_session_data *sd)
+{
+	const struct PACKET_CZ_REQ_ZENY_FROM_MAIL *rPacket = RFIFOP(fd, 0);
+	struct PACKET_ZC_ACK_ZENY_FROM_MAIL *sPacket;
+	int8 result;
+
+	result = rodex->get_zeny(sd, rPacket->MailID);
+
+	WFIFOHEAD(fd, sizeof(*sPacket));
```
4144: server must send packets only if compiled with supported version. Need add #ifdef

src/map/clif.c
```diff
@@ -19097,6 +19098,381 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+
+void clif_rodex_request_items(int fd, struct map_session_data *sd)
+{
+	const struct PACKET_CZ_REQ_ITEM_FROM_MAIL *rPacket = RFIFOP(fd, 0);
+	struct PACKET_ZC_ACK_ITEM_FROM_MAIL *sPacket;
+	int8 result;
+
+	result = rodex->get_items(sd, rPacket->MailID);
+
+	WFIFOHEAD(fd, sizeof(*sPacket));
```
4144: server must send packets only if compiled with supported version. Need add #ifdef

src/map/rodex.c
```diff
@@ -0,0 +1,531 @@
...
+
+/// Request if character with given name exists and returns information about him
+/// @param sd : The player who's requesting
+/// @param name : The name of the character to check
+/// @param base_level : Reference to return the character base level, if he exists
+/// @param char_id : Reference to return the character id, if he exists
+/// @param class : Reference to return the character class id, if he exists
+bool rodex_check_player(struct map_session_data *sd, const char *name, int *base_level, int *char_id, short *class)
+{
+	nullpo_retr(false, sd);
```
4144: here both nullpo useless

src/map/packets.h
```diff
@@ -3731,6 +3731,23 @@ packet(0x96e,-1,clif->ackmergeitems);
 	packet(0x0923,36,clif->pStoragePassword,0);
 	packet(0x0a27,8); // ZC_RECOVERY2
 	packet(0x09f7,75); // ZC_PROPERTY_HOMUN_2
+	packet(0x09E8,11,clif->rodex_open_mailbox); // CZ_OPEN_MAILBOX
```
guilherme-gm: Thanks for the reference, but that's weird, 0x0a08 is on 2014-04 client but it doesn't have RoDEX and its length is really different (8 in this date compared to 26 in newer ones), what should I do in these cases?

4144: Search packet with correct size. Often size changed and this mean packet used for some thing else. You need only packets with size like in you pr

src/map/intif.c
```diff
@@ -2270,6 +2271,236 @@ void intif_parse_Itembound_ack(int fd) {
...
+	int target_char_id = RFIFOL(fd, 6);
+	short target_class = RFIFOW(fd, 10);
+	int target_level = RFIFOL(fd, 12);
+	char name[NAME_LENGTH];
+
+	safestrncpy(name, RFIFOP(inter_fd, 16), NAME_LENGTH);
+
+	if (reqchar_id <= 0)
+		return;
+	else
```
Smokexyz: There's no need for this `else`.

src/map/rodex.c
```diff
@@ -0,0 +1,531 @@
...
+// Maximun number of messages that can be sent in one day
+#define DAILY_MAX_MAILS 100
+
+struct rodex_interface rodex_s;
+struct rodex_interface *rodex;
+
+/// Checks if RoDEX System is enabled in the server
+bool rodex_isenabled(void)
+{
+	if (battle_config.feature_rodex != true)
```
Smokexyz: `feature_rodex` is an integer type, so checking it against a boolean would be wrong.

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+{
+	int i, count = 0;
+	struct rodex_message msg = { 0 };
+	struct SqlStmt *stmt = SQL->StmtMalloc(inter->sql_handle);
+	struct SqlStmt *stmt_items = SQL->StmtMalloc(inter->sql_handle);
+
+	nullpo_retr(false, stmt);
+	nullpo_retr(false, stmt_items);
+
+	nullpo_retr(-1, mails);
```
MishimaHaruna: This leaks memory by not freeing the multiple SqlStmt that were allocated. This happens in many other places through this function

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+
+	switch (opentype) {
+	case RODEX_OPENTYPE_MAIL:
+		if (SQL_ERROR == SQL->StmtPrepare(stmt,
+			"SELECT `id`, `sender_name`, `sender_id`, `receiver_name`, `receiver_id`, `receiver_accountid`,"
+			"`title`, `body`, `zeny`, `type`, `is_readed`, `send_date`, `expire_date`, `weight`"
+			"FROM `rodex_mail` WHERE `expire_date` > '%d' AND `receiver_id` = '%d' AND `id` > '%"PRId64"'"
```
MishimaHaruna: This query uses `expire_date` (which isn't part of an index) in the WHERE condition.

sql-files/main.sql
```diff
@@ -916,3 +918,53 @@ CREATE TABLE IF NOT EXISTS `storage` (
...
+  `opt_val4` SMALLINT(5) NOT NULL DEFAULT '0',
+	`expire_time` INT(11) UNSIGNED NOT NULL DEFAULT '0',
+	`bound` TINYINT(1) UNSIGNED NOT NULL DEFAULT '0',
+	`unique_id` BIGINT(20) UNSIGNED NOT NULL DEFAULT '0',
+	PRIMARY KEY (`id`),
+  KEY `mail_id` (`mail_id`)
+) ENGINE=InnoDB;
+
+CREATE TABLE IF NOT EXISTS `rodex_mail` (
+	`id` BIGINT(20) NOT NULL AUTO_INCREMENT,
```
MishimaHaruna: Can you rename this to `mail_id`, for consistency with the other table (we usually tend to re-use the column names to avoid confusion)

sql-files/main.sql
```diff
@@ -916,3 +918,53 @@ CREATE TABLE IF NOT EXISTS `storage` (
...
+	`sender_name` VARCHAR(30) NOT NULL COLLATE 'utf8_unicode_ci',
+	`sender_id` INT(11) NOT NULL,
+	`receiver_name` VARCHAR(30) NOT NULL COLLATE 'utf8_unicode_ci',
+	`receiver_id` INT(11) NOT NULL,
+  `receiver_accountid` INT(11) NOT NULL,
+	`title` VARCHAR(50) NOT NULL COLLATE 'utf8_unicode_ci',
```
MishimaHaruna: I think the character set shouldn't be a column property, but rather a database (or at most, table) property, unless the client requires this to be utf8_unicode_ci (which I'm not aware of)

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+			return -1;
+		}
+		break;
+
+	case RODEX_OPENTYPE_ACCOUNT:
+		if (SQL_ERROR == SQL->StmtPrepare(stmt,
+			"SELECT `id`, `sender_name`, `sender_id`, `receiver_name`, `receiver_id`, `receiver_accountid`,"
+			"`title`, `body`, `zeny`, `type`, `is_readed`, `send_date`, `expire_date`, `weight`"
+			"FROM `rodex_mail` WHERE "
+			"`expire_date` > '%d' AND `receiver_accountid` = '%d' AND `id` > '%"PRId64"'"
```
MishimaHaruna: Here as well, `expire_date` isn't an index

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+			SQL->StmtFree(stmt);
+			return -1;
+		}
+		break;
+
+	case RODEX_OPENTYPE_ACCOUNT:
+		if (SQL_ERROR == SQL->StmtPrepare(stmt,
+			"SELECT `id`, `sender_name`, `sender_id`, `receiver_name`, `receiver_id`, `receiver_accountid`,"
+			"`title`, `body`, `zeny`, `type`, `is_readed`, `send_date`, `expire_date`, `weight`"
+			"FROM `rodex_mail` WHERE "
```
MishimaHaruna: `rodex_mail` is hardcoded here, while I see it being configurable elsewhere

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+			SQL->StmtFree(stmt);
+			return -1;
+		}
+		break;
+
+	case RODEX_OPENTYPE_RETURN:
+		if (SQL_ERROR == SQL->StmtPrepare(stmt,
+			"SELECT `id`, `sender_name`, `sender_id`, `receiver_name`, `receiver_id`, `receiver_accountid`,"
+			"`title`, `body`, `zeny`, `type`, `is_readed`, `send_date`, `expire_date`, `weight`"
+			"FROM `rodex_mail` WHERE (`sender_id` = '%d' AND `expire_date` <= '%d' AND `send_date` + '%d' > '%d' AND `id` > '%"PRId64"')"
```
MishimaHaruna: `sender_id`, `expire_date`, `send_date` aren't indexes

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+		msg.type &= ~MAIL_TYPE_NPC;
+#endif
+
+		++count;
+		VECTOR_ENSURE(*mails, 1, 1);
+		VECTOR_PUSH(*mails, msg);
+		memset(&msg, 0, sizeof(struct rodex_message));
+	}
+
+	SQL->FreeResult(inter->sql_handle);
```
MishimaHaruna: The query was (actually, the queries were) executed using prepared statements. This means you don't need to `->FreeResult()` but rather `->StmtFreeResult()`.

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+
+	nullpo_retr(-1, mails);
+
+	switch (opentype) {
+	case RODEX_OPENTYPE_MAIL:
+		// Prepares the statement to retrieve first page of mail box
+		if (SQL_ERROR == SQL->StmtPrepare(stmt,
+			"SELECT `id`, `sender_name`, `sender_id`, `receiver_name`, `receiver_id`, `receiver_accountid`,"
+			"`title`, `body`, `zeny`, `type`, `is_readed`, `send_date`, `expire_date`, `weight`"
+			"FROM `rodex_mail` WHERE `expire_date` > '%d' AND `receiver_id` = '%d'"
```
MishimaHaruna: `expire_date` isn't an index in `rodex_mail`

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+		}
+		break;
+
+	case RODEX_OPENTYPE_ACCOUNT:
+		// Prepares the statement to retrieve first page of account box
+		if (SQL_ERROR == SQL->StmtPrepare(stmt,
+			"SELECT `id`, `sender_name`, `sender_id`, `receiver_name`, `receiver_id`, `receiver_accountid`,"
+			"`title`, `body`, `zeny`, `type`, `is_readed`, `send_date`, `expire_date`, `weight`"
+			"FROM `rodex_mail` WHERE "
+			"`expire_date` > '%d' AND `receiver_accountid` = '%d'"
```
MishimaHaruna: `expire_date` isn't an index

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+			return -1;
+		}
+		break;
+
+	case RODEX_OPENTYPE_RETURN:
+		// Prepares the statement to retrieve first page of returned box
+		if (SQL_ERROR == SQL->StmtPrepare(stmt,
+			"SELECT `id`, `sender_name`, `sender_id`, `receiver_name`, `receiver_id`, `receiver_accountid`,"
+			"`title`, `body`, `zeny`, `type`, `is_readed`, `send_date`, `expire_date`, `weight`"
+			"FROM `rodex_mail` WHERE (`sender_id` = '%d' AND `expire_date` <= '%d' AND `send_date` + '%d' > '%d')"
```
MishimaHaruna: `sender_id`, `expire_date`, `send_date` aren't indexes

Jedzkie:
New update from kRO:

a

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+}
+
+/*==========================================
+* Checks if there are new mails
+*------------------------------------------*/
+void mapif_rodex_sendhasnew(int fd, int char_id, bool has_new)
+{
+	WFIFOHEAD(fd, 7);
+	WFIFOW(fd, 0) = 0x3897;
+	WFIFOL(fd, 2) = char_id;
```
guilherme-gm: can you show me an example of how to do this assertion? thanks.

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+
+void mapif_parse_rodex_send(int fd)
+{
+	struct rodex_message msg;
+
+	if(RFIFOW(fd,2) != 4 + sizeof(struct rodex_message))
+		return;
+
+	memcpy(&msg, RFIFOP(fd,4), sizeof(struct rodex_message));
+	if(msg.receiver_id > 0 || msg.receiver_accountid > 0)
```
guilherme-gm: in normal use it shouldn't be, in case it happens, msg.id will stay in 0 and `false` will be sent as result to whoever made this request

src/char/int_rodex.c
Smokexyz: This would be for example - `Assert_retv(char_id > 0);`

src/char/int_rodex.c
guilherme-gm: this function is called when a new message is sent to a player, at this moment it doesn't have an ID (it's not in the databse yet), after this check it saves the message in the database (i.e. insert in the db), returning the message id. and inter_rodex->savemessage does a saving in the database

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+				msg.items[i].item = it;
+				msg.items_count++;
+			}
+		}
+
+		if (msg.items_count <= 0) {
+			msg.type &= ~MAIL_TYPE_ITEM;
+		}
+
+		if (msg.zeny <= 0) {
```
guilherme-gm: yeah, it shouldn't be negative, but I keep the variable signed right? I read in another PR that we should avoid unsigned vars and in this case, Zeny is also signed in map-server.

src/char/int_rodex.c
Smokexyz: Ah that line 557 totally escaped me due to the comments.

src/char/int_rodex.c
Smokexyz: Yes please ignore this comment. We're apparently not using `unsigned` for anything other than bitmasks.

guilherme-gm:
Thanks for the reviews, I just pushed the fixes, I hope I got everything you pointed out

Smokexyz:
@guilherme-gm There are still some pending un-checked comments -
eg. https://github.com/HerculesWS/Hercules/pull/1686#discussion-diff-111676875R555
If you can fix them it'll be nice. If not, it's totally okay for me... thanks for working on and releasing this awesome feature. 👍  Hope to see more.

src/map/date.c
```diff
@@ -77,6 +77,15 @@ int date_get_sec(void)
 	return lt->tm_sec;
 }
 
+// Returns YYYYMMDD of now
+int date_get_date(void)
+{
+	return
+		date_get_year() * 10000 +
```
4144: calling 3 times date functions here too slow, each this function will call time(NULL)
better call in once in date_get_date and calculate date from it

src/map/packets.h
```diff
@@ -2222,7 +2222,7 @@ packet(0x96e,-1,clif->ackmergeitems);
 	packet(0x0436,4,clif->pDull); // CZ_GANGSI_RANK
 #endif
 
-// 2012-09-25aRagexe
+//2013-03-20Ragexe (Judas + Yommy)
```
4144: wrong comment. here really 2012-09-25aRagexe

src/map/packets.h
```diff
@@ -2590,9 +2590,9 @@ packet(0x96e,-1,clif->ackmergeitems);
 	packet(0x0436,4,clif->pDull); // CZ_GANGSI_RANK
 #endif
 
-// 2013-04-17aRagexe
```
4144: dont remove version comments

src/map/packets.h
```diff
@@ -2602,7 +2602,7 @@ packet(0x96e,-1,clif->ackmergeitems);
 // changed packet sizes
 #endif
 
-// 2013-04-24aRagexe
+//2013-08-07Ragexe (Shakto)
```
4144: wrong comment

src/map/packets.h
```diff
@@ -2697,7 +2697,7 @@ packet(0x96e,-1,clif->ackmergeitems);
 	packet(0x0436,4,clif->pDull); // CZ_GANGSI_RANK
 #endif
 
-// 2013-08-07aRagexe
+//2013-08-14aRagexe - Themon
```
4144: wrong comment

src/map/packets.h
```diff
@@ -2741,7 +2741,7 @@ packet(0x96e,-1,clif->ackmergeitems);
 	packet(0x0896,8,clif->pDull); // CZ_JOIN_BATTLE_FIELD
 #endif
 
-// 2013-08-14aRagexe
+// 2013-10-30aRagexe
```
4144: wrong comment

src/map/packets.h
```diff
@@ -2895,9 +2895,12 @@ packet(0x96e,-1,clif->ackmergeitems);
 	packet(0x085C,36,clif->pStoragePassword,0);
 	packet(0x0363,8,clif->pDull); // CZ_JOIN_BATTLE_FIELD
 	packet(0x087b,4,clif->pDull); // CZ_GANGSI_RANK
+	packet(0x09f5,11,clif->pRodexDeleteMail); // CZ_REQ_DELETE_RODEX
```
4144: this packets already added in 20131218
you should change only clif->pDull to correct values

4144: all 3 packets

src/map/packets.h
```diff
@@ -2895,9 +2895,12 @@ packet(0x96e,-1,clif->ackmergeitems);
...
+	packet(0x09ee,11,clif->pRodexNextMaillist); // CZ_REQ_NEXT_RODEX
+	packet(0x09ef,11,clif->pRodexRefreshMaillist); // CZ_REQ_REFRESH_RODEX
 #endif
 
-// 2013-12-18bRagexe
+// 2013-12-23cRagexe - Yommy
```
4144: wrong comment

src/map/packets.h
```diff
@@ -2942,9 +2945,10 @@ packet(0x96e,-1,clif->ackmergeitems);
 	packet(0x08A4,36,clif->pStoragePassword,0);
 	packet(0x0363,8,clif->pDull); // CZ_JOIN_BATTLE_FIELD
 	packet(0x0436,4,clif->pDull); // CZ_GANGSI_RANK
+	packet(0x09ea,11,clif->pRodexReadMail); // CZ_REQ_READ_RODEX
 #endif
 
-// 2013-12-23bRagexe
+// 2013-12-30aRagexe - Yommy
```
4144: wrong comment

4144: wrong comment

src/map/packets.h
```diff
@@ -2983,15 +2987,16 @@ packet(0x96e,-1,clif->ackmergeitems);
 	packet(0x091D,36,clif->pStoragePassword,0);
 	packet(0x087e,4,clif->pDull); // CZ_GANGSI_RANK
 	packet(0x093e,8,clif->pDull); // CZ_JOIN_BATTLE_FIELD
+	packet(0x09ec,-1,clif->pRodexSendMail); // CZ_REQ_SEND_RODEX
 #endif
 
-// 2013-12-30aRagexe
+// 2014 Packet Data
```
4144: wrong comment

src/map/packets.h
```diff
@@ -2983,15 +2987,16 @@ packet(0x96e,-1,clif->ackmergeitems);
...
 #if PACKETVER >= 20131230
-// new packets
+
 	packet(0x09ec,-1,clif->pDull/*,XXX*/); // CZ_REQ_SEND_RODEX
 	packet(0x09ed,3); // ZC_ACK_SEND_RODEX
 	packet(0x09f7,75); // ZC_PROPERTY_HOMUN_2
-// changed packet sizes
+// 2014-01-15eRagexe
```
4144: wrong comment

MishimaHaruna:
Is this still ongoing @guilherme-gm ? If so, could you please rebase it and resolve the conflicts when you address @4144 's code review? I'll then give a final pass of review and possibly merge

guilherme-gm:
@MishimaHaruna yes, sorry for the delay, the end of the semester at university is taking too much of my time, I'll try to update this until Friday

src/map/clif.c
```diff
@@ -19104,6 +19105,478 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+	rodex->add_item(sd, idx, (int16)rPacket->count);
+}
+
+void clif_rodex_add_item_result(struct map_session_data *sd, int16 idx, int16 amount, int8 result) {
+#if PACKETVER >= 20140416
+	struct PACKET_ZC_ADD_ITEM_TO_MAIL *packet;
+	int fd, j;
+
+	nullpo_retv(sd);
+
```
guilherme-gm: hmm, this is a result function and it can be called with the invalid idx when an invalid idx is received on rodex.c, in this case with an error result code, do I need to check it there too? (you can see what I'm saying in rodex.c at line 86, inside rodex_add_item)

guilherme-gm:
I just pushed the fixes, except for [https://github.com/HerculesWS/Hercules/pull/1686#discussion_r118718219](this onel). @Smokexyz those diffs that you showed seems to be fixed on mine, maybe you read the old version?

Thanks again!

vBrenth:
Hi @guilherme-gm Please give us update.

guilherme-gm:
@vBrenth I just pushed the update a couple hours ago

src/map/clif.c
```diff
@@ -19114,6 +19115,484 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+void clif_parse_rodex_read_mail(int fd, struct map_session_data *sd) __attribute__((nonnull(2)));
+void clif_parse_rodex_read_mail(int fd, struct map_session_data *sd)
+{
+	const struct PACKET_CZ_REQ_READ_MAIL *rPacket = RFIFOP(fd, 0);
+
+	rodex->read_mail(sd, rPacket->MailID);
+}
+
+void clif_rodex_read_mail(struct map_session_data *sd, int8 opentype, struct rodex_message *msg) {
```
4144: "{" should be on new line

src/map/date.c
```diff
@@ -77,6 +77,20 @@ int date_get_sec(void)
...
+// Returns YYYYMMDD of now
+int date_get_date(void)
+{
+	time_t t;
+	struct tm * lt;
+	t = time(NULL);
+	lt = localtime(&t);
+
+	return
+		(lt->tm_year+1900) * 10000 +
```
4144: need space before and after plus sign

src/map/intif.c
```diff
@@ -2399,6 +2400,217 @@ void intif_parse_Itembound_ack(int fd) {
...
+	WFIFOL(inter_fd, 6) = account_id;
+	WFIFOL(inter_fd, 10) = flag;
+	WFIFOB(inter_fd, 11) = opentype;
+	WFIFOQ(inter_fd, 12) = mail_id;
+	WFIFOSET(inter_fd, 20);
+
+	return 0;
+}
+
+void intif_parse_RequestRodexOpenInbox(int fd) {
```
4144: "{" should be on new line

src/map/packets.h
```diff
@@ -2983,12 +2984,13 @@ packet(0x96e,-1,clif->ackmergeitems);
 	packet(0x091D,36,clif->pStoragePassword,0);
 	packet(0x087e,4,clif->pDull); // CZ_GANGSI_RANK
 	packet(0x093e,8,clif->pDull); // CZ_JOIN_BATTLE_FIELD
+	packet(0x09ec,-1,clif->pRodexSendMail); // CZ_REQ_SEND_RODEX
```
4144: here need remove this line. because packet already was added and changed by you some lines below

src/map/rodex.c
```diff
@@ -0,0 +1,581 @@
...
+{
+	if (battle_config.feature_rodex == 1)
+		return true;
+
+	return false;
+}
+
+/// Checks and refreshes the user daily number of Stamps
+/// @param sd : The player who's being checked
+void rodex_refresh_stamps(struct map_session_data *sd) {
```
4144: "{" should be in new line

src/map/rodex.c
```diff
@@ -0,0 +1,581 @@
...
+/// @param body : Mail message
+/// @param title : Mail Title
+/// @param zeny : Amount of zeny attached
+/// Returns result code:
+/// 	RODEX_SEND_MAIL_SUCCESS = 0,
+///		RODEX_SEND_MAIL_FATAL_ERROR = 1,
+///		RODEX_SEND_MAIL_COUNT_ERROR = 2,
+///		RODEX_SEND_MAIL_ITEM_ERROR = 3,
+///		RODEX_SEND_MAIL_RECEIVER_ERROR = 4
+int rodex_send_mail(struct map_session_data *sd, const char *receiver_name, const char *body, const char *title, int64 zeny)
```
4144: probably need check zeny to negative value?

src/map/rodex.c
```diff
@@ -0,0 +1,581 @@
...
+	}
+
+	if (total_zeny > sd->status.zeny || total_zeny < 0) {
+		rodex->clean(sd, 1);
+		return RODEX_SEND_MAIL_FATAL_ERROR;
+	}
+
+	rodex_refresh_stamps(sd);
+
+	if (sd->sc.data[SC_DAILYSENDMAILCNT]->val2 >= DAILY_MAX_MAILS) {
```
4144: need check sd->sc.data[SC_DAILYSENDMAILCNT] for null

src/map/rodex.c
```diff
@@ -0,0 +1,581 @@
...
+		rodex->clean(sd, 1);
+		return RODEX_SEND_MAIL_COUNT_ERROR;
+	}
+
+	sc_start2(NULL, &sd->bl, SC_DAILYSENDMAILCNT, 100, sd->sc.data[SC_DAILYSENDMAILCNT]->val1, sd->sc.data[SC_DAILYSENDMAILCNT]->val2 + 1, INFINITE_DURATION);
```
4144: and here

src/map/clif.c
```diff
@@ -19114,6 +19115,484 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+{
+	const struct PACKET_CZ_SEND_MAIL *rPacket = RFIFOP(fd, 0);
+	int8 result;
+
+	if (rPacket->TextcontentsLength + rPacket->Titlelength > rPacket->PacketLength - sizeof(*rPacket))
+		result = RODEX_SEND_MAIL_FATAL_ERROR;
+	else if (rPacket->TextcontentsLength > RODEX_BODY_LENGTH || rPacket->Titlelength > RODEX_TITLE_LENGTH)
+		result = RODEX_SEND_MAIL_FATAL_ERROR;
+	else
+		result = rodex->send_mail(sd, rPacket->receiveName, &rPacket->string[rPacket->Titlelength], rPacket->string, rPacket->zeny);
```
4144: before call rodex->send_mail, need all char* parameters copy with safestrncpy into temporary variables. Because client may send crafted corrupted packet without null char and end of each string.

src/char/int_rodex.c
```diff
@@ -0,0 +1,629 @@
...
+		|| SQL_ERROR == SQL->StmtBindColumn(stmt, 12, SQLDT_INT, &msg.expire_date, 0, NULL, NULL)
+		|| SQL_ERROR == SQL->StmtBindColumn(stmt, 13, SQLDT_INT, &msg.weight, 0, NULL, NULL)
+		) {
+		SqlStmt_ShowDebug(stmt);
+		SQL->StmtFree(stmt);
+		return -1;
+	}
+
+	// Read mails
+	while (SQL_SUCCESS == SQL->StmtNextRow(stmt)) {
```
guilherme-gm: for future reference, I'm splitting the packets when sending to map (on `mapif_rodex_sendinbox`)

src/map/rodex.c
guilherme-gm: @4144  `rodex_refresh_stamps` one line before will init it if it's null, I think there's no way for it to become null one line after, do I really need to recheck?

4144: i think better check. because init happened in other function. and from time it can be changed

guilherme-gm:
@4144 Fixed, Thanks!

vBrenth:
@guilherme-gm Please do the request change! I am really rooting for this one.

guilherme-gm:
@vBrenth Did I miss something? because I think I finished fixing everything pointed out until now

4144:
@guilherme-gm if you see to comments log, you will see what at least some issues not fixed.

guilherme-gm:
@4144 I checked both conversation and "Files Changed" tab, read all the comments and checked the code, I only found one note of a missing space on an `if`.
I think these comments were fixed already, github just didn't collapsed them because the fix wasn't in the same line or other things changed. (e.g.: "Document return value" -> the function doesn't have a return value anymore)

src/map/clif.c
```diff
@@ -19114,6 +19115,498 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+
+void clif_parse_rodex_checkname(int fd, struct map_session_data *sd) __attribute__((nonnull(2)));
+void clif_parse_rodex_checkname(int fd, struct map_session_data *sd)
+{
+	const struct PACKET_CZ_CHECKNAME *rPacket = RFIFOP(fd, 0);
+	int char_id = 0, base_level = 0;
+	short class = 0;
+	char name[24];
+	
+	safestrncpy(name, rPacket->Name, 24);
```
4144: probably here need use constant NAME_LENGTH?

src/map/packets.h
```diff
@@ -2942,13 +2942,14 @@ packet(0x96e,-1,clif->ackmergeitems);
 	packet(0x08A4,36,clif->pStoragePassword,0);
 	packet(0x0363,8,clif->pDull); // CZ_JOIN_BATTLE_FIELD
 	packet(0x0436,4,clif->pDull); // CZ_GANGSI_RANK
+	packet(0x09ea,11,clif->pRodexReadMail); // CZ_REQ_READ_RODEX
```
4144: Was in previous review already
This line in duplicated for line 2952.
This is non shuffle packets and cant be added into shuffle packets.
In line 2952 it added already

guilherme-gm:
@4144 Fixed. Thanks for the review

src/char/int_rodex.c
```diff
@@ -0,0 +1,535 @@
...
+	has_new = inter_rodex->hasnew(char_id, account_id);
+	mapif->rodex_sendhasnew(fd, char_id, has_new);
+}
+
+/*==========================================
+ * Update/Delete mail
+ *------------------------------------------*/
+void mapif_parse_rodex_updatemail(int fd)
+{
+	int64 mail_id = RFIFOL(fd,2);
```
4144: missing space after ","

src/map/clif.c
```diff
@@ -19114,6 +19115,498 @@ unsigned short clif_parse_cmd_optional( int fd, struct map_session_data *sd ) {
...
+}
+
+void clif_rodex_remove_item_result(struct map_session_data *sd, int16 idx, int16 amount)
+{
+#if PACKETVER >= 20140521
+	struct PACKET_ZC_ACK_REMOVE_ITEM_MAIL *packet;
+	int fd;
+
+	nullpo_retv(sd);
+	Assert_retv(idx >= 0 || idx < MAX_INVENTORY);
```
4144: Look like wrong condition
Need check: idx >= 0 **&&** idx < MAX_INVENTORY

guilherme-gm:
@4144 fixed

vBrenth:
@guilherme-gm thanks for keeping this pull request active :) i really want to try this out :)

4144:
@guilherme-gm now looks fine, but need also squash all commits into one and rebase.

guilherme-gm:
@4144 okay, thanks! is there a way to squash when I have "merge branch" commits in the middle? I'm not very skilled with git

4144:
i think simply drop merge commits.
But if you cant do this, probably some one who will push your pr will squash for you.
Wait first for @MishimaHaruna review

vBrenth:
@MishimaHaruna Hello, Please review this one :)
@4144 Thank you! you are the best!

I hope someone will squash it* Hahaha even i dont know the thing.

MishimaHaruna:
Reviewed 22 of 50 files at r1, 3 of 19 files at r2, 1 of 11 files at r3, 15 of 18 files at r4, 3 of 7 files at r5, 3 of 7 files at r7, 1 of 3 files at r10, 1 of 1 files at r11, 2 of 2 files at r12.
Review status: all files reviewed at latest revision, 108 unresolved discussions.

---

*[src/char/int_rodex.c, line 56 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KhnRS8k2usvfX0-6jPm-r1-56:-KogaSxwKZZRIybemZKm:b-uv1z5u) ([raw file](https://github.com/herculesws/hercules/blob/c18432e0e5df740819e6f8f394fd2a69a9d7f32b/src/char/int_rodex.c#L56)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

This leaks memory by not freeing the multiple SqlStmt that were allocated. This happens in many other places through this function
</blockquote></details>

(this is still open - the StmtMalloc should be moved to after this nullpo check)

---

*[src/char/int_rodex.c, line 226 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-Koge7XszWy-6EEKGWLD:-Koge7XszWy-6EEKGWLE:b-cpv1tm) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/char/int_rodex.c#L226)):*
> ```C
> 
> /// Checks player name and retrieves some data
> static bool inter_rodex_checkname(char name[NAME_LENGTH], int *target_char_id, short *target_class, int *target_level)
> ```

It's not recommended to use the `array[length]` syntax in function arguments. This would give the false illusion that `sizeof(name)` returns the size of the array (while in fact it returns the size of the pointer, since functions can't take arrays as arguments but only pointers).

This should be changed to:

```
static bool inter_rodex_checkname(const char *name, int *target_char_id, short *target_class, int *target_level)
```

(yes, const too, since it's not meant to be modified)

For reference: https://lkml.org/lkml/2015/9/3/428

---

*[src/char/int_rodex.c, line 275 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogcYnWltayXyuC0Oux:-KogcYnWltayXyuC0Ouy:b-4q87nr) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/char/int_rodex.c#L275)):*
> ```C
> 		"('%s', '%d', '%s', '%d', '%d', '%s', '%s', '%d', '%d', '%d', '%d', '%d', '%d')",
> 		rodex_db, sender_name, msg->sender_id, receiver_name, msg->receiver_id, msg->receiver_accountid,
> 		title, body, (int)msg->zeny, msg->type, msg->is_read == true ? 1 : 0, msg->send_date, msg->expire_date, msg->weight)) {
> ```

Is there any reason to cast this to int here? (what's the point of it being int64, if it's cast to int before saving?)

---

*[src/common/mmo.h, line 25 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-Kogff8ruGJ-DqwvlQhJ:-Kogff8ruGJ-DqwvlQhK:by4tfju) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/common/mmo.h#L25)):*
> ```objc
> 
> #include "config/core.h"
> #include "common/db.h"
> ```

Already included below

---

*[src/map/clif.c, line 19130 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-Kogg7ddsU1A6DttW71S:-Kogg7ddsU1A6DttW71T:ba1x1f2) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/clif.c#L19130)):*
> ```C
> }
> 
> void clif_rodex_open_write_mail(int fd, const char receiver_name[NAME_LENGTH], int8 result)
> ```

`const char *receiver_name` (see comment int int_rodex.c)

---

*[src/map/clif.c, line 19241 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KoghBS3C_rFUL-44NcC:-KoghBS3C_rFUL-44NcD:b-edg5v2) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/clif.c#L19241)):*
> ```C
> }
> 
> void clif_rodex_checkname_result(struct map_session_data *sd, int char_id, short class_, int base_level, char name[NAME_LENGTH])
> ```

`const char *name` (see comment in int_rodex.c)

---

*[src/map/clif.c, line 19282 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KoghTuAt_FPixaDcUNo:-KoghTuAt_FPixaDcUNp:bn8571) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/clif.c#L19282)):*
> ```C
> 		char rname[NAME_LENGTH];
> 		char title[RODEX_TITLE_LENGTH];
> 		char body[RODEX_BODY_LENGTH];
> ```

These should be initialized to empty, so that we won't send garbage from the server memory to the client (which might contain anything, including fragments of passwords or other confidential data).

```
char rname[NAME_LENGTH] = "";
char title[RODEX_TITLE_LENGTH] = "";
char body[RODEX_BODY_LENGTH] = "";
```

---

*[src/map/clif.c, line 19345 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KoghziGHE5JKsZoIVEM:-KoghziGHE5JKsZoIVEN:b-skmecg) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/clif.c#L19345)):*
> ```C
> 			strncpy(inner->SenderName, msg->sender_name, sizeof(msg->sender_name));
> 		}
> 		else {
> ```

nitpick: `} else {` (all on the same line)

---

*[src/map/clif.c, line 19399 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogiA7Z1v5LRxMXSSEV:-KogiA7Z1v5LRxMXSSEW:bjdxo3y) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/clif.c#L19399)):*
> ```C
> 			strncpy(inner->SenderName, msg->sender_name, sizeof(msg->sender_name));
> 		}
> 		else {
> ```

`} else {`

---

*[src/map/clif.c, line 19471 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogiZx8zyz7jMUYh4af:-KogiZx8zyz7jMUYh4ag:b-j5vybv) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/clif.c#L19471)):*
> ```C
> 		item->IsDamaged = (it->attribute & ATTR_BROKEN) != 0 ? 1 : 0;
> 		item->refiningLevel = it->refine;
> 		item->slot.card[0] = it->card[0];
> ```

Like the item options below, the card slots are controlled by `MAX_SLOTS`. This should go in a for() loop.

---

*[src/map/map.h, line 25 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-Kogld5G9eV3LmBaeDYk:-Kogld5G9eV3LmBaeDYl:bak5vdg) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/map.h#L25)):*
> ```objc
> 
> #include "map/atcommand.h"
> #include "map/rodex.h"
> ```

Is there a reason for this `#include`? It's better to avoid unnecessary includes in .h files. If a .c file needs it, then that file should include it.

---

*[src/map/rodex.c, line 141 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogncHW6dQGgTfQmEB5:-KogncHW6dQGgTfQmEB6:b5fidas) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/rodex.c#L141)):*
> ```C
> 	sd->rodex.tmp.weight += sd->inventory_data[idx]->weight * amount;
> 	if (is_stack == false) {
> 		memcpy(&sd->rodex.tmp.items[i].item, &sd->status.inventory[idx], sizeof(sd->status.inventory[0]));
> ```

`sd->rodex.tmp.items[i].item = sd->status.inventory[idx]` (with the benefit of better error messages if the type of one of them changes and becomes incompatible with the other)

---

*[src/map/rodex.c, line 469 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogqrSLW8yu-ktTw7yg:-KogqrSLW8yu-ktTw7yh:b-avz1yt) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/rodex.c#L469)):*
> ```C
> 	}
> 
> 	if (empty_slots < msg->items_count) {
> ```

If the items are stackable, this check can be less strict.

Speaking of stackable items, it should also check that any added items won't exceed the stack limits

---

*[src/map/rodex.c, line 481 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogrCE7cYREzm2LpMy7:-KogrCE7cYREzm2LpMy8:b5pwb3a) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/rodex.c#L481)):*
> ```C
> 		}
> 
> 		if (pc->additem(sd, it, it->amount, LOG_TYPE_MAIL) != 0) {
> ```

If this fails after at least an item was added, there's an inconsistent situation. The added items were correctly cleared from the message, but the message wasn't saved (no intif functions called before returning). This sounds like a recipe for item duplication.

---

*[src/map/rodex.c, line 572 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogrgOgTzq5i720nU4O:-KogrgOgTzq5i720nU4P:bay7smw) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/rodex.c#L572)):*
> ```C
> }
> 
> void rodex_defaults(void)
> ```

Can you add the `rodex_init()` and `rodex_final()` functions, and call them from map.c (even if they don't do anything right now)? We usually define them even if they're currently unused, so that if a plugin needs to hook into them to initialize/finalize any custom data, they're available.

---

*[src/map/script.c, line 23260 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogsxAssV8hIcAN3Ycy:-KogsxAssV8hIcAN3Ycz:b-uslg20) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/script.c#L23260)):*
> ```C
> 		data = script_getdata(st, 7);
> 
> 		if (data_isreference(data)) {
> ```

This is true even if the variable that was passed isn't an array. I'd advise against making functions that receive arrays as arguments, since this is unreliable. You never know if the caller wanted to pass the entire array, or just one item from an array.

How about passing the items in sequence, as separate variables? (item_id_1, amount_1, refine_1, card0_1, card1_1, card2_1, card3_1, item_id_2, amount_2, ...). There could be a variant of this function that only takes id/amount, and one that takes all the other arguments as well. I realize it's ugly, but I have nothing better to suggest

---

*[src/map/script.c, line 23285 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-Kogu-aSxKynTVoc73G7:-Kogu-aSxKynTVoc73G8:bezkejs) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/script.c#L23285)):*
> ```C
> 
> 				for (i = 0; i < item_count; ++i) {
> 					int item_id = (int)(long)script->get_val2(st, reference_uid(arr_id, arr_idx + i), reference_getref(data));
> ```

What's the purpose of two typecasts here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1686)*

vBrenth:
@guilherme-gm 💃

guilherme-gm:
@MishimaHaruna Thanks for the review. about the script command, what do you think of:
`rodex_sendmail(sender, .., item1, amount1, item2, amount2,...)` and `rodex_sendmail2(sender, .., item1, amount1, card1_1, card2_1,..., item2, amount2,...)` ? (and also one `_acc` version of each of them).

mekolat:
@guilherme-gm why have two commands when you could have one? also I think it should take arrays rather than a huge list of arguments, ie something like
```c
rodex_sendmail(<sender>, <receiver>, <item array>, <amount array>, <identify array>, <refine array>, <attribute array>, <card1 array>, <card2 array>, <card3 array>, <card4 array>);
```

guilherme-gm:
@mekolat check the end of Haru's last review. Currently it uses arrays.

vBrenth:
Any updates?

guilherme-gm:
@MishimaHaruna can you give your opinion about the command? Thanks :)

vBrenth:
@MishimaHaruna please review this, so the achievement of @Smokexyz  can progress thank you :'(

MishimaHaruna:
Sorry for the delay in getting back to you @guilherme-gm (processing the recent requests in a FIFO)

> `rodex_sendmail(sender, .., item1, amount1, item2, amount2,...)` and `rodex_sendmail2(sender, .., item1, amount1, card1_1, card2_1,..., item2, amount2,...)` ? (and also one `_acc` version of each of them).

I think the approach is reasonable. It's not especially pretty, but unfortunately I can't think of anything better to suggest with the current limitations we have in the scripting engine.
Please go ahead with that

guilherme-gm:
No problem, thanks Haru! I'll try to update until sunday.

guilherme-gm:
@MishimaHaruna I think I fixed everything that you've pointed. Thanks for the review!

vBrenth:
@MishimaHaruna

guilherme-gm:
Review status: 28 of 48 files reviewed at latest revision, 116 unresolved discussions.

---

*[src/char/int_rodex.c, line 56 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KhnRS8k2usvfX0-6jPm-r1-56:-KqyDdui_oOwWx-BzZNg:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/c18432e0e5df740819e6f8f394fd2a69a9d7f32b/src/char/int_rodex.c#L56)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

(this is still open - the StmtMalloc should be moved to after this nullpo check)
</blockquote></details>

Done.

---

*[src/char/int_rodex.c, line 64 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KhnRS8k2usvfX0-6jPm-r1-64:-KqyE3Rg2Tjaq5XO4soE:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/c18432e0e5df740819e6f8f394fd2a69a9d7f32b/src/char/int_rodex.c#L64)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

`expire_date` isn't an index in `rodex_mail`
</blockquote></details>

Done.

---

*[src/char/int_rodex.c, line 79 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KhnRS8k2usvfX0-6jPm-r1-79:-KqyE4NPJisyTY2-CDsX:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/c18432e0e5df740819e6f8f394fd2a69a9d7f32b/src/char/int_rodex.c#L79)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

`expire_date` isn't an index
</blockquote></details>

Done.

---

*[src/char/int_rodex.c, line 93 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KhnRS8k2usvfX0-6jPm-r1-93:-KqyE4tTGzOE87a4VGF5:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/c18432e0e5df740819e6f8f394fd2a69a9d7f32b/src/char/int_rodex.c#L93)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

`sender_id`, `expire_date`, `send_date` aren't indexes
</blockquote></details>

Done.

---

*[src/char/int_rodex.c, line 186 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KhnRS8k2usvfX0-6jPm-r1-186:-KqyE1H0gPDUVNfDSd2K:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/c18432e0e5df740819e6f8f394fd2a69a9d7f32b/src/char/int_rodex.c#L186)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

The query was (actually, the queries were) executed using prepared statements. This means you don't need to `->FreeResult()` but rather `->StmtFreeResult()`.
</blockquote></details>

Done.

---

*[src/char/int_rodex.c, line 210 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KhnRS8k2usvfX0-6jPm-r1-210:-KqyDpsP9wuw1k81yq7P:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/c18432e0e5df740819e6f8f394fd2a69a9d7f32b/src/char/int_rodex.c#L210)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

This query uses `expire_date` (which isn't part of an index) in the WHERE condition.
</blockquote></details>

Done.

---

*[src/char/int_rodex.c, line 223 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KhnRS8k2usvfX0-6jPm-r1-223:-KqyDxJbSQ9lW1SYBpVm:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/c18432e0e5df740819e6f8f394fd2a69a9d7f32b/src/char/int_rodex.c#L223)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

`rodex_mail` is hardcoded here, while I see it being configurable elsewhere
</blockquote></details>

Done.

---

*[src/char/int_rodex.c, line 224 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KhnRS8k2usvfX0-6jPm-r1-224:-KqyDqrAB6qo-p1qGtL1:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/c18432e0e5df740819e6f8f394fd2a69a9d7f32b/src/char/int_rodex.c#L224)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

Here as well, `expire_date` isn't an index
</blockquote></details>

Done.

---

*[src/char/int_rodex.c, line 237 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KhnRS8k2usvfX0-6jPm-r1-237:-KqyDyWpIxCJDvLr4xlD:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/c18432e0e5df740819e6f8f394fd2a69a9d7f32b/src/char/int_rodex.c#L237)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

`sender_id`, `expire_date`, `send_date` aren't indexes
</blockquote></details>

Done.

---

*[src/char/int_rodex.c, line 226 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-Koge7XszWy-6EEKGWLD:-KqyEbR5-qGY9RPX2L_u:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/char/int_rodex.c#L226)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

It's not recommended to use the `array[length]` syntax in function arguments. This would give the false illusion that `sizeof(name)` returns the size of the array (while in fact it returns the size of the pointer, since functions can't take arrays as arguments but only pointers).

This should be changed to:

```
static bool inter_rodex_checkname(const char *name, int *target_char_id, short *target_class, int *target_level)
```

(yes, const too, since it's not meant to be modified)

For reference: https://lkml.org/lkml/2015/9/3/428
</blockquote></details>

Done.

---

*[src/char/int_rodex.c, line 275 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogcYnWltayXyuC0Oux:-KqyESQnIQ2782YWEpop:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/char/int_rodex.c#L275)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

Is there any reason to cast this to int here? (what's the point of it being int64, if it's cast to int before saving?)
</blockquote></details>

Done.

---

*[src/common/mmo.h, line 25 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-Kogff8ruGJ-DqwvlQhJ:-KqyEcT1JUNHZ8pUj7EC:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/common/mmo.h#L25)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

Already included below
</blockquote></details>

Done.

---

*[src/map/clif.c, line 19130 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-Kogg7ddsU1A6DttW71S:-KqyEg5eQnElzbObt7wq:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/clif.c#L19130)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

`const char *receiver_name` (see comment int int_rodex.c)
</blockquote></details>

Done.

---

*[src/map/clif.c, line 19241 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KoghBS3C_rFUL-44NcC:-KqyEfaus4FaqBPhA7qn:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/clif.c#L19241)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

`const char *name` (see comment in int_rodex.c)
</blockquote></details>

Done.

---

*[src/map/clif.c, line 19282 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KoghTuAt_FPixaDcUNo:-KqyEhp8H2JzCCi_8m0L:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/clif.c#L19282)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

These should be initialized to empty, so that we won't send garbage from the server memory to the client (which might contain anything, including fragments of passwords or other confidential data).

```
char rname[NAME_LENGTH] = "";
char title[RODEX_TITLE_LENGTH] = "";
char body[RODEX_BODY_LENGTH] = "";
```

</blockquote></details>

Done.

---

*[src/map/clif.c, line 19345 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KoghziGHE5JKsZoIVEM:-KqyEj1HQoxw0iMrFpsi:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/clif.c#L19345)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

nitpick: `} else {` (all on the same line)
</blockquote></details>

Done.

---

*[src/map/clif.c, line 19399 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogiA7Z1v5LRxMXSSEV:-KqyEjYIRNSJSvT6P2MK:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/clif.c#L19399)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

`} else {`
</blockquote></details>

Done.

---

*[src/map/clif.c, line 19471 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogiZx8zyz7jMUYh4af:-KqyEkJAiUcDbb3ZjNWM:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/clif.c#L19471)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

Like the item options below, the card slots are controlled by `MAX_SLOTS`. This should go in a for() loop.
</blockquote></details>

Done.

---

*[src/map/rodex.c, line 141 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogncHW6dQGgTfQmEB5:-KqJf7_EgIdjfLQU0mNK:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/rodex.c#L141)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

`sd->rodex.tmp.items[i].item = sd->status.inventory[idx]` (with the benefit of better error messages if the type of one of them changes and becomes incompatible with the other)
</blockquote></details>

Done.

---

*[src/map/rodex.c, line 469 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogqrSLW8yu-ktTw7yg:-KqyEqBwic-T5xXS3bmc:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/rodex.c#L469)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

If the items are stackable, this check can be less strict.

Speaking of stackable items, it should also check that any added items won't exceed the stack limits
</blockquote></details>

Done.

---

*[src/map/rodex.c, line 481 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogrCE7cYREzm2LpMy7:-KqyEx1c7E94ayry1vKH:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/rodex.c#L481)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

If this fails after at least an item was added, there's an inconsistent situation. The added items were correctly cleared from the message, but the message wasn't saved (no intif functions called before returning). This sounds like a recipe for item duplication.
</blockquote></details>

Done.

---

*[src/map/rodex.c, line 572 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogrgOgTzq5i720nU4O:-KqyEzvAsHl1ILCaggpM:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/rodex.c#L572)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

Can you add the `rodex_init()` and `rodex_final()` functions, and call them from map.c (even if they don't do anything right now)? We usually define them even if they're currently unused, so that if a plugin needs to hook into them to initialize/finalize any custom data, they're available.
</blockquote></details>

Done.

---

*[src/map/script.c, line 23260 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KogsxAssV8hIcAN3Ycy:-KqyF0HNYnIOSGWLNXDa:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/script.c#L23260)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

This is true even if the variable that was passed isn't an array. I'd advise against making functions that receive arrays as arguments, since this is unreliable. You never know if the caller wanted to pass the entire array, or just one item from an array.

How about passing the items in sequence, as separate variables? (item_id_1, amount_1, refine_1, card0_1, card1_1, card2_1, card3_1, item_id_2, amount_2, ...). There could be a variant of this function that only takes id/amount, and one that takes all the other arguments as well. I realize it's ugly, but I have nothing better to suggest
</blockquote></details>

Done.

---

*[src/map/script.c, line 23285 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-Kogu-aSxKynTVoc73G7:-KqyF0sUREjhfj4Ib78Q:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/script.c#L23285)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

What's the purpose of two typecasts here?
</blockquote></details>

Done.

---

*[sql-files/upgrades/2017-04-15--13-00.sql, line 21 at r1](https://reviewable.io:443/reviews/herculesws/hercules/1686#-KhnRS8j4xwjQ33CLPFQ-r1-21:-KqyF2MOIcUqBbVQKd59:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/c18432e0e5df740819e6f8f394fd2a69a9d7f32b/sql-files/upgrades/2017-04-15--13-00.sql#L21)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

This upgrade isn't really necessary, please include it in the other (at the time of merging this pull request, the item options are already implemented)
</blockquote></details>

Done.

---

*[src/map/map.h, line 25 at r12](https://reviewable.io:443/reviews/herculesws/hercules/1686#-Kogld5G9eV3LmBaeDYk:-KqyF4itnefMLNslhrQr:b-896fix) ([raw file](https://github.com/herculesws/hercules/blob/7a93850d2ebe392222f2357079da4e56e22a33c9/src/map/map.h#L25)):*
<details><summary><i>Previously, MishimaHaruna (Haru) wrote…</i></summary><blockquote>

Is there a reason for this `#include`? It's better to avoid unnecessary includes in .h files. If a .c file needs it, then that file should include it.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1686)*

MishimaHaruna:
Reviewed 1 of 18 files at r4, 21 of 21 files at r13.
Review status: all files reviewed at latest revision, 90 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1686)*

MishimaHaruna:
Thank you very much @guilherme-gm and everyone else who contributed. I believe this is finally ready to merge now!
Since this is a pretty big (and amazing) new feature, in the unfortunate case that something slipped past us, we'll fix it in a follow-up.

PR #1690, opened 2017-04-18
Add missing non shuffle packets for 2013 ragexe client

4144:
### Pull Request Prelude

### Changes Proposed

Add missing non shuffle packets for 2013 ragexe client.

Move some packets added in wrong versions to past versions.

Non destructive changes for existing implemented packets. In other words it change nothing for already implemented packets, but fix possible missing packets issues.

### DIFF
--- a/src/map/packets.h
+++ b/src/map/packets.h
@@ -2802,6 +2802,13 @@ packet(0x96e,-1,clif->ackmergeitems);
 	packet(0x09da,-1); // ZC_ACK_GUILDSTORAGE_LOG
 #endif
 
+// 2013-10-16aRagexe
+#if PACKETVER >= 20131016
+// new packets
+// changed packet sizes
+	packet(0x09d9,6,clif->pDull/*,XXX*/); // CZ_REQ_GUILDSTORAGE_LOG
+#endif
+
 // 2013-10-30aRagexe
 #if PACKETVER >= 20131030
 // new packets
--- a/src/map/packets.h
+++ b/src/map/packets.h
@@ -2229,6 +2229,17 @@ packet(0x96e,-1,clif->ackmergeitems);
 // changed packet sizes
 #endif
 
+// 2013-03-06aRagexe
+#if PACKETVER >= 20130306
+// new packets
+	packet(0x09a6,12); // ZC_BANKING_CHECK
+	packet(0x09a7,14,clif->pDull/*,XXX*/); // CZ_REQ_BANKING_DEPOSIT
+	packet(0x09a8,4); // ZC_ACK_BANKING_DEPOSIT
+	packet(0x09a9,14,clif->pDull/*,XXX*/); // CZ_REQ_BANKING_WITHDRAW
+	packet(0x09aa,4); // ZC_ACK_BANKING_WITHDRAW
+// changed packet sizes
+#endif
+
 //2013-03-20Ragexe (Judas + Yommy)
 #if PACKETVER >= 20130320
 	// Shuffle Start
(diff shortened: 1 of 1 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1690)

src/map/packets.h
```diff
@@ -2229,6 +2229,17 @@ packet(0x96e,-1,clif->ackmergeitems);
 // changed packet sizes
 #endif
 
+// 2013-03-06aRagexe
+#if PACKETVER >= 20130306
+// new packets
+	packet(0x09a6,12); // ZC_BANKING_CHECK
```
Smokexyz: Aren't code-styling changes followed in packets.h? (space after commas)

Jedzkie: Hmm How about changing all of the small letters to capital ?

src/map/packets.h
```diff
@@ -2299,6 +2299,16 @@ packet(0x96e,-1,clif->ackmergeitems);
 	// New Packets End
 #endif
 
+#if PACKETVER >= 20130320
+// new packets
+// changed packet sizes
+	packet(0x09a7,10,clif->pBankDeposit,2,4,6); // CZ_REQ_BANKING_DEPOSIT
```
MishimaHaruna: If the 2013-03-20 client has these packets, then we should probably also update the condition to enable the banking system, since it still checks it against the 20130724 PACKETVER. See battle.c line 7592:

```
#if PACKETVER < 20130724
	if( battle_config.feature_banking ) {
		ShowWarning("conf/map/battle/feature.conf banking is enabled but it requires PACKETVER 2013-07-24 or newer, disabling...\n");
		battle_config.feature_banking = 0;
	}
#endif
```

4144: packets.h mean only packet added/changed in packets len table in client.
But is client using packet at add date or bit after no one know.
This mean for any changes like this need manual testing with clients.
I will left this to some one else, because in wine very low amount of clients working. i cant test them all.

src/map/packets.h
4144: style in packets.h was always own. probably better left it as is.
About case, some times it small, some times in big. For checking tools here no difference. For humans too. My tool using always low case.

src/map/packets.h
Asheraf: Tested with all clients starting from `2013-03-20eRagexe` till `2013-07-17cRagexe` and banking not working.

4144: @Asheraf you enabled other functions what was disabled for this versions?
Because in code can be many #ifdefs related to this.
Or you mean banking in client not implemented?

Asheraf: I did change the check that haru talked about also i tried to `@send 0x09a6` to open the bank window but didn't work on those clients.

Jedzkie:
As far as i know, if the release date of the feature is for example 20130724, you can only use the feature starting on that client date abd onwards. I dont know the packet if they add it advance before it released on sakray. @Asheraf try to use RagexeRE if it works.

4144:
@Jedzkie yes, but most time gravity devs logic like this.
1. add new packets, or change existing for future usage.
2. use new packets.
and client release can be after 1 or after 2. this mean some times packets exists, but they unused. This mean client can receive it (in map server) but do nothing.

And packets.h show only add packet dates and not features itself. This mean if packet date and feature date in hercules different, this mean probably feature date can be updated or may not.
In each case here need testing.

My prs only fixes for packets.h and not to features dates.

4144:
After this pr i will also create pr with packets for 2016 and after for 2017 clients

MishimaHaruna:
Ok, I'll be merging this then, thank you for the tests/clarification!

PR #1698, opened 2017-04-19
Added job checks for @bodystyle command

Jedzkie:

### Pull Request Prelude

### Changes Proposed

**Affected Branches:** master

[//]: # (Master? Slave?)

**Issues addressed:**

Other jobs can still use this command even they don't have body styles. This addition can prevent client crashes.

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -1714,7 +1714,21 @@ ACMD(bodystyle)
 
 	memset(atcmd_output, '\0', sizeof(atcmd_output));
 
-	if (!*message || sscanf(message, "%d", &body_style) < 1) {
+	if ((sd->job & MAPID_THIRDMASK) != MAPID_GUILLOTINE_CROSS
+	 && (sd->job & MAPID_THIRDMASK) != MAPID_GENETIC
+	 && (sd->job & MAPID_THIRDMASK) != MAPID_MECHANIC
+	 && (sd->job & MAPID_THIRDMASK) != MAPID_ROYAL_GUARD
+	 && (sd->job & MAPID_THIRDMASK) != MAPID_ARCH_BISHOP
+	 && (sd->job & MAPID_THIRDMASK) != MAPID_RANGER
+	 && (sd->job & MAPID_THIRDMASK) != MAPID_WARLOCK
+	 && (sd->job & MAPID_THIRDMASK) != MAPID_SHADOW_CHASER
+	 && (sd->job & MAPID_THIRDMASK) != MAPID_MINSTRELWANDERER
+	 ) {
+		clif->message(fd, msg_fd(fd, 35)); // This job has no alternate body styles.
+		return false;
+	}
+
+	if (*message == '\0' || sscanf(message, "%d", &body_style) < 1) {
 		sprintf(atcmd_output, "Please, enter a body style (usage: @bodystyle <body ID: %d-%d>).", MIN_BODY_STYLE, MAX_BODY_STYLE);
 		clif->message(fd, atcmd_output);
 		return false;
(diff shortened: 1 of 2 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1698)

src/map/atcommand.c
```diff
@@ -1714,6 +1714,15 @@ ACMD(bodystyle)
 
 	memset(atcmd_output, '\0', sizeof(atcmd_output));
 
+	if (!((sd->job&MAPID_THIRDMASK) == MAPID_GUILLOTINE_CROSS || (sd->job&MAPID_THIRDMASK) == MAPID_GENETIC
```
Smokexyz: `sd->job & ` for all here. As well as removing the `if (!(` and adding it as ` == 0` at the end of the statement.

Jedzkie: Done.

MishimaHaruna: @Smokexyz you can't replace this with just one bitmask (see the definitions for the MAPID values lower than MAPID_1_1_MAX: they aren't bitmask-compatible)

This needs to be done similar to how @Jedzkie originally did. If you want to get rid of the `!`, you can do the following:

```
if ((sd->job & MAPID_THIRDMASK) != MAPID_GUILLOTINE_CROSS
 && (sd->job & MAPID_THIRDMASK) != MAPID_GENETIC
 && (sd->job & MAPID_THIRDMASK) != MAPID_MECHANIC
 && (sd->job & MAPID_THIRDMASK) != MAPID_ROYAL_GUARD
 && (sd->job & MAPID_THIRDMASK) != MAPID_ARCH_BISHOP
 && (sd->job & MAPID_THIRDMASK) != MAPID_RANGER
 && (sd->job & MAPID_THIRDMASK) != MAPID_WARLOCK
 && (sd->job & MAPID_THIRDMASK) != MAPID_SHADOW_CHASER
 && (sd->job & MAPID_THIRDMASK) != MAPID_MINSTRELWANDERER
) {
	clif->message(fd, msg_txt(35));	// This job has no alternate body styles.
	return false;
}
```

Smokexyz: @MishimaHaruna Hmm, I meant for the code-styling, spaces between bitwise binary operators. And the as discussed with @Jedzkie on IRC, the statement you suggested is the same as `!(` or `() == 0`.

MishimaHaruna: It's not exactly the same. Check the types (I'll pretend that boolean is a type)

The line before this PR was:

```
// original
! (int == int || int == int || int == int)
// simplified:
! (bool || bool || bool)
// further simplified:
! bool
```

which is pretty good, except for the ugly `(! (long expression here))` that has the problem of being counter-intuitive and makes it very easy not to see the negation.

You want to change it to:

```
// suggested line:
(int || int || int) == 0
// simplified:
bool == 0
```

which contains several types "formal" mismatch. You're doing a logical OR between integers instead of booleans. You're doing a comparison between a boolean and zero.

What I suggest is, instead:

```
// suggested line:
int != int && int != int && int != int
// simplified:
bool && bool && bool
// simplified again:
bool
```

The types there are all correct.

My reasoning assumes the following (which isn't technically true in C, since it doesn't really support the boolean type, but please pretend with me that it does, at least at a formal level)

- operator `==` and `!=` take two `int` and return a `bool`
- operator `==` and `!=` take two `bool` and return a `bool`
- using operator `==` or `!=` between a `bool` and an `int` is an error
- operator `||` and `&&` take two `bool` and return a `bool`
- using operator `||` or `&&` between one or more `int` is an error
- operator `&` and `|` take two `int`, return an `int`
- operator `!` takes a `bool` and returns a `bool`

Smokexyz: Uhm, you may verify the return types here - http://icecube.wisc.edu/~dglo/c_class/operators.html

If you're assuming that boolean is a type, you might as well accept it for its real values as false (0) and true (1). All of the boolean expressions return an integer 0 or 1. CPU registers for conditional flags are set as 0 or 1 and not true or false.

The expression - `(sd->job&MAPID_THIRDMASK) == MAPID_GUILLOTINE_CROSS` for example would return 
1) `((int) == int)`
2) `(1 or 0)` and not `(true or false)`

And so the statement above isn't wrong, neither is `() == false` or `() == true` so long as they are both equal to either 0 and 1 respectively.

MishimaHaruna: I stated that boolean isn't a type, but I'd like the conditions to be formally correct, regardless of the C type system.

I insist that the check is incorrect, regardless of its compliance with the C standard. Comparing the result of a logical or against zero makes no sense, even if it returns the correct result.

I'll be fixing it myself before merging.

PR #1722, opened 2017-04-27
Dismemberment of `storage_data` from `mmo_charstatus`.

Smokexyz:

### Pull Request Prelude

### Changes Proposed

Add `storage_data` reception, parsing and sending to/from the map-server.
Remove loading and saving of `storage_data` through char.c
Re-declaration of structure storage_data as a vector.
Re-code of portions in the map-server using `storage_data`.
A new approach is taken by saving the loaded storage data from sql into memory for the duration of the session, thereby removing the need of querying the database to re-load all items everytime a storage save routine is issued from the map-server.
Saving of storage items is done through a new function that significantly reduces the number of queries compared to `char_memitemdata_tosql()`, and therefore run-time speed. This method could potentially reduce the number of update and delete queries from `MAX_STORAGE` (which could be >= 600) times to literally `1`.
Storage items are stored in a dynamically allocated array and handled accordingly.
`struct mmo_charstatus` size reduces by `34,800 bytes`.

**Affected Branches:** Master

[//]: # (Master? Slave?)

**Issues addressed:** #1669

- [x] Needs more testing.


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1722)

4144:
see travis log. look like wrong sign in format strings

Smokexyz:
@4144 
> see travis log. look like wrong sign in format strings
Irrelevant as of https://travis-ci.org/HerculesWS/Hercules/builds/226807632

src/char/int_storage.c
```diff
@@ -240,13 +392,126 @@ int mapif_save_guild_storage_ack(int fd, int account_id, int guild_id, int fail)
...
+	Assert_ret(fd > 0);
+	Assert_ret(account_id > 0);
+
+	cp_stor = (struct storage_data *) idb_ensure(inter_storage->account_storage, account_id, inter_storage->ensure_account_storage);
+
+	count = payload_size/sizeof(struct item);
+
+	VECTOR_INIT(p_stor.item);
+
+	if (count > 0) {
```
MishimaHaruna: Doesn't this also need to run if count is 0? (storage that used to contain items and has just been emptied by the owner)

src/plugins/HPMHooking/HPMHooking_char.Hooks.inc
```diff
@@ -8032,29 +8032,29 @@ int HP_inter_quest_parse_frommap(int fd) {
 	return retVal___;
 }
 /* inter_storage_interface */
-int HP_inter_storage_tosql(int account_id, struct storage_data *p) {
+int HP_inter_storage_tosql(int account_id, struct storage_data *cp, struct storage_data *p) {
```
4144: probably better split auto generated HPM files into separate commit

src/char/int_storage.c
```diff
@@ -174,15 +304,37 @@ int inter_storage_guild_storage_fromsql(int guild_id, struct guild_storage* p)
...
 int inter_storage_sql_init(void)
 {
+	inter_storage->account_storage = idb_alloc(DB_OPT_RELEASE_DATA);
+
 	return 1;
 }
 // storage data finalize
 void inter_storage_sql_final(void)
 {
+	inter_storage->account_storage->destroy(inter_storage->account_storage, NULL);
```
MishimaHaruna: This should call a destructor function that calls VECTOR_CLEAR() on each remaining entry before it gets autoreleased

src/common/mmo.h
```diff
@@ -479,8 +480,9 @@ struct status_change_data {
 };
 
 struct storage_data {
-	int storage_amount;
-	struct item items[MAX_STORAGE];
+	bool save; //< save flag.
+	uint32 aggregate; //< total item count.
```
MishimaHaruna: Can you make this a signed variable instead? I don't really want to see any unsigned used for any reason other than bitmasks or shifts (unless it's something that the client forces us to do)

src/char/int_storage.c
```diff
@@ -41,24 +41,144 @@ struct inter_storage_interface inter_storage_s;
...
+		Sql_ShowDebug(inter->sql_handle);
+
+	StrBuf->Destroy(&buf);
+
+	/**
+	 * Sync map-storage with inter-storage
+	 */
+	VECTOR_CLEAR(cp->item);
+	VECTOR_ENSURE(cp->item, VECTOR_LENGTH(p->item), 1);
+	VECTOR_PUSHARRAY(cp->item, VECTOR_DATA(p->item), VECTOR_LENGTH(p->item));
```
Smokexyz: I think querying the table for a final sync would be better because there is no proper way to get the `id` field of the table after insertion (without splitting the insert queries per item), so I'll go with that.

src/char/int_storage.c
```diff
@@ -41,24 +41,144 @@ struct inter_storage_interface inter_storage_s;
...
+	StrBuf->Init(&buf);
+
+	/**
+	 * Compare and update items, if any.
+	 */
+	for (i = 0; i < VECTOR_LENGTH(cp->item); i++) {
+		struct item *cp_it = &VECTOR_INDEX(cp->item, i);
+		struct item *p_it = NULL;
+
+		ARR_FIND(0, VECTOR_LENGTH(p->item), j,
```
Smokexyz: Oops, this slipped my mind. Added.

src/char/int_storage.c
Smokexyz: Oops, this one escaped me as well. Thank you. Added.

src/char/int_storage.c
Smokexyz: I went ahead and added more than the required number of lines into the clause. Fixed and added.

src/common/mmo.h
Smokexyz: Okay, fixed.

src/map/storage.c
```diff
@@ -212,21 +231,29 @@ int storage_additem(struct map_session_data* sd, struct item* item_data, int amo
...
-	sd->status.storage.items[n].amount -= amount;
-	if( sd->status.storage.items[n].amount == 0 )
-	{
-		memset(&sd->status.storage.items[n],0,sizeof(sd->status.storage.items[0]));
-		sd->status.storage.storage_amount--;
-		if( sd->state.storage_flag == STORAGE_FLAG_NORMAL )
-			clif->updatestorageamount(sd, sd->status.storage.storage_amount, MAX_STORAGE);
+	Assert_retr(1, n >= 0 && n < VECTOR_LENGTH(sd->storage.item));
+
+	nullpo_retr(1, (it = &VECTOR_INDEX(sd->storage.item, n)));
```
MishimaHaruna: no need to use `nullpo_retr()` here, `VECTOR_INDEX()` can't return `NULL` (well, it theoretically can, but only for `n == 0 && VECTOR_DATA(foo) == NULL`, but that can't be the case)

src/map/storage.c
```diff
@@ -212,21 +231,29 @@ int storage_additem(struct map_session_data* sd, struct item* item_data, int amo
...
+
+	Assert_retr(1, amount <= it->amount);
+
+	Assert_retr(1, it->nameid > 0);
```
MishimaHaruna: There are places where this might get called even if `it->nameid` is 0 (example: `@clearstorage`). It's fine to keep the assertion here, but if that's the case, you need to make sure it's not 0 on the caller's side. Else change this assertion to an if

src/map/storage.c
```diff
@@ -212,21 +231,29 @@ int storage_additem(struct map_session_data* sd, struct item* item_data, int amo
...
+	if (it->amount == 0) {
+		memset(it, 0, sizeof(struct item));
+		clif->updatestorageamount(sd, --sd->storage.aggregate, MAX_STORAGE);
 	}
-	if( sd->state.storage_flag == STORAGE_FLAG_NORMAL )
-		clif->storageitemremoved(sd,n,amount);
+
+	sd->storage.save = true;
+
+	clif->storageitemremoved(sd, n, amount);
```
MishimaHaruna: In the case of `@clearstorage` or the itemcheck, you shouldn't send this packet (that's the reason for the `STORAGE_FLAG_NORMAL` check in the original code: if the client isn't showing the storage, we shouldn't send an update packet)

src/map/storage.c
```diff
@@ -356,12 +391,26 @@ int storage_storagegettocart(struct map_session_data* sd, int index, int amount)
  *------------------------------------------*/
 void storage_storageclose(struct map_session_data* sd)
 {
+	int i = 0;
+
 	nullpo_retv(sd);
 
+	Assert_retv(sd->state.storage_flag == STORAGE_FLAG_NORMAL);
```
MishimaHaruna: There are some cases that call this function just to ensure that the storage is saved, even if it's not open in the client (item check, `@clearstorage`). This assertion would fail

src/char/int_storage.c
```diff
@@ -41,24 +41,142 @@ struct inter_storage_interface inter_storage_s;
...
+		}
+
+		StrBuf->Printf(&buf, "('%d', '%d', '%d', '%u', '%d', '%d', '%d', '%u', '%d', '%"PRIu64"'",
+				account_id, p_it->nameid, p_it->amount, p_it->equip, p_it->identify, p_it->refine,
+				p_it->attribute, p_it->expire_time, p_it->bound, p_it->unique_id);
+		for (j = 0; j < MAX_SLOTS; ++j)
+			StrBuf->Printf(&buf, ", '%d'", p_it->card[j]);
+		for (j = 0; j < MAX_ITEM_OPTIONS; ++j)
+			StrBuf->Printf(&buf, ", '%d', '%d'", p_it->option[j].index, p_it->option[j].value);
+		StrBuf->Printf(&buf, ")%s", total_inserts > 0 ? ", " : "");
```
Smokexyz: Uhm, same thing no?

src/char/int_storage.c
```diff
@@ -41,24 +41,142 @@ struct inter_storage_interface inter_storage_s;
...
+		if (j < VECTOR_LENGTH(p->item)) {
+			int k = 0;
+			if (memcmp(cp_it, p_it, sizeof(struct item)) != 0) {
+				if (total_updates == 0) {
+					StrBuf->Printf(&buf, "REPLACE INTO `%s` (`id`, `account_id`, `amount`, `equip`, `identify`, `refine`, `attribute`", storage_db);
+					for (k = 0; k < MAX_SLOTS; k++)
+						StrBuf->Printf(&buf, ", `card%d`", k);
+					for (k = 0; k < MAX_ITEM_OPTIONS; k++)
+						StrBuf->Printf(&buf, ", `opt_idx%d`, `opt_val%d`", k, k);
+					StrBuf->Printf(&buf, ", `expire_time`, `bound`, `unique_id`) VALUES");
```
Smokexyz: Changed to AppendStr.

src/char/int_storage.c
```diff
@@ -41,24 +41,142 @@ struct inter_storage_interface inter_storage_s;
...
+		Sql_ShowDebug(inter->sql_handle);
+
+	/**
+	 * Handle deletions, if any.
+	 */
+	if (total_deletes > 0) {
+		StrBuf->Clear(&buf);
+		StrBuf->Printf(&buf, "DELETE FROM `%s` WHERE `id` IN (", storage_db);
+		for (i = 0; i < total_deletes; i++) {
+			StrBuf->Printf(&buf, ", '%d'", delete[i]);
```
Smokexyz: Fixed, comma will not be added on the first iteration.

src/char/int_storage.c
```diff
@@ -41,24 +41,142 @@ struct inter_storage_interface inter_storage_s;
...
+		}
+		StrBuf->Printf(&buf, ");");
```
Smokexyz: Changed to AppendStr.

src/map/intif.c
```diff
@@ -439,6 +439,131 @@ int intif_request_registry(struct map_session_data *sd, int flag)
...
+		return;
+
+	if (sd->storage.save == false)
+		return;
+
+	len = 8 + VECTOR_LENGTH(sd->storage.item) * sizeof(struct item);
+
+	WFIFOHEAD(inter_fd, len);
+	WFIFOW(inter_fd, 0) = 0x3011;
+	WFIFOW(inter_fd, 2) = (uint16) len;
```
Smokexyz: Since there is no need of sending zeroed entries, we can cache the available items before sending a copy.

src/map/pc.c
```diff
@@ -10218,21 +10222,23 @@ int pc_checkitem(struct map_session_data *sd)
...
+			struct item *it = &VECTOR_INDEX(sd->storage.item, i);
 
-			if (!id)
+			if (it->nameid == 0)
 				continue;
 
-			if( id && !itemdb_available(id) ) {
-				ShowWarning("Removed invalid/disabled item id %d from storage (amount=%d, char_id=%d).\n", id, sd->status.storage.items[i].amount, sd->status.char_id);
-				storage->delitem(sd, i, sd->status.storage.items[i].amount);
+			if (!itemdb_available(id)) {
```
Smokexyz: Yup, has been fixed.

src/map/storage.c
```diff
@@ -148,61 +152,76 @@ int compare_item(struct item *a, struct item *b)
...
+	if (i == VECTOR_LENGTH(sd->storage.item)) {
+		VECTOR_ENSURE(sd->storage.item, 1, 1);
+		VECTOR_PUSH(sd->storage.item, *item_data);
+
+		it = &VECTOR_INDEX(sd->storage.item, i);
+	} else {
+		it = &VECTOR_INDEX(sd->storage.item, i);
+	}
+
+	*it = *item_data;
```
Smokexyz: Yes, thanks, has been fixed.

src/map/storage.c
Smokexyz: Makes sense, removed the nullpo check here.

src/map/storage.c
Smokexyz: This seemed to be the case everywhere else except `@clearstorage`, and I don't see the point in deleting a zeroed entry so I've added checks on the caller's side.

src/map/storage.c
Smokexyz: I think I removed this because I had earlier added the check at the beginning of the function. Re-added because it was removed.

src/map/storage.c
Smokexyz: Yes, not sure how that bit works so I'm going to remove this line here.

Smokexyz:
@MishimaHaruna I will add `sd->state.itemcheck` in another commit once this base change is finalised. And then HPM changes after that.

src/char/int_storage.c
MishimaHaruna: Umm, no, the comma you need to control is the one at the beginning of the line, not the one at the end:

```
INSERT INTO `storage` ( ... ) VALUES  -- beginning, only printed once
( ... ) -- this one lacks a comma
( ... ),
( ... ),
( ... ),
( ... ), -- there's an extra comma here
```

vs.

```
INSERT INTO `storage` ( ... ) VALUES
( ... ) -- no comma
, ( ... )
, ( ... )
, ( ... )
, ( ... )
```

(the line endings are added just for the example purpose, the query is all in one line)

Smokexyz: Ah, of course. 🙄

src/map/storage.c
MishimaHaruna: Along with the assignment you had hidden inside the nullpo ;)

Please restore the assignment, or `it` will be NULL for the rest of the function

src/map/storage.c
MishimaHaruna: Yup, adding the checks on the caller's side sounds good. There's no point in calling this function to delete an item that doesn't exist to begin with

MishimaHaruna:
@Smokexyz looks good, only the two points I commented on are still open (other than itemcheck and HPM)

src/map/storage.c
Smokexyz: 🤦‍♂️

mekolat:
you might want to squash your commits

Smokexyz:
@MishimaHaruna Squashed and updated. Changes are mentioned in the commit message.
I will add the HPM edits after approval.

src/map/pc.c
```diff
@@ -10185,103 +10185,101 @@ int pc_checkitem(struct map_session_data *sd)
...
-	if (sd->state.itemcheck) { // check for invalid(ated) items
-		int id;
-		for (i = 0; i < MAX_INVENTORY; i++) {
-			id = sd->status.inventory[i].nameid;
+	if (sd->itemcheck != PCCHECKITEM_NONE) { // check for invalid(ated) items
+		int id = 0;
 
-			if (!id)
-				continue;
+		if (sd->itemcheck & PCCHECKITEM_INVENTORY) {
```
MishimaHaruna: After each check is complete, you should unset the corresponding flag from `sd->itemcheck` (i.e. `sd->itemcheck &= ~PCCHECKITEM_FOO`)

src/map/clif.c
```diff
@@ -9348,21 +9351,25 @@ void clif_parse_LoadEndAck(int fd, struct map_session_data *sd) {
 
 	if(sd->vd.cloth_color)
 		clif->refreshlook(&sd->bl,sd->bl.id,LOOK_CLOTHES_COLOR,sd->vd.cloth_color,SELF);
+
 	if (sd->vd.body_style)
 		clif->refreshlook(&sd->bl,sd->bl.id,LOOK_BODY2,sd->vd.body_style,SELF);
-	// item
-	clif->inventorylist(sd);  // inventory list first, otherwise deleted items in pc->checkitem show up as 'unknown item'
```
MishimaHaruna: I think this comment is relevant to keep

src/map/clif.c
```diff
@@ -9338,7 +9341,7 @@ void clif_parse_LoadEndAck(int fd, struct map_session_data *sd) {
 	sd->state.warping = 0;
 	sd->state.dialog = 0;/* reset when warping, client dialog will go missing */
 
-	// look
+	//< Character Looks
```
MishimaHaruna: There's an extra `<` sign here. Please note the syntax recognized by doxygen:

```
int foo; ///< Description of foo (three slashes and a left arrow to point to the variable declaration to the left)
```

or

```
/// Description of foo (three slashes and no arrow to refer to the following declaration)
int foo;
```

or

```
/**
 * Description of foo
 * (c-style block comment starting with slash star star, also refers to the following declaration)
 */
int foo;
```

src/map/intif.c
```diff
@@ -439,6 +439,140 @@ int intif_request_registry(struct map_session_data *sd, int flag)
...
+	sd->storage.aggregate = storage_count; //< Total items in storage.
+
+	for (i = 0; i < storage_count; i++) {
+		const struct item *it = RFIFOP(fd, 8 + i * sizeof(struct item));
+		VECTOR_PUSH(sd->storage.item, *it);
+	}
+
+	sd->storage.received = true; //< Mark the storage state as received.
+	sd->storage.save = false; //< Initialize the save flag as false.
+}
```
MishimaHaruna: This seems a good place to re-attempt a second itemcheck, in case the storage wasn't available yet at the time the first itemcheck runs (this time it'll only run the storage part, since the other flags have been unset already)

src/map/pc.c
```diff
@@ -10181,99 +10185,101 @@ int pc_checkitem(struct map_session_data *sd)
...
-				storage->delitem(sd, i, sd->status.storage.items[i].amount);
-				storage->close(sd);
-				continue;
-			}
+				if (!itemdb_available(id)) {
+					ShowWarning("pc_checkitem: Removed invalid/disabled item id %d from storage (amount=%d, char_id=%d).\n", id, it->amount, sd->status.char_id);
+
+					storage->delitem(sd, i, it->amount);
+
+					storage->close(sd);
```
MishimaHaruna: You need to defer this `storage->close()` to the end of the loop (and only if at least an item was deleted).

`storage->close()` will compact the array, so if you do it during the loop, the index will be invalidated.

src/map/pc.c
Smokexyz: Is this necessary? It is not really treated as a state in the code. There are no checks for whether the player is in the state either.

src/map/clif.c
Smokexyz: But if it recognizes `//`, and it should, anything after it would be treated as a comment no?

src/map/clif.c
Smokexyz: This would be obvious because `pc_delitem` in `pc_checkitem` sends `clif_delitem`, which requires the client to have the inventory list before asking it to delete something. But okay, I'll put it back.

src/map/pc.c
MishimaHaruna: Umm, the point of this flag is so that the check doesn't get unnecessarily repeated (it was the same before this pull request as well, it'd unset the flag once the check was completed). After this pull request it becomes more important to flag it, so that you can safely call the function twice, without repeating the check unnecessarily (and wastefully): once when the character logs in, and the second time when the storage is initially received.

src/map/clif.c
MishimaHaruna: No, the `//` (normal comment) is ignored, only `///` (one extra slash) is recognized by doxygen. And `///<` has the special meaning I described.

Now that I re-read my message, it looks like I encouraged to put a doxygen comment here, but that wasn't my intent (tiredness plays weird tricks, sorry). What I meant to say here is that the notation you used, would make it look like an attempt to insert a doxygen comment, while it isn't. You can just remove the extra `<` and have a normal comment here (it's not documenting a variable after all)

Smokexyz:
@MishimaHaruna Updated with changes.

src/map/intif.c
```diff
@@ -512,10 +518,13 @@ void intif_send_account_storage(const struct map_session_data *sd)
 	int len = 0, i = 0;
 	VECTOR_VAR(struct item, items);
 
-	if (intif->CheckForCharServer())
-		return;
+	nullpo_retv(sd);
 
-	if (sd->storage.save == false)
+	// Assert that at this point in the code, both flags are true.
+	Assert_retv(sd->storage.save == true);
```
officialwedevgames: I receive an error after testing this.

```
(04/13/2017 05:06:24) [ Error ] : --- failed assertion --------------------------------------------
(04/13/2017 05:06:24) [ Error ] : c:\users\wedevgames\desktop\ro test servers\herc_04252017\src\map\intif.c:524: 'sd->storage.save == 1' in function `unknown'
(04/13/2017 05:06:24) [ Error ] : --- end failed assertion ----------------------------------------
```

officialwedevgames:
I receive this error while trying to store 1 blue potion using @storeall. Im sure that my sql is updated.

```
[SQL]: DB error - Out of range value adjusted for column 'nameid' at row 1
[Debug]: at c:\users\wedevgames\desktop\ro test servers\herc_04252017\src\char\i
nt_storage.c:152 - INSERT INTO `storage`(`account_id`, `nameid`, `amount`, `equi
p`, `identify`, `refine`, `attribute`, `expire_time`, `bound`, `unique_id`, `car
d0`, `card1`, `card2`, `card3`, `opt_idx0`, `opt_val0`, `opt_idx1`, `opt_val1`,
`opt_idx2`, `opt_val2`, `opt_idx3`, `opt_val3`, `opt_idx4`, `opt_val4`) VALUES (
'2000000', '-13108', '-13108', '3435973836', '1', '0', '0', '80', '204', '327694
49418861860', '-13108', '-13108', '-13108', '-13108', '14608', '1476', '24', '-2
1196', '-8192', '32509', '0', '-13108', '-13108', '-13108')
```

Smokexyz:
@officialwedevgames Thanks for the testing. Updated with fixes.

MishimaHaruna:
Reviewed 5 of 17 files at r1, 1 of 3 files at r3, 1 of 6 files at r4, 3 of 12 files at r5, 6 of 8 files at r6, 4 of 4 files at r7, 7 of 7 files at r8.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[src/map/intif.c, line 506 at r7](https://reviewable.io:443/reviews/herculesws/hercules/1722#-KkjTAqxxlmrMuIz-CG3:-KkjTAqyZ2F0johD8v_w:b-oew7b1) ([raw file](https://github.com/herculesws/hercules/blob/83c368ced9766d4055271aeb86ae0f30ae51bd86/src/map/intif.c#L506)):*
> ```C
> 
> 	if (battle->bc->item_check != PCCHECKITEM_NONE) { // Check and flag items for inspection.
> 		sd->itemcheck = (enum pc_checkitem_types) battle->bc->item_check;
> ```

No need to reset this flag, or it'll run all the checks again. Just let it run the remaining checks only, this line can be removed

---

*[src/map/intif.c, line 539 at r7](https://reviewable.io:443/reviews/herculesws/hercules/1722#-KkjTmbf_L2uTRrXlGwz:-KkjTmbf_L2uTRrXlGx-:b-1ag4z) ([raw file](https://github.com/herculesws/hercules/blob/83c368ced9766d4055271aeb86ae0f30ae51bd86/src/map/intif.c#L539)):*
> ```C
> 		if (VECTOR_INDEX(sd->storage.item, i).nameid == 0)
> 			continue;
> 		memcpy(WFIFOP(inter_fd, 8 + i * sizeof(struct item)), &VECTOR_INDEX(sd->storage.item, i), sizeof(struct item));
> ```

This is still leaving holes in the sent data. Why not just use two index variables? Loop with `i`, but put the data at `8 + c * sizeof(struct item)`, increment `c` only when the memcpy occurred.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1722)*

Smokexyz:
@MishimaHaruna Fixed, couldn't update hooks from my laptop. Could be left to the API?

MishimaHaruna:
I believe this is ni a good state to merge now. Pinging @4144 to mark the changes as approved if it looks good

MishimaHaruna:
Reviewed 8 of 8 files at r9, 7 of 7 files at r10.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/herculesws/hercules/1722)*

MishimaHaruna:
I'm merging this now. Any further changes will be made as follow-up commits.

Thank you very much to @Smokexyz and the reviewers. This was a pretty big one.

PR #1731, opened 2017-05-09
Revised *heal command.

marky291:
### Pull Request Prelude

- [x ] I have followed [proper Hercules code styling][code].
- [x ] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x ] I am aware that this PR will be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

**Issues addressed:**

No longer use command when dead, 
Will now display a message of non use when dead.
No longer display [HP/SP recovered when] removing health, instead [HP/SP Modified]
Improved readability and approach for future editing.

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -1101,7 +1101,7 @@ ACMD(heal)
 {
 	int hp = 0, sp = 0; // [Valaris, FOV]
 	bool recovered = true; // Player not damaged
-	
+
 	sscanf(message, "%12d %12d", &hp, &sp);
 
 	// some overflow checks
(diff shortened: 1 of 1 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1731)

src/map/atcommand.c
```diff
@@ -1094,59 +1094,69 @@ ACMD(kami)
...
 		return true;
 	}
 
-	if ( hp < 0 && sp <= 0 ) {
-		status->damage(NULL, &sd->bl, -hp, -sp, 0, 0);
-		clif->damage(&sd->bl,&sd->bl, 0, 0, -hp, 0, BDT_ENDURE, 0);
-		clif->message(fd, msg_fd(fd,156)); // HP or/and SP modified.
-		return true;
+	// if hp or sp is defined for recovery.
+	if (hp >= 0 || sp >= 0) {
```
MishimaHaruna: I think these checks should be `> 0` rather than `>= 0`. If 0 is passed, there's nothing to heal and this branch shouldn't be entered.

src/map/atcommand.c
```diff
@@ -1094,59 +1094,69 @@ ACMD(kami)
...
+		if (hp > 0 && sp > 0) {
+			status->heal(&sd->bl, hp, sp, 0);
+		} else {
+			if (hp >= 0)
```
MishimaHaruna: Same as above, this should be `hp > 0`. If it's zero, no healing will take place.

src/map/atcommand.c
```diff
@@ -1094,59 +1094,69 @@ ACMD(kami)
...
+				status->heal(&sd->bl, hp, 0, 0);
+			if (sp >= 0)
```
MishimaHaruna: Same as above, this should be `sp > 0`. If it's zero, no healing will take place.

src/map/atcommand.c
```diff
@@ -1094,59 +1094,69 @@ ACMD(kami)
...
+		recovered = false; //negative values applied.
 	}
 
-	if ( sp ) {
-		if (sp > 0)
-			status->heal(&sd->bl, 0, sp, 0);
-		else
-			status->damage(NULL, &sd->bl, 0, -sp, 0, 0);
+	// if we use '@heal 20 -20', we are using damage, which means we didnt recover HP/SP.
+	if (recovered)
```
MishimaHaruna: According to the code style guidelines, if the `else` body has braces, then the `if` body also should.

The only valid possibilities are:

```
if (foo) {
    /* ... */
} else {
    /* ... */
}
```

or

```
if (foo)
    /* ... */;
else
    /* ... */;
```

but not a mix of the two

mekolat:
...why close it?

@MishimaHaruna want me to take this one up?

MishimaHaruna:
@mekolat please do if you feel like it. It won't let me reopen the PR because the source repository was deleted, but the commits are still accessible

marky291:
Sorry, thought it was worthless, so I gave up with it.

Issue #1726, opened 2017-05-02
custom atcommands and group permissions


mekolat:
`bindatcmd()` uses the deprecated GM level system instead of using permission groups.

A solution could be to add a new buildin to dynamically change permissions, so you'd do for example:
```c
bindatcmd("foobar", "NPC::OnUseCommand"); // register @foobar
add_group_command("foobar", 2, true, false); // allow group id 2 to use @foobar
add_group_command("foobar", 5, true, true); // allow group id 5 to use @foobar and #foobar
```

thoughts?

VMGlab:
Isn't it the same with <group level> and <group level char> on the current bindatcmd?
*bindatcmd("command", "<NPC object name>::<event label>"{, <group level>, <group level char>, <log>})
but for better and easier and cleaner way, I'll second this.

mekolat:
on bindatcmd the parameters are min level for @ and min level for # but that's minimum level, not an array of group IDs, so it basically forces you to give group X and up the permission

MishimaHaruna:
Great idea, I second it

There's one use-case that it doesn't cover (but it might be unnecessary, if it's not a very common case). Assuming that if a group inherits another, the command availability is also inherited, groups.conf allows to exclude an inherited command by setting it to false in the child.
The only way I can think to support that would be to use 1, 0, -1 instead of true/false, but that's a bit ugly.

mekolat:
@MishimaHaruna the initial idea was to make `add_group_command()` bypass inheritance, so you can explicitly set true or false for each group. If you do want to keep the inheritance in place though, it could still work:

let's say `<group b>` and `<group c>` both inherit from `<group a>`, you could do the following:
```c
bindatcmd("foobar", "NPC::OnUseCommand"); // register @foobar
add_group_command("foobar", <group a>, true, false); // allow <group a> to use @foobar, and implicitly allow <group b> and <group c> since they inherit
add_group_command("foobar", <group b>, false, false); // explicitly forbid <group b>
```
which would yield this permission tree:
* `<group a>` is `true`
    * `<group b>` is `false`
    * `<group c>` is `true`

mekolat:
@MishimaHaruna here's an implementation that disregards inheritance: https://github.com/HerculesWS/Hercules/pull/1736

MishimaHaruna:
[...]

PR #1736, opened 2017-05-15
new buildin: add_group_command

mekolat:

### Pull Request Prelude

### Changes Proposed
* new buildin: [add_group_command](https://github.com/HerculesWS/Hercules/issues/1726)

**Issues addressed:** #1726 
**Type:** [enhancement](https://github.com/HerculesWS/Hercules/labels/type%3Aenhancement)
**Compopents:** [core](https://github.com/HerculesWS/Hercules/labels/component%3Acore), [script engine](https://github.com/HerculesWS/Hercules/labels/component%3Acore%3Ascriptengine), [documentation](https://github.com/HerculesWS/Hercules/labels/component%3Adocumentation)

<br>

---
If desired I could also make a `add_group_permission()` command

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -7773,6 +7773,21 @@ OnAtcommand:
 
 ---------------------------------------
 
+*add_group_command("<command>", <group id>, <use on self>, <use on other>)
+
+Allows to explicitly change the command permissions for a specific group.
+
+This command bypasses group inheritance, which means groups inheriting from
+the specified <group id> will NOT inherit the specified permission. You should
+use add_group_command() for every group you want to give permission to.
+
+Example:
+	bindatcmd("foobar", "NPC::OnUseCommand", 99, 99, 0); // define the command
+	add_group_command("foobar", 2, true, false); // allow group 2 to use @foobar
+	add_group_command("foobar", 5, true, true); // allow group 5 to use @foobar and #foobar
+
+---------------------------------------
+
 *unbindatcmd("command")
 
 This command will unbind a NPC event label from an atcommand.
@@ -7790,7 +7805,8 @@ scripts-atcommands this way.
 *can_use_command("<command>"{, <account id>})
 
 Checks if the attached or specified player can use the specified
-atcommand and returns true or false accordingly.
+atcommand and returns true or false accordingly. Works for both
+built-in atcommands and custom atcommands.
 
 ---------------------------------------
 
--- a/src/map/atcommand.h
+++ b/src/map/atcommand.h
@@ -79,8 +79,10 @@ struct AtCommandInfo {
 struct atcmd_binding_data {
 	char command[ATCOMMAND_LENGTH];
 	char npc_event[ATCOMMAND_LENGTH];
-	int group_lv;
-	int group_lv_char;
+	int group_lv; // DEPRECATED
+	int group_lv_char; // DEPRECATED
+	char *at_groups; // quick @commands "can-use" lookup
+	char *char_groups; // quick @charcommands "can-use" lookup
 	bool log;
 };
 
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -21499,6 +21499,43 @@ BUILDIN(issit)
 	return true;
 }
 
+BUILDIN(add_group_command)
+{
+	AtCommandInfo *acmd_d;
+	struct atcmd_binding_data *bcmd_d;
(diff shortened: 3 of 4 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1736)

mekolat:
@4144 anything needs changing?

doc/script_commands.txt
```diff
@@ -7773,6 +7773,17 @@ OnAtcommand:
 
 ---------------------------------------
 
+*add_group_command("<command>", <group id>, <use on self>, <use on other>)
+
+Allows to explicitly change the command permissions for a specific group.
```
MishimaHaruna: As I mentioned in #1726 can you point out explicitly that the group inheritance is disregarded?

src/map/atcommand.h
```diff
@@ -79,8 +79,10 @@ struct AtCommandInfo {
 struct atcmd_binding_data {
 	char command[ATCOMMAND_LENGTH];
 	char npc_event[ATCOMMAND_LENGTH];
-	int group_lv;
-	int group_lv_char;
+	int group_lv; // DEPRECATED
+	int group_lv_char; // DEPRECATED
+	char *at_groups; // quick @commands "can-use" lookup
```
MishimaHaruna: Why is this a `char *` if it stores an array of `bool`? (side note: to avoid ambiguity, this should be a VECTOR, so that it self-documents that it's a vector rather than a pointer to a single entry)

src/map/script.c
```diff
@@ -21540,6 +21572,8 @@ BUILDIN(bindatcmd)
 		atcommand->binding[i]->group_lv = group_lv;
 		atcommand->binding[i]->group_lv_char = group_lv_char;
 		atcommand->binding[i]->log = log;
+		CREATE(atcommand->binding[i]->at_groups, char, db_size(pcg->db));
```
MishimaHaruna: `db_size(pcg->db)` doesn't seem enough. Group IDs aren't sequential, but arbitrarily assigned. By default we ship with a group having ID = 99, but we have less than 99 groups.

src/map/atcommand.h
mekolat: I found this odd too, but I thought it was intended because of https://github.com/mekolat/Hercules/blob/master/src/map/atcommand.h#L73

src/map/script.c
mekolat: What do you suggest then? looping through every member of the db? I thought it was correct because already checked elsewhere: https://github.com/mekolat/Hercules/blob/master/src/map/pc_groups.c#L146

doc/script_commands.txt
mekolat: sure, will do

mekolat: @MishimaHaruna done; Is this better?

src/map/script.c
MishimaHaruna: Hmm that assertion doesn't ensure that there are no holes.

What we have right now is:

- Group ID 0: Player
- Group ID 1: Super Player
- Group ID 2: Support
- Group ID 3: Script Manager
- Group ID 4: Event Manager
- Group ID 10: Law Enforcement
- Group ID 99: Admin

There are 7 groups, but their IDs don't go from 0 to 6. And of course you can't access foo[99] in an array of [7].

There's already a function that converts a group ID to a group index though. This can make use of that. Look into `pcg->get_idx()` (I believe you only need to update `BUILDIN(add_group_command)` to make use of that, everything else is already using it. Once that's done, this line can stay as it is)

src/map/atcommand.h
MishimaHaruna: Hmm you're right, I didn't notice we already have that in use.

I still don't think it's a really good design (mostly because we already had bugs in the past, caused by this kind of array pointers being used as normal pointers), but it can stay for now.

Side note: you can access an already-known index of a VECTOR just like you can with an array (i.e. `VECTOR_INDEX(vector, i)` - and yes, it's uglier, but prevents people from accessing it as `*vector`, thinking it only has one record).

In any case, feel free to ignore this for now, it's beyond the scope of this pull request.

mekolat:
@MishimaHaruna if you don't need any other changes, I think this PR is ready to merge

MishimaHaruna:
Ok, looks good, thank you very much

Issue #1667, opened 2017-04-08
[DIS] array syntactic sugar


mekolat:
I wanna hear your thoughts before I start implementing some script shorthands so we can all agree on the syntax.

### array shorthand methods
* `.@var = 1, 2, 3, 4, 5;`
    * shorthand for `setarray .@var[0], 1, 2, 3, 4, 5;`

* `.@var = 1...5;`
    * shorthand for `setarray .@var[0], 1, 2, 3, 4, 5;`

* `.@var = ...#var2;`
    * passes every element of `#var2` as an argument to the array shorthand method
    * shorthand for `copyarray(.@var[0], #var[0], getarraysize(#var[0]));`

* `some_buildin(...array);`
    * passes every element of `array` as arguments to `some_buildin`

### destructuring
* `.@a, .@b, .@c = .@d, .@e, .@f;`
    * shorthand for `.@a = .@d; .@b = .@e; .@c = .@f;`

* `.@a, .@b, .@c = ....@d;`
    * shorthand for `.@a = .@d[0]; .@b = .@d[1]; .@c = .@d[2];`

### math shorthand methods
* `5 ** 2`
    * shorthand for `pow(5, 2)`

Murilo-BiO:
Something that would be great is multidimensional arrays. Is it possible to make with the current script engine?

Ridley8819:
@Murilo-BiO 
'eventInfo = [
        "reward": [
                "id": Apple,
                "amount": 20
        ]
];'

Murilo-BiO:
@Ridley8819  sent an example that I made in a conversation about changes in script engine, I suggested the multidimensional arrays with more readable keys like in PHP arrays

Smokexyz:
I'm going to have to disagree on all due to lack of flexibility or unnecessary processing for simple function calls. I strongly believe that code should be easily understandable. Some of the syntaxes such as `...` imply "magic" happening. There is no magic in code (so to speak).

Might I  suggest considering the following - 
`.@var = [1, 2, 3, 4, 5];` for `setarray .@var[0], 1, 2, 3, 4, 5;`.
and so if ever implementing multi-dimensional arrays, we could have the following variant(s) -
`.@var = [ [1, 2], [3, 4], [5, 0]];` and so on.

as for the math shorthand method, the python method is ok but I'd prefer the good old `a^b`.

4144:
If add brackets, better add it like in C.
this mean
```
[...]

PR #1739, opened 2017-05-17
implementation of the exponentiation operator

mekolat:

### Pull Request Prelude

### Changes Proposed
This is an implementation of the exponentiation operator (from #1667) with left-assignment.

* `pow(5, 3)` => `5 ** 3`
* `pow(pow(5, 3), 2)` => `5 ** 3 ** 2`
* `pow(5, pow(3, 2))` => `5 ** (3 ** 2)`
* `.@var = pow(.@var, 3)` =>`.@var **= 3`

The exponentiation operator has higher precedence than the multiplication operator and lower precedence than the unary operators. This means:

* `5 * 2 ** 3` is equal to `5 * (2 ** 3)`
* `5 ** 2 * 3` is equal to `(5 ** 2) * 3`

<br>

**Related issues:** #1667
**Type:** [enhancement](https://github.com/HerculesWS/Hercules/labels/type%3Aenhancement)
**Compopents:** [script engine](https://github.com/HerculesWS/Hercules/labels/component%3Acore%3Ascriptengine), [documentation](https://github.com/HerculesWS/Hercules/labels/component%3Adocumentation)

---

### Questions
- Does the operator use the right level of precedence?
- Should the `pow()` buildin be marked as deprecated?
- Should existing scripts be converted so they use the exponentiation operator?

### DIFF
--- a/db/re/item_db.conf
+++ b/db/re/item_db.conf
@@ -6419,7 +6419,7 @@ item_db: (
 		bonus2 bIgnoreDefRate,RC_DemiPlayer,25;
 		bonus bUnbreakableWeapon,0;
 		if(getrefine()>5) {
-			bonus2 bAddRace,RC_DemiPlayer,pow(min(getrefine(), 14)-3,2);
+			bonus2 bAddRace,RC_DemiPlayer,((min(getrefine(), 14)-3) ** (2));
 			bonus2 bIgnoreDefRate,RC_DemiPlayer,5;
 		}
 		if(getrefine()>8) {
@@ -9130,7 +9130,7 @@ item_db: (
 		bonus2 bIgnoreDefRate,RC_DemiPlayer,20;
 		bonus bUnbreakableWeapon,0;
 		if(getrefine()>5) {
-			bonus2 bAddRace,RC_DemiPlayer,pow(min(getrefine(),14)-4,2);
+			bonus2 bAddRace,RC_DemiPlayer,((min(getrefine(),14)-4) ** (2));
 			bonus2 bIgnoreDefRate,RC_DemiPlayer,5;
 		}
 		if(getrefine()>8) autobonus "{ bonus bAspdRate,100; }",70,3000,BF_WEAPON,"{ specialeffect2 EF_POTION_BERSERK; }";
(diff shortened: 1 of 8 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1739)

Emistry:
`Should the pow() buildin be marked as deprecated?`
I believe not to mark it as deprecated.

Just like `Set( )` ... they still have their own use-case where the new/current operator cant work in these theircases...

MishimaHaruna:
> Does the operator use the right level of precedence?

Yes, in mathematics, exponentiation takes precedence over multiplication, so I agree with the precedence you assigned.

The problem is that it seems to be generally recognized as being right-associative (see [wikipedia](https://en.wikipedia.org/wiki/Order_of_operations)), so:

```
a ** b ** c
```

should equal to

```
a ** (b ** c)
```

... and I don't remember having any support for this in the parser right now. (I wonder how troublesome it would be to add it, and whether it's worth the effort, or it's better to just keep it left-associative)

> Should the pow() buildin be marked as deprecated?

I think it should. I don't think we'd have cases where it's still necessary to use the pow() command once this is implemented, and I don't like having duplicate facilities.

> Should existing scripts be converted so they use the exponentiation operator?

Definitely. There shouldn't be many I believe.

mekolat:
>The problem is that it seems to be generally recognized as being right-associative

I have no clue how I would make it right-associative without adding extraneous code... Is this a deal-breaker?

---

> I don't think we'd have cases where it's still necessary to use the pow() command once this is implemented, and I don't like having duplicate facilities.

Ok, I'll mark `pow` as deprecated; Is using `BUILDIN_DEF_DEPRECATED` enough or should I explicitly add a `DeprecationWarning()` to `pow` to tell the user to use the exponentiation operator?

---

> There shouldn't be many I believe.

Last time I checked there was only two scripts using it.

MishimaHaruna:
Yes, let's keep it right associative, I don't have any ideas that wouldn't make parsing even uglier than it is. It's not deal-breaker, we can perhaps add a warning in the documentation, to make sure people will notice the left-associativity mention in the precedence table, even if they're used to other languages where this is right-associative.

db/re/item_db.conf
```diff
@@ -6419,7 +6419,7 @@ item_db: (
 		bonus2 bIgnoreDefRate,RC_DemiPlayer,25;
 		bonus bUnbreakableWeapon,0;
 		if(getrefine()>5) {
-			bonus2 bAddRace,RC_DemiPlayer,pow(min(getrefine(), 14)-3,2);
+			bonus2 bAddRace,RC_DemiPlayer,((min(getrefine(), 14)-3) ** (2));
```
mekolat: I was too lazy to go through the whole database so I used `sed`, hence the extraneous parentheses

mekolat:
@MishimaHaruna if you don't need any other changes, I think this PR is ready to merge

PR #1742, opened 2017-05-19
Add support for most released 2017 RE clients

4144:
### Pull Request Prelude

### Changes Proposed

Add support for most released 2017 RE clients.

Add shuffle packets.
Add new packet ids in login and char server.

### DIFF
--- a/src/login/lclif.p.h
+++ b/src/login/lclif.p.h
@@ -56,7 +56,9 @@ enum login_packet_id {
 	//PACKET_ID_CA_SSO_LOGIN_REQa       = 0x825a, /* unused */
 
 	// AC (Login to Client)
+
 	PACKET_ID_AC_ACCEPT_LOGIN         = 0x0069,
+	PACKET_ID_AC_ACCEPT_LOGIN2        = 0x0ac4,
 	PACKET_ID_AC_REFUSE_LOGIN         = 0x006a,
 	PACKET_ID_SC_NOTIFY_BAN           = 0x0081,
 	PACKET_ID_AC_ACK_HASH             = 0x01dc,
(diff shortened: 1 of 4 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1742)

src/login/lclif.p.h
```diff
@@ -56,7 +56,12 @@ enum login_packet_id {
 	//PACKET_ID_CA_SSO_LOGIN_REQa       = 0x825a, /* unused */
 
 	// AC (Login to Client)
+
+#if PACKETVER < 20170315
 	PACKET_ID_AC_ACCEPT_LOGIN         = 0x0069,
+#else
+	PACKET_ID_AC_ACCEPT_LOGIN         = 0x0ac4,
```
MishimaHaruna: Is this meant to have the same name? They usually add a new name when they use a new ID (i.e. `AC_ACCEPT_LOGIN2` or so)

dastgir: We don't know the name, so It's better to rename to AC_ACCEPT_LOGIN2

4144: because we dont know name may be better use PACKET_ID_AC_ACCEPT_LOGIN?
Because this packet can be 3 or other number

MishimaHaruna: Hmm if we don't know the name, I'd go for `PACKET_ID_AC_ACCEPT_LOGIN_` to mark it as temporary name. I'd rather not reuse an existing name, as it could cause problems with existing code that relies on the other ID.

4144: Hmm, if rename then probably better add 2. because _ looks bad,

4144: But here other issue. If add another packet name, need add ifdefs inside lclif.c. For now it clean from ifdefs

PR #1760, opened 2017-06-03
Optimisation of char_memitemdata_to_sql()

Smokexyz:
Implements the storage saving algorithm for inventory, cart and guild storage tables.
Total queries to a table in any call would be no more than 4 per call, replacing the original algorithm that could make a lot more through single update queries per call. This significantly reduces the run time speed for saving/loading of item data from the game server.

### Pull Request Prelude

### Changes Proposed

Optimisation of char_memitemdata_to_sql()
    Implements the storage saving algorithm for inventory, cart and guild storage tables.
    Total queries to a table in any call would be no more than 4 per call, replacing the original algorithm that could make a lot more through single update queries per call. This significantly reduces the run time speed for saving/loading of item data from the game server.

**Affected Branches:** master

### DIFF
--- a/src/char/char.c
+++ b/src/char/char.c
@@ -430,7 +430,7 @@ int char_mmo_char_tosql(int char_id, struct mmo_charstatus* p)
 
 	//map inventory data
 	if( memcmp(p->inventory, cp->inventory, sizeof(p->inventory)) ) {
-		if (!chr->memitemdata_to_sql(p->inventory, MAX_INVENTORY, p->char_id, TABLE_INVENTORY))
+		if (!chr->memitemdata_to_sql(p->inventory, p->char_id, TABLE_INVENTORY))
 			strcat(save_status, " inventory");
 		else
 			errors++;
@@ -438,7 +438,7 @@ int char_mmo_char_tosql(int char_id, struct mmo_charstatus* p)
 
 	//map cart data
 	if( memcmp(p->cart, cp->cart, sizeof(p->cart)) ) {
-		if (!chr->memitemdata_to_sql(p->cart, MAX_CART, p->char_id, TABLE_CART))
+		if (!chr->memitemdata_to_sql(p->cart, p->char_id, TABLE_CART))
 			strcat(save_status, " cart");
 		else
 			errors++;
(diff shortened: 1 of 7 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1760)

src/char/char.c
```diff
@@ -765,123 +755,196 @@ int char_memitemdata_to_sql(const struct item items[], int max, int id, int tabl
...
+	case TABLE_INVENTORY:     tablename = inventory_db;     selectoption = "char_id";    has_favorite = true; break;
+	case TABLE_CART:          tablename = cart_db;          selectoption = "char_id";    break;
+	case TABLE_GUILD_STORAGE: tablename = guild_storage_db; selectoption = "guild_id";   break;
+	default:
+		ShowError("Invalid table name!\n");
+		Assert_retr(-1, table);
+	}
+
+	cp_items = aCalloc(max, sizeof(struct item));
+	deletes = aCalloc(max, sizeof(struct item));
```
MishimaHaruna: This is a memory leak in the case the table is currently empty. Can you move the allocation (and actually, the variable itself) to that block?

src/char/char.c
```diff
@@ -765,123 +755,196 @@ int char_memitemdata_to_sql(const struct item items[], int max, int id, int tabl
...
+	matched_p = aCalloc(max, sizeof(bool));
+
+	StrBuf->Init(&buf);
+
+	/**
+	 * If the storage table is not empty, check for items and replace or delete where needed.
+	 */
+	if (chr->getitemdata_from_sql(cp_items, max, guid, table) > 0) {
+		for (i = 0; i < max; i++) {
```
MishimaHaruna: Optimization: `chr->getitemdata_from_sql()` returns the actual data size, so there's no need to loop through the trailing empty values until `max` here.

src/char/char.c
```diff
@@ -1194,89 +1255,13 @@ int char_mmo_char_fromsql(int char_id, struct mmo_charstatus* p, bool load_every
...
-	//`cart_inventory` (`id`,`char_id`, `nameid`, `amount`, `equip`, `identify`, `refine`, `attribute`, `card0`, `card1`, `card2`, `card3`, expire_time`, `bound`, `unique_id`)
-	StrBuf->Clear(&buf);
-	StrBuf->AppendStr(&buf, "SELECT `id`, `nameid`, `amount`, `equip`, `identify`, `refine`, `attribute`, `expire_time`, `bound`, `unique_id`");
-	for (j = 0; j < MAX_SLOTS; ++j)
-		StrBuf->Printf(&buf, ", `card%d`", j);
-	for (j = 0; j < MAX_ITEM_OPTIONS; ++j)
-		StrBuf->Printf(&buf, ", `opt_idx%d`, `opt_val%d`", j, j);
-	StrBuf->Printf(&buf, " FROM `%s` WHERE `char_id`=? LIMIT %d", cart_db, MAX_CART);
+	/* read inventory [Smokexyz/Hercules] */
+	if (chr->getitemdata_from_sql(p->inventory, MAX_INVENTORY, p->char_id, TABLE_INVENTORY) > 0)
```
MishimaHaruna: Note: `chr->getitemdata_from_sql()` doesn't clear the unused trailing part of the buffer (which we later expect to have `nameid = 0`. This should be ensured either before the call, or inside the function (up to you where as long as it's consistent)

src/char/char.c
```diff
@@ -1194,89 +1255,13 @@ int char_mmo_char_fromsql(int char_id, struct mmo_charstatus* p, bool load_every
...
-		if (SQL_ERROR == SQL->StmtBindColumn(stmt, 10 + MAX_SLOTS + i * 2, SQLDT_INT16, &tmp_item.option[i].index, 0, NULL, NULL)
-			|| SQL_ERROR == SQL->StmtBindColumn(stmt, 11 + MAX_SLOTS + i * 2, SQLDT_INT16, &tmp_item.option[i].value, 0, NULL, NULL))
-			SqlStmt_ShowDebug(stmt);
-
-	for( i = 0; i < MAX_CART && SQL_SUCCESS == SQL->StmtNextRow(stmt); ++i )
-		memcpy(&p->cart[i], &tmp_item, sizeof(tmp_item));
-	
-	strcat(t_msg, " cart");
+	/* read cart [Smokexyz/Hercules] */
+	if (chr->memitemdata_to_sql(p->cart, MAX_CART, p->char_id, TABLE_CART) > 0)
```
MishimaHaruna: Did you mean `chr->getitemdata_from_sql`?

src/char/char.c
```diff
@@ -765,123 +755,196 @@ int char_memitemdata_to_sql(const struct item items[], int max, int id, int tabl
...
+	cp_items = aCalloc(max, sizeof(struct item));
+	deletes = aCalloc(max, sizeof(struct item));
+	matched_p = aCalloc(max, sizeof(bool));
+
+	StrBuf->Init(&buf);
+
+	/**
+	 * If the storage table is not empty, check for items and replace or delete where needed.
+	 */
+	if (chr->getitemdata_from_sql(cp_items, max, guid, table) > 0) {
```
Smokexyz: Max is actually documented as `Max amount of items to be pulled into the list.` for the function that reads from the database. Here, it is the job of the caller to inform the function of the size of the array it is dealing with. Since the inventory, cart and guild storage arrays are not dynamic in size, the assigned value from the caller is passed here (ones defined in macros). The function cannot work on an array (dynamic or static) without knowing its size, whether checked using macros or passed by values to the function. And so similarly the getter function relies upon the caller (this function or elsewhere) to inform it of the size of the array being passed by reference. And the array could be of variable or static length.

src/char/char.c
```diff
@@ -696,63 +696,53 @@ int char_mmo_char_tosql(int char_id, struct mmo_charstatus* p)
...
+	switch (table) {
 	case TABLE_INVENTORY:     tablename = inventory_db;     selectoption = "char_id";    has_favorite = true; break;
 	case TABLE_CART:          tablename = cart_db;          selectoption = "char_id";    break;
-	case TABLE_STORAGE:       tablename = storage_db;       selectoption = "account_id"; break;
 	case TABLE_GUILD_STORAGE: tablename = guild_storage_db; selectoption = "guild_id";   break;
 	default:
-		ShowError("Invalid table name!\n");
-		Assert_retr(1, tableswitch);
+		ShowError("char_getitemdata_from_sql: Invalid table name!\n");
+		Assert_retr(-1, table);
```
Smokexyz: If 0 ever becomes an invalid ID, this assertion check should be changed in that particular commit as it is would be more relevant there.

src/char/char.c
MishimaHaruna: Yes, I agree that the array passed to `getitemdata_from_sql` can be of variable length (and the caller decides the max amount of data to pull. But if the array passed to `memitemdata_to_sql` was of dynamic size, and you were trying to save an almost empty inventory (i.e. 1 item, passing `max = 1`), then this specific call to getitemdata_to_sql would propagate `max = 1`, which will fail to load data from the database, if there's more than one item saved.
I have no problems with `getitemdata_to_sql`'s `max` being the maximum size of the pulled data, but `memitemdata_to_sql`'s `max` is defined as the array size (which, in the case of a dynamically sized array, might be much smaller than the maximum acceptable array size).
As such, I believe it should mention that the array must be large enough to contain the maximum acceptable amount of items (i.e. statically sized array), otherwise the incremental saving algorithm will silently and horribly (item duplication) fail...

src/char/char.c
MishimaHaruna: Well, let's put this in different terms. I'll expand the Assert_retr() macro in this code snippet:


```
enum inventory_table_type {
	TABLE_INVENTORY, // implicit 0
	TABLE_CART,
	TABLE_STORAGE,
	TABLE_GUILD_STORAGE,
};

switch (table) {
case TABLE_INVENTORY:
    tablename = inventory_db;
    selectoption = "char_id";
    has_favorite = true; break;
case TABLE_CART:
    tablename = cart_db;
    selectoption = "char_id";
    break;
case TABLE_GUILD_STORAGE:
    tablename = guild_storage_db;
    selectoption = "guild_id";
    break;
default:
    ShowError("char_getitemdata_from_sql: Invalid table name!\n");
    if (!table) {
        nullpo->assert_report(__FILE__, __LINE__, __func__, "table", "failed assertion");
        return(-1);
    }
 ```

Now, this is clearly not intended. The code returns only if `table` is zero. Now, a good static analyzer could detect the if as useless and warn about it. A good compiler could optimize it out. It's still wrong. And a change in the enum would subtly change the function behavior in a very hard to detect way (think what happens if one day we decide to add `TABLE_UNKNOWN = 0` at the top of the enum for example)

Smokexyz: Yes, this is wrong, from the start (from the old memitemdata code), but it will not bother execution of the code in this PR. But okay I'll address the issue here.

src/char/char.c
Smokexyz: >But if the array passed to memitemdata_to_sql was of dynamic size, and you were trying to save an almost empty inventory (i.e. 1 item, passing max = 1), then this specific call to getitemdata_to_sql would propagate max = 1, which will fail to load data from the database, if there's more than one item saved.

Hmm... Why would it fail to load data if max passed to it is 1? (given that there are items in the database as you say). `memitemdata_to_sql`'s max shouldn't be defined as anything else other than the array size. Which would indeed be the max size (or used size) of a dynamic array as well, and sending anything more than that (for a dynamic sized array) would be unnecessary.

Smokexyz:
@MishimaHaruna updated, please revise.

MishimaHaruna:
Ok, I believe this is ready now. I'm merging it, thank you!

PR #1786, opened 2017-06-16
Make the recovery() command more versatile

### LINKED ISSUE #1784: Suggestion: Add alive() script command
This is a suggestion that we should add the **alive()** script command to Hercules. It functions same as the @alive atcommand, but you can trigger through script engine without having to use `atcommand("@alive")`
The reason for this is sometimes you need to resurrect a character through script, but don't want to display the @alive msgs.

Maybe you can even add optional param for Account ID or Char ID.

```
// *alive(<account id>{, <char id>})
// This function revives the specified character of the
// account id specified. You can also pass the char_id
// to check for both account and char id. Returns true
// on success, false on failure.
BUILDIN(alive)
{
	struct map_session_data *sd = map->id2sd(script_getnum(st, 2));

	if (script_hasdata(st, 3) && sd != NULL
	 && sd->status.char_id != script_getnum(st, 3))
		sd = NULL;

	if (sd == NULL)
	{
		script_pushint(st, 0);
		return false;
	}

	if (!status->revive(&sd->bl, 100, 100))
		return false;

	clif->skill_nodamage(&sd->bl, &sd->bl, ALL_RESURRECTION, 4, 1);
	script_pushint(st, 1);
	return true;
}

BUILDIN_DEF(alive,"i?"),
```

I tested on my own server using `alive(getcharid(CHAR_ID_ACCOUNT))` it seems to work okay.

mekolat:

### Pull Request Prelude

### Changes Proposed
* it can now be used on a single player
* it can now be used on a whole map, or area
* it is backward-compatible

---
closes #1784 

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -13565,7 +13565,7 @@ BUILDIN(failedremovecards)
 			clif->delitem(sd, i, 1, DELITEM_MATERIALCHANGE);
 			clif->additem(sd, i, 1, 0);
 			pc->equipitem(sd, i, sd->status.inventory[i].equip);
-		}		
+		}
 	}
 	clif->misceffect(&sd->bl, 2);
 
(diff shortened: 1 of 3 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1786)

src/map/script.c
```diff
@@ -15043,22 +15043,50 @@ BUILDIN(dispbottom)
...
+				int16 x1 = script_getnum(st, 3);
+				int16 y1 = script_getnum(st, 4);
+				int16 x2 = script_getnum(st, 5);
+				int16 y2 = script_getnum(st, 6);
+				map->foreachinarea(script->buildin_recovery_sub, m, x1, y1, x2, y2, BL_PC);
+			} else {
+				map->foreachinmap(script->buildin_recovery_sub, m, BL_PC);
+			}
+		} else {
+			script->buildin_recovery_pc_sub(map->id2sd(script_getnum(st, 2)), NULL);
```
MishimaHaruna: It's better to split this into two parts: first retrieve the sd pointer (you can use `script->id2sd()`, which is a wrapper for `map->id2sd()` that already displays an error message on the console in case of failure), and then call the `script->buildin_recovery_pc_sub()` function only in case of success, in order to avoid the null pointer report in case of logged off character

MishimaHaruna:
Merged, thank you!

PR #1809, opened 2017-07-17
use /usr/bin/env bash in tools

mekolat:

### Pull Request Prelude

### Changes Proposed

Changes the interpreter directive in tools to use `/usr/bin/env bash` rather than `/bin/bash`. Apparently it's incompatible with BSD otherwise

### DIFF
--- a/tools/ci/retry.sh
+++ b/tools/ci/retry.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # This file is part of Hercules.
 # http://herc.ws - http://github.com/HerculesWS/Hercules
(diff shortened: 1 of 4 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1809)

MishimaHaruna:
This is indeed true, some BSD systems don't come with bash in `/bin/bash`, but install it as a local or optional package (if at all).

Using `/usr/bin/env bash` as shebang can be a security concern in some cases (since it relies on the user's `$PATH`), but I don't find it harmful in this case.

Thank you for the PR, I'm merging it

PR #1812, opened 2017-07-31
add option to restrict party leader change to same map

mekolat:

### Pull Request Prelude

### Changes Proposed

Simple implementation of #1804.
For current installations it will use the old behaviour and for new installations it will use the official behaviour.

### DIFF
--- a/conf/map/battle/party.conf
+++ b/conf/map/battle/party.conf
@@ -41,6 +41,12 @@ party_update_interval: 1000
 // 1: Athena - bar is updated with the party map dots (up to 1 second delay)
 party_hp_mode: 0
 
+// When changing party leader, should the new leader be in the same map
+// as the current leader. (Note 1)
+// true: Aegis - They must be in the same map
+// false: Athena - They can be in different maps
+party_change_leader_same_map: true
+
 // When 'Party Share' item sharing is enabled in a party,
 // announce in the party which party-member received the item and what's he received? (Note 1)
 show_party_share_picker: true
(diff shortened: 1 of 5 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1812)

conf/messages.conf
```diff
@@ -289,7 +289,8 @@
 287: You cannot change party leaders in this map.
 //Missing stuff for @killer related commands.
 288: You are no longer killable.
-//289-290 FREE
+289: Target character must be in the same map.
+//290 FREE
```
Jedzkie: I suggest to use msgstring for 20120410 clients onwards and use `message.conf` on lower client. And also please use the same message text on msgstringtable.txt

mekolat: Why `20120410`? The packet was added in `20060424` so I don't think we even need to check version at all

Jedzkie: Hmm, Not working on 2010 ~ 2011 clients. NO MSG is showing. I'll check later what is the exact date of the client supports this feature.

mekolat: The packet (`0x291`) [is in](https://github.com/HerculesWS/Hercules/blob/master/src/map/packets.h#L1028) the `PACKETVER >= 20060424` section though and the comment says `2006-04-24aSakexe`

4144: packets.h show at what version packet was added to packets table into client. But after this date for some time packet can be unused or even will be removed and added after. This mean if packet added but unused, it's safe to send this packet to client, but client will ignore it.

Actually packet 0x0291 was added in 2006-10-30, and was not changed and not removed for all the time. (but client may use or not use it for other things)

I am not fixing packets.h for old clients because packets itself can be messy here and with packets.h at same time need fix also clif.c

MishimaHaruna:
Is the discussion about 2010~2011 clients still relevant, or is it resolved?

If it's still relevant, one reason for the 'NO MSG' could be that the string isn't present in those older clients executables (or in the associated msgstringtable) and, while packet `0x291` exists and works as intended, there is no string with ID `0x82e`.

MishimaHaruna:
Marking this as accepted for the next release, but we'll need a follow-up patch in case the 2010~2011 clients end up needing special attention.

Jedzkie:
@MishimaHaruna @mekolat exact client date of this feature is 2012-03-07aRagexe

PR #1817, opened 2017-08-10
Save near 200 places in arrays with size SC_MAX.

4144:
SC_X conatants can be with any numbers.

### Pull Request Prelude

### Changes Proposed

Save near 200 places in arrays with size SC_MAX.

### DIFF
--- a/src/map/status.h
+++ b/src/map/status.h
@@ -844,7 +844,7 @@ typedef enum sc_type {
 	SC_DRESS_UP,
 
 	// Rodex
-	SC_DAILYSENDMAILCNT = 824,
+	SC_DAILYSENDMAILCNT,
 #ifndef SC_MAX
 	SC_MAX, //Automatically updated max, used in for's to check we are within bounds.
 #endif


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1817)

guilherme-gm:
If I recall correctly SC_DAILYSENDCOUNT is used by client to show the amount of messages you can send in the current day, so it uses that fixed ID

dastgir:
Afaik, SC doesn't depend on client, might be something else?

guilherme-gm:
hmm SI does then? I thought the server sents sc ID to the client (and it would be a problem there: https://github.com/HerculesWS/Hercules/blob/master/src/map/rodex.c#L59 ) but seems to be fine, my bad

dastgir:
I don't know how rodex works for that.
SI: Status Icon (the one displayed on right side)

guilherme-gm:
Oh I see, didn't know that, thanks!

PR #1821, opened 2017-08-12
WIP: Deprecate hardcoded variables

mekolat:

### Pull Request Prelude

### Changes Proposed

The aim of this PR is to remove hardcoded variables and allow to explicitly specify the variables, or when not possible, reduce the scope to the lowest possible (ie `.@` vars).

### To-Do

- [ ] Update all buildins that use hardcoded variables so they allow to specify a variable instead
- [ ] Make edge cases (ie the `~=` operator) use the lowest possible scope
    - [ ] Add a `use_deprecated_variables` config flag and set to `true` by default when not specified and to `false` in the .conf file. This means new installations will use the low-scope (`.@`) variables by default (because present in the conf file) and old installations will continue to use the high-scope (`$@`) variables for backward-compatibility
- [ ] Update existing scripts
- [ ] Update the documentation


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1821)

mekolat:
* I cannot reduce the scope of `$@p%i$` in `npc_chat.c` because it does not have access to the script state (`st`)
    * I could make the player specify a variable when using `buildin_defpattern` and then save the variable uid and a pointer to its `reg_db` inside `struct pcrematch_entry` so that it can be accessed from `npc_chat.c`

src/map/npc_chat.c
```diff
@@ -388,27 +388,49 @@ int npc_chat_sub(struct block_list* bl, va_list ap)
...
+
+				} else {
+					const char *var_name = ".@regexmatch$";
+					int var_key = script->add_str(var_name);
+					ARGREC_ALLOC(r);
+
+					for (i = 0; i < r; i++) {
+						char val[255];
+						libpcre->copy_substring(msg, offsets, r, i, val, sizeof(val));
+						ARGREC_SET(var_name, var_key, i, val);
```
mekolat: @MishimaHaruna finally I opted for a simple solution:

1. A vector is allocated to hold the scope variables
```c
ARGREC_ALLOC(size);
```

2. The variables are pushed to the vector
```c
ARGREC_SET(".@variable$", script->add_str(".@variable$"), index, "value");
...
```

3. The vector is passed to `run_script_and_inject()`
```c
script->run_npc_inject(script, pos, rid, oid, &ARGREC);
```

4. `run_script_and_inject()` allocates a new `script_state` and iterates through the vector, then adds the variables to the stack, and then frees the vector and runs the script

mekolat:
All the hardcoded global vars have been deprecated; I will now begin deprecating hardcoded PC vars so they use scope vars

mekolat:
@MishimaHaruna since we now have a way to inject variables into the scope we could allow passing arguments to events/timers; is this desirable?

The syntax could be something like this:
```c
addtimer(<ticks>, "<npc>::<event>", <account id>{, <variable>[<index>] ...});
```

Which would allow to do things like:
```c
addtimer(5000, "NPC::OnMyEvent", .@account_id,
	.@foo, .@bar[50], .@baz$);
```

The above would add a timer to the player and store a pointer to a `struct argrec_t` which would contain the variables to inject. After `5000` ticks, `NPC::OnMyEvent` is called and the variables from `argrec_t` are injected into the new scope, with the same name and index they had when passed to addtimer

Issue #1523, opened 2016-12-03
npctalk with empty NPC_NAME


Asheraf:
Git Hash: e1453434af18b30119114077dfc52d27ad196cef
using the following script

```
prontera,150,150,0	script	#blalalla	4_M_COOKIE,{
	npctalk(_("Some Random Words"));
	end;
}
```

would result in-game the npctalk command show the npc hidden name:
`#blalalla : Some Random Words`

Ridley8819:
There are other issues with npctalk too. Sometimes it ignores the the first letters' capital and then all are small letters, I can't exactly tell how to reproduce, but if you  have several npctalk lines starting with a capital letter, it will occur. 

On Topic: Totally agree for an option to disable the npc name

Emistry:
or we could follow rAthena approach here.
https://github.com/rathena/rathena/commit/321b1d22e819023f40fe72128bb25d4177d83970
`npctalk` no longer append the npcname unless it's provided in the text.

Ridley8819:
Makes honestly less sense to me, since you want the npc name in 90% when you use it. An additional TRUE statement to disable it sounds more userfriendly to me

dastgir:
I agree with ridley, optional parameter to disable npc name seems quite good, and there wont be any trouble from user side too.

Ragn0:
According to Aegis chinese script for Horror Toy Factory Instance, every single npctalk command used in there have manually writed the npc name with no exception.

Because of that, it seems that official script does only show the text provided between quotation marks and does never show any other thing that is not writed in there, even the npc name.

This has been more evident because the npc #fac1bs has npctalk command and texted talked by npc shows a name, but npc has not any visible name making them show the hidden name instead, as Asheraf explains in this issue.

PR #1831, opened 2017-08-25
WIP: Add hide name option to npctalk() and unittalk()

Asheraf:

### Pull Request Prelude

### Changes Proposed

Since #1571 have been inactive for a long time i'm opening a new pr.. feel free to review and ask for changes if necessary and i'll update it.

This PR adds GID as a parameter for `getgroupid()`
also adds option to hide names from both `npctalk()` and `unittalk()`

[//]: # (Master? Slave?)

**Issues addressed:**
 #1523 

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -3364,11 +3364,13 @@ Return true if successful, otherwise it will return false.
 
 ---------------------------------------
 
-*getgroupid()
+*getgroupid({<GID>})
 
 This function will return the id of player group the account to which the 
 invoking player belongs.
 
+If <GID> is passed, it will return the value of the specified Player instead of
+the attached Player. If the Player is not found, it will return -1.
 ---------------------------------------
 
 *gettimetick(<type>)
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -6489,12 +6489,16 @@ Examples:
 *unitwarp(<GID>, <Mapname>, <x>, <y>)
 *unitattack(<GID>, <Target ID>)
 *unitstop(<GID>)
-*unittalk(<GID>, <Text>)
+*unittalk(<GID>, <Text>{, show_name})
 *unitemote(<GID>, <Emote>)
 
 Okay, these commands should be fairly self explaining.
 For the emotions, you can look in db/constants.conf for prefixes with e_
 PS: unitwarp() supports a <GID> of zero, which causes the executor of the
+PS: unittalk() can receive a third parameter:
+	show_name:
+		true: Shows Unit name like "UnitName : Message" (default)
+		false: Hides Unit name
 script to be affected. This can be used with OnTouchNPC to warp
 monsters:
 
@@ -6608,12 +6612,16 @@ A debug message also shows on the console when no events are triggered.
 
 ---------------------------------------
 
-*npctalk("<message>"{, "<npc name>"})
+*npctalk("<message>"{, "<npc name>"{, <show_name>}})
+
+show_npcname values:
+	true: shows npc name (default)
+	false: hide npc name
 
 This command will display a message to the surrounding area as if the NPC 
 object running it was a player talking - that is, above their head and in 
-the chat window. The display name of the NPC will get appended in front of 
-the message to complete the effect.
+the chat window. If display is false the name of the NPC will get appended in front of
+the message to complete the effect otherwise the npc name will not be shown.
 
 	// This will make everyone in the area see the NPC greet the character
 	// who just invoked it.
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -15288,12 +15288,13 @@ BUILDIN(message)
 
 /*==========================================
  * npctalk (sends message to surrounding area)
- * usage: npctalk "<message>"{,"<npc name>"};
+ * usage: npctalk("<message>"{, "<npc name>"{, <show_name>}});
  *------------------------------------------*/
 BUILDIN(npctalk)
 {
 	struct npc_data* nd;
 	const char *str = script_getstr(st,2);
+	bool show_name = true;
 
 	if (script_hasdata(st, 3)) {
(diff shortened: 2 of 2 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1831)

mekolat:
# needs testing

Asheraf:
@mekolat i have tested your changes on `2017-05-17aRagexeRE` and `2013-08-07aRagexe` they work fine but i believe the default value for target in `unittalk()` should be `AREA` (everyone in the area) instead of everyone except the attached player

mekolat:
@Asheraf, the hardcoded target was `AREA_CHAT_WOC`, so I kept the same to have it behave the same way. The problem with `AREA` is that it uses the view area size rather than the [chat area size](https://github.com/HerculesWS/Hercules/pull/1562/files#diff-85d990dde8caf2d72d1e2b0f7dcf9786R454) also it still sends to the attached player, but via another packet

Asheraf:
@mekolat for some reason the part of packet `0x8e` is no longer working after this PR.
I have tested on fbc66fa0982a9d7a4ae03377b2eb2b3cfa30c90c
```
unittalk(getnpcid(0), "Hello");
```
this would show the message to everyone in the area including the player attached to the npc.

apter applying this PR same command does not show the message to the attached player.

src/map/clif.c
```diff
@@ -8924,31 +8924,54 @@ void clif_slide(struct block_list *bl, int x, int y)
 /// 008d <packet len>.W <id>.L <message>.?B
 void clif_disp_overhead(struct block_list *bl, const char *mes)
 {
-	unsigned char buf[256]; //This should be more than sufficient, the theoretical max is CHAT_SIZE + 8 (pads and extra inserted crap)
+	struct PACKET_ZC_NOTIFY_CHAT *p;
+	uint32 max_len = CHAT_SIZE_MAX - sizeof(*p);
```
MishimaHaruna: Can you change this to `int` instead? (so that it's the same type as `int mes_len`)

`int max_len = CHAT_SIZE_MAX - (int)sizeof(*p);`

src/map/clif.c
```diff
@@ -8924,31 +8924,54 @@ void clif_slide(struct block_list *bl, int x, int y)
...
-		safestrncpy(WBUFP(buf,4), mes, mes_len);
-		clif->send(buf, WBUFW(buf,2), bl, SELF);
+	if (bl->type == BL_PC)
+		clif->notify_playerchat(bl, mes);
+}
+
+void clif_notify_playerchat(struct block_list *bl, const char *mes)
+{
+	struct PACKET_ZC_NOTIFY_PLAYERCHAT *p;
+	uint32 max_len = CHAT_SIZE_MAX - sizeof(*p);
```
MishimaHaruna: Same as above

src/map/script.c
```diff
@@ -19941,12 +19941,13 @@ BUILDIN(unitstop) {
 
 /// Makes the unit say the message
 ///
-/// unittalk(<unit_id>,"<message>"{, show_name});
+/// unittalk(<unit_id>,"<message>"{, show_name{, <send_target>}});
```
4144: missing target_id in this help string

AnnieRuru:
how about showscript ?
just recently encountered a topic in rathena which had this problem
https://rathena.org/board/topic/115773-npctalk-problems/

Asheraf:
@AnnieRuru yes same thing can be done with `showscript()` as the packet structure is pretty much the same.

AnnieRuru:
seems like my tortoisegit having problem on the 3rd patch
have to apply by hand

doc/script_commands.txt
```diff
@@ -6611,18 +6611,49 @@ Examples:
...
 	unitwarp(0, "this", -1, -1);
 
+	// hide the npc name from the text
+	unittalk(getnpcid(0), "foobar", false);
+
+	// display by npc to everyone
+	unittalk(getnpcid(0), "foobar", true);
+
+	// display by npc to npc
+	unittalk(getnpcid(0), "foobar", true, SELF);
```
AnnieRuru: have you try this line in-game ?
what does it mean by `display by npc to npc` ?
test in-game and it does nothing

also tried defpattern, npctalk doesn't trigger the defpattern
because it only run when player talk to npc, so what's the point of having npc talk to itself ?

Asheraf: This will send the message where only the npc itself can see it, this is useless but for now I didn't want to restrict the command and kept the behavior the same across all unit types however I would like to see @MishimaHaruna input on this, should it be restricted to BL_PC only or it's fine like this.

AnnieRuru: I'm also proposing the format 
*unittalk(<GID>, <Text>{, <flag>{, <account ID>}})

old behavior, send to all with npc name
unittalk getnpcid(0), "blah";

send to all without npc name
unittalk getnpcid(0), "blah", false;

send the chat message from npc to only 1 player only
unittalk getnpcid(0), "blah", true, getcharid(3);

not really sure why you want the monster/pet/npc to hear them
unless we are going to change defpattern to allow it trigger from non-player sources ?

MishimaHaruna: Is this "by npc to npc" still true in the current version of the PR? Wouldn't it send the message to the attached player instead? (I think defaulting to the attached player instead of the specified GID, would cover more use cases)

Am I misreading it, or this is already handled here?
```
		if (target == SELF && (target_bl == NULL || bl == target_bl)) {
			clif->notify_playerchat(bl, StrBuf->Value(&sbuf));
		} else {
			clif->disp_overhead(bl, StrBuf->Value(&sbuf), target, target_bl);
		} 
```

Asheraf: @MishimaHaruna actually in current version this is a mistake `clif->notify_playerchat` is a player only packet and should not cover the case of other units.

PR #1833, opened 2017-08-27
Add NoAutoloot mapflag

Emistry:

### Pull Request Prelude

### Changes Proposed

Autoloot are useful and popular among the players. 
However, sometimes when we try to restrict the autoloot usage at certain map, we can only rely on `mapflag nocommand` to disable the autoloot, or some tricky ways of using the npc script (`bindatcmd`/`OnPCLoadMapEvent`..etc).
With this mapflag, server has more control over the autoloot usage and prevent player abuse the autoloot feature when the server doesn't want them to use it at certain maps.

**Affected Branches:** 

- [x] : Master

**Issues addressed:**
Enable server to disable autoloot settings for certain maps.


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1833)

src/map/atcommand.c
```diff
@@ -3906,6 +3906,8 @@ ACMD(mapinfo)
 		strcat(atcmd_output, msg_fd(fd,1096)); // PartyLock |
 	if (map->list[m_id].flag.guildlock)
 		strcat(atcmd_output, msg_fd(fd,1097)); // GuildLock |
+	if (map->list[m_id].flag.noautoloot)
+		strcat(atcmd_output, "No Autoloot |"); // NoViewID |
```
phaicm: Change comment appropriately

Emistry: lol, will change it

src/map/atcommand.c
```diff
@@ -5710,6 +5718,12 @@ ACMD(autolootitem)
 	int i;
 	int action = 3; // 1=add, 2=remove, 3=help+list (default), 4=reset
 
+					// Disable autoloot
```
dastgir: Extra tab

src/map/mob.c
```diff
@@ -1903,7 +1903,7 @@ void mob_item_drop(struct mob_data *md, struct item_drop_list *dlist, struct ite
 	if( sd == NULL ) sd = map->charid2sd(dlist->third_charid);
 
 	if( sd
-		&& (drop_rate <= sd->state.autoloot || pc->isautolooting(sd, ditem->item_data.nameid))
+		&& ((drop_rate <= sd->state.autoloot || pc->isautolooting(sd, ditem->item_data.nameid)) && !map->list[sd->bl.m].flag.noautoloot)
```
dastgir: Does not need this, instead you can edit
pc_isautolooting to return false if player is on noautoloot

src/map/pc.c
```diff
@@ -5561,6 +5561,13 @@ int pc_setpos(struct map_session_data* sd, unsigned short map_index, int x, int
 		pc->setrestartvalue(sd,1);
 	}
 
+	if (map->list[m].flag.noautoloot) { // Disable autoloot
```
dastgir: Does not need this, as this would be doing unnecessary setting of variables to 0, even if they are 0.
pc_isautolooting should be enough

dastgir: Also, this would mean.
Warping from noautoloot map to other maps where autoloot is allowed would make autoloot list empty.

src/map/mob.c
Emistry: but i think the `pc_isautolooting` doesnt cover the `sd->state.autoloot`

src/map/pc.c
Emistry: removed this part, auto loot list will be remain upon enter/leave the map with `noautoloot` mapflag.

src/map/atcommand.c
Emistry: removed.

dastgir:
Should we do the way it is now(not allowing to turn autoloot on/off or add items) or allow autoloot commands but also display message that autoloot is off in current map?

dastgir:
Also just a message when warping to new map would be good. To let player know autoloot is off.

Emistry:
> Should we do the way it is now(not allowing to turn autoloot on/off or add items) or allow autoloot commands but also display message that autoloot is off in current map?
I think its good to turn it off. Prevent player asking "Why I can use `@autoloot` but its not looting any items.

> Also just a message when warping to new map would be good. To let player know autoloot is off.
erm, I think should be fine without any message to notify them.... just like any other mapflags (nosave, nocommand) we doesn't display any message to alert them that these commands are disabled when they enter the map.

MishimaHaruna:
The way this is implemented doesn't offer a great user experience in my opinion. Some critical points that I'd like to be addressed before merging this:

- If an user has autoloot enabled, and wants to disable it while in a map with this flag, they won't be able to. They instead see a message saying that auto looting items is disabled in the current map (but that's not what they were trying to do to begin with).
- Assuming that a server implements an OnPCLoginEvent that enables autoloot for the users, it'll fail if they log in while in a map with this flag (and won't turn on if they later leave that map).

I'd personally prefer to have the ability to toggle the setting regardless of whether the map allows autolooting. Since autoloot is a global setting attached to the player (rather than attached to the map), then maps shouldn't disallow any changes to the setting, but just disable its effect.

Emistry:
Removed the restriction to use the atcommands in the map that has this mapflag.

Emistry:
Aw, didn't realize that. Thank you and I have updated it.

MishimaHaruna:
Thank you, I'm merging it now

PR #1854, opened 2017-10-09
Change skill_init_unit_layout to report skill id instead of index

Asheraf:

### Pull Request Prelude

### Changes Proposed

Reporting the index to the user when there is a mistake in the database is not helpful and could be confusing sometimes as the user might search for an id that does not exist.

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -18984,7 +18984,8 @@ void skill_init_unit_layout (void)
 
 void skill_init_unit_layout_unknown(int skill_idx)
 {
-	ShowError("unknown unit layout at skill %d\n", skill_idx);
+	Assert_retv(skill_idx >= 0 && skill_idx < MAX_SKILL_DB);
+	ShowError("unknown unit layout at skill %d\n", skill->dbs->db[skill_idx].nameid);
 }
 
 int skill_block_check(struct block_list *bl, sc_type type , uint16 skill_id)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1854)

src/map/skill.c
```diff
@@ -18881,7 +18881,7 @@ void skill_init_unit_layout (void)
 				}
 				break;
 			default:
-				skill->init_unit_layout_unknown(i);
+				skill->init_unit_layout_unknown(skill->dbs->db[i].nameid);
```
4144: probably better here send i as before.
but in function skill_init_unit_layout_unknown do like this:
```
ShowError("unknown unit layout at skill %d\n", skill->dbs->db[skill_idx].nameid);
```

Asheraf: Done

src/map/skill.c
```diff
@@ -18984,7 +18984,7 @@ void skill_init_unit_layout (void)
 
 void skill_init_unit_layout_unknown(int skill_idx)
 {
-	ShowError("unknown unit layout at skill %d\n", skill_idx);
+	ShowError("unknown unit layout at skill %d\n", skill->dbs->db[skill_idx].nameid);
```
MishimaHaruna: How about adding an assertion (to ensure that skill_idx is between 0 and MAX_SKILL_DB) before this call? (just in case the function is called in third party code, passing an invalid index)

PR #1890, opened 2017-10-28
Overflow correction in visual function bg_message

lucaslsb:
### Changes Proposed
Correction of the problem mentioned [IN THIS TOPIC.](http://herc.ws/board/topic/15409-bug-bg_message/)
Apparently there was a memory leak in the dynamic allocation. I compared previous versions of the emulator and found that the message size was received as a parameter, and that all entries per parameter were +1 the original size of the string.
I kept the length definition inside the function and incremented the variable. I have run the tests and the messages are now displayed normally.

**Affected Branches:** 
Master

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -16533,6 +16533,9 @@ void clif_bg_message(struct battleground_data *bgd, int src_id, const char *name
 		return;
 
 	len = (int)strlen(mes);
+#if PACKETVER <= 20120716
+	len += 1;
+#endif
 	Assert_retv(len <= INT16_MAX - NAME_LENGTH - 8);
 	buf = (unsigned char*)aMalloc((len + NAME_LENGTH + 8)*sizeof(unsigned char));
 


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1890)

dastgir:
I guess while combining those chat things into one feature, this one might be left.
@Haru
Maybe we should use clif->process_chat_message here?

4144:
You tested with different clients?
I remember was same fix for other packets, and was need add zero char at end for packet version < 20151001 only.

Probably same should be here?

lucaslsb:
Yes, I conducted tests on clients > 20100730. 
The problem persisted in everyone, with lesser or greater intensity.

4144:
need test on > 20151001

EyesOfAHawk:
great job lucas

Asheraf:
After few tests, this problem exist only in clients `PACKETVER <= 20120716` , tests on 20130306 shows that battle message works correctly. (I cannot test clients in between those two dates)

EyesOfAHawk:
this fixed it for me in 20141022

MishimaHaruna:
@dastgir we're already processing the incoming `clif->pBattleChat()` messages with `clif->process_chat_message()` -- this refers to the outgoing messages instead (`clif->bg_message()`)

Waiting for tests with any client > 20151001 before merging

EyesOfAHawk:
seems workign good on 20151029

xtreamprince:
tested this for 2013-12-23c client, no luk
error

lucaslsb:
Sorry for the delay in the answers.
@Asheraf , changes made in d2c1f88

EyesOfAHawk:
How come `20120716`? This was an issue in 20141022 for me o.o

Asheraf:
@EyesOfAHawk im not sure of the conditions you're testing with but for me i tested +20 clients from 2009 until 2017 the affected ones were all `<= 20120716`.

shadow-mo0n:
Client: 20130807
screen shot 2018-02-12 at 5 28 32 pm

Issue #1531, opened 2016-12-21
Pincode isn't been saved on database.


gilbertobitt:
Hercules commit **9795505**
Hexed Version: ** 2016-02-03aRagexeRE**
[SQL]: DB error - Out of range value for column 'pincode_change' at row 1
[Debug]: at account_sql.c:654 - UPDATE `login` SET `userid`=?,`user_pass`=?,`sex`=?,`email`=?,`group_id`=?,`state`=?,`unban_time`=?,`expiration_time`=?,`logincount`=?,`lastlogin`=?,`last_ip`=?,`birthdate`=?,`character_slots`=?,`pincode`=?,`pincode_change`=? WHERE `account_id` = '2000000'
[Info]: Char load request (150000)

Asheraf:
Hercules support only clients  up to 20151216, your client may have differences in the pincode system.

gilbertobitt:
thank's! so i will disable it for now!

gilbertobitt:
@Asheraf  can u give me a link to a 20151216 ragexe Unpacket?

Ridley8819:
Side note:
On kRO the numbers doesnt switch anymore (during Input). I didnt check if this is client or server related or since when it does behave like this

leomoyses:
Same error :(

Can't get it to work even on 2014-10-22 or 2015-05-13

MishimaHaruna:
This might actually be our fault, not the client's. Does anyone know if the pincode feature ever worked at all? (honestly, I'd never use it personally, it looks really annoying)

I see that we're building the query with a parameter of type `SQLDT_LONG`, despite the variable we pass to it is an unsigned int:
https://github.com/HerculesWS/Hercules/blob/master/src/login/account_sql.c#L651
https://github.com/HerculesWS/Hercules/blob/master/src/login/account.h#L55

Now, this was probably written on a windows machine or on a 32 machine, and it seemed to work correctly, but it won't work at all on 64 bit machines running a UNIX-based OS (most common use case of Hercules), where `sizeof(long) != sizeof(unsigned int)`.
Note (for those that aren't aware): when passing a type other than `SQLDT_STRING`, `SQLDT_ENUM` or `SQLDT_BLOB`, the last argument to `SQL->StmtBindParam()` is ignored, so the `sizeof(acc->pincode_change)` won't be considered but the hardcoded size of `SQLDT_LONG` (which evaluates to `sizeof(long)` is considered instead.

[...]

PR #1901, opened 2017-11-11
Correct and sanitize SQL bind types

MishimaHaruna:
### Pull Request Prelude

### Changes Proposed

Through the codebase there are several calls to the StmtBind functions with incorrect data types. This works on some platforms, and fails on others. An example of failure, caused by the different size of `long` on different platforms, is described in #1531.

This PR checks the calls to the StmtBind functions, changes the data type to the correct one for each, and adds an assertion inside the function, so that it won't fail silently if called with the wrong data type. Detection is currently solely based on the size (so there is still a chance of passing a signed data type while specifying an unsigned one, or pass a different type that has the same size), but it is enough to avoid issues such as the one from the linked issue.

Note: Existing third party code may experience failures at run-time if it calls `SQL->StmtBindColumn()` or `SQL->StmtBindParam()` passing 0 as size, as it was common practice in the past.

**Affected Branches:** 

- master
- stable

**Issues addressed:**

- #1531 

Errors are detected at runtime. There is currently no way to detect them at compile time. This is beyond the scope of this PR in any case.

### DIFF
--- a/src/char/char.c
+++ b/src/char/char.c
@@ -755,7 +755,7 @@ int char_getitemdata_from_sql(struct item *items, int max, int guid, enum invent
 
 	stmt = SQL->StmtMalloc(inter->sql_handle);
 	if (SQL_ERROR == SQL->StmtPrepareStr(stmt, StrBuf->Value(&buf))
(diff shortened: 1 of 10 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1901)

Issue #1909, opened 2017-11-23
failed assertion


xBeret:
[//]: # (**********************************)
[//]: # (** Fill in the following fields **)
[//]: # (**********************************)

[//]: # (Note: Lines beginning with syntax such as this one, are comments and will not be visible in your report!)

### Issue Prelude

[//]: # (Please complete these mandatory steps and check the following boxes by putting an `x` inside the brackets _before_ filing your issue)

-  [x] I have not modified the source prior to reproducing this issue.
-  [x] I am using the latest version of Hercules.
-  [x] I am aware that this report will be closed or deleted if it becomes obvious that I am stating the false.

### Description

When using a skills or attacking some monster occurs failed assertion

![Image of Yaktocat](http://i63.tinypic.com/2lag4f6.png)

[//]: # (Description of the problem or issue at length.)
[//]: # (Please specify any battle configuration related to the components of this issue that have been changed from the default values. This will allow quicker determination of the cause of the problem.)

**Branch(es):**
- [x] master
- [ ] other

**Hercules rev. hash/commit:** 

Git revision src: c1c8d6b7424b6673bd444b0905171c06602feaa4

[//]: # (Copy the first 3 lines of the login-server, char-server or map-server startup.)
[//]: # ( [Info]: Hercules 64-bit for Mac OS X )
[//]: # ( [Info]: Git revision src: 'a5918b329ca0826b04dca32ede783586403f58db' )
[//]: # ( [Info]: Git revision scripts: 'a5918b329ca0826b04dca32ede783586403f58db' )

dastgir:
@4144 @MishimaHaruna

MishimaHaruna:
Confirmed. Adding assertion reports:

```
[Error]: --- failed assertion --------------------------------------------
[Error]: skill.c:179: 'idx != 0' in function `skill_get_ele'
[Error]: 0   map-server                          0x0000000109e5f346 assert_report + 150
[Error]: 1   map-server                          0x0000000109dc5525 skill_get_ele + 85
[Error]: 2   map-server                          0x0000000109e1c0d5 status_check_skilluse + 2069
[Error]: 3   map-server                          0x0000000109e4d39d unit_attack + 205
[Error]: 4   map-server                          0x0000000109c9eadc clif_parse + 396
[Error]: 5   map-server                          0x0000000109e6e03a do_sockets + 618
[Error]: 6   map-server                          0x0000000109e76b4a main + 442
[...]

PR #1910, opened 2017-11-24
Allow using the skill DB accessors on non-skill attacks (skill_id = 0)

MishimaHaruna:
### Pull Request Prelude

### Changes Proposed

This suppresses an assertion in the skill DB accessors when called with `skill_id = 0` (such as during normal attacks). In such case, they'll now return an appropriate zero value, as it happened before 2b4bfa5d0.

**Affected Branches:** 

- master
- stable (v2017.11.19)

**Issues addressed:**

- #1909 

N/A

### Notes

This fixes a regression in v2017.11.19 that makes the map server console very chatty, potentially causing slowdows as the server prints multiple stack traces to the console for every attack. I suggest an uplift to the stable branch (and a v2017.11.19-1 release) once this is merged to master.

Review for uplifting to stable:
- Regression caused by: 2b4bfa5d0
- Impact if declined performance degradation as multiple stack traces are printed to the console, potentially several times each second (during normal attacks)
- Risks: low/none, as this only suppresses assertions in known safe cases (restoring the behavior that existed before the above mentioned commit)

### DIFF
--- a/src/map/battle.h
+++ b/src/map/battle.h
@@ -54,6 +54,7 @@ struct status_data;
 
 enum {
 	// Flag of the final calculation
+	BF_NONE       = 0x0000,
 	BF_WEAPON     = 0x0001,
 	BF_MAGIC      = 0x0002,
 	BF_MISC       = 0x0004,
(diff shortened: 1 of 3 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1910)

PR #1929, opened 2017-12-09
Add support for packet ZC_PROGRESS_ACTOR

Asheraf:

### Pull Request Prelude

### Changes Proposed

This PR adds support for Packet ZC_PROGRESS_ACTOR (0x09D1) also introduce new buildin `progressbar_unit` to use progress bar with given gid.

**Affected Branches:** 

master

[//]: # (Master? Slave?)

- [x] test with ro clients

note: after test the new packet doesn't work with player units however im keeping support for it in the command in case future clients support it

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -9914,7 +9914,6 @@ void clif_parse_Hotkey(int fd, struct map_session_data *sd) {
 
 /// Displays cast-like progress bar (ZC_PROGRESS).
 /// 02f0 <color>.L <time>.L
-/* TODO ZC_PROGRESS_ACTOR <account_id>.L */
 void clif_progressbar(struct map_session_data * sd, unsigned int color, unsigned int second)
 {
 	int fd;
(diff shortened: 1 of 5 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1929)

src/map/clif.c
```diff
@@ -9943,6 +9942,31 @@ void clif_progressbar_abort(struct map_session_data * sd)
...
+	nullpo_retv(bl);
+
+	p.PacketType = progressbarunit;
+	p.GID = bl->id;
+	p.color = color;
+	p.time = time;
+	clif->send(&p, sizeof(p), bl, AREA);
+#else
+	ShowWarning("clif_progressbar_unit: Using progressbar with units available for PACKETVER >= 20130821 only.");
+	return;
```
4144: return here can be removed

Emistry:
Just curious, why split to a new script command just for the new additional GID parameter?
how about merge both together?

4144:
look like this is two different progress bar types.

dastgir:
I guess, having seperate command is good for this case.
As both of them uses different packet, and newer packet is incompatible with old clients.

dastgir:
Alternatively, we could maybe merge that.

@4144 how about following cases:
1) New Client and Account ID Provided: Use New packet
2) New Client without account id: use old packet
3) old client with account id: attach character and send old packet
4) old client without account id: send old packet

So we can have one command to do the work (less confusion)

Asheraf:
@Emistry 
@dastgir i made it at first like that but to reduce the confusion between giving an id or no and the other stuff it was suggested to me to make it this way due to 2 main points

- The packets exist in different client dates (merging them would require having check if the id exist... and having a client check for an argument is just confusion++).
- The functionality of the buildin is different (the new packet is just an animation).

PR #1937, opened 2017-12-20
Enable always asserts and nullpo checks.

4144:

### Pull Request Prelude

### Changes Proposed

Enable always asserts and nullpo checks.
Before this options was depend on debug configure flag.
But this options is must because without it server may crash in any exploits.

### DIFF
--- a/src/common/nullpo.h
+++ b/src/common/nullpo.h
@@ -23,15 +23,10 @@
 
 #include "common/hercules.h"
 
-// enabled by default on debug builds
-#if defined(DEBUG) && !defined(NULLPO_CHECK)
+// if need disable nullpo checks this line can be commented
 #define NULLPO_CHECK
-#endif
-
-// Skip assert checks on release builds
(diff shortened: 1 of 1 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1937)

Smokexyz:
This is confusing. I don't think anyone has decided on what a proper release configuration Hercules should run in (unless I missed out on it). And for the most part it heavily depends on those assertion/nullpo checks. I don't think anyone has an in-depth low-level understanding of the software either, to guarantee that it could run without the `nullpo` checks. So for now I approve of this.

4144:
Without nullpo checks server open for any exploits crashes. Many issues happened even without exploits, if disabled nullpo checks.
This mean it must be permanent enabled
or for all possible exploits need replace nullpo to if (but this is almost impossible)

4144:
And still exists way to disable checks, but for this manually need edit header.

PR #1941, opened 2017-12-23
Add proper MySQL Lib/Header checks

Smokexyz:
Updated configure.ac with better mysql checks library and header file checks.

### Pull Request Prelude

### Changes Proposed

**Affected Branches:** master

[//]: # (Master? Slave?)

**Issues addressed:** 
Fixes a compile issue in macOS High Sierra when searching for mysql header files.

- [x] test cross-compatibility through travis etc.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1500,18 +1500,21 @@ fi
 
 MYSQL_OLD_LIBS="$LIBS" ; LIBS="$LIBS $MYSQL_LIBS"
 MYSQL_OLD_CPPFLAGS="$CPPFLAGS" ; CPPFLAGS="$CPPFLAGS $MYSQL_CFLAGS"
-AC_CHECK_FUNC([mysql_init], [HAVE_MYSQL="yes"])
-AC_CHECK_HEADER([mysql.h], [], [HAVE_MYSQL=""])
+# Note that for AC_CHECK_FUNC to succeed, LIBS has to be correctly defined.
+AC_CHECK_FUNC(mysql_init, [
+		HAVE_MYSQL="yes"
+	], [
+		AC_MSG_ERROR([MySQL Library not found or incompatible])
+		HAVE_MYSQL=""
+	])
+AC_CHECK_HEADER([mysql/mysql.h], [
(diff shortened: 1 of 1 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1941)

4144:
look like this checks will not works for mariadb or other mysql forks if use native client lib.

4144:
for check default ubuntu/debian configurations you can push in gitlab hercules repository and see is CI works.
But even if all passed this can be no enough.

Smokexyz:
I've tested this on macOS, it compiles with mysql, mariadb and percona-server. I'll check on gitlab.

4144:
I mean for native client lib can be named libCOOLNAMEclient.
and your code cant detect this.
But current one support any lib names.

Smokexyz:
@4144 https://gitlab.com/HerculesWS/Hercules/pipelines/15445760
looks like its passing? 

I think autoconf could internally handle such names given the pre-defined ones that it uses? Also, do we need to stay that diverse? 😛

4144:
Look like debian using by default standard libmysqlclient or using auto detection.
In other words you not removed existing code for adding correct mysql client and add check for hardcoded libmysqlclient.

Probably better try use hardcoded name only if existing code failed?

Smokexyz:
> In other words you not removed existing code for adding correct mysql client and add check for hardcoded libmysqlclient.

Uhh clarify please?

Smokexyz:
I've updated the commit. In the existing code, checking for headers fails on macOS because it doesn't update `HAVE_MYSQL`. The PR replaces the if/else checks, utilising the already provided closures arguments as well.

4144:
If it not works on macosx, this mean missing some packages.
Current code works for any os and any mysql forks. Need in patch script mysql_config.

Only limitation here what if fork renamed mysql_config (for example to mariadb_config), need manually give new script name to configure script.

4144:
And your code add additional linking to hardcoded to mysql client lib. I think it totally useless here

PR #1959, opened 2018-01-21
Req. Weapon Class Correction for Spiral Pierce

changelets:
Hi,

According to RateMyServer and iRO Wiki, the Req. Weapon Class for Spiral Pierce is One-handed Sword,  Two-handed Sword, One-handed Spear or Two-handed Spear

### DIFF
--- a/db/pre-re/skill_db.conf
+++ b/db/pre-re/skill_db.conf
@@ -11750,6 +11750,8 @@ skill_db: (
 			Lv10: 45
 		}
 		WeaponTypes: {
+			1HSwords: true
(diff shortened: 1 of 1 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/1959)

Murilo-BiO:
On pre-re it is only possible to use the skill with spears, only in Renewal that is possible to use swords.

changelets:
Oh, that is true. Thanks for the information!

PR #2040, opened 2018-05-13
WIP: Add the Evol Online hashtable buildins

mekolat:
Adds the hashtable buildins from Evol Online (now The Mana World), which make use of strdb to provide a key-value store to the script engine.


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2040)

AnnieRuru:
also -1
reminds me of queue iterator in some ways .... but seems this allows for script variable

I wonder, if do a `@reloadscript` will it refresh/delete the variables like it stated in the documentation ?

mekolat:
There's 2 ways I could change the hash table to thwart bad scripting practices:

1.
    * only allow `htnew()` to be used in `OnInit:`, which would force scripters to keep track of the table IDs in order to use them. <br><br> OR

    * allow `htnew()` anywhere, but instead of using IDs, hash tables would be identified by strings, so you can clearly do `htnew("my_table");` and later use it with `htput("my_table", "my key", data);` for example
        * if the table already exists when calling htnew, it will reuse the old one but clear it


<br>
2. make the iterators expire with the scope

<br><br>

---

@AnnieRuru yes, the hash tables are bound to the script engine, so they reset when `@reloadscript` is used

mekolat:
I have made a plugin so I'm closing this PR

&gt;&gt; https://github.com/HerculesWS/StaffPlugins/tree/master/meko

PR #2061, opened 2018-06-04
Allow OnNPCKillEvent to run on monsters with event label

AnnieRuru:
### Pull Request Prelude

### 2 Issues addressed
1. https://github.com/rathena/rathena/issues/1949
2. https://github.com/rathena/rathena/issues/1278
http://herc.ws/board/topic/15799-killedrid-variable-dont-work-with-script-command-monster/#comment-87230

### 2 Changes Proposed
1st. yes aleos has a point that OnNPCKillEvent shouldn't trigger with monster has event labels
however there are also cases that we want OnNPCKillEvent to trigger both of them together

let's say for example [my mission board script](https://rathena.org/board/topic/109667-mission-board-made-by-annieruru/?do=findComment&comment=344774) or [tr0n's quest board](https://github.com/rathena/rathena/blob/master/npc/custom/quests/questboard.txt)
members keep on complaining that when they setup a board nearby an instanced npc
the board doesn't trigger the kill for them, because those instanced npc already has event label attached

and the reason why I make this configurable so when members change the false into true
they should also change how their script going to work
... just like changing the `check_gotocount` means you don't have to use `*freeloop` as often

2nd. currently monster with event labels doesn't save the `killedrid` variable
suddenly remember this bug, and just add 2 lines in the process

### Affected Branches
* Master

none

### DIFF
--- a/conf/map/script.conf
+++ b/conf/map/script.conf
@@ -59,6 +59,9 @@ script_configuration: {
 	// Defaults to INT_MAX.
 	//input_max_value: 2147483647
 	input_max_value: 10000000
+
+	// Allow OnNPCKillEvent to also run monster with event labels ?
+	kill_mob_event_type: false
 }
 
 import: "conf/import/script.conf"
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -2684,18 +2684,26 @@ int mob_dead(struct mob_data *md, struct block_list *src, int type) {
 				mercenary->kills(sd->md);
 		}
 
-		if( md->npc_event[0] && !md->state.npc_killmonster ) {
-			if( sd && battle_config.mob_npc_event_type ) {
-				pc->setparam(sd, SP_KILLERRID, sd->bl.id);
-				npc->event(sd,md->npc_event,0);
-			} else if( mvp_sd ) {
-				pc->setparam(mvp_sd, SP_KILLERRID, sd?sd->bl.id:0);
-				npc->event(mvp_sd,md->npc_event,0);
-			} else
-				npc->event_do(md->npc_event);
-		} else if( mvp_sd && !md->state.npc_killmonster ) {
-			pc->setparam(mvp_sd, SP_KILLEDRID, md->class_);
-			npc->script_event(mvp_sd, NPCE_KILLNPC); // PCKillNPC [Lance]
+		if (!md->state.npc_killmonster) {
+			if (md->npc_event[0]) {
+				if (sd && battle_config.mob_npc_event_type) {
+					pc->setparam(sd, SP_KILLEDRID, md->class_);
+					pc->setparam(sd, SP_KILLERRID, sd->bl.id);
+					npc->event(sd,md->npc_event,0);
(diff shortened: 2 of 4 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2061)

dastgirp:
I too believe this was much needed feature.

AnnieRuru:
since I have some time now, just write something here to explain whats happen
this post will be announce on the forum

---------------------------------------------------------------------
when the OnNPCKillEvent was introduced years ago (2004-2005?), at the time, 
monsters were only spawn on the fields and dungeon
monsters that has event labels are only in job changer quest
(eg: I still remember at the time taekwon was still developing)

the decision to make OnNPCKillEvent doesn't trigger with event labels is because
at the time monster were only spawn on the fields and dungeon
so they don't want the OnNPCKillEvent to trigger by the monsters inside job changer quest
which was intended behavior, since we only want it to trigger on the fields/dungeon

but things start to change later, with the introduction of instance script (2009?)
all instance script has event labels, and some monsters can only be found inside instance
and this quickly diminish the purpose of OnNPCKillEvent ... 
which at the time said `monster can only found on field and dungeon`

---------------------------------------------------------------

so what happen now is ... how to change our script ...
since most of the event map (including the job changer) has `nosave` mapflag
(I still remember `*getmapflag` was introduce in 2008?)
we can use `if (getmapflag(strcharinfo(PC_MAP), mf_nosave))` to block triggering from event map

which means, yeah, many of the existing scripts that uses OnNPCKillEvent,
you have to understand what it really does now,
as they were working properly before might break now after you make this change

for example, bot-killer script,
some bot-killer script use OnNPCKillEvent: without any kind of checks
but if you decided to change the `kill_event_script_type` from false into true
you have to add these 2 lines
```
if (getmapflag(strcharinfo(PC_MAP), mf_nosave))
end;
```
to block your anti-bot script from triggering from event maps
you don't want players to file complain ... ->
they are killing monsters in event map (devil square?) and suddenly anti-bot pops up, don't ya ?

--------------------------------------------------

conclusion, yeah this is a big change,
that's why rathena ... or aleos didn't implement this system back then,
[at the time I thought he made the right decision](https://rathena.org/board/topic/109667-mission-board-made-by-annieruru/#comment-344774)
he also probably understood if make such change without telling the community can break their scripts

MishimaHaruna:
The only thing I'm worried about is that, to my understanding, there isn't an easy way for a script to implement the previous behavior once the configuration flag for this is enabled.
We should perhaps add a new type to `getunitdata()` to retrieve whether it's a naturally spawning monster, before merging this

Kenpachi2k13:
Indeed.
I think this should be done together with fixing #2692.

AnnieRuru:
ok let me take some time to read that issue from kenpachi 1st

MrKeiKun:
I feel since https://github.com/HerculesWS/Hercules/pull/2821 got merged, can this PR now can be merged too.

PR #2081, opened 2018-06-15
Adds setparam(), change optional param for readparam()

EyesOfAHawk:
[//]: # (**********************************)
[//]: # (** Fill in the following fields **)
[//]: # (**********************************)

[//]: # (Note: Lines beginning with syntax such as this one, are comments and will not be visible in your report!)

### Pull Request Prelude

[//]: # (Thank you for working on improving Hercules!)

[//]: # (Please complete these steps and check the following boxes by putting an `x` inside the brackets _before_ filing your Pull Request.)

- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR will be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

[//]: # (Describe at length, the changes that this pull request makes.)
1. Adds BUILDIN(setparam), which sets the parameter of an account id. Reason is it is currently not possible to set stuff like StatusPoint of a character unless they are attached to the script.
2. Adds functionality so BUILDIN(readparam) can take either a Character Name or account id as the optional parameter.

**Affected Branches:** 

[//]: # (Master? Slave?)
master

**Issues addressed:**

[//]: # (Issue Tracker Number if any.)
N/A

### Known Issues and TODO List

[//]: # (Insert checklist here)
[//]: # (Syntax: - [ ] Checkbox)
N/A

[//]: # (**NOTE** Enable the setting "[√] Allow edits from maintainers." when creating your pull request if you have not already enabled it.)

[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style

PR #2084, opened 2018-06-15
standardize the multi-form buildin definitions

mekolat:
see https://github.com/HerculesWS/Hercules/pull/2081#discussion_r195763559

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -2756,7 +2756,9 @@ save points. Available information types are:
 
 ---------------------------------------
 
-*getcharip({"<character name>"|<account id>|<char id>})
+*getcharip({"<character name>"})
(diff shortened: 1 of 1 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2084)

AnnieRuru:
curious ... what is this line doing ...
[``*groupranditem(<item_id/constant>)``](https://github.com/HerculesWS/Hercules/blob/stable/doc/script_commands.txt#L5306)
``*getrandgroupitem(<item_id/constant>, <quantity>)``

mekolat:
instead of `<account id>` we might want to use `<GID: account>` to make clear that account IDs are also GIDs

PR #2088, opened 2018-06-16
Fixed monsters not disappearing from some clients upon death

guilherme-gm:

### Pull Request Prelude

### Changes Proposed

This fixes an issue where a dead monster stay on the user map until he somehow reloads the map (e.g. through refresh/login out/warp). This happens when the user goes out of the area range of the monster in a couple miliseconds between monster dead and the actual packet that removes the monster.

This PR adds a new area that is big enough to allow `clearunit` to reach those clients.

**Affected Branches:** 

[//]: # (Master? Slave?)
master

**Issues addressed:**

I think there isn't a github issue

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -434,6 +434,7 @@ static bool clif_send(const void *buf, int len, struct block_list *bl, enum send
 	struct battleground_data *bgd = NULL;
 	int x0 = 0, x1 = 0, y0 = 0, y1 = 0, fd;
 	struct s_mapiterator* iter;
+	int area_size;
 
 	if( type != ALL_CLIENT )
 		nullpo_ret(bl);
@@ -472,13 +473,18 @@ static bool clif_send(const void *buf, int len, struct block_list *bl, enum send
 
 		case AREA:
 		case AREA_WOSC:
+		case AREA_DEAD:
 			if (sd && bl->prev == NULL) //Otherwise source misses the packet.[Skotlex]
 				clif->send (buf, len, bl, SELF);
 			/* Fall through */
 		case AREA_WOC:
 		case AREA_WOS:
+			if (type == AREA_DEAD)
+				area_size = DEAD_AREA_SIZE;
+			else
+				area_size = AREA_SIZE;
 			nullpo_retr(true, bl);
-			map->foreachinarea(clif->send_sub, bl->m, bl->x-AREA_SIZE, bl->y-AREA_SIZE, bl->x+AREA_SIZE, bl->y+AREA_SIZE,
+			map->foreachinarea(clif->send_sub, bl->m, bl->x - area_size, bl->y - area_size, bl->x + area_size, bl->y + area_size,
 				BL_PC, buf, len, bl, type);
 			break;
 		case AREA_CHAT_WOC:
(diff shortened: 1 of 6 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2088)

src/map/clif.c
```diff
@@ -453,13 +454,18 @@ bool clif_send(const void* buf, int len, struct block_list* bl, enum send_target
 
 		case AREA:
 		case AREA_WOSC:
+		case AREA_DEAD:
 			if (sd && bl->prev == NULL) //Otherwise source misses the packet.[Skotlex]
 				clif->send (buf, len, bl, SELF);
 			/* Fall through */
 		case AREA_WOC:
 		case AREA_WOS:
+			if (type == AREA_DEAD)
```
4144: probably better add AREA_DEAD into battle.conf like other areas?

AnnieRuru: I thought this is just a 1 line edit ?
just clif->send ... AREA_SIZE *2

what's the point of having this as a setting...

guilherme-gm: Initially I thought about making it a config as you suggested, but I think it'll be a bit confusing and cause issues, because if people make AREA_DEAD smaller than AREA_SIZE, or not big enough, the issue will be back.
Letting it there without a config and using a value that will be safe for pretty much any area_size will make people think better before changing it (if someone ever wants to change that).

4144: @AnnieRuru point what option can be changed
for some one 2x area is too big for some one else too small.
Any hard coded value is bad.

@guilherme-gm if you afraid what some one can set wrong value, you can add comment about what value here better to set. Or even not add any value to config, and left it default value by default, and custom if some one add line into config.

src/map/clif.c
```diff
@@ -453,13 +454,18 @@ bool clif_send(const void* buf, int len, struct block_list* bl, enum send_target
...
+				area_size = 2 * AREA_SIZE;
+			else
+				area_size = AREA_SIZE;
 			nullpo_retr(true, bl);
-			map->foreachinarea(clif->send_sub, bl->m, bl->x-AREA_SIZE, bl->y-AREA_SIZE, bl->x+AREA_SIZE, bl->y+AREA_SIZE,
+			map->foreachinarea(clif->send_sub, bl->m, bl->x- area_size, bl->y- area_size, bl->x+ area_size, bl->y+ area_size,
```
skyleo: Can you add a space between those operators please?

src/map/clif.c
skyleo: I agree with @4144 let people set the AREA_DEAD themselves and document it in a way that people will choose reasonable numbers, like the recommended one 2 * AREA_SIZE. (Which it can be by default anyways)

This would also make this function more aesthetically pleasing than it would be with this PR's change so far.

skyleo:
Side-note:
This bug can still persist on some clients older than 2014 even with AREA_DEAD set to map-wide range.

4144:
@skyleo probably this is other bug or some client limitation/feature.
Current bug was debugged to hercules not send remove packet. But here can be other issues. For example client may ignore packets from long distance etc

Issue #2011, opened 2018-04-04
please deprecate *strmobinfo script command


AnnieRuru:
woohoo ~ I see many script command getting squash ..... yesh ...
there's 1 more ...

*strmobinfo <-- should be deprecate
*getmonsterinfo <-- everyone use this one ~

the only good thing about `strmobinfo` is ...
it can display both IRO name and Aegis name
but `getmonsterinfo` can only display Aegis name
so we can just add another constant `MOB_ENAME` or `MOB_INAME` ?? for `getmonsterinfo`

PR #2089, opened 2018-06-16
Deprecate *strmobinfo script command

AnnieRuru:
### Pull Request Prelude

### Issues addressed
My own https://github.com/HerculesWS/Hercules/issues/2011 ~

### Changes Proposed
1. Deprecate `*strmobinfo` script command
2. `*getmonsterinfo` gets `MOB_ENAME` field
the reason its `MOB_ENAME` but not `MOB_INAME` is because of `--ja--` and `--en--`

### Affected Branches
* Master

as you can see on the 4th patch, some script uses `--ja--` some use `--en--`
so which one to be standardize ?


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2089)

src/map/script.c
```diff
@@ -17931,18 +17931,22 @@ BUILDIN(delmonsterdrop) {
...
-	int mob_id;
+	int mob_id, type;
 
 	mob_id = script_getnum(st,2);
+	type = script_getnum(st,3);
+
 	if (!mob->db_checkid(mob_id)) {
 		ShowError("buildin_getmonsterinfo: Wrong Monster ID: %i\n", mob_id);
-		if ( !script_getnum(st,3) ) //requested a string
+		if (type == 0 || type == 23) //requested a string
```
4144: please add enum with type constants in C side.
And probably map this enum values to script engine. and remove MOB_ constants from constants.conf

4144: Instead adding constants in C and constants.conf you can add constant values only C side.
And add mapping between C and script constants, see function script_hardcoded_constants.
Here no constant values, only names. And because this impossible to add mistake in constants mapping.

All constants you can see in doc/constants.md

4144: for 1. after add new constants in script_hardcoded_constants, from constant.conf duplicates must be removed, or server will show warnings always. and ci will fail.
for 2. document in script_commands.txt what for all constants need see constants.md
All constants ALREADY in constants.md

Asheraf: @AnnieRuru what 4144 is going for here is:
Mapping of constants should be unified between the source code and the scripting engine therefore when an enum is introduced we move the constant definition to the source code in-order to map them with the enum values.

AnnieRuru: ok understood now, making another PR to address this issue

AnnieRuru:
> and also move the constants from ``db/` folder into into source?

you mean do like rathena did ?
https://github.com/rathena/rathena/blob/master/src/map/script_constants.hpp
that's an interesting subject, if the core developers here gives the green light, I also vote for a `yes`

--------------------------------------------------------------

> about the MOB_ENAME, why not its named MOB_JNAME?

because MOB_NAME is actually `mob->db(class_)->jname`
there are still 2 more fields that listed as name
which is `mob->db(class_)->jname` and [`mob->db(class_)->sprite`](https://github.com/HerculesWS/Hercules/blob/stable/src/map/mob.c#L4257)
I didn't include the sprite name in this PR, because I think that's useless ?

-----------------------------------------------

> in mob_db we have the JNAME field but not ENAME or kNAME or iNAME etc...
MOB_NAME -> default name
MOB_JNAME > optional name / JNAME.

silently change their script ? this is also what I have in mind
so the default MOB_NAME change into `mob->db(class_)->name` ... hmm ... sounds better to me too

this is the exact reason I want to deprecate `*strmobinfo` script command
after all, I also prefer to standardize in all scripts, as you can see the description in the 5th patch
some uses MOB_ENAME, some uses MOB_NAME

--------------------------------------------------

> another random thought, maybe drop the getmobdrop too? and merge into getmonsterinfo ?

[`getmobdrops`](https://github.com/HerculesWS/Hercules/blob/stable/doc/script_commands.txt#L3899) generate an array ... hell no ...
I think that's different thing though

PR #2093, opened 2018-06-18
Fix SCRIPT_MAX_ARRAYSIZE

AnnieRuru:
[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style
### Pull Request Prelude
- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR will be closed if the above-mentioned criteria are not fulfilled.

## Issue no.1
SCRIPT_MAX_ARRAYSIZE is UINT_MAX -1 ...
so let's try it out
```
prontera,161,185,5	script	test	1_F_MARIA,{
	.@test[2147483648] = 1;
	end;
}
```
oops ~ already got an error
```
[Warning]: script error in file 'npc/zzz.txt' line 2 column 9
    parse_simpleexpr: overflow detected, capping value to INT_MAX
     1: {
*    2:         .@test[2147483648] = 1;
        ~~~~~~~~~~~~~~~^
     3:         end;
     4: }
```
and no need to say that it cannot be a negative value
```
[Warning]: script:getelementofarray: index out of range (-1)
[Debug]: Data: variable name='.@test' index=0
[Debug]: Source (NPC): test9 at prontera (161,185)
```
this means ... the element in array can only hold between 0~2147483647

## Issue No. 2
let's try another thing
```
prontera,161,185,5	script	test9	1_F_MARIA,{
	.@test[0x7FFFFFFF] = 1;
	dispbottom getarraysize(.@test) +"";
	end;
}
```
return ... WHAT ?? `2147483648` ?? shouldn't this throw error ?
```
prontera,161,185,5	script	test9	1_F_MARIA,{
	.@test[0x7FFFFFFF] = 1;
	dispbottom ( getarraysize(.@test) +0 )+"";
	end;
}
```
now this is better, OVERFLOW !! return `-2147483648`

in other words, the maximum array limit should be INT_MAX -1,
because when needed to display in `*getarraysize`, it shouldn't hit the INT limit

### Changes Proposed
Fix `SCRIPT_MAX_ARRAYSIZE`

### Affected Branches
* Master

### Known Issues and TODO List
```
prontera,161,185,5	script	test9	1_F_MARIA,{
	.@test[0x7FFFFFFF] = 1;
	dispbottom ( getarraysize(.@test) +100 )+"";
	end;
}
```
try this out !

PR #2094, opened 2018-06-18
Add *array_find script command

AnnieRuru:
### Pull Request Prelude

### Issues addressed
Fix my hunger for https://github.com/rathena/rathena/pull/3071 ... grrr ...
there is a [`ARR_FIND` macro](https://github.com/HerculesWS/Hercules/blob/stable/src/common/db.h#L986) in the source code, might as well turn it into a script command as well
reminds me of [`*swap`](https://github.com/HerculesWS/Hercules/pull/1012) script command ~

### Changes Proposed
Add *array_find script command

### Tested with
```
prontera,155,185,5	script	test#1	1_F_MARIA,{
	input .@input;
	setarray .@a[5], 2,4,6,8,10;
	.@test1 = array_find( .@a[1], .@input );
	dispbottom .@test1 +"";
	end;
}

prontera,158,185,5	script	test#2	1_F_MARIA,{
	setarray .@map$, "prontera", "payon", "izlude", "yuno", "alberta", "geffen";
	.@test$ = .@map$[2];
//	.@test2 = array_find( .@map$, .@test$ );
	.@test2 = array_find( .@map$, "yuno" );
	dispbottom .@test2 +"";
	end;
}
```

### Affected Branches
* Master

This is weird, rathena version and @mekolat version return -1 when value not found
but the source code return the array size when value not found ...
which one is correct ??
I personally prefer return array size because this is just a script command without having to use loop
```
	for ( .@i = 0; .@i < getarraysize(.@test); ++.@i )
		if ( .@test == 3 )
			break;
	if ( .@i == getarraysize(.@test) )
		dispbottom "not found";
```
<----------- SIMPLIFY AS ----------->
```
	.@i = array_find( .@test, 3 );
	if ( .@i == getarraysize(.@test) )
		dispbottom "not found";
```


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2094)

Emistry:
imo, return `-1` are more reasonable. any value above -1 are valid index for array. It could be confusing if it returned the array size, which other people end up thinking its the legit index.

and the script command should tell it return index of the array.

AnnieRuru:
but our ARR_FIND macro in the source code return size of array length
https://github.com/HerculesWS/Hercules/blob/stable/src/common/db.h#L986
1 easy example can easily understand is here
https://github.com/HerculesWS/Hercules/blob/stable/src/map/guild.c#L241
it is `if (i == g->max_member)` not `if (i == -1)`

if we actually want to change it into -1,
then the ARR_FIND macro should also return -1 as well ?
then it will become another BIG PR again

Asheraf:
Although I don't mind the addition of this but it will encourage bad practice of having giant arrays and going though them with linear search. @MishimaHaruna

src/map/script.c
```diff
@@ -7496,6 +7496,86 @@ BUILDIN(getelementofarray)
...
+		ShowError("buildin_array_find: not a variable\n");
+		script->reportdata(data);
+		script_pushnil(st);
+		st->state = END;
+		return false;
+	}
+
+	start_index = reference_getindex(data);
+	if (start_index < 0 || start_index > SCRIPT_MAX_ARRAYSIZE) {
+		ShowError("buildin_array_find: starting index out of range (%d)\n", start_index);
```
mekolat: the script engine already ensures that `.@array[-1]` is illegal and that the index isn't above the max, so this check here is redundant and will never be triggered

src/map/script.c
```diff
@@ -7496,6 +7496,86 @@ BUILDIN(getelementofarray)
...
+		st->state = END;
+		return false;
+	}
+
+	id = reference_getid(data);
+	name = reference_getname(data);
+	ref = reference_getref(data);
+
+	array_size = script->array_highest_key(st, st->rid ? script->rid2sd(st) : NULL, name, ref);
+	if (array_size >= SCRIPT_MAX_ARRAYSIZE) { // edge case
```
mekolat: likewise, this here would never be triggered

src/map/script.c
```diff
@@ -7496,6 +7496,86 @@ BUILDIN(getelementofarray)
...
+		script_pushnil(st);
+		st->state = END;
+		return false;
+	}
+
+	if (is_string_variable(name)) {
+		const char *temp, *value;
+
+		if (script_isinttype(st, 3)) {
+			ShowError("buildin_array_find: both sides must be same integer or string type.\n");
```
mekolat: not necessarily, here you could convert your integer to a string with sprintf

src/map/script.c
```diff
@@ -7496,6 +7496,86 @@ BUILDIN(getelementofarray)
...
+			temp = aStrdup(script->get_val2(st, reference_uid(id, i), ref));
+			script_removetop(st, -1, 0);
+			if (!strcmpi(value, temp))
+				break;
+		}
+	} else {
+		int32 temp, value;
+
+		if (script_isstringtype(st, 3)) {
+			ShowError("buildin_array_find: both sides must be same integer or string type.\n");
```
mekolat: here you could convert with atoi

src/map/script.c
```diff
@@ -7496,6 +7496,86 @@ BUILDIN(getelementofarray)
...
+
+		value = script_getnum(st, 3);
+		for (i = start_index; i < array_size; ++i) {
+			temp = (int32)h64BPTRSIZE(script->get_val2(st, reference_uid(id, i), ref));
+			script_removetop(st, -1, 0);
+			if (value == temp)
+				break;
+		}
+	}
+	script_pushint(st, i);
```
mekolat: here you could do `i == array_size ? -1 : i`

src/map/script.c
AnnieRuru: ok

src/map/script.c
AnnieRuru: no,
if we treat this command as a shorten for-loop, it will look like this
```
prontera,155,185,5	script	kdjfhskfhd	1_F_MARIA,{
	setarray .@a,2,4,6,8,10;
	for ( .@i = 0; .@i < getarraysize(.@a); ++.@i )
		if ( .@a == "2" )
			break;
	end;
}
```
which cause this error
```
[Error]: script:op_2: invalid data for operator C_EQ
[Debug]: Data: number value=2
[Debug]: Data: string value="2"
[Debug]: Source (NPC): kdjfhskfhd at prontera (155,185)
```
if we silently convert integer into string means promoting bad scripting habits

src/map/script.c
AnnieRuru: same as above, no conversion

src/map/script.c
AnnieRuru: ok, so everyone love it return -1,
I guess the PHP searcharray also return -1 then ?

src/map/script.c
AnnieRuru: well when I made this script command, I haven't made [SCRIPT_MAX_ARRAYSIZE PR](https://github.com/HerculesWS/Hercules/pull/2093) yet
it is ACTUALLY this condition made me do that PR

but yeah, if that PR is merge, this condition becomes redundant

AnnieRuru:
why plugin ?
I remember during eathena days Toastofdoom made that sprintf/explode/implode source release
and Ai4rei helped fix and optimized those codes
after Jman founded rathena, Epoque merge these script commands into rathena repo

in my eyes, if something that is useful for the community, like that string manipulation script commands
it will eventually merge into repo anyways

EyesOfAHawk:
+1 I'm using meko's script for these but would be much better if they are in the repo as script commands.

PR #2108, opened 2018-06-25
Mapif cleanup

MishimaHaruna:
### Pull Request Prelude

- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR will be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

This PR moves the `mapif_*` functions that are spread through the char server source files, to mapif.c, where they belong. In the process, some conflicts between the return types in the interface and those in the actual functions are fixed (the nonmatching prototypes were not bothering the linker before).

Some non-mapif functionality has been either split or moved to the appropriate `inter_*` or `char` interfaces.

This turned to be necessary for a subsequent pull request I'll submit, that will make many functions `static`, for plugin compatibility (and as a follow-up to #2079).

**Affected Branches:** 

- master
- stable

**Issues addressed:**

No issue present, but necessary for a subsequent PR, as described above.

### Known Issues and TODO List

This is far from being a complete cleanup of the mapif code, which is really due for some serious refactoring. I find it pointless to clean it up any further right now, since it'll get simplified once we officially drop support for multiple map-servers. That's addressed by #1826 and well out of the scope of this PR.

[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style

PR #2112, opened 2018-06-30
Staticfunctions

MishimaHaruna:
### Pull Request Prelude

### Changes Proposed

**Note**: This PR depends on (and contains all of) #2108 - it should be merged after it.

All the functions that allowed for it, have been changed to `static`, in order to avoid issues with plugins on Linux now that we no longer use `RTLD_DEEPBIND`.

When a plugin redefines a symbol with the same name as a non-static function in the core and tries to use it, in some cases it ends up using the one exported by the core. This can be observed, for example, in the db2sql plugin, that defines the following two global variables:

```
/// Backup of the original item_db parser function pointer.
int (*itemdb_readdb_libconfig_sub) (struct config_setting_t *it, int n, const char *source);
/// Backup of the original mob_db parser function pointer.
int (*mob_read_db_sub) (struct config_setting_t *it, int n, const char *source);
```

and, at the time it tries to use them, the linker binds them to the respective functions with the same names found in the core.

The issue is reproducible in Debian 9 with GCC 6.3, and possibly in many other cases.

By making the core functions `static`, I decrease the chance that a symbol name is reused by mistake (the names of static functions aren't exported outside their compilation unit).

Plugin writers should nonetheless use care when naming their variables, to avoid any similar issues.

**Affected Branches:** 

- master

**Issues addressed:**

None, but this is a follow-up to #2079 

At the current state, it's impossible to detect the described issue at compile time, or even at link time. I couldn't find any way to ask the dynamic linker whether there were duplicated names that could cause problems. For the time being, we'll have to rely on the plugin developer's good care to avoid duplicated global names.


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2112)

PR #2130, opened 2018-07-13
reload the const db on script reload

mekolat:
* removes the ability to set params from the const db, which has been deprecated [since 2016](https://github.com/HerculesWS/Hercules/commit/98c1de6a66535311491899db52dc5aed52d98484#diff-0dec96292c5f60451efa76decb262afaR2353)
    * now params can only be [**hardcore**](https://github.com/HerculesWS/Hercules/commit/98c1de6a66535311491899db52dc5aed52d98484) 🤘
* makes the const db reload on script reload


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2130)

PR #2134, opened 2018-07-16
Update int fields to using int32

4144:
### Pull Request Prelude

### Changes Proposed

Update int fields to using int32.
This should allow use for old client item id up to 65535 and for new clients up to 2147483648
Introduced new field MAX_ITEM_ID for limit max allowed item id. (And MAX_ITEMDB used only for size of static item db. Faster works but using too much memory)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2134)

PR #2146, opened 2018-07-25
bake sprintf into debugmes

mekolat:
I frequently find myself doing things like `debugmes(sprintf("foo=%d, bar=%d", .@foo, .@bar));` so I thought "why not simplify this?".

```c
debugmes("Hello %s!", "World"); // => Hello World!
```

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -12349,9 +12349,18 @@ static BUILDIN(getstatus)
  *------------------------------------------*/
 static BUILDIN(debugmes)
 {
-	const char *str;
-	str=script_getstr(st,2);
-	ShowDebug("script debug : %d %d : %s\n",st->rid,st->oid,str);
+	struct StringBuf buf;
+	StrBuf->Init(&buf);
+
+	if (!script->sprintf_helper(st, 2, &buf)) {
+		StrBuf->Destroy(&buf);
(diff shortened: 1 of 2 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2146)

src/map/script.c
```diff
@@ -12349,9 +12349,18 @@ static BUILDIN(getstatus)
...
+
+	if (!script->sprintf_helper(st, 2, &buf)) {
+		StrBuf->Destroy(&buf);
+		script_pushint(st, 0);
+		return false;
+	}
+
+	ShowDebug("script debug : %d %d : %s\n", st->rid, st->oid, StrBuf->Value(&buf));
+	StrBuf->Destroy(&buf);
+	script_pushint(st, 1);
```
AnnieRuru: whats the `script_pushint` doing here ?
its not mention anywhere in the documentation

I guess its to return value based on getting use of sprintf_helper ?

mekolat: it's to return true/false whether or not the message is printed to the console, because `sprintf_helper` can fail

mekolat: there you go @AnnieRuru, I updated the docs

PR #2158, opened 2018-07-28
fix getd not setting the type to C_NAME for new variables

mekolat:
consider the following:
```c
.@var = 1;
set(getd(".@var"), 2);
```
here everything works fine, because on the first line `.@var` is defined, so the script engine gives it type `C_NAME`, but if the first line is omitted,  `.@var` will not be defined and `getd` will give it type `C_NIL` (default type from `add_str`)

this creates issues when the script engine checks the type of the reference:
```c
set(getd(".@var"), 2);
getdatatype(getd(".@var"));
```
```
script:getdatatype: Unknown reference type!
Data: reference name='.@var' type=C_NAME
Please report this!!! - script->str_data.type=C_NIL
```

The solution is simple: always give type `C_NAME` when using `getd`.
`setd` is not affected by this bug.

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -17784,14 +17784,26 @@ static BUILDIN(getd)
 	char varname[100];
 	const char *buffer;
 	int elem;
+	int id;
 
 	buffer = script_getstr(st, 2);
 
 	if (sscanf(buffer, "%99[^[][%d]", varname, &elem) < 2)
 		elem = 0;
 
+	id = script->search_str(varname);
+
+	if (id < 0) {
+		id = script->add_str(varname);
+		script->str_data[id].type = C_NAME;
+	} else if (script->str_data[id].type != C_NAME) {
+		ShowError("script:getd: `%s` is already used by something that is not a variable.\n", varname);
+		st->state = END;
+		return false;
+	}
+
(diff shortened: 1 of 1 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2158)

src/map/script.c
```diff
@@ -17784,14 +17784,18 @@ static BUILDIN(getd)
...
 	const char *buffer;
 	int elem;
+	int id;
 
 	buffer = script_getstr(st, 2);
 
 	if (sscanf(buffer, "%99[^[][%d]", varname, &elem) < 2)
 		elem = 0;
 
+	id = script->add_str(varname);
```
4144: what if varname already added with other type?
assign C_NAME will break other things.

Probably better check is varname not added yet, and if it not added, add it and set type to C_NAME?

4144:
btw you may not define all vars before code.

Old visual studio support finally was removed and now we not depend on very old C version. can be used normal code style with vars declaration where they should be used and not before code like it must be in old C code.

4144:
for example ``id = script->search_str(varname);`` can be replaced to ``int id = script->search_str(varname);`` etc

PR #2167, opened 2018-08-03
add some missing bounds checking for map->list in map.c

Helianthella:
see http://herc.ws/board/topic/16205-mapc-error/?do=findComment&comment=88875

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -3543,6 +3543,8 @@ static int battle_range_type(struct block_list *src, struct block_list *target, 
 
 static int battle_adjust_skill_damage(int m, unsigned short skill_id)
 {
+	Assert_ret(m >= 0 && m < map->count);
+
 	if( map->list[m].skill_count ) {
 		int i;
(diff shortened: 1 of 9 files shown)


src/map/map.c
```diff
@@ -449,6 +449,8 @@ static int map_count_oncell(int16 m, int16 x, int16 y, int type, int flag)
 	struct block_list *bl;
 	int count = 0;
 
+	Assert_ret(m < 0 || m >= map->count);
```
4144: look like all assert checks is wrong.
condition in assert should be for correct values. but here look like for invalid values.

mekolat: yeah sorry, I noticed after pushing...

I guess I shouldn't work while tired haha

src/map/battle.c
```diff
@@ -3543,6 +3543,8 @@ static int battle_range_type(struct block_list *src, struct block_list *target,
 
 static int battle_adjust_skill_damage(int m, unsigned short skill_id)
 {
+	Assert_ret(m >= 0 && m < map->count);
```
4144, line 3546: i think map id can be 0

4144: same in other checks

PR #2179, opened 2018-08-13
Adds extra option to 'mob_npc_event_type' for who did most damage, unless Emperium.

EyesOfAHawk:
### Pull Request Prelude

### Changes Proposed

An extra option will be added to `mob_npc_event_type` so that people can choose if they want NPC event to trigger for most damage, except on Emperium mob.

**Affected Branches:** 
- [x] master
- [x] stable

**Issues addressed:**

N/A


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2179)

dastgirp:
I guess type 0 behaviour should be changed and add a check for woe mapflag and emperium.

Wait for inputs from other devs

Emistry:
imo, the setting are weird.
It should be able to restrict to happen for certain monster or at least within certain maps.
The current behaviour make it trigger for all monsters lol ,even if the char doesn't stay within the same map with the monster that make the event triggered.

if really needd, I did vote for an additional new field in mob_db to allow/disallow the event trigger, having emperium hardcoded doesn't sound good for long term goals. Too much things are hardcoded because of emperium itself, make it us so hard to customize anything that related with emperium and guardian etc.

or maybe might as well consider to revamp the way to attach and trigger monster's events ??

dastgirp:
@Emistry the code does ignore the players who are dead or not in same map or logged out.
But I sort of agree with configurable things, probably should have something which would not have performance hit.

PR #2184, opened 2018-08-21
Stops tsd client crashing if party inviter is disguised.

bWolfie:

### Pull Request Prelude

### Changes Proposed
Stops disguised party invites if disguised. If the party inviter is disguised, they can crash anyone's client just by inviting them to their party (not sure if it will crash on partylock or if player already has party).

Confirmed to crash the target's client in 20160113.

**Affected Branches:** 

[//]: # (Master? Slave?)
- [x] master
- [x] stable

**Issues addressed:**

- [ ] Test in other clients to see if it is a packetver related thing.

### DIFF
--- a/conf/messages.conf
+++ b/conf/messages.conf
@@ -298,7 +298,8 @@
 287: You cannot change party leaders in this map.
 //Missing stuff for @killer related commands.
 288: You are no longer killable.
-//289-290 FREE
+289: You cannot invite players while disguised.
+//290 FREE
 291: Weather effects will disappear after teleporting or refreshing.
 292: Killer state reset.
 //Item Bind System
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -12536,6 +12536,12 @@ static void clif_parse_PartyInvite(int fd, struct map_session_data *sd)
 		return;
 	}
 
+	// If inviter is disguised, crashes target's client
+	if (sd->disguise != -1) {
+		clif->messagecolor_self(sd->fd, COLOR_RED, msg_sd(sd, 289)); // You cannot invite players while disguised.
+		return;
+	}
+
 	t_sd = map->id2sd(RFIFOL(fd,2));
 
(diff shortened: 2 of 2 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2184)

4144:
caption said prevent invite if party leader disguised, but pr really prevent invites if party inviter disguised.
what is correct caption or code?

dastgirp:
If someone else can test for which client it crashes, it would be great.

MrKeiKun:
> 
> 
> If someone else can test for which client it crashes, it would be great.

yep crashing here using current Hercules~

4144:
look like need rebase and conflict fixing

Kenpachi2k13:
I just tested these 3 clients:

- **2016-01-13aRagexeRE**: Crashes upon being invited by a disguised character.
- **2016-01-13bRagexeRE**: Doesn't Crash upon being invited by a disguised character.
- **2016-01-13cRagexeRE**: Doesn't Crash upon being invited by a disguised character.

So this issue obviously is caused by a faulty client and not by Hercules code.

Kenpachi2k13:
I did some testing with `2016-01-13cRagexeRE` and the issue isn't related to the invitation itself.
The actual problem is opening the party window. As of yet I wasn't able to find the real culprit since the crash doesn't occur every time the party window is opened. Actually it doesn't always happen under identical circumstances. :disappointed:

However, prevent disguised characters from inviting other characters doesn't fix anything and thus this PR brings no benefit.

skyleo:
This crash is caused by a missing check in the client for when getting the Actor Object by AID what type of object it is. They think it's of type "0" instead of "1" but both have different vftables and different adresses / offsets for their HP bar information, which leads to an memory access violation for when you have disguised people in your party.

Can pretty much only be solved by modifying the client with your own asm. :)

PR #2189, opened 2018-08-25
Change the language specification to C11 in the autoconf-based builds

MishimaHaruna:
### Pull Request Prelude

### Changes Proposed

Since we use certain C11 features, and all the officially supported *nix systems and respective compilers support the C11 standard, I'm changing the language standard specification from `-std=c99` to -std=c11` in the autoconf script. This requires at least GCC 4.7 (we currently recommend at least 4.9 in any case.

There's a special handling for GCC 4.6 (which we still support, even if we discourage it), to use `-std=c1x` instead (as documented at https://gcc.gnu.org/wiki/C11Status)

Compilers that don't support `-std=c11` will still fall back to to `-std=c99`. The Visual Studio projects aren't affected by this change, either.

Using the chance, I also included const-safe variants of some string manipulation functions (`strchr()`, `strrchr()`, `strstr()`), to make up for shortcoming of the C standard (and have better constness checks at compile time)

**Affected Branches:** 

- master

**Issues addressed:**

N/A

None


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2189)

PR #2193, opened 2018-08-26
buildin_mes shorthand

mekolat:
Makes building_mes use an empty string if none is given:
```c
mes();  =>  mes("");
```


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2193)

Helianthella:
adding something to a milestone at the very last instant doesn't seem like a great idea

Asheraf:
I added it to the milestone just for it to be logged, this is a minor change so I believe it should be okay.

PR #2196, opened 2018-08-28
deprecate the UNITTYPE_ constants, and the commands that use them

### LINKED ISSUE #1677: deprecate UNITTYPE_ constants, switch to hardcoded BL_ constants
For the sake of simplicity I believe we should deprecate the `UNITTYPE_` constants so we only have the `BL_` constants.
* The `BL_` constants have the same values as in the `bl_type` enum
* The `BL_` constants can be bitmasked, enabling to pass more than one type to a command
* The `BL_` constants are hardcoded in script.c so if the enum ever changes they will still be up-to-date

Helianthella:
If we make `getmapxy()` accept `BL_` constants then we can't make it backwards-compatible with `UNITTYPE_` constants, because if an integer is given instead of a constant there's no way to know if the value is intended to be one of `UNITTYPE_` or `BL_`, so it's better to just deprecate the whole command altogether, since it's already integrated into `getunitdata()`. Also `getmapxy()` has some weird behaviour, such as finding the location of a pet by the name of its owner (WTF?) and is quite messy in general.

closes #1677 

<br> <br> <br>

- - -

<sub>BTW this is meko, I changed by username</sub>

### DIFF
--- a/npc/other/Global_Functions.txt
+++ b/npc/other/Global_Functions.txt
@@ -407,7 +407,7 @@ function	script	Time2Str	{
 //== Function F_ShuffleNumbers =============================
 // Generate a set of numbers in random order that the numbers are not repeated
 //    callfunc "F_ShuffleNumbers",<start num>,<last num>,<output array>{,<count>};
-// Examples: 
+// Examples:
 //    callfunc("F_ShuffleNumbers", 0, 5, .@output)      // possible output 4,1,3,2,0,5
 //    callfunc("F_ShuffleNumbers", -5, 1, .@output)     // possible output -3,-5,-4,-2,-1,1,0
 //    callfunc("F_ShuffleNumbers", 0, 100, .@output, 5) // possible output 9,55,27,84,33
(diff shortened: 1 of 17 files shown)


npc/other/Global_Functions.txt
```diff
@@ -438,3 +438,13 @@ function	script	F_ShuffleNumbers	{
 function	script	F_MesColor	{
 	return sprintf("^%06X", min(getarg(0), 0xFFFFFF));
 }
+
+// getunitlocation(<unit id>, <map variable>, <x variable>, <y variable>)
+//     syntax sugar for getunitdata
+function	script	getunitlocation	{
```
Helianthella: this is a drop-in replacement for `getmapxy()` for old scripts

Emistry:
why not just fix and update the `getmapxy` instead? from a 1 line of codes it has now basically become 5+ lines... and lost some of the feature from `getmapxy` for example it able to check if the server able to retrieve the unit location. Of course it could be done for the new `getunitlocation()` but why create so much hassle just for a changes like this? It has no backward-compatibility for older Herc too.

anyway
```
function	script	getunitlocation	{
```
should have be
```
function	script	F_GetUnitLocation	{
```

Helianthella:
@emistry it is not possible to make `getmapxy` support `BL_` constants without breaking backward-compatibility. there is simply no easy fix

dastgirp:
I believe deprecating getmapxy would be bad, since it's used even in old scripts (since eA) and updating everything would be troublesome. Or might be not as handy.

Can there be some kind of processing to detect if BL_ constant was used or UNITTYPE_ was used and throw deprecation accordingly

Helianthella:
@dastgirp yes, there can, but as stated in the first post if a scripter decides to pass a number instead of a constant then there's no way to know which enum it belongs to

Helianthella:
this would cause too much chaos, so I'm closing it

PR #2227, opened 2018-09-28
only update zeny achievements and logs for non-zero transactions

Helianthella:

### Pull Request Prelude

### Changes Proposed

When a player does a zeny transaction, `achievement_validate_zeny` is called to see if the transaction triggered the completion of an achievement objective, but this function expects `z < 0 || z > 0`, so we should either remove the non-zero assert or only call it when we have a non-zero zeny transaction.

This PR fixes http://herc.ws/board/topic/16336--

### DIFF
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -4506,14 +4506,15 @@ static int pc_payzeny(struct map_session_data *sd, int zeny, enum e_log_pick_typ
 	sd->status.zeny -= zeny;
 	clif->updatestatus(sd,SP_ZENY);
 
-	achievement->validate_zeny(sd, -zeny); // Achievements [Smokexyz/Hercules]
+	if (zeny > 0) {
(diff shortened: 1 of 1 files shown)


src/map/pc.c
```diff
@@ -4506,14 +4506,15 @@ static int pc_payzeny(struct map_session_data *sd, int zeny, enum e_log_pick_typ
 	sd->status.zeny -= zeny;
 	clif->updatestatus(sd,SP_ZENY);
 
-	achievement->validate_zeny(sd, -zeny); // Achievements [Smokexyz/Hercules]
+	if (zeny > 0) {
```
dastgirp: if it's 0, do we even need to process above functions?
particularly:
```
sd->status.zeny -= zeny;
clif->updatestatus(sd,SP_ZENY);
```

PR #2248, opened 2018-10-03
Minor fixes to Asura Strike

### LINKED ISSUE #1239: Asura Strike
1. The direction changes upon using Asura Strike on a target, instead of facing your point of direction it will face on the target.
   Hercules Video:
   https://youtu.be/fE3ZpHP9hdo

2. When using Asura Strike in town (or not pvp map)  it will show "Skill has failed" twice.
   Hercules Video:
   https://youtu.be/AzeF21Jgm_g

3. On non-pvp maps (no damage): In official caster should be falling 3 cells apart from his current position but here in hercules only 2 cells on non-pvp maps.
Hercules Video:
https://youtu.be/vxUvu6pj1zk

guilherme-gm:

### Pull Request Prelude

### Changes Proposed

These are small fixes to how Asura Strike skill behaves. This fixes:
* After using Asura Strike the caster doesn't end up looking to its target anymore
* When Asura Strike fails, it doesn't shows "Skill failed" twice in chat
* When Asura Strike fails, the caster will now walk 3 cells from where he originally was (instead of 2)

**Affected Branches:** 
master

**Issues addressed:**
Fixes #1239

Special thanks to @kyeme

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -5613,6 +5613,8 @@ static int skill_castend_id(int tid, int64 tick, int id, intptr_t data)
 
 	// Use a do so that you can break out of it when the skill fails.
 	do {
+		bool is_asura = (ud->skill_id == MO_EXTREMITYFIST);
+
 		if(!target || target->prev==NULL) break;
 
 		if(src->m != target->m || status->isdead(src)) break;
@@ -5845,7 +5847,8 @@ static int skill_castend_id(int tid, int64 tick, int id, intptr_t data)
 			ud->skill_lv = ud->skilltarget = 0;
 		}
 
-		if (src->id != target->id)
+		// Asura Strike caster doesn't look to their target in the end
+		if (src->id != target->id && !is_asura)
 			unit->setdir(src, map->calc_dir(src, target->x, target->y));
 
 		map->freeblock_unlock();
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -5882,7 +5882,8 @@ static int skill_castend_id(int tid, int64 tick, int id, intptr_t data)
 				clif->slide(src,src->x,src->y);
 				clif->spiritball(src);
 			}
-			clif->skill_fail(sd, ud->skill_id, USESKILL_FAIL_LEVEL, 0, 0);
+			// "Skill Failed" message was already shown when checking that target is invalid
+			//clif->skill_fail(sd, ud->skill_id, USESKILL_FAIL_LEVEL, 0, 0);
 		}
 	}
 
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -5873,16 +5873,26 @@ static int skill_castend_id(int tid, int64 tick, int id, intptr_t data)
 		if (target && target->m == src->m) {
 			//Move character to target anyway.
 			int dir, x, y;
+			int dist = 3; // number of cells that asura caster will walk
+
 			dir = map->calc_dir(src,target->x,target->y);
-			if( dir > 0 && dir < 4) x = -2;
-			else if( dir > 4 ) x = 2;
-			else x = 0;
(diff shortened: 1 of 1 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2248)

src/map/skill.c
```diff
@@ -5831,7 +5833,8 @@ static int skill_castend_id(int tid, int64 tick, int id, intptr_t data)
 			ud->skill_lv = ud->skilltarget = 0;
 		}
 
-		if (src->id != target->id)
+		// Asura Strike caster doesn't look to its target in the end
+		if (src->id != target->id && is_asura == false)
```
dastgirp:
```suggestion
		if (src->id != target->id && !is_asura)
```

guilherme-gm:
@dastgirp can you check the updated version? thanks!

src/map/skill.c
```diff
@@ -5856,19 +5859,23 @@ static int skill_castend_id(int tid, int64 tick, int id, intptr_t data)
...
+			else if (dir > 4) x = dist;
 			else x = 0;
-			if( dir > 2 && dir < 6 ) y = -2;
-			else if( dir == 7 || dir < 2 ) y = 2;
+			if (dir > 2 && dir < 6) y = -dist;
+			else if (dir == 7 || dir < 2) y = dist;
 			else y = 0;
-			if (unit->movepos(src, src->x+x, src->y+y, 1, 1)) {
+
+			if (unit->movepos(src, src->x + x, src->y + y, 1, 1)) {
```
Kenpachi2k13: `unit->movepos(src, src->x + x, src->y + y, 1, 1) == 1`
or
`unit->movepos(src, src->x + x, src->y + y, 1, 1) != 0`

guilherme-gm:
@Kenpachi2k13 fixed and rebased. Thanks!

src/map/skill.c
skyleo: Alright. Please NEVER use `== 1` when correcting a condition that mistreats an int as a boolean, ~~UNLESS you're 100% sure that it will never be value besides 1, and even then~~ it's a bad idea, because custom code might expect this condition here to still work the same way it was before.

This should be `!= 0`.

skyleo: Opened an issue about this matter https://github.com/HerculesWS/Hercules/issues/2629, which shall be closed with a commit fixing this.

Kenpachi2k13: Since you highlighted me, I'll throw in my opinion about this...

I don't know where you see a mistreatment of the integer type here. The relation between 1/0 and true/false in C is just implicit. If `== true` was used, I'd agree with you.

`unit_movepos()` returns 0 on error and 1 on success, so checking for `== 1` is absolutely valid.
If custom code introduces a new return value, the one who adds it is responsible for validating all dependencies.

> There needs to be a compile error then.
How would you do this?

\**********
**But in the end, you are right. If `!= 0` is used, all possible side effects are excluded.** :+1:
\**********

guilherme-gm: Ah, true, i totally agree.. I didn't thought about custom code when I did the change, sorry...
I may open a PR fixing it later today (~12 hours from now), but if you guys want to ship a follow up release earlier, feel free to do it before me. Thanks!

skyleo: With mistreatment of integer type as bool I simply refer to implicit comparisons such as `if (some_int)` which makes you think that it's checking for `== true`.

Seems like I was really sleepy yesterday and saw a return 2 where there wasn't any. So there are at least no side-effects in our code-base, not that I know of any. 😅 
But yeah let's use `== 0` and `!= 0` for stuff like this when just fixing up implict comparison on non-bool types. Essentially chaging those comparisons and only looking at specific return values should be done in a separate commit at least, so that it's more noticeable.

Issue #203, opened 2013-10-24
@recallall suggestion


Streusel:
Either, modify @recallall or create a new command such as @recallmost.
The purpose of the new or modified command would be to disregard venders (buyer & seller) and chat room operators/users, possibly also users that are currently in a NPC chat as they might else get stuck when used during a quest, and teleport/warp everyone else to the person who used the command.

PR #2269, opened 2018-10-07
Minor logic change to @recallall to only warp players which can act

shouxian92:

### Pull Request Prelude

### Changes Proposed

Adds an additional conditional statement before setting a particular player's position. Re-used  `pc_cant_act` which covers all the conditions mentioned in the issue. In addition, it also checks for players who are active in the past 300 seconds.

Since i don't see any configuration variable storing a value that denotes something like -> **how many seconds before we determine a user is idle?** I'm not sure if I should introduce a new global variable since such a timeout is only used for this command.

**Affected Branches:** 
- master

**Issues addressed:**
-  https://github.com/HerculesWS/Hercules/issues/203

[//]: https://github.com/HerculesWS/Hercules/issues/203


doc/atcommands.txt
```diff
@@ -786,7 +786,7 @@ Additionally, @follow will track the player's movements until turned off.
 @recallall
 
 Warps a character to your position.
-Additionally, @recallall will recall the entire server.
+Additionally, @recallall will recall the entire server, except for players who are Vending, in an NPC dialog or in a Chatroom.
```
dastgirp: You forgot to include idle in description
```@recallall will recall entire server except players who are vending, interacting with NPC, in chatroom or idle```

File: src/map/atcommand.c [hunk 2]
```diff
@@ -3552,9 +3552,9 @@ ACMD(recallall)
 		if (sd->status.account_id != pl_sd->status.account_id && pc_get_group_level(sd) >= pc_get_group_level(pl_sd)) {
 			if (pl_sd->bl.m == sd->bl.m && pl_sd->bl.x == sd->bl.x && pl_sd->bl.y == sd->bl.y)
 				continue; // Don't waste time warping the character to the same place.
-			if (pl_sd->bl.m >= 0 && map->list[pl_sd->bl.m].flag.nowarp && !pc_has_permission(sd, PC_PERM_WARP_ANYWHERE))
+			if (pl_sd->bl.m >= 0 && map->list[pl_sd->bl.m].flag.nowarp && !pc_has_permission(sd, PC_PERM_WARP_ANYWHERE)) {
 				count++;
-			else {
+			} else if (!pc_cant_act(pl_sd) && DIFF_TICK(sockt->last_tick, (pl_sd)->idletime) <= 300) { // Perform the warp only if all the conditions are met (pc_cant_act + idletime lesser than 5 min)
```
dastgirp: add to count if they can't be warped?
since below it displays message ` Because you are not authorized to warp from some maps, %d player(s) have not been recalled.` It would be showing incorrect value if count is not increased

dastgirp:
> how many seconds before we determine a user is idle?

That depends on implementation perspective, we just record the last interaction in sd->idletime, and it depends on the developer to determine idletime for each mod been made (since they can be different). I don't think having a global configuration would be good idea for that

shouxian92:
Thanks for your comments @dastgirp, I've updated the PR. PTAL 😃

Emistry:
just a suggestion, maybe allow GM to self define the `active duration`? this way we wont need to hardcode the value of `300` in the source, and enable more flexibility to define idle duration.
Ex: `@recallall 60`
recall all players who hasn't been idle for the past 60 seconds.

shouxian92:
> just a suggestion, maybe allow GM to self define the `active duration`? this way we wont need to hardcode the value of `300` in the source, and enable more flexibility to define idle duration.
> Ex: `@recallall 60`
> recall all players who hasn't been idle for the past 60 seconds.

This sounds like a great suggestion. I'll probably go along the lines of -> let the GM define the number of seconds, else default to 300.

src/map/atcommand.c
```diff
@@ -3537,7 +3537,7 @@ ACMD(recallall)
 {
 	struct map_session_data* pl_sd;
 	struct s_mapiterator* iter;
-	int count;
+	int unauthorised_count, idle_count;
```
4144: better move this variabled definition into lines where they assigned first time. to line 3549

PR #2307, opened 2018-10-23
Added countnameditem script command.

kisuka:
### Pull Request Prelude

### Changes Proposed

Adds a script command to return the number of signed items a character has in their inventory.

Currently you have to do the following for this same functionality:

```
.@slot3 = getcharid(0) & 65535;
.@slot4 = getcharid(0) >> 16;
countitem2(<item ID>, 1, 0, 0, 254, 0, .@slot3, .@slot4)
```
This pull request changes that so you can simply do this:
```
countnameditem(<item ID>, {<character id/name>});
```

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -5340,6 +5340,30 @@ Check getitem2() to understand the arguments of the function.
 
 ---------------------------------------
 
+*countnameditem(<item id>, "<character name>")
+*countnameditem(<item id>, <character ID>)
+*countnameditem("<item name>", "<character name>")
+*countnameditem("<item name>", <character ID>)
+
+This function will return the number of signed items for the specified item ID
+that the invoking character has in their inventory.
+
+	mes("[Item Checker]");
+	if (countnameditem(Apple) > 0) {
+		mes("You have an apple with your name on it!");
+	} else {
+		mes("You do not have an apple with your name on it.");
+	}
+	next();
+	mes("[Item Checker]");
+	mesf("You have %d apples with John's name on it!", countnameditem(Apple, "John"));
+	close();
+
+Like getnameditem(), this function will also accept an 'english name' from the
+database as an argument.
+
+---------------------------------------
+
 *groupranditem(<item id>)
 
 Returns the item_id of a random item picked from the item container specified. There
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -7714,6 +7714,65 @@ static BUILDIN(countitem2)
 }
 
 /*==========================================
+ * countnameditem(item ID, { <Char Name / ID> })
+ * returns number of named items.
+ *------------------------------------------*/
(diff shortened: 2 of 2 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2307)

kisuka:
> could we have extra parameters for count named item of other specific char id that exists in player inventory too?
> example: count how many named `Apple` that exists in PlayerA inventory are named `PlayerB's Apple`?
> 
> ```
> countnameditem(<item ID>, {<character id/name>, <target_char_id/name>});
> ```
> use-case may not be wide, but seem useful if we could have it to count item that named using other player name in different player inventory.

Sure. I'll need to tweak the code a bit but should be possible.

kisuka:
Hmmm might be a bit more complex than I previously thought. So... even in the current state, the inscribed character has to be online for the count to be successful.

I'm not seeing any existing methods in script.c that allows the look up of offline character data. Should we possibly add a new function within pc.c that allows searching for a character's basic information even if they're offline?

src/map/script.c
```diff
@@ -7713,6 +7713,66 @@ static BUILDIN(countitem2)
...
+		id = itemdb->exists(script_getnum(st, 2));
+	}
+
+	if (id == NULL) {
+		ShowError("buildin_countnameditem: Invalid item '%s'.\n", script_getstr(st, 2));  // returns string, regardless of what it was
+		script_pushint(st, 0);
+		return false;
+	}
+
+	for (i = 0; i < MAX_INVENTORY; i++) {
```
4144: i better define inside for ``for (int i = ...``

kisuka: Updated

doc/script_commands.txt
```diff
@@ -5340,6 +5340,30 @@ Check getitem2() to understand the arguments of the function.
...
+
+	mes("[Item Checker]");
+	if (countnameditem(Apple) > 0) {
+		mes("You have an apple with your name on it!");
+	} else {
+		mes("You do not have an apple with your name on it.");
+	}
+	next;
+	mes("[Item Checker]");
+	mes("You have "+countnameditem(Apple, "John")+" apples with John's name on it!");
```
4144: If you want support for translations, you cant split sentenses.
Need use sprintf or other formatting functions.

Example: ``mes(sprintf(_$("You have %d apples with John's name on it!"), countnameditem(Apple, "John"))``

But this is docs, because this i not sure is it really need here.

dastgirp: Or 
```suggestion
mesf("You have %d apples with John's name on it!", countnameditem(Apple, "John"));
```

MishimaHaruna: I agree that docs should show the recommended code style, since this is where people learn/copy from. The suggested `mesf()` variant is the cleanest/most appropriate

PR #2309, opened 2018-10-25
Support for monster drops with Random Options

guilherme-gm:
### Pull Request Prelude

#### Change 1 : Support to random option drop
This PR adds support for dropping items that contains Random Options in them (used by some official features -- e.g: Crimsom Weapons).

A new database file was included: `db/option_drop_group.conf` that allows users to create groups of options with their rates.

These groups are then linked on `mob_db.conf` drop entries, I have changed a bit the Drops to provide an alternate syntax: `AegisName: (chance, "GROUP_NAME")`.

I didn't like how I coded the part that picks an option for a slot (I'll make a comment in this line), but I couldn't think of another way until now, so I'm open fo suggestions on that.

#### Change 2 : Moved mobs drop data to `mob_drop` structure (as suggested in PR review)

**Issues addressed:** 
I think there isn't one.

Thanks to @Ridley8819 for explaining me how this was supposed to work, @Asheraf and @MishimaHaruna for giving suggestions on the structure.


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2309)

src/map/mob.c
```diff
@@ -1903,12 +1903,41 @@ static int mob_ai_hard(int tid, int64 tick, int id, intptr_t data)
...
+			int idx = 0;
+			int min, max;
+			int count;
+
+			if (rnd() % 10000 >= options->options_rate[i])
+				continue;
+
+			// count avoids a too long loop that would cause lag. if after 10 full iterations it still fails
+			// it'll stop in one random index in the next one.
+			count = 10 * options->options[i].option_count + (rnd() % options->options[i].option_count);
```
guilherme-gm: I didn't like this count, I think it kind of breaks the idea of rates, but I couldn't really think of a better way. One idea was to make something like we do with boxes, where we put how many times that item appears in the set.

src/map/mob.h
```diff
@@ -160,8 +177,8 @@ struct mob_db {
 	short range2,range3;
 	short race2; // celest
 	unsigned short lv;
-	struct { int nameid,p; } dropitem[MAX_MOB_DROP];
-	struct { int nameid,p; } mvpitem[MAX_MVP_DROP];
+	struct { int nameid, p; struct mob_drop_option *options; } dropitem[MAX_MOB_DROP];
+	struct { int nameid,p; struct mob_drop_option *options; } mvpitem[MAX_MVP_DROP];
```
dastgirp: Maybe make new structure with these variables and use that structure here?

src/map/mob.c
```diff
@@ -5254,6 +5452,20 @@ static void mob_reload(void)
 	}
 	mob->item_drop_ratio_other_db->clear(mob->item_drop_ratio_other_db, mob->final_ratio_sub);
 
+	// Clear drops that has option info. [KirieZ]
+	for (i = 0; i < MAX_MOB_DB; i++) {
+		int j;
```
4144: j better define inside for at line 5460

src/map/mob.c
```diff
@@ -5302,8 +5514,19 @@ static int do_init_mob(bool minimal)
 static void mob_destroy_mob_db(int index)
 {
 	struct mob_db *data;
+	int i;
```
4144: i better define inside for at line 5523

tlacson7:
`        CC      HPMHooking.c (CHAR)
        PLUGIN  HPMHooking_char
        CC      HPMHooking.c (LOGIN)
        PLUGIN  HPMHooking_login
        CC      HPMHooking.c (MAP)
In file included from HPMHooking.c:228:0:
HPMHooking/HPMHooking_map.Hooks.inc: In function ‘HP_mob_setdropitem’:
HPMHooking/HPMHooking_map.Hooks.inc:51236:3: warning: passing argument 2 of ‘HPMHooks.source.mob.setdropitem’ makes pointer from integer without a cast [enabled by default]
   retVal___ = HPMHooks.source.mob.setdropitem(nameid, qty, data);
   ^
HPMHooking/HPMHooking_map.Hooks.inc:51236:3: note: expected ‘struct mob_drop_option *’ but argument is of type ‘int’
HPMHooking/HPMHooking_map.Hooks.inc:51236:3: warning: passing argument 3 of ‘HPMHooks.source.mob.setdropitem’ makes integer from pointer without a cast [enabled by default]
HPMHooking/HPMHooking_map.Hooks.inc:51236:3: note: expected ‘int’ but argument is of type ‘struct item_data *’
HPMHooking/HPMHooking_map.Hooks.inc:51236:3: error: too few arguments to function ‘HPMHooks.source.mob.setdropitem’
make[1]: *** [../../plugins/HPMHooking_map.so] Error 1
`

4144:
@tlacson7 this is probably because hpm hooks not updated.
open ``tools/HPMHookGen`` and run ``make`` after it will update hpm hooks, and you can try to build it

src/map/mob.c
```diff
@@ -1903,12 +1903,41 @@ static int mob_ai_hard(int tid, int64 tick, int id, intptr_t data)
...
 	drop->item_data.amount = qty;
 	drop->item_data.identify = data ? itemdb->isidentified2(data) : itemdb->isidentified(nameid);
+	
+	// Set item options [KirieZ]
+	if (options != NULL) {
+		int i;
+		
+		for (i = 0; i < options->options_count; i++) {
+			int idx = 0;
+			int min, max;
```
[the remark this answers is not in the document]
guilherme-gm: `min` and `max` are used in the same scope, if I declare them on line 1935/1936, wouldn't it break some compilers? I recall that was the reason for us to declare all variables at the beggining of a block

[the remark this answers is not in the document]
4144: not anymore. visual studio 2002 was already dropped. even new visual studio still not support all C features, but vars now you can define in any place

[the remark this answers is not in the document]
Asheraf: @guilherme-gm the last compiler supported by hercules that didn't have C99 variable declaration style was MSVC 11 (Visual Studio 2012) which was dropped back in v2018.07.29

src/map/mob.c
```diff
@@ -3997,6 +4026,164 @@ static void mob_read_db_mvpdrops_sub(struct mob_db *entry, struct config_setting
...
+	//	up to MAX_ITEM_OPTION
+	//)
+
+	if (!libconfig->setting_lookup_int(drop, "Rate", value)) {
+		ShowWarning("mob_read_db_drop_options: Missing item \"%s\" drop rate in monster %d, skipping.\n", name, entry->mob_id);
+		*value = 0;
+		return NULL;
+	}
+
+	drop_option = (struct mob_drop_option *) malloc(sizeof(struct mob_drop_option));
```
Asheraf: use `aMalloc` macro here since that helps with statistics.

guilherme-gm:
Hello everybody, sorry for the delay.
Recently I was talking to @Ridley8819 and we noticed that this structure will make mob_db file even bigger than it already is, so I'm thinking on moving the option settings to a separate (`option_drop_group.conf`) file, this not only would reduce the file size when options get added, but also allows us to reuse the same group of options. The structure would be like that:

mob_db.conf:
```
Drops: {
    AegisName: chance
    // or
    AegisName: [Drop Rate, Option group]
}

// Example:
Drops: {
    Apple: 7000
    Knife: [1000, OPT_GROUP_SOMEITEMS] // So knife will use OPT_GROUP_SOMEITEMS from option_drop_group
}
```

option_drop_group: (structure that is in this PR already)
```
{
    Group: "Group_Name"
    Options: (            options per slot
        {
            Rate: chance of option 1  (int)

            // Option possibilites for this slot
            // min/max value : int, defaults to 0
            // rate : int, 100 = 1% if not set, will be 100%/number of possibiltiies
            OptionName1: value
            // or
            OptionName1: [min value, max value]
            // or
            OptionName1: [min value, max value, rate]
            // ...  (up to MAX_MOB_DROP_OPTIONS)
        },
        // ... (up to MAX_ITEM_OPTION)
    )
}

// Example:
{
    Group: "OPT_GROUP_SOMEITEMS"
    Options: (        // options per slot
        {
            Rate: 5000
            RACE_TOLERACE_DEVIL: 10000
        },
    )
}
```

what you guys think?

AnnieRuru:
yes having a separate group structure is better

just want to make sure ... this thing is entirely custom ... unofficial right ?
like .... we also don't support item drop with equipment refined ... from source

guilherme-gm:
Official servers does have drops with random options (e.g. Crimsom Weapons), but the implementation itself is solely based on the information I could find (which isn't much). So I made it as customizable as possible to try to cover whatever is needed once those items are implemented.

Asheraf:
@guilherme-gm Groups would be nice :)

AnnieRuru:
oh ok, http://www.playragnarok.com/news/updatedetail.aspx?id=280&p=1
its under Episode 15.2

asdf

Ps: ~I actually thought it has something to do with makeitem3 ... seems I was wrong~
btw all makeitem2 in hercules/npc are slash-out

guilherme-gm:
PR updated, I did some major changes to support the extra db file and changed a bit how it is stored (I believe this way will save more memory when compared to the old approach), I tried to remember all the changes requested while fixing it and hopefully I haven't forgot anything. Thanks for your reviews :)

src/map/mob.c
```diff
@@ -3860,6 +3900,192 @@ static inline int mob_parse_dbrow_cap_value(int class_, int min, int max, int va
...
+	if (count <= 1) { // 1 = Rate
+		ShowWarning("mob_read_optdrops_optslot: Option %d of %s group doesn't contain any possible options, skipping.\n", n + 1, group);
+		return false;
+	}
+
+	struct optdrop_group_optslot *entry = &(mob->opt_drop_groups[group_id].optslot[n]);
+	entry->options = aCalloc(sizeof(struct optdrop_group_option), count);
+	
+	int idx = 0,
+		i = 0;
```
4144: look like extra tab here

src/map/mob.c
```diff
@@ -3860,6 +3900,192 @@ static inline int mob_parse_dbrow_cap_value(int class_, int min, int max, int va
...
+	struct config_setting_t *opt = NULL;
+	bool calc_rate = false;
+	while (i < count && (opt = libconfig->setting_get_elem(optslot, i))) {
+		++i;
```
4144: better extract code from inside while into separate function

src/map/mob.c
```diff
@@ -3940,6 +4166,35 @@ static uint32 mob_read_db_mode_sub(struct mob_db *entry, struct config_setting_t
...
+/**
+ * Process an entry of mob/mvp drops that contains a random option drop group.
+ * 
+ * @param entry : mob db entry being read (used in error messages)
+ * @param item_name : AegisName of the item in this entry (used in error messages)
+ * @param drop : drop data entry
+ * @param drop_rate : used to return the entry drop_rate
+ * @returns a reference to the opt_drop_group to be used when creating this item drop
+ */
+static struct optdrop_group *mob_read_db_drops_option(struct mob_db *entry, const char *item_name, struct config_setting_t *drop, int * drop_rate)
```
4144: change `` int * drop_rate `` to ``int *drop_rate``

src/map/mob.h
```diff
@@ -152,6 +152,36 @@ struct spawn_info {
...
+	int max; //< Maximun value when this option drops
+	int rate; //< Chance of dropping this option
+};
+
+/**
+ * Information of possible options that will fill
+ * one option slot
+ */
+struct optdrop_group_optslot {
+	uint8 option_count; //< Number of options in *options
```
4144: probably better here use int

src/map/mob.h
```diff
@@ -152,6 +152,36 @@ struct spawn_info {
...
+	struct optdrop_group_option *options; //< Array of possible options
+};
+
+/**
+ * A group of options to be random picked when
+ * dropping an item
+ */
+struct optdrop_group {
+	uint8 optslot_count; //< How many option slots are configured by this group
```
4144: and here int too

4144:
according ci, server crashing with memory manager

db/option_drop_groups.conf
```diff
@@ -0,0 +1,53 @@
...
+			OptionName: [min value, max value]
+			// or
+			OptionName: [min value, max value, chance]
+			// ...  (as many as you want)
+		},
+		// ... (up to MAX_ITEM_OPTION)
+	),
+**************************************************************************/
+}
+)
```
AnnieRuru: no new line ?

doc/option_drop_group.md
```diff
@@ -0,0 +1,97 @@
...
+		// ...  (as many as you want)
+	},
+	// ... (up to MAX_ITEM_OPTION)
+),
+```
+
+### `Group Name Constant`
+This is the group name, it is how this group is referenced in other files
+(e.g. mob_db). It must be globally unique, as it is a server constant, and
+must contain only letters, numbers and "_".
```
AnnieRuru: Github Markdown error
underscore _ will change text italic
just adding a space will remove the italic font

src/map/mob.c
guilherme-gm: like that?

4144: while in old function, but all inside while in new one. (probably ++i still in old function)
this need for simpler code and for plugins to allow change something. for example change only reading min, max fields or add custom fields here and without changing anything else.

guilherme-gm: I think it is correct now

src/map/mob.c
```diff
@@ -3860,6 +3900,212 @@ static inline int mob_parse_dbrow_cap_value(int class_, int min, int max, int va
...
+ * @param entry : memory db entry for current slot
+ * @param idx : index of entry where this option should be inserted at
+ * @param calc_rate : if rates should be recalculated after reading all entries
+ * @param slot : option group slot being read (for messages)
+ * @param group : option group being read (for messages)
+ * @return true if it successfully read the entry, false otherwise
+ */
+static bool mob_read_optdrops_option(struct config_setting_t *option, struct optdrop_group_optslot *entry, int *idx, bool *calc_rate, int slot, const char *group)
+{
+	nullpo_retr(false, option);
```
4144: here need check all pointers parameters

src/map/mob.c
```diff
@@ -3965,9 +4238,18 @@ static void mob_read_db_mvpdrops_sub(struct mob_db *entry, struct config_setting
 			i++;
 			continue;
 		}
-		if (mob->get_const(drop, &i32) && i32 >= 0) {
-			value = i32;
+		
+		struct optdrop_group *drop_option = NULL;
```
hemagx: move it to the top of the loop, where other variables are declared.

src/map/mob.c
```diff
@@ -3965,9 +4238,18 @@ static void mob_read_db_mvpdrops_sub(struct mob_db *entry, struct config_setting
...
+		if (config_setting_is_number(drop)) {
+			// Setting is a number, item doesn't contain options
+			if (mob->get_const(drop, &i32) && i32 >= 0) {
```
hemagx: What happens if i32 is a negative value? perhaps throwing an error and put a default value.

src/map/mob.c
```diff
@@ -4024,16 +4307,26 @@ static void mob_read_db_drops_sub(struct mob_db *entry, struct config_setting_t
 			i++;
 			continue;
 		}
-		if (mob->get_const(drop, &i32) && i32 >= 0) {
-			value = i32;
+
+		struct optdrop_group *drop_option = NULL;
```
hemagx: move it to the top of the loop, where other variables are declared.

src/map/mob.c
```diff
@@ -4024,16 +4307,26 @@ static void mob_read_db_drops_sub(struct mob_db *entry, struct config_setting_t
...
+		if (config_setting_is_number(drop)) {
+			// Setting is a number, item doesn't contain options
+			if (mob->get_const(drop, &i32) && i32 >= 0) {
```
hemagx: What happens if i32 is a negative value? perhaps throwing an error and put a default value.

src/map/mob.c
guilherme-gm: hmm, I am a bit confused now, on an earlier review it was requested to move the declaration to where it is used, and now the opposite?

4144: old way was correct. now we not using outdated visual studio, and code should follow good style

4144: separate declaration is useless and error prone. as many as possible vars must be used and smaller scope

src/map/mob.c
```diff
@@ -1900,15 +1900,52 @@ static int mob_ai_hard(int tid, int64 tick, int id, intptr_t data)
...
+		int idx = 0;
+		while (count > 0 && rnd() % 10000 >= options->optslot[i].options[idx].rate) {
+			idx = (idx + 1) % options->optslot[i].option_count;
+			--count;
+		}
+
+		item->option[i].index = options->optslot[i].options[idx].id;
+
+		int min = options->optslot[i].options[idx].min;
+		int max = options->optslot[i].options[idx].max;
```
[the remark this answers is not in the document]
4144: if vars used only here or/and below, declaration must be here too

src/map/mob.c
```diff
@@ -3860,6 +3896,214 @@ static inline int mob_parse_dbrow_cap_value(int class_, int min, int max, int va
...
+	if (count <= 1) { // 1 = Rate
+		ShowWarning("mob_read_optdrops_optslot: Option %d of %s group doesn't contain any possible options, skipping.\n", n, group);
+		return false;
+	}
+
+	struct optdrop_group_optslot *entry = &(mob->opt_drop_groups[group_id].optslot[n]);
+	entry->options = aCalloc(sizeof(struct optdrop_group_option), count);
+	
+	int idx = 0,
+	    i = 0;
```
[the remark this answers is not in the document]
4144: this is wrong. here extra tab only

src/map/mob.c
```diff
@@ -3860,6 +3896,214 @@ static inline int mob_parse_dbrow_cap_value(int class_, int min, int max, int va
...
+	// Structure:
+	//	{
+	//		Rate: chance of option 1 (int)
+	//		OptionName1: value
+	//		OptionName2: [min, max]
+	//		OptionName3: [min, max, rate]
+	//      ....
+	//	}
+
+	int drop_rate; // The rate for this option to be dropped (Rate field)
```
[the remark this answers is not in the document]
4144: wrong

src/map/mob.c
```diff
@@ -3860,6 +3896,214 @@ static inline int mob_parse_dbrow_cap_value(int class_, int min, int max, int va
...
+	if (!libconfig->setting_lookup_int(optslot, "Rate", &drop_rate)) {
+		ShowWarning("mob_read_optdrops_optslot: Missing option %d rate in group %s, skipping.\n", n, group);
+		return false;
+	}
+
+	int count = libconfig->setting_length(optslot);
```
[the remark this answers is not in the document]
4144: wrong too

src/map/mob.c
```diff
@@ -3860,6 +3896,214 @@ static inline int mob_parse_dbrow_cap_value(int class_, int min, int max, int va
...
+	if (count <= 1) { // 1 = Rate
+		ShowWarning("mob_read_optdrops_optslot: Option %d of %s group doesn't contain any possible options, skipping.\n", n, group);
+		return false;
+	}
+
+	struct optdrop_group_optslot *entry = &(mob->opt_drop_groups[group_id].optslot[n]);
```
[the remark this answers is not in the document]
4144: @hemagx please dont add bad comments. and try check current hercules style

src/map/mob.c
```diff
@@ -3860,6 +3896,214 @@ static inline int mob_parse_dbrow_cap_value(int class_, int min, int max, int va
...
+	}
+	entry->option_count = idx;
+	mob->opt_drop_groups[group_id].optslot_count++;
+	mob->opt_drop_groups[group_id].optslot_rate[n] = drop_rate;
+	
+	// If there're empty rates, calculate them
+	if (calc_rate == true) {
+		int j = 0;
+
+		for (j = 0; j < idx; ++j) {
```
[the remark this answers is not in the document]
4144: here correct comment. scope should be as smaller as possible

src/map/mob.c
```diff
@@ -1900,15 +1900,52 @@ static int mob_ai_hard(int tid, int64 tick, int id, intptr_t data)
...
+	nullpo_retv(item);
+	nullpo_retv(options);
+	
+	for (int i = 0; i < options->optslot_count; i++) {
+		if (rnd() % 10000 >= options->optslot_rate[i])
+			continue;
+
+		// count avoids a too long loop that would cause lag. if after 10 full iterations it still fails
+		// it'll stop at one random index in the next iteration.
+		int count = 10 * options->optslot[i].option_count + (rnd() % options->optslot[i].option_count);
```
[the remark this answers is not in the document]
guilherme-gm: The comment says 10 full iterations + random:
`if after 10 full iterations it still fails` = `10 * options->optslot[i].option_count`
`it'll stop at one random index in the next iteration` = `+ (rnd() % options->optslot[i].option_count)`

so, 10 full iterations (a full iteration = running through all indexes) + a random index.

Maybe rephrasing it? I'm open for suggestions

I like the idea of moving to a battle config, will do that.

src/map/mob.c
guilherme-gm: `value` by default is 0 (a couple lines above), if `i32` is negative, `value` will stay on 0, and on line 4253 (a couple lines below) there is a check for value <= 0 where it throws a warning that this drop chance is invalid. Keeping it like that will handle both if/else cases of `is_number` check in the same place.

src/map/mob.c
guilherme-gm: same as in 4245, being negative keeps `value` to 0 and throws a warning on line 4322.

src/map/mob.c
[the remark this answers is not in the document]
hemagx: @4144 that's not what I'm talking about, there's no reason to have those variables in first place.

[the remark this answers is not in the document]
guilherme-gm: @hemagx doesn't it improve readability? I actually decided to make them because writting them 3 times would make that line below quite polluted and hard to read:

```
item->option[i].value = min + (rnd() % (max - min + 1));
```
vs:
```
item->option[i].value = options->optslot[i].options[idx].min + (rnd() % (options->optslot[i].options[idx].max - options->optslot[i].options[idx].min + 1));
```

I could use 2 lines too:
```
item->option[i].value = options->optslot[i].options[idx].min +
                        (rnd() % (options->optslot[i].options[idx].max - options->optslot[i].options[idx].min + 1));
````

File: src/map/mob.c [hunk 26]
```diff
@@ -3860,6 +3896,214 @@ static inline int mob_parse_dbrow_cap_value(int class_, int min, int max, int va
...
+	
+	entry->options[*idx].id = opt_id;
+	entry->options[*idx].min = min;
+	entry->options[*idx].max = max;
+	entry->options[*idx].rate = opt_rate;
+
+	if (entry->options[*idx].rate == 0)
+		*calc_rate = true;
+
+	(*idx)++;
```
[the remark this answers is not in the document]
guilherme-gm: I think it is needed, according to: https://en.cppreference.com/w/c/language/operator_precedence
`*idx++` will be parsed as `*(idx++)` (which is wrong)

just in case, I tried and if I remove the parenthesis I get compiler warnings about value not being used (which makes sense, as it would be simply taking the value in the address `idx+1` and doing nothing)

src/map/mob.c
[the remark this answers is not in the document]
guilherme-gm: I rephrased the comment and added the battle conf, hopefully it is ok now

src/map/mob.c
```diff
@@ -3940,6 +4184,35 @@ static uint32 mob_read_db_mode_sub(struct mob_db *entry, struct config_setting_t
...
+ */
+static struct optdrop_group *mob_read_db_drops_option(struct mob_db *entry, const char *item_name, struct config_setting_t *drop, int *drop_rate)
+{
+	nullpo_retr(NULL, entry);
+	nullpo_retr(NULL, item_name);
+	nullpo_retr(NULL, drop);
+	nullpo_retr(NULL, drop_rate);
+
+	// [Drop Rate, Option Group]
+	*drop_rate = libconfig->setting_get_int_elem(drop, 0);
```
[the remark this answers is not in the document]
guilherme-gm: Changed to read the same way it is read on normal drops without option group, if it is still invalid, it will be detected/handled like it always did

src/map/mob.c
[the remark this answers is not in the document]
4144: first way also should be faster (with min, max vars)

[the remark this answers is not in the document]
AnnieRuru: https://github.com/HerculesWS/Hercules/pull/1012#discussion_r48851863
Haru did say its ok to do it for improve readability

I also after improve readability rather than performance sometimes,
because having a high readability scripts can save human's time

guilherme-gm:
@hemagx @4144 @Asheraf can you review it again? thanks!

src/map/mob.c
```diff
@@ -3860,6 +3901,209 @@ static inline int mob_parse_dbrow_cap_value(int class_, int min, int max, int va
...
+ * @param n : slot index
+ * @param group_id : Group index
+ * @param group : group name (used in messages)
+ * @return true if it succesfully read, false otherwise
+ */
+static bool mob_read_optdrops_optslot(struct config_setting_t *optslot, int n, int group_id, const char *group)
+{
+	nullpo_retr(false, optslot);
+	nullpo_retr(false, group);
+
```
4144: need add here Assert_ret for group_id or check group_id before used as array index

4144: same for n

src/map/mob.c
```diff
@@ -3940,6 +4184,47 @@ static uint32 mob_read_db_mode_sub(struct mob_db *entry, struct config_setting_t
...
+		ShowError("mob_read_db_optdrops: Missing option drop group name on item \"%s\" in monster %d, skipping.\n", item_name, entry->mob_id);
+		return NULL;
+	}
+
+	int opt_id;
+	if (script->get_constant(group_name, &opt_id) == false) {
+		ShowError("mob_read_db_optdrops: Invalid option drop group \"%s\" on item \"%s\" in monster %d, does this group really exists? Skipping...\n", group_name, item_name, entry->mob_id);
+		return NULL;
+	}
+
```
4144: here need check opt_id for valid range from 0 to size of opt_drop_groups - 1

PR #2319, opened 2018-11-13
Remove minicore

MishimaHaruna:
### Pull Request Prelude

### Changes Proposed

This PR removes the (currently unused, and with no plans for future use) "minicore" feature, and all the related complications that it required in the makefiles. Compilation time on systems using the GNU Makefiles should also be marginally improved, since they were needlessly compiling `common_mini.a` even if it wasn't linked by any tool. Xcode and Visual Studio aren't affected, since they weren't compiling the minicore at all.

The last minicore user was the old mapcache tool, which has been removed in #1552 (almost two years ago). If we'll need something similar in the future, it should probably be redesigned in a different way (i.e. by making the source files in src/common more modular).


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2319)

PR #2332, opened 2018-12-04
Fix sealed bapho problem

carloshenrq:
### Pull Request Prelude

### Changes Proposed
There is a situation where you can invoke more than one sealed Baphomet.

To reproduce:

1. Go to the second floor of catacombs and before you summon, as party leader, stop on the dialog "The bottom of the Main Altar trembles furiously.".
2. Change party leader to another member, so, he can talk with the seal in the middle too. He stops on the same dialog "The bottom of the Main Altar trembles furiously."
3. Keep doing it until all party is on the same dialog. All of them proceeds the dialog to invoke the sealed Baphomet, and we'll have a lot of sealed Baphomet unsealed.

So, this PR is to don't allow players to abuse the instance using this way.
I've just added instance status check and if the character is truly the party leader.

**Issues addressed:** 
Related to instances #1582, not directly reported.

### DIFF
--- a/npc/instances/SealedShrine.txt
+++ b/npc/instances/SealedShrine.txt
@@ -1108,13 +1108,15 @@ OnInstanceInit:
 		next();
 		mes("The bottom of the Main Altar trembles furiously.");
 		next();
-		specialeffect(EF_METEORSTORM);
-		specialeffect(EF_METEORSTORM);
-		mesf("[%s]", strcharinfo(PC_NAME));
-		mes("Watch out! Something... Something is coming.");
-		'ins_baphomet = 6;
-		donpcevent(instance_npcname("ins_2f_hero_broad")+"::OnEnable");
-		disablenpc(instance_npcname("The Main Altar#ss"));
+		if ('ins_baphomet == 5 && getpartyleader(.@party_id, 2) == getcharid(CHAR_ID_CHAR)) {
+			specialeffect(EF_METEORSTORM);
+			specialeffect(EF_METEORSTORM);
+			mesf("[%s]", strcharinfo(PC_NAME));
+			mes("Watch out! Something... Something is coming.");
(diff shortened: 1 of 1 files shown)


carloshenrq:
@Asheraf or @4144 can you guys review it? Thank you =]

npc/instances/SealedShrine.txt
```diff
@@ -1108,13 +1108,15 @@ OnInstanceInit:
...
 		mes("The bottom of the Main Altar trembles furiously.");
 		next();
-		specialeffect(EF_METEORSTORM);
-		specialeffect(EF_METEORSTORM);
-		mesf("[%s]", strcharinfo(PC_NAME));
-		mes("Watch out! Something... Something is coming.");
-		'ins_baphomet = 6;
-		donpcevent(instance_npcname("ins_2f_hero_broad")+"::OnEnable");
-		disablenpc(instance_npcname("The Main Altar#ss"));
+		if ('ins_baphomet == 5 && getpartyleader(.@party_id, 2) == getcharid(CHAR_ID_CHAR)) {
```
Emistry: since you already have the instance variable checking, the party leader checking should be no longer necessary as it only will trigger once during the instance.
```suggestion
		if ('ins_baphomet == 5) {
```

dastgirp:
Better yet, we should disable change party leader in instance maps, can anyone check official behaviour?

Asheraf:
^ as dastgir said, changing party leader inside the instance should not be possible.

carloshenrq:
Block changing party leader doesn't fix the problem. If I'm allowed changing party leader in the instance... the problem persists. Anyway, do whatever you guys want.

MishimaHaruna:
Merging this, regardless of whether we're going to block changing party leaders - an additional sanity check in the script won't hurt.

PR #2355, opened 2019-01-22
Allows BUILDIN(equip) to specify an inventory index to equip.

EyesOfAHawk:
### Pull Request Prelude

### Changes Proposed
Adds an extra **'index'** argument to BUILDIN(equip).
i.e. `equip(<item id>{, <index>})`

The reason for this change is it allows you to the option to choose the specific inventory index to equip. Of course, you need to know the inventory index.

Currently I am using this in a mass refiner script. Since you need to equip an item to refine it, I was having trouble due to the equip() command always equipping the lowest index (I believe that's what it was doing).

Here is a sample in which all the different versions of 'Knife' are cycled through and equipped every 1 second.
```
.@id = Knife;
getinventorylist();
for (.@i = 0; .@i < @inventorylist_count; .@i++) {
	if (@inventorylist_id[.@i] == .@id) {
		.@Index[.@j] = .@i;
		.@j++;
	}
}

.@count = countitem(.@id);
for (.@i = 0; .@i < .@count; .@i++) {
	equip(.@id, .@Index[.@i]);
	sleep2(1000);
}
end;
```

**Issues addressed:** 
N/A


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2355)

AnnieRuru:
I think don't need, `*equip2` already seems enough

>Currently I am using this in a mass refiner script`

 ... this is not an official feature ...

>Since you need to equip an item to refine it, I was having trouble due to the equip() command always equipping the lowest index (I believe that's what it was doing)

I am sure `*successrefitem` is meant to upgrade the equipped item ...



well, just my 2 cents

PS: when you say mass refiner script, I was thinking about using delitem2/getitem2
hmmm ... we don't have getitem3 ??

4144:
i not sure probably better create command ``equipIdx``? Because if you know index, id is useless

AnnieRuru:
erm ... but if the item already refined, it went from knife +0 into knife +10
isn't the index changed ? because its entirely different item

4144:
if get list non refined items before start refining, item with this it cant be refined in this loop two times.

I mean one item refine do remove/add item, and it reusing existing item index or adding item in new index, it not affect other items in one loop iteration

4144:
Another option for refine without add/remove item is abusing packet ZC_ENCHANT_EQUIPMENT script command enchantitem.
Normally it only add/remove cards in equipped item. But i can abuse it for change almost any items properties. Is this feature should be added?

AnnieRuru:
that sounds AWESOME ...

but wait, can it change the ItemID as well ?
like change from knife into Spear ...? without deleting existing item ?
seems like it can cause weight issue or something
```
struct PACKET_ZC_ENCHANT_EQUIPMENT {
	int16 packetType;
	int16 wearState;
	int16 cardSlot;
#if PACKETVER_MAIN_NUM >= 20181121 || PACKETVER_RE_NUM >= 20180704 || PACKETVER_ZERO_NUM >= 20181114
	int32 itemId;
#else
	int16 itemId;
#endif
	int8 equipFlag;
} __attribute__((packed));
```
enlighten me, but I didn't see the refine flag there, nor the random option ...

4144:
item id i think can be changed too. But i not sure is client update all things if change id.
As i said before, for change other fields than cards need abuse client features.
With item options bit hard, because need send many packets for change them.
All magic happend in client if use offset smaller or bigger than 0 to 4.

EyesOfAHawk:
After successrefitem(), it seems the item doesn't change index and I can continue to loop through all them in my inventory.

doc/script_commands.txt
```diff
@@ -5622,6 +5625,37 @@ Examples:
...
+//This will attempt to equip the item at inventory index 45.
+	equipidx(45);
+
+//This will attempt to equip a 1201 (Knife) every 1 second until it
+//has checked every inventory item.
+	getinventorylist();
+	for (.@i = 0; .@i < ; .@i++) {
+		if (@inventorylist_id[.@i] == Knife) {
+			equipidx(.@i, Knife);
+			sleep2(1000);
```
AnnieRuru: this is consider a trick, and shouldn't include in the documentation
inside the documentation has to teach proper scripting methods,
so no sleep2 inside a refiner script
example: what happens if the player disconnect during the refining process ?

Just remove the examples will do

`for (.@i = 0; .@i < ; .@i++) {` <--- missing `@inventorylist_count` ... nonono just remove this example

src/map/script.c
```diff
@@ -16473,21 +16473,28 @@ static BUILDIN(unequip)
...
+	if (sd == NULL) {
+		script_pushint(st, 0);
+		return true;
+	}
+
+	nameid = script_getnum(st, 2);
+	if ((item_data = itemdb->exists(nameid)) == NULL) {
+		ShowError("buildin_equip: Invalid Item ID (%d).\n",nameid);
+		script_pushint(st, 0);
+		return true;
```
AnnieRuru: `return false;`

the condition above, (sd == NULL) has to `return true;` because `script->rid2sd` has `script->reportsrc(st);` in it

src/map/script.c
```diff
@@ -16515,6 +16522,44 @@ static BUILDIN(autoequip)
...
+	if (i < sd->status.inventorySize) {
+		if (script_hasdata(st, 3))
+			nameid = script_getnum(st, 3);
+		else
+			nameid = sd->status.inventory[i].nameid == nameid;
+
+		if ((item_data = itemdb->exists(nameid)) == NULL) {
+			ShowError("buildin_equipidx: Invalid Item ID (%d).\n", nameid);
+			script_pushint(st, 0);
+			return true;
```
AnnieRuru: `return false;`

src/map/script.c
```diff
@@ -16515,6 +16522,44 @@ static BUILDIN(autoequip)
...
+		}
+
+		if (sd->status.inventory[i].nameid == nameid && sd->status.inventory[i].equip == 0) {
+			script_pushint(st, pc->equipitem(sd, i, item_data->equip));
+			return true;
+		} else {
+			ShowDebug("buildin_equipidx: Item ID (%d) either not found at index (%d) or cannot be equipped.\n", nameid, i);
```
AnnieRuru: ShowWarning

EyesOfAHawk:
I added a check so index isn't out of range. Is this correct or unnecessary?

```
	if (i < 0 || i >= sd->status.inventorySize) {
		ShowError("buildin_equipidx: Index (%d) should be from 0-%d.\n", i, sd->status.inventorySize);
		script_pushint(st, 0);
		return false;
	}
```

4144:
check is correct

src/map/script.c
```diff
@@ -16515,6 +16515,53 @@ static BUILDIN(autoequip)
 	return true;
 }
 
+/**
+ * equipidx(<index>{, <item id>})
+ */
+static BUILDIN(equipidx)
+{
+	int nameid = 0, i = script_getnum(st, 2);
+	struct item_data *item_data;
```
4144: you can move item_data variable inside if, because it used only inside is

src/map/script.c
```diff
@@ -16515,6 +16515,53 @@ static BUILDIN(autoequip)
...
+		return true;
+	}
+
+	if (i < 0 || i >= sd->status.inventorySize) {
+		ShowError("buildin_equipidx: Index (%d) should be from 0-%d.\n", i, sd->status.inventorySize - 1);
+		script_pushint(st, 0);
+		return false;
+	}
+
+	if (i < sd->status.inventorySize) {
```
4144: here check useless, because i was checked before

src/map/script.c
```diff
@@ -16515,6 +16515,53 @@ static BUILDIN(autoequip)
...
+		if (script_hasdata(st, 3))
+			nameid = script_getnum(st, 3);
+		else
+			nameid = sd->status.inventory[i].nameid == nameid;
```
4144: here look like error. Probably you want left only code ``nameid = sd->status.inventory[i].nameid``

src/map/script.c
```diff
@@ -16515,6 +16515,53 @@ static BUILDIN(autoequip)
 	return true;
 }
 
+/**
+ * equipidx(<index>{, <item id>})
+ */
+static BUILDIN(equipidx)
+{
+	int nameid = 0, i = script_getnum(st, 2);
```
4144: nameid can be moved close to usage (inside if)

src/map/script.c
```diff
@@ -16515,6 +16522,53 @@ static BUILDIN(autoequip)
...
+
+		if (sd->status.inventory[i].nameid == nameid && sd->status.inventory[i].equip == 0) {
+			script_pushint(st, pc->equipitem(sd, i, item_data->equip));
+			return true;
+		}
+
+		if (sd->status.inventory[i].nameid != nameid)
+			ShowWarning("buildin_equipidx: Item ID (%d) not found at index (%d) (found Item ID %d).\n", nameid, i, sd->status.inventory[i].nameid);
+		else if (sd->status.inventory[i].equip != 0)
+			ShowWarning("buildin_equipidx: Item ID (%d) at index (%d) cannot be equipped.\n", sd->status.inventory[i].nameid, i);
```
AnnieRuru: I just tested this in-game, I have a priest, trying to equip a knife, it runs this line

IT IS OK to have this check,
but we don't have something like `@inventorylist_canequip` or `*getequipcanequipthis`
so should we suppress this warning ?

AnnieRuru:
another things concern me is assassin class can equip weapons in both hands, 
accessory can equip in EQI_ACC_L or EQI_ACC_R

hmm ... if you couldn't do it, can leave at it is, after all nobody actually complain about this issue

EyesOfAHawk:
Thanks for the input guys. I am busy for a few days I will update once I'm free.

EyesOfAHawk:
Updated. Though I made this bit `if (pc->equipitem(sd, i, item_data->equip) == 0) {` return true instead of false, since I don't see reason to stop script execution if it failed to equip (we're already returning 0 in that instance).

src/map/script.c
```diff
@@ -16522,6 +16522,48 @@ static BUILDIN(autoequip)
...
+	if (item_data == NULL) {
+		ShowError("buildin_equipidx: Invalid Item ID (%d).\n", nameid);
+		script_pushint(st, 0);
+		return false;
+	}
+
+	if (pc->equipitem(sd, i, item_data->equip) == 0) {
+		ShowWarning("buildin_equipidx: Item ID (%d) at index (%d) cannot be equipped.\n", nameid, i);
+		script_pushint(st, 0);
+		return true;
```
AnnieRuru: wrong, `return true;` or `return false;` is almost the same thing
`return false;` is equal to `script->reportsrc(st);` + `return true;`
read this PR https://github.com/HerculesWS/Hercules/pull/868

what you are doing here is put up a ShowWarning without telling where this warning came from
should be `return false;` here

============== OR =================

suppress this warning, since we don't have a script command to check either this job can equip this item or not

src/map/script.c
```diff
@@ -16473,21 +16473,28 @@ static BUILDIN(unequip)
...
-	{
-		ShowError("wrong item ID : equipitem(%d)\n",nameid);
+	if (sd == NULL) {
+		script_pushint(st, 0);
+		return true;
+	}
+
+	nameid = script_getnum(st, 2);
+	if ((item_data = itemdb->exists(nameid)) == NULL) {
+		ShowError("buildin_equip: Invalid Item ID (%d).\n",nameid);
```
AnnieRuru: `(%d).\n", nameid);`
forget a space ?

src/map/script.c
EyesOfAHawk: so to be clear, it will not stop execution of script correct

EyesOfAHawk:
Okay done. Let me know if anything else is needed.

doc/script_commands.txt
```diff
@@ -5622,6 +5625,24 @@ Examples:
...
+
+Returns 1 if the item was successfully equipped.
+Returns 0 if the item failed to equip.
+
+See getinventorylist() for an explanation on retrieving an inventory index.
+
+Examples:
+
+//This will attempt to equip the item at inventory index 45.
+	equipidx(45);
```
AnnieRuru: bad example 
```
//	This will prioritize equip a rental Knife
	getinventorylist();
	for (.@i = 0; .@i < @inventorylist_count; ++.@i) {
		if (@inventorylist_id[.@i] == Knife && @inventorylist_expire[.@i]) {
			equipidx(.@i);
			break;
		}
	}
```

src/map/script.c
```diff
@@ -16473,21 +16473,28 @@ static BUILDIN(unequip)
...
-	nameid=script_getnum(st,2);
-	if((item_data = itemdb->exists(nameid)) == NULL)
-	{
-		ShowError("wrong item ID : equipitem(%d)\n",nameid);
+	if (sd == NULL) {
+		script_pushint(st, 0);
+		return true;
+	}
+
+	nameid = script_getnum(st, 2);
```
4144: nameid can be defined here
``int nameid = script_getnum(st, 2);``

AnnieRuru:
```
/*=================================================
 * Checks if the player can equip the item at index n in inventory.
 * Returns 0 (no) or 1 (yes).
 *------------------------------------------------*/
static int pc_isequip(struct map_session_data *sd, int n)
```
try introduce a new script command `*canequip` to check if that item can be equip ?

doc/script_commands.txt
```diff
@@ -5622,6 +5625,30 @@ Examples:
...
+
+See getinventorylist() for an explanation on retrieving an inventory index.
+
+Examples:
+
+//	This will prioritize equip a rental Knife
+	getinventorylist();
+	for (.@i = 0; .@i < @inventorylist_count; ++.@i) {
+		if (@inventorylist_id[.@i] == Knife && @inventorylist_expire[.@i]) {
+			equipidx(.@i);
```
AnnieRuru: thanks to dastgir idea, that just push another array of the index,
that means this example has to change accordingly

```suggestion
 		equipidx(@inventorylist_idx[.@i]);
```

PR #2358, opened 2019-01-26
Fix show_mob_info not working

AnnieRuru:
### Pull Request Prelude

### Issues addressed
https://rathena.org/board/topic/118076-monster-level-not-showing/

### Changes Proposed
Fix show_mob_info not working

### Affected Branches
* Master

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -9334,7 +9334,7 @@ static void clif_charnameack(int fd, struct block_list *bl)
 				//can parse it. [Skotlex]
 				if (str_p != mobhp) {
 					*(str_p-3) = '\0'; //Remove trailing space + pipe.
-					memcpy(packet.party_name, mobhp, NAME_LENGTH);
+					safesnprintf(packet.name, NAME_LENGTH, "%s %s", md->name, mobhp);
 				}
 			}
 		}


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2358)

Emistry:
Since its fixing the `show_mob_info`, do you think its possible to include some enhancement?
add new option to display other info like `element / race / size` using the guild name area for monster?

AnnieRuru:
@dastgirp hahaha ~ didn't test this thoroughly

@Emistry doesn't seems can do like that now,
because using party_name and guild_name area for monsters doesn't seems to work
maybe we should find out which packet change causing this problem
and the name area only has 23 characters ....

dastgirp:
@AnnieRuru are you sure it's not because of /showname?

AnnieRuru:
ok be honest, I didn't know about the /showname

without this patch, just type /showname will display correctly
so how do we actually solve this with default newer client without typing /showname ?

Emistry:
> @Emistry doesn't seems can do like that now,
because using party_name and guild_name area for monsters doesn't seems to work
maybe we should find out which packet change causing this problem

seem this done in the SEAR emulator, seem workable .... I did once long time ago, but it look ugly hence i dropped it and hope someone with better knowledge could do it xD

their format 
```
Monster Name (HP/MaxHP)   <--- Name/Party column
[Element: XXX] (Race:XXXX)    <--- Guildname / Position column
```

PR #2367, opened 2019-02-06
Add new server define constant MAX_ITEM_ID

AnnieRuru:
### Pull Request Prelude

### Issues addressed
we use to do like this
```
prontera,155,185,5	script	kjsdkf	1_F_MARIA,{
	mes "input Item ID";
	next;
	if ( input( .@itemid, 501, 32768 ) ) { // item ID was smallint(6)
//	if ( input( .@itemid, 501, MAX_ITEM_ID ) ) {
		mes "invalid range";
		close;
	}
	if ( getitemname( .@itemid ) == "null" ) {
		mes "invalid Item ID";
		close;
	}
	mes "OK";
	close;
}
```

### Changes Proposed
add MAX_ITEM_ID as the limit of Item ID is no longer 32768
https://github.com/HerculesWS/Hercules/blob/stable/src/map/itemdb.h#L41

### Affected Branches
* Master

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -716,6 +716,7 @@ MAX_BANK_ZENY		- Maximum Zeny in the bank
 MAX_BG_MEMBERS		- Maximum BattleGround members
 MAX_CHAT_USERS		- Maximum Chat users
 MAX_REFINE			- Maximum Refine level
+MAX_ITEM_ID			- Maximum Item ID
 MAX_MENU_OPTIONS    - Maximum NPC menu options
 MAX_MENU_LENGTH     - Maximum NPC menu string length
 
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -25843,6 +25843,7 @@ static void script_hardcoded_constants(void)
 	script->set_constant("MAX_BG_MEMBERS",MAX_BG_MEMBERS,false, false);
 	script->set_constant("MAX_CHAT_USERS",MAX_CHAT_USERS,false, false);
 	script->set_constant("MAX_REFINE",MAX_REFINE,false, false);
+	script->set_constant("MAX_ITEM_ID",MAX_ITEM_ID,false, false);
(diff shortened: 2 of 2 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2367)

4144:
Item id limit can be from any to 2 147 483 648.
In your selected line for new clients it set to  0x20000 (131 072), because bigger values for now useless.
For old clients it defined some lines below as 0xFFFF aka 65535.

About pr, yes adding this constant can be good

AnnieRuru:
SOMETHING NOT RIGHT

see this line
....EDITING....

4144:
MAX_INVENTORY depend on packet version and in docs it can be in two versions at same time

AnnieRuru:
I hope there is a way to silently change the constants.md file in your computer though
like when compiling ... it change these values in your computer ... or something
... just suggestion

Emistry:
I think the [doc/constants.md](https://github.com/HerculesWS/Hercules/blob/stable/doc/constants.md) are just a general list of available constants and sample value for reference.
Most people would probably just refer to this [db/constants.conf](https://github.com/HerculesWS/Hercules/blob/stable/db/constants.conf) instead.

I never once checked the value or the variables i need from the [doc/constants.md](https://github.com/HerculesWS/Hercules/blob/stable/doc/constants.md).
Especially those server defines which the server owner should have know themselves when they setup/compile the server.
Just saying.

4144:
constants.md can be also regenerated with constdb2doc plugin

AnnieRuru:
LMAO !! just notice the `constdb2doc` plugin
so it means everyone needs to load this plugin to make an update to the `constants.md` file
just tested, it works

@Emistry ,
I have `constants.md` file open all the time because this file also contains item/mob db constants
btw you seems like out from scripting for a few years now,
rathena script engine has changed a lot though

PR #2386, opened 2019-02-25
Add config for party leave behavior

EyesOfAHawk:
### Pull Request Prelude

### Changes Proposed

_Note: Was unsure of the best way to go about this._

Some time ago the behavior was changed for when a party leader leaves the party. I feel a config option should have been added to retain the old behavior.

**Old:** Disband the party.
**Current:** Pass the party leader to another player.
**New:** Uncomment `#define PARTY_LEADER_LEAVE` in `src/config/classes/general.h`

Perhaps others can input on a better way to implement a config option.

**Issues addressed:** 

N/A


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2386)

AnnieRuru:
imo, should move this config into conf folder

although our `party_share_level` is define inside `conf/common/inter-server.conf`

src/config/classes/general.h
```diff
@@ -49,6 +49,12 @@
  **/
 //#define DEVOTION_REFLECT_DAMAGE
 
+/**
+ * when enabled, parties will be disbanded if its leader leaves the party
+ * uncomment to enable
+ **/
+//#define PARTY_LEADER_LEAVE
```
4144: better add configuration option for this

AnnieRuru:
actually I'm thinking, why not just add an atcommand for it ?
we already have `party` `changeleader` `partyoption`  ...
so just add `partydisband` something like that ...

EyesOfAHawk:
> actually I'm thinking, why not just add an atcommand for it ?
> we already have `@party` `@changeleader` `@partyoption` ...
> so just add `@partydisband` something like that ...

Yeah that sounds better.

Emistry:
`@partydisband` doesn't sound like it solve the issue. 
It create confusion with the party disband/leaving behaviour.
it could end up `@partydisband` and `/leave` work differently: Imagine you end up with such behaviour
```
@partydisband - leave party and remove the party (originally should behave like /leave)
/leave - leave party (and currently it would assign leader to another party member)
```
not to mention if the party manipulation script commands are planned to be added in future, it create unnecessary workaround to handling both behaviour simultaneously.

beside if player want to change/switch party leader, they could just use the `@changeleader` atcommand ?

So, adding a configuration in `battle/conf/party.conf` would most likely a more feasible option here, to switch the party disband behaviour.

EyesOfAHawk:
I changed it to a battleconf option. Let me know what you think.

src/char/int_party.c
```diff
@@ -447,42 +447,54 @@ static bool inter_party_change_option(int party_id, int account_id, int exp, int
...
-	p->size--;
-	if (j == p->min_lv || j == p->max_lv || p->family) {
-		if(p->family) p->family = 0; //Family state broken.
-		inter_party->check_lv(p);
+	if (p->party.member[i].leader && option == 1) {
+		p->party.member[i].account_id = 0;
+		for (j = 0; j < MAX_PARTY; j++) {
+			if (!p->party.member[j].account_id)
+				continue;
+			mapif->party_withdraw(party_id, p->party.member[j].account_id, p->party.member[j].char_id);
```
AnnieRuru: what happens if the party member is offline ?
isn't that ... this configuration should just disband the party ?
you still kick them one by one though ... and expect them all to be online ...

### tested
this modification doesn't check offline party members
if I have member A leave party while member B is offline, it shows a message for disband the party
but in-fact, the party leader pass to member B

EyesOfAHawk: thats why i said i dont understand it completely. need help. can you take over PR?

PR #2388, opened 2019-02-26
Add optional parameter for *gettimestr

AnnieRuru:
### Pull Request Prelude

### Issues addressed
we have `*getcalendartime` script command, but why our `*gettimestr` isn't update for it?

### Changes Proposed
Add optional parameter for `*gettimestr`

### Tested with
```
prontera,155,185,5	script	sdkfjhsdkf	1_F_MARIA,{
	mes(gettimestr("%Y-%m/%d %H:%M:%S", 21));
	mes(gettimestr("%Y-%m/%d %H:%M:%S", 21, getcalendartime(0,0)));
	close;
}
```
which is stated in the documentation

### Affected Branches
* Master

none
in fact this is mostly a copy-paste from rathena repo

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -3536,9 +3536,9 @@ Examples :
 This function will return a string containing time data as specified by
 the format string.
 
-This uses the C function 'strfmtime', which obeys special format
+This uses the C function 'strftime', which obeys special format
 characters. For a full description see, for example, the description of
-'strfmtime' at http://www.delorie.com/gnu/docs/glibc/libc_437.html
+'strftime' at http://www.delorie.com/gnu/docs/glibc/libc_437.html
 All the format characters given in there should properly work.
 Max length is the maximum length of a time string to generate.
 
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -3531,7 +3531,7 @@ Examples :
 
 ---------------------------------------
 
-*gettimestr(<format string>, <max length>)
+*gettimestr(<format string>, <max length>{, <timestamp>})
 
 This function will return a string containing time data as specified by
 the format string.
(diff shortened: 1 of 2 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2388)

AnnieRuru:
curious, isn't that after using `Malloc` should always use `Free` ?
in this script command, has `aMalloc` but doesn't have `aFree` ... isn't that will cause memory leak ?

4144:
if string next pushed to script engine as string and not cont, script engine will free it if need remove from memory

AnnieRuru:
hmm ... for some reason this isn't merge ...

Emistry:
btw, cant we make the script command auto calculate the `max length` parameter? by default its just the same length of the `format` parameter. Aren't this `max length` parameter currently look redundant and not necessary?

AnnieRuru:
hmm ... the person who made this script command probably has die hard habit from C language
`char message[CHAT_SIZE_MAX];`

then how do you want to modify it ? this is actually a copy paste from rathena, **literally**
https://github.com/rathena/rathena/blob/78eed02273414a4332f9c088ad7e795a2903ea5f/doc/script_commands.txt#L3155

if we make the `<max length>` optional parameter, that would become ...
`*gettimestr(<format string>{, <max length>{, <timestamp>})`
means still have to input the max length ... 
or ...
`*gettimestr(<format string>{, <max length>})`
`*gettimestr(<format string>{, <timestamp>})`
both fields are INT type, its not like integer/string that we could switch it easily

Emistry:
> some script already using it...

that doesn't meant we aren't allow to change it or update it? Both rA/Herc and any other emulators has done it countless time. So its not a restriction.

what I meant in previous post are totally remove the `max length` field.
End Result: `*gettimestr(<format string>{, <timestamp>})`

For example: changing the code to auto determine the length ?
```
fmtstr=script_getstr(st,2);
maxlen=script_getnum(st,3);
```
into
```
fmtstr=script_getstr(st,2);
maxlen=(int)strlen(fmtstr);
```

> changing this behavior silently is big no-no ...

that's why changelog/commit history exists for a reason. For user to read/keep track of what has been changed. Every milestone release come with a changelog for each release.

AnnieRuru:
just noticed the documentation says `strfmtime`
when I ask google, says -> Do you mean `strftime` ?
LOL

PR #2389, opened 2019-02-26
Add *msgtxt script command to read from conf/message.conf

AnnieRuru:
### Pull Request Prelude

### Changes Proposed
Add `*msgtxt` script command to read a line from `conf/message.conf`

### Tested with
```c
prontera,155,185,5	script	sdkfjhsdkf	1_F_MARIA,{
	input .@a, 0, 1000;
	#CASHPOINTS += .@a;
	mesf(msgtxt(505), .@a, #CASHPOINTS); // Gained 100 cash points. Total 100 points.
	dispbottom sprintf( msgtxt(505), .@a, #CASHPOINTS ), C_SPRINGGREEN;
	msgtxt(5000);
	close;
}
```

### Affected Branches
* Master

### Additional context
we have HULD, which allows to members to ~machine~ translate their npc files
this should probably reduce the work load to translate part of the npc stuffs

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -2825,6 +2825,13 @@ If the character is sitting, it will return true, otherwise (standing) it will r
 In case no player is specified, the function will return the state of the attached player.
 
 ---------------------------------------
+
+*msgtxt(<number>)
+
+This function return the string defined from 'conf/message.conf'.
+return "??" if not found.
+
+---------------------------------------
 //=====================================
 2.1 - Item-Related Commands
 //=====================================
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -24920,6 +24920,19 @@ static BUILDIN(msgtable2)
 	return true;
 }
 
+static BUILDIN(msgtxt)
+{
+	int num = script_getnum(st, 2);
+	
+	if (num < 0 || num >= MAX_MSG) {
+		ShowWarning("buildin_msgtxt: Invalid range %d. Min: %d Max: %d\n", num, 0, MAX_MSG);
+		return false;
+	}
+
+	script_pushstrcopy(st, msg_txt(num));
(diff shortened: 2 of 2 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2389)

AnnieRuru:
After giving it some thought, I think this might be a bad implementation
because this might encourage members to add their own ID into `conf/message.conf` file

and if WE developers introduce new atcommand, and use up the free space,
these people that already having their own CUSTOM message will clash with ours

@Asheraf what do you think about it ?

Asheraf:
@AnnieRuru good point, resolving conflict in messages ids can be annoying, lets wait for @MishimaHaruna review.

MishimaHaruna:
I agree with your point @AnnieRuru - I never really liked how messages.conf is indexed by ID, and the mess it causes when a new value is added to it if you had customized it (i.e. you'll have to search through the entire code for that number, to make sure that you replace it with a different ID everywhere).

If the purpose of this `msgtxt()` is just to use existing message IDs (in order to use the same string that's already in use somewhere in the source), then I think we can merge this as is. I can see some use cases for this (especially when implementing some atcommands with scripts, or when showing `dispbottom()` or similar messages to the user as part of login scripts and such.

If, instead, the purpose is to let people move to messages.conf all their their NPC strings, then absolutely not, let's rethink this over. The messages.conf design is pretty bad for this kind of use.

AnnieRuru:
accidentally miss-click the milestone

yes, the original plan was to help reduce the workload for those people who use HULD heavily
however this current design is flawed ... 

at this moment I dunno how to make a better design than this system ...
close this one then ...

maybe some time later when the new HULD come out, we may come out with better design

PR #2391, opened 2019-02-26
setunitdata/getunitdata clean up. Phase 1

AnnieRuru:
[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style
### Pull Request Prelude
- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR will be closed if the above-mentioned criteria are not fulfilled.

### Issues addressed
I am so depress to see these 2 script commands broken in Hercules ...
NO KIDDING, 70% of them are broken ...

### Changes Proposed
1. Deprecate `UDT_MAPIDXY` constant 
2. Deprecate `UDT_WALKTOXY` constant 
3. Remove all the `(int)` in the documentation

### Affected Branches
* Master

### Known Issues and TODO List
I made a FATAL mistakes for having 1 BIG Pull Request like this ->https://github.com/HerculesWS/Hercules/pull/2080
and having nobody review my work

so this time I'm breaking into MULTIPLE Pull Request and hopefully ... somebody will take notice
https://github.com/HerculesWS/Hercules/compare/stable...AnnieRuru:34-setunitdata

PR #2395, opened 2019-02-27
Deprecate *getunitname script command

AnnieRuru:
After I give it some thought, I think better split the `*getunitname` deprecation out from https://github.com/HerculesWS/Hercules/pull/2391

-----------------------------------

### Pull Request Prelude

### Issues addressed
we have 2 script command do the same thing, `*rid2name` and `*getunitname`

### Changes Proposed
deprecate `*getunitname` because rid2name has been exist since eathena times
also minor fix rid2name to support elementals

### Affected Branches
* Master

we have to decide which one to deprecate,
better don't have 2 script commands do exact same thing again, like this https://github.com/HerculesWS/Hercules/pull/2089

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -10165,9 +10165,16 @@ returns -1 if value could not be retrieved.
 
 *getunitname(<GID>)
 
+    @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+    @ /!\ This command is deprecated @
+    @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+
 Retrieve the name of a unit.
 
 returns "Unknown" if the value could not be retrieved.
+
+This command is deprecated and it should not be used in new scripts, as it is
+likely to be removed at a later time. Please use rid2name instead.
 
 ---------------------------------------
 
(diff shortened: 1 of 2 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2395)

EyesOfAHawk:
getunit... commands would be more consistent in documentation, as there are many commands of that name and they can be grouped together.

Plus we already set the precedent by deprecating `*specialeffect2()` which was used in many official scripts.

Asheraf:
I don't feel deprecating `getunitname()` is a good idea, It's more popular than `rid2name()` and consistent with the other data retrieval commands.

AnnieRuru:
..... I was lost in making getmercinfo/getpetinfo/gethominfo/geteleminfo,

actually I want to keep both

getunitname will be real time name, and rid2name will be database name

example 1.
if player having `@fakename`,
getunitname will display fakename
rid2name will display real character name

example 2
`monster("this", -1,-1,"BOSSS",PORING,1)`
getunitname will display `BOSSS`
rid2name will display `Poring`

I read through the source code and found getunitname call the function status_get_name,
which was use in clif ..... to display in client-side

.... another thing worth mention
DO NOT use `getnameditem Knife, getunitname(getcharid(CHAR_ID_ACCOUNT));`
because if the player under `@fakename`,
`*getunitname` script command will return fakename of the character
has to use rid2name


_..... ok go back complete my getpetinfo script command ...._

PR #2413, opened 2019-03-25
(WIP) Add skill_check_invalid_level for plugin

AnnieRuru:
### Pull Request Prelude

### Issues addressed
` skill TF_HIDING, -1;` <-- actually gives you hiding skill level 255

### Changes Proposed
Add skill_check_invalid_level for plugin
http://herc.ws/board/topic/16616-portal-skill/?do=findComment&comment=91018

### Affected Branches
* Master

none

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -13961,6 +13961,14 @@ static bool skill_is_combo(int skill_id)
 	return false;
 }
 
+static int skill_check_invalid_level(int skill_id, int skill_lv)
+{
+	if (skill_lv < 1 || skill_lv > MAX_SKILL_LEVEL)
+		return 1;
+
+	return 0;
+}
+
 static int skill_check_condition_castbegin(struct map_session_data *sd, uint16 skill_id, uint16 skill_lv)
 {
 	struct status_data *st;
@@ -14154,7 +14162,7 @@ static int skill_check_condition_castbegin(struct map_session_data *sd, uint16 s
 			}
 	}
 
-	if( skill_lv < 1 || skill_lv > MAX_SKILL_LEVEL )
+	if (skill->check_invalid_level(skill_id, skill_lv) == 1)
 		return 0;
 
 	require = skill->get_requirement(sd,skill_id,skill_lv);
@@ -21554,6 +21562,7 @@ void skill_defaults(void)
 	skill->cast_fix_sc = skill_castfix_sc;
 	skill->vf_cast_fix = skill_vfcastfix;
 	skill->delay_fix = skill_delay_fix;
+	skill->check_invalid_level = skill_check_invalid_level;
 	skill->check_condition_castbegin = skill_check_condition_castbegin;
 	skill->check_condition_castend = skill_check_condition_castend;
 	skill->consume_requirement = skill_consume_requirement;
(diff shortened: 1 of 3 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2413)

4144:
you know, what server may have buffer overflow issue if remove or change this check?
probably better add two functions one check for valid skill level if need check only level, and other checks for levels array bound if need access this array

AnnieRuru:
> you know, what server may have buffer overflow issue if remove or change this check?

well, plugin = custom modification ... means we can do anything NOT official
if buffer overflow, we can again use another plugin to fix this issue lol
Note -> this PR is official, check only skill level 1-10

> probably better add two functions one check for valid skill level if need check only level, and other checks for levels array bound if need access this array

eh ? in plugin we can do like
```c
	addHookPre( skill, check_invalid_level, skill_check_invalid_level_pre );
```
```c
int skill_check_invalid_level_pre( int skill_id, int skill_lv ) {
	if ( skill_id == 2993 ) {
		hookStop();
		return 0;
	}
	return 0;
}
```
since already push the skill_id in the function already

or maybe I didn't get your meaning ?

PS: just noticed in this PR I check 0-10, it should be 1-10, fixed

4144:
I mean for example in plugin you set max level to 11 or 100, if you start server with this it may crash or corrupt memory. Level checks for reason because calculated values stored in array with size 10 only, and skill level here is index to access this arrays

4144:
and I suggest add another function what validate level if need access to this arrays

AnnieRuru:
you mean like some array use things like
skill[skill_lv].value = something ...
and skill_lv above 10 will crash server ?

ok, now I understand your problem, but I still don't get the way how to solve it
seems you have the idea, but I don't understand your technique
can give an example ?

4144:
validate level variable just before it access level as index in array. You can search where MAX_SKILL_LEVEL used in arrays in headers. Search for string ``[MAX_SKILL_LEVEL]``
And before each usage of this arrays need check is index correct

4144:
hm, i not checked code. In some places it already validated. If used macro skill_get_lvl_idx for convert level to index, this places already safe

4144:
For example in ``skill_get_range``
But need check all herc code what this arrays never used directly as ``array[skill_level]``

src/map/skill.c
```diff
@@ -13961,6 +13961,14 @@ static bool skill_is_combo(int skill_id)
 	return false;
 }
 
+static int skill_check_invalid_level(int skill_id, int skill_lv)
```
dastgirp: use bool instead of int?

AnnieRuru:
closing this, now I understand what 4144 means (last year don't understand)

https://github.com/HerculesWS/Hercules/commit/3119f6e757ec0c0b43a312954a5a3c09b27bbaf3
yes, hercules source code actually has protection against custom MAX_SKILL_LEVEL
for example on the topic I link, instead of doing a source edit, the actual way is like 4144 and haru did, 
**change MAX_SKILL_LEVEL in skill.h and recompile**
then this PR is totally invalid,
because coding a custom skill with max level 50 without changing MAX_SKILL_LEVEL will corrupt the memory

PR #2419, opened 2019-04-01
Dehardcode limits for attack related calculated parameters

4144:
### Pull Request Prelude

### Changes Proposed
Return from attack related status functions int variable.
Move min and max limits for attack related functions into battle/limits.conf


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2419)

AnnieRuru:
shouldn't all go for `signed int`, or all go for `unsigned int` ?
I think `signed int` is the way to go ...

4144:
ah sorry, i set limits same as it was before my changes

4144:
now should be real int/uint limits

4144:
about signed or unsigned, i left type as it was before my changes. i not sure is if change to signed can add issues or not

4144:
sorry was another issue with limits. configurable options allowed only in signed int limits.
Now pr should works fine

src/map/status.c
```diff
@@ -4448,7 +4448,7 @@ static int status_base_amotion_pc(struct map_session_data *sd, struct status_dat
 	return amotion;
 }
 
-static unsigned short status_base_atk(const struct block_list *bl, const struct status_data *st)
+static unsigned int status_base_atk(const struct block_list *bl, const struct status_data *st)
```
MishimaHaruna: What do you think about changing all of these to just `int`? The only reason why they were unsigned was to double their range (which wasn't a great idea to begin with), and that's not necessary anymore now.
For values that go into arithmetic expressions, I'd rather have signed variables, so that compilers won't think we want a modulo operation, and will be more keen on emitting warnings if there's a risk of overflows or underflows (and will avoid to do unsafe unsigned promotions in arithmetic expressions involving them)

4144: I left types as it was before but yes i think signed is better.
What about limits. allow negative values too?

MishimaHaruna: I can't think of an use case for negative values in the limits, I'd keep positive even if signed (unless there are cases I can't think of?)

4144: for base stats negative probably impossible, but if continue same limits to other stats, may be some calculated stat can have negative bonus.
but main question is here safe allow negative values at all?

MishimaHaruna: Ah I see your point - yeah, it's possible that in some stats a negative value could make sense. On the libconfig and the battle_config sides, there is no problem supporting negative numbers. Whether that's fine for the actual battle code, that's a much more difficult question, I'd not risk it.

For the settings that are part of this PR, I think a lower cap of 0 makes sense, I'd keep it this way.

4144: what you think about variables itself in status_data. should this vars left as ``unsigned int`` or ``int``?

MishimaHaruna: Sorry, I missed this message before .I'd personally change everything that gets used in arithmetic calculations to int, so that includes the vars in status_data

4144: ok, then this is probably for other pr

PR #2424, opened 2019-04-08
Storage item list packet size fix and minor clif.c fixes

hemagx:
### Pull Request Prelude

### Changes Proposed
1. Fixed a packet size underflow in storage list sending which started with client supporting int32 for item id and added static assert to detect this problem if ever happened again.
2. Fixed direct calls to clif_send in clif.c
3. Add missing nullpo and assert to clif_send

**Issues addressed:** 
None

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -2977,28 +2977,22 @@ static void clif_storageItems(struct map_session_data *sd, enum inventory_type t
 	nullpo_retv(sd);
 	nullpo_retv(items);
 
-	int i = 0;
-	struct item_data *id;
+	for (int i = 0, normal_count = 0, equip_count = 0; i < items_length; ++i) {
+		struct item_data *itd;
 
-	do {
-		int normal = 0, equip = 0, k = 0;
+		if (items[i].nameid == 0)
+			continue;
 
-		for( ; i < items_length && k < 500; i++, k++ ) {
+		itd = itemdb->search(items[i].nameid);
 
-			if( items[i].nameid <= 0 )
-				continue;
+		if (!itemdb->isstackable2(itd))
+			clif->item_equip(i + 1, &storelist_equip.list[equip_count++], &items[i], itd, itd->equip);
+		else
+			clif->item_normal(i + 1, &storelist_normal.list[normal_count++], &items[i], itd);
 
-			id = itemdb->search(items[i].nameid);
-
-			if( !itemdb->isstackable2(id) ) //Non-stackable (Equippable)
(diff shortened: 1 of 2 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2424)

src/map/clif.c
```diff
@@ -2980,10 +2980,16 @@ static void clif_storageItems(struct map_session_data *sd, enum inventory_type t
 	int i = 0;
 	struct item_data *id;
 
+#define STORAGE_ITEMS_MAX_ITERATION 500
```
4144: probably better on the fly calculate max possible items per packet for current packet version?
and move this define to one of headers?

hemagx: That would be better to calculate for each packer version indeed, the reason i did not put the define in header is because i want to keep the scope of the define for this function only.

hemagx: I will rewrite the function to be more flexible.

hemagx:
@4144 Please check the new code, i have not tested it yet, but i will test it and rebase the branch tomorrow after my work shift ~

hemagx:
> 
> 
> can you also fix storage names?
> Current code strip last byte from name
> need use name len: `NAME_LENGTH + 1`

Is not NAME_LENGTH = (23 + 1), it should be enough to use it? no? or the client doesn't require null-terminator anymore?

4144:
ah sorry, issue was in ZC_INVENTORY_START, but i thinked in clif_storageItems

4144:
i will fix ZC_INVENTORY_START by self

hemagx:
Okay splendid!

hemagx:
@4144 Fixed some minor issues and rebased, please check he new code.

src/map/clif.c
```diff
@@ -2977,28 +2977,22 @@ static void clif_storageItems(struct map_session_data *sd, enum inventory_type t
...
+	for (int i = 0, normal_count = 0, equip_count = 0; i < items_length; ++i) {
+		struct item_data *itd;
 
-			if( items[i].nameid <= 0 )
-				continue;
+		if (items[i].nameid == 0)
+			continue;
 
-			id = itemdb->search(items[i].nameid);
+		itd = itemdb->search(items[i].nameid);
```
4144: better put here also ``struct item_data *itd = ...``

4144: better change to ``struct item_data *itd = itemdb->search(items[i].nameid);``

PR #2426, opened 2019-04-09
Expand *getinventorylist() script command - favorite field

Emistry:
### Pull Request Prelude

- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
- return an array `@inventorylist_favorite`
- indicate whether an item in the inventory are located inside favorite tab or not.

**Issues addressed:** 
Allow npc script to detect if an item are located in favorite tab. 

Sample:
```
prontera,155,181,5	script	sample	4_F_KAFRA1,{
	getinventorylist;
	mes "Favourite Items";
	for (.@i = 0; .@i < @inventorylist_count; .@i++) {
		if (@inventorylist_favorite[.@i]) 
			mes "~ "+ getitemname(@inventorylist_id[.@i]);
	}
	close;
}
```
sample

[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style

PR #2394, opened 2019-02-27
Adds BUILDIN(delitemidx), which deletes an item at the given index.

bWolfie:
### Pull Request Prelude

- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

Adds new script command delitemidx().
`delitem(<index>{, <account id>})`

An important command which will allow scripters to delete an item at a specific index. Why is this important? Now we can check an item for cards, if it's a rental and other useful information `getinventorylist()` provides us.

**Issues addressed:** 

Currently if you have a quest which has an item requirement which can also be a Rental on your server, it will just eat up that rental item with no regard. This can be problematic if that rental item is a recolour, effectively allowing you to make the item with a rental.

**Hypothetical example:**
- Recolouring Beret (Black, Blue, Red, Yellow) quest requires Beret and 10,000z.
- You can rent a Beret for 1z.
- If you give it a rental Beret, you can still complete the quest. Now you can get basically unlimited recoloured Berets as it only costs 1z to rent.

[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style

PR #2427, opened 2019-04-09
Add *setfavoriteitemidx & *autofavoriteitem script commands - alter item favorite state

Emistry:
### Pull Request Prelude

### Changes Proposed
- `setfavoriteitemidx()` set an item as favorite item or not
- if an item is set to favorite item, it will be moved into favorite
tab, else move out from favorite tab.
- only non-equipped item can adjust the favorite item state.
- `autofavoriteitem()` will auto set an item as favorite state or not whenever obtained. (affect globally)

-----------------------------

Sample Script: `autofavoriteitem(<itemID>, <flag>)`
<details>

```
prontera,155,181,4	script	Sample	4_F_KAFRA1,{
	.@item_id = 909;
	mes getitemname(.@item_id);
	// mes "auto favorite: "+getiteminfo2(.@item_id, 21);
	.@flag = select("disable favorite", "enable favorite") - 1;
	autofavoriteitem(.@item_id, .@flag);
	close;
}
```
</details>

Sample Script: `setfavoriteitemidx(<idx>, <flag>)`
<details>

```
prontera,155,181,4	script	Sample	4_F_KAFRA1,{
	do {
		getinventorylist;
		.@menu$ = "";
		for (.@i = 0; .@i < @inventorylist_count; .@i++) {
			.@c = (@inventorylist_card1[.@i] > 0) + (@inventorylist_card2[.@i] > 0) + (@inventorylist_card3[.@i] > 0) + (@inventorylist_card4[.@i] > 0);
			.@menu$ += "[idx- "+@inventorylist_idx[.@i]+"] " + getitemname(@inventorylist_id[.@i]);
			if (.@c) 
				.@menu$ += "["+.@c+" cards]";
			.@menu$ += " "+ (@inventorylist_favorite[.@i] ? "^0055FF- favorite^000000":"") + ":";
		}
		.@i = select(.@menu$) - 1;
		// based on inventory index
		setfavoriteitemidx(@inventorylist_idx[.@i], !@inventorylist_favorite[.@i]);
	} while (true);
	close;
}
```

`@inventorylist_favorite[.@i]` [from this PR](https://github.com/HerculesWS/Hercules/pull/2426), but not really dependant on it. Since it can be just set with the value `0 or 1`. Bonus if have it.
</details>

**Issues addressed:** 
None

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -24681,6 +24681,49 @@ static BUILDIN(consolemes)
 	return true;
 }
 
+static BUILDIN(setfavoriteitemidx)
+{
+	struct map_session_data *sd = script_rid2sd(st);
+	int idx = script_getnum(st, 2);
+	int value = script_getnum(st, 3);
+
+	if (sd == NULL) {
+		ShowError("buildin_setfavoriteitemidx: No player attached.\n");
+		return false;
+	}
+
+	if (idx < 0 || idx >= sd->status.inventorySize) {
+		ShowError("buildin_setfavoriteitemidx: Invalid inventory index %d (min: %d, max: %d).\n", idx, 0, (sd->status.inventorySize - 1));
+		return false;
+	} else if (sd->inventory_data[idx] == NULL || sd->inventory_data[idx]->nameid <= 0) {
+		ShowWarning("buildin_setfavoriteitemidx: Current inventory index %d has no data.\n", idx);
+		return false;
+	} else if (sd->status.inventory[idx].equip > 0) {
+		ShowWarning("buildin_setfavoriteitemidx: Cant change favorite flag of an equipped item.\n");
+		return false;
+	} else {
+		sd->status.inventory[idx].favorite = cap_value(value, 0, 1);
+		clif->favorite_item(sd, idx);
+	}
+
+	return true;
+}
+
+static BUILDIN(autofavoriteitem)
+{
+	int nameid = script_getnum(st, 2);
+	int flag = script_getnum(st, 3);
+	struct item_data *item_data;
+
+	if ((item_data = itemdb->exists(nameid)) == NULL) {
+		ShowError("buildin_autofavoriteitem: Invalid item '%d'.\n", nameid);
+		return false;
+	}
+
+	item_data->flag.auto_favorite = cap_value(flag, 0, 1);
+	return true;
+}
+
 /** place holder for the translation macro **/
 static BUILDIN(_)
 {
@@ -26059,6 +26102,8 @@ static void script_parse_builtin(void)
 
 		BUILDIN_DEF(closeroulette, ""),
 		BUILDIN_DEF(openrefineryui, ""),
+		BUILDIN_DEF(setfavoriteitemidx, "ii"),
+		BUILDIN_DEF(autofavoriteitem, "ii"),
 	};
 	int i, len = ARRAYLENGTH(BUILDIN);
 	RECREATE(script->buildin, char *, script->buildin_count + len); // Pre-alloc to speed up
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -3192,6 +3192,30 @@ runs of getcartinventorylist().
 
 ---------------------------------------
 
+*setfavoriteitemidx(<idx>, <flag>)
+
+This function will set an item in inventory as favorite or not.
+If its favorite item, it will be moved to favorite tab, else move out from favorite tab.
+Note: Cant change favorite flag of an equipped item.
+
+Valid Parameters: 
+	<idx>    - inventory index, refer *getinventorylist()
+	<flag>   - true/false (true = favorite item, false = otherwise)
+
+---------------------------------------
+
+*autofavoriteitem(<item_id>, <flag>)
+
+This function will auto set an item as favorite when <item_id> is obtained. 
+If its favorite item, it will be auto moved to favorite tab, else move out from favorite tab.
+This setting affect not only attached player, but also everyone player globally.
+
+Valid Parameters: 
+	<item_id> - item ID
+	<flag>    - true/false (true = favorite item, false = otherwise)
+
+---------------------------------------
+
 *cardscnt()
 
 This function will return the number of cards inserted into the weapon
--- a/src/map/itemdb.h
+++ b/src/map/itemdb.h
@@ -521,7 +521,8 @@ struct item_data {
 		unsigned no_refine : 1; // [celest]
 		unsigned delay_consume : 1;     ///< Signifies items that are not consumed immediately upon double-click [Skotlex]
 		unsigned trade_restriction : 9; ///< Item trade restrictions mask (@see enum ItemTradeRestrictions)
-		unsigned autoequip: 1;
+		unsigned autoequip : 1;
+		unsigned auto_favorite : 1;
 		unsigned buyingstore : 1;
 		unsigned bindonequip : 1;
 		unsigned keepafteruse : 1;
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -4778,6 +4778,13 @@ static int pc_additem(struct map_session_data *sd, const struct item *item_data,
 
 	sd->weight += w;
 	clif->updatestatus(sd,SP_WEIGHT);
+
+	// auto-favorite
+	if (data->flag.auto_favorite > 0) {
+		sd->status.inventory[i].favorite = 1;
+		clif->favorite_item(sd, i);
+	}
+
 	//Auto-equip
 	if(data->flag.autoequip)
 		pc->equipitem(sd, i, data->equip);


src/map/script.c
```diff
@@ -24559,6 +24559,28 @@ static BUILDIN(getcalendartime)
...
+{
+	struct map_session_data *sd = script_rid2sd(st);
+	int idx = script_getnum(st, 2);
+	int value = script_getnum(st, 3);
+
+	if (sd == NULL) {
+		ShowError("buildin_setfavoriteitem: no player attached.\n");
+		return false;
+	}
+	else if (idx < 0 || idx >= sd->status.inventorySize || sd->status.inventorySize <= 0) {
```
hemagx: ``` } else if (x) { ```

hemagx: Also there's no need to check the inventory size, the index check is enough to validate it

src/map/script.c
```diff
@@ -24559,6 +24559,28 @@ static BUILDIN(getcalendartime)
...
+		ShowError("buildin_setfavoriteitem: Invalid Inventory Index %d (min: %d, max: %d).\n", idx, 0, (sd->status.inventorySize - 1));
+		return false;
+	}
+	else if (sd->status.inventory[idx].equip == 0) {
```
hemagx: ``` } else if (x) { ```

Emistry: done

src/map/script.c
Emistry: done

src/map/script.c
```diff
@@ -24559,6 +24559,26 @@ static BUILDIN(getcalendartime)
...
+static BUILDIN(setfavoriteitem)
+{
+	struct map_session_data *sd = script_rid2sd(st);
+	int idx = script_getnum(st, 2);
+	int value = script_getnum(st, 3);
+
+	if (sd == NULL) {
+		ShowError("buildin_setfavoriteitem: no player attached.\n");
+		return false;
+	} else if (idx < 0 || sd->status.inventorySize <= 0) {
```
dastgirp: check "idx" with max size?
because this may cause crash if idx is not proper.
e.g:
```
idx = 100
sd->status.inventorySize = 60
condition: if (100 < 0 || 60 < 0) => False
then it would go below at line 24574 and crash.

Comment (Emistry) [reply src/map/script.c]: done

File: src/map/script.c [hunk 4]
```diff
@@ -24559,6 +24559,26 @@ static BUILDIN(getcalendartime)
...
+static BUILDIN(setfavoriteitem)
+{
+	struct map_session_data *sd = script_rid2sd(st);
+	int idx = script_getnum(st, 2);
+	int value = script_getnum(st, 3);
+
+	if (sd == NULL) {
+		ShowError("buildin_setfavoriteitem: no player attached.\n");
+		return false;
+	} else if (idx < 0 || sd->status.inventorySize <= 0 || idx >= sd->status.inventorySize) {
```
AnnieRuru: what's the point of having `sd->status.inventorySize <= 0` ?
isn't this always between 100 and MAX_INVENTORY ?

Emistry: checking condition for if player has no item in inventory

AnnieRuru: no, `sd->status.inventorySize` is equivalent to `*getInventorySize` script command
`sd->status.inventorySize` is the maximum amount of different items the player can hold

for example, if I have 10 items inside my inventory, 
`sd->status.inventorySize` is still 100, and can only be expand with `*expandInventory` script command

Emistry: i see. thx for the info. updated it.

AnnieRuru:
btw, why using inventory index for this ?
https://github.com/AnnieRuru/Release/blob/master/plugins/favorite_tab/favorite_tab_0.1.c
I did it with ItemID instead of inventory index ...

hmm ... so ... 
using ItemID can ... when pick up this ItemID, can immediately move to favorite tab
using inventory index ... when there are multiple ItemID ...
-> example equipment with same ID, only that equipment will move to favorite tab ...
hmmm ....

AnnieRuru:
how about split into 3 script commands
similar to `*equip` `*equipidx` `*autoequip`
so there will be `*setfavoriteitem` `*setfavoriteitemidx` and `*autofavoriteitem` ?

Emistry:
for itemID as parameter,
it doesn't take the correct/desired item if the character has multiple same itemID or item with different refine/itembonus/cards/etc _(I am sure you had experienced same issue in the past using the `autoequip`, delitem`, and etc)_
using `getinventorylist` could have avoided all that since all info of that item can be obtained and check during the script execution, it even pick the correct and actual inventory index that you intended to set to favorite state.
its possible to implement with itemID/index as parameter, but dont really see the needs for it.
for multiple new script command for the purpose of separating parameter index or itemid? I did vote for -1 for this.
imo, using itemID as parameter are less flexible/convenient than inventory index.

AnnieRuru:
> its possible to implement with itemID/index as parameter, but dont really see the needs for it.
for multiple new script command for the purpose of separating parameter index or itemid? I did vote for -1 for this.
imo, using itemID as parameter are less flexible/convenient than inventory index.

com'on, you didn't read the topic link in rathena forum
https://rathena.org/board/topic/118227-consumables-in-favorites/
Seravy-sensei wants the blue-gemstone in the favorite tab forever
this can only done with `*autofovariteitem`, set 717 as permanent item in favorite tab

... even if I said I can always release as plugin on the forum,
I also wish this script command name as `*setfavoriteitemidx` instead
to avoid confusion with my release later,
as your script command clearly only move item by inventory index

Emistry:
`setfavoriteitem` I have added support for item_id, if the value provided are higher than inventory size, it will be treated as item_id.
But still, I don't recommend using itemID since its lack of accuracy to determine the item in case player has multiple same item id in their inventory.

`autofavourite` I have added this to auto set the item as favorite state if player obtained it.
This is actually a good idea to implement it. The use-case are way common if compared to `setfavoriteitem`.

Kindly help re-review the PR. 
Thank you in advance.

Emistry:
the patch check for `if (nameid >= ITEMID_RED_POTION || nameid >= sd->status.inventorySize) {`.
this is also one of the reason I don't wish it support item ID, since its silly whenever a player try to create an item that shouldn't fall below "minimum" item id, and gravity is brilliant enough to start itemid at 501. lol


btw, kindly  dont use "request changes" to add a comment without really need anything to changes in the code? lol

AnnieRuru:
> gravity is brilliant enough to start itemid at 501

LOL !! I actually want to say gravity is stupid enough to not start the ID from 1 or 0

> btw, kindly dont use "request changes" to add a comment without really need anything to changes in the code? lol

yes, because we can use ~item_avail~ `Sprite` field to create itemID gravity not using
1-500 fall within the possible range that any server can use these ID range
means your code actually create conflicts for those server that actually uses these ID range

Emistry:
I am leaving the restriction here since there are no official way to tell the `min_item_id` that a server is using, unless they self-defined the MIN_ITEM_ID or something equivalent. 
Not to mention the current `max_item_id` support up to millions, I dont really see a point using `itemID <= 500` as their custom item id.

if I to changed the code to give priority for legit `item_id` checking, it would lost the function to alter the state based on inventory index if an item ID exists with the same number as the inventory index. 
Which at the end, resulting to force creating another equivalent script command to support it.
<details>

```
	// if item id are provided
	if (id != NULL) {
		idx = pc->search_inventory(sd, nameid);
		if (idx == INDEX_NOT_FOUND) {
			ShowError("buildin_setfavoriteitem: Item %d not found.\n", nameid);
			return false;
		}
	}
```
</details>

this is why itemID as parameter aren't preferable in the beginning, it has lack accuracy, and create unnecessary works, which originally could be avoided if they use `getinventolist()` to get the `inventory_idx` value to use with this script command.

AnnieRuru:
> Which at the end, resulting to force creating another equivalent script command to support it.

I have been saying separate into `*setfavoriteitem` and `*setfavoriteitemidx`
if your idea actually accurate, that also means the PR about [*delitemidx](https://github.com/HerculesWS/Hercules/pull/2394) and [*equipidx](https://github.com/HerculesWS/Hercules/pull/2355) shouldn't even exist
and should alter existing `*delitem` and `*equip` script command instead, which create further confusion

ok now wait 1 month until Haru come back and review this

MishimaHaruna:
I also think that having a command that can take either an index or a nameid is a recipe for problems.

My main concerns are:
- A customized setting, or a future Gravity update could push the minimum item ID to < 500
- A customized setting/client or a future Gravity update could push the maximum inventory size to > 500
- A malfunctioning script that expects to have an item ID in a variable could have the wrong value (or it could have a typo in a variable or constant, resulting into a `0`), which could be silently accepted by the script as an inventory index, instead of throwing a runtime error and aborting the script because of an invalid nameid.

I'd suggest to accept either the nameid or the index (and have separate commands if you want to accept both). It's not nice to have several commands that do the same thing, but it seems to me the approach that's less prone to errors.

Emistry:
Then I will just proceed with original plan, only accept `inventory_index`.
imo, using itemID as parameter are less flexible/convenient than inventory index.

Asheraf:
Following up on what Haru daid:
   - Gravity already have an item with id = 1
```
	[1] = {
		unidentifiedDisplayName = "머리띠",
		unidentifiedResourceName = "눈가리개",
		unidentifiedDescriptionName = {},
		identifiedDisplayName = "마력의 별",
		identifiedResourceName = "마력의별",
		identifiedDescriptionName = {},
		slotCount = 0,
		ClassNum = 0
	},
```

- Inventory Expansion feature allows for > 500 inventory size (herc still have some packet limits but client support it normally)

PR #1733, opened 2017-05-13
Fix map-server crash when using npcspeed on a npc not on a map

chriser-:
### Pull Request Prelude

- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
Check for unit data `(nd->ud != NULL)` before adjust the NPC speed. 
Throw warnings if `nd->ud == NULL`.

**Issues addressed:** 
Prevent map-server from crashing when adjust the NPC speed.

Sample NPC Script:
```
-	script	dummynpc	-1,{
	OnInit:
		debugmes "set speed to 100";
		npcspeed 100;
		debugmes "set speed done";
		end;
}
```
Result:
sample

After fixed:  throw warnings but didnt crash map-server
sample2

[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style

PR #2430, opened 2019-04-10
Fix map-crash when floating npc use *npcspeed script command

Emistry:
### Pull Request Prelude

### Changes Proposed
Check for unit data `(nd->ud != NULL)` before adjust the NPC speed. 
Throw warnings if `nd->ud == NULL`.

**Issues addressed:** 
script command `npcspeed`, `npcwalkto`, `npcstop`, `unitwalk`, `unitwarp`, `unitstop` are not applicable for floating npc (without coordinate/location).
Fix map-crash due to unit_data `NULL`.

Sample NPC Script:
```
-	script	dummynpc	-1,{
	OnInit:
		debugmes "set speed to 100";
		npcspeed 100;
		debugmes "set speed done";
		end;
}
```
Result:
sample

After fixed:  throw warnings but didnt crash map-server
sample2

Related PR https://github.com/HerculesWS/Hercules/pull/1733

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -15910,37 +15910,41 @@ static BUILDIN(npctalk)
 // change npc walkspeed [Valaris]
 static BUILDIN(npcspeed)
 {
-	struct npc_data* nd;
-	int speed;
-
-	speed = script_getnum(st,2);
-	nd = map->id2nd(st->oid);
+	struct npc_data *nd = map->id2nd(st->oid);
+	int speed = script_getnum(st, 2);
 
 	if (nd != NULL) {
 		unit->bl2ud2(&nd->bl); // ensure nd->ud is safe to edit
+		if (nd->ud == NULL) {
+			ShowWarning("buildin_npcspeed: floating NPC don't have unit data.\n");
+			return false;
+		}
 		nd->speed = speed;
 		nd->ud->state.speed_changed = 1;
 	}
 
 	return true;
 }
+
 // make an npc walk to a position [Valaris]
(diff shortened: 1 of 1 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2430)

4144:
@AnnieRuru  this code may not crash server but corrupt memory

probably better fix it in all functions

AnnieRuru:
at least it doesn't cause memory leak ...
well, if just add in the same lines in those 2 script commands, simple I think

Emistry:
the reason it doesn't crash are probably due to : 
`npcwalkto()` actually have the checking `if( ud == NULL) return 0;` added inside `unit_walktoxy()`
and
`npcstop()` actually have the checking `if(!ud ..... ) return 0;` added inside `unit_stop_walking()`

AnnieRuru:
how about just adding 
```c
		if (nd->ud == NULL) {
			ShowWarning("buildin_npcwalkto: floating NPC don't have unit data.\n");
			return false;
		}
```
instead of let the script command fail silently ?

hemagx:
Emistry, can you separate syntax fixing from the crash fixing? it's sort of impossible to review as it is right now.
separating each in it's respective commit would be a lot better.

Emistry:
huh? like how? reset then redo to create 2 different commits? 
aren't at the end it still gonna end up become 1 commit that fix the same thing?
beside its not a huge commit to begin with, most of it just the `if (ud == NULL)` checking.

PR #2439, opened 2019-04-14
Update dropall atcommand drop item based on type.

Emistry:
### Pull Request Prelude

### Changes Proposed
- Enable to drop items based on item type.

**Issues addressed:** 
Unable to drop items based on item type.

Reference: https://github.com/RagEmu/Renewal/commit/fa865ca07f43e5154c5253393640ec3f28da4a79

### DIFF
--- a/doc/atcommands.txt
+++ b/doc/atcommands.txt
@@ -692,9 +692,23 @@ Repairs all broken items in your inventory.
 
 ---------------------------------------
 
-@dropall
+@dropall {<item type>}
 
-Drops all inventory and equipped items onto the floor.
+Drops all items based on the item type.
+
+Valid item types:
+    -1 = All Items (default)
+     0 = Healing Items
+     2 = Useable Items
+     3 = Etc Items
+     4 = Weapons
+     5 = Armors
+     6 = Cards
+     7 = Pet Eggs
+     8 = Pet Armors
+    10 = Ammunition Items
+    11 = Delayed-Consumable Items
+    18 = Cash Items
 
 ---------------------------------------
 
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -5277,18 +5277,43 @@ ACMD(follow)
 }
 
 /*==========================================
- * @dropall by [MouseJstr]
- * Drop all your possession on the ground
+ * @dropall by [MouseJstr] and [Xantara]
+ * Drop all your possession on the ground based on item type
  *------------------------------------------*/
 ACMD(dropall)
 {
-	for (int i = 0; i < sd->status.inventorySize; i++) {
-		if (sd->status.inventory[i].amount) {
-			if(sd->status.inventory[i].equip != 0)
-				pc->unequipitem(sd, i, PCUNEQUIPITEM_RECALC|PCUNEQUIPITEM_FORCE);
-			pc->dropitem(sd,  i, sd->status.inventory[i].amount);
+	int type = -1;
+	int count = 0;
+
+	if (message[0] != '\0') {
+		type = atoi(message);
+		if (!((type >= IT_HEALING && type <= IT_DELAYCONSUME) || type == IT_CASH || type == -1)) {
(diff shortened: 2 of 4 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2439)

src/map/atcommand.c
```diff
@@ -5277,18 +5277,58 @@ ACMD(follow)
...
- *------------------------------------------*/
+* @dropall by [MouseJstr] and [Xantara]
+* Drop all your possession on the ground based on item type
+*------------------------------------------*/
 ACMD(dropall)
 {
-	for (int i = 0; i < sd->status.inventorySize; i++) {
+	int type = -1;
+	int i, count = 0;
+	struct item_data *item_data = NULL;
```
4144: item_data used only inside ``if``, this mean variable must be moved into ``if`` block.
``i`` variable should be inside ``for``: ``for (int i = ...``
other vars probably should be moved too

Emistry: hmm, aren't previously all variables shall move to the top? something like to avoid confusion/ease to spot the variable for other dev to may modify it in future.

anyway, updated.

4144: in past it was because old visual studio.
But correct way to use closer variable scope

doc/atcommands.txt
```diff
@@ -692,9 +692,23 @@ Repairs all broken items in your inventory.
...
 
-@dropall
-
-Drops all inventory and equipped items onto the floor.
+@dropall {<item type>}
+
+Drops all items based on the item type.
+
+Valid item types:
+    -1 = All Items (default)
```
hemagx: Wrong indention?

src/map/atcommand.c
```diff
@@ -5277,18 +5277,57 @@ ACMD(follow)
...
- *------------------------------------------*/
+* @dropall by [MouseJstr] and [Xantara]
+* Drop all your possession on the ground based on item type
+*------------------------------------------*/
 ACMD(dropall)
 {
+	int type = -1;
+	int count = 0;
+
+	if (message[0]) {
```
hemagx: ```if (message[0] != '\0')```

src/map/atcommand.c
```diff
@@ -5277,18 +5277,57 @@ ACMD(follow)
...
+		type = atoi(message);
+		switch (type) {
```
hemagx: why a switch instead of a range check? this switch can be changed to
```
if (!((type >= IT_HEALING && type <= IT_DELAYCONSUME) || type == IT_CASH || type == -1)) {
	clif->message(fd, msg_fd(fd, 1500));
	clif->message(fd, msg_fd(fd, 1501));
	return false;
}
```

hemagx:
Also please change your target branch to master, it's impossible to merge right now as it's target stable branch.

src/map/atcommand.c
Emistry: copy paste from ragemu, perhaps readability.
anyway, updated to `if (...)`

src/map/atcommand.c
Emistry: done

doc/atcommands.txt
Emistry: done

src/map/atcommand.c
```diff
@@ -5277,18 +5277,43 @@ ACMD(follow)
 }
 
 /*==========================================
- * @dropall by [MouseJstr]
- * Drop all your possession on the ground
- *------------------------------------------*/
+* @dropall by [MouseJstr] and [Xantara]
```
hemagx: Sorry i did not notice this before, please put back the space before the comments.
Afterwards, it should be ready to merge :)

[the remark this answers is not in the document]
Emistry: done

Emistry: done

Issue #1955, opened 2018-01-16
Mushroom Event not working properly


xtreamprince:
hello i got problems with mushroom event
![untitled](https://user-images.githubusercontent.com/22543759/34990434-176f4928-faec-11e7-92fc-fb6af0098546.jpg)

after all mushroom killed, it should end the event.

xtreamprince:
Look like i'm using old version of script.

xtreamprince:
same problem with latest version of script

guilherme-gm:
Can you check if this is still happening? It's working fine here (latest master)

Emistry:
checked and work fine 
![img](https://i.imgur.com/GVmLy45.png)

xtreamprince:
Well it start to work fine after reload script, after some time when no one playing this event, it got stuck & says 0 mushroom left.

Emistry:
hmm, I think I found the issue.
you are probably did `@killmonster` during the event. 
with this trick, only I able to duplicate the similar scenario that you show above.

![pic](https://i.imgur.com/J5tuh7X.png)

PR #2442, opened 2019-04-14
Fix issue #1955

Emistry:
### Pull Request Prelude

### Changes Proposed
- fix error due to event terminated by GM using `@killmonster` command.
pic

**Issues addressed:** 
https://github.com/HerculesWS/Hercules/issues/1955

### DIFF
--- a/npc/custom/events/mushroom_event.txt
+++ b/npc/custom/events/mushroom_event.txt
@@ -50,9 +50,12 @@ OnMinute10:	// Start time (every hour)
 
 OnMobKilled:
 	set .Spawn, .Spawn - 1;
-	getitem .Prize, .Amount;
-	if (.Spawn) announce "[ "+strcharinfo(PC_NAME)+" ] has killed a Mushroom. There are now "+.Spawn+" Mushroom(s) left.",bc_map;
-	else {
+	if (playerattached() != 0) {
+		getitem .Prize, .Amount;
+		if (.Spawn)
+			announce "[ "+strcharinfo(PC_NAME)+" ] has killed a Mushroom. There are now "+.Spawn+" Mushroom(s) left.",bc_map;
+	}	
+	if (!.Spawn) {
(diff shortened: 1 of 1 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2442)

npc/custom/events/mushroom_event.txt
```diff
@@ -50,9 +50,12 @@ OnMinute10:	// Start time (every hour)
 
 OnMobKilled:
 	set .Spawn, .Spawn - 1;
-	getitem .Prize, .Amount;
-	if (.Spawn) announce "[ "+strcharinfo(PC_NAME)+" ] has killed a Mushroom. There are now "+.Spawn+" Mushroom(s) left.",bc_map;
-	else {
+	if (playerattached()) {
```
hemagx: according to documentations and code this function is a non-boolean, it always return an integer value and 0 on failure, a ```!= 0``` check should be placed instead.

Emistry: done.

npc/custom/events/mushroom_event.txt
```diff
@@ -50,9 +50,12 @@ OnMinute10:	// Start time (every hour)
 
 OnMobKilled:
 	set .Spawn, .Spawn - 1;
-	getitem .Prize, .Amount;
-	if (.Spawn) announce "[ "+strcharinfo(PC_NAME)+" ] has killed a Mushroom. There are now "+.Spawn+" Mushroom(s) left.",bc_map;
-	else {
+	if (playerattached() > 0) {
```
hemagx: Uh, it should be a ```!= 0``` not ```> 0```
The source returns 0 on failure, which means that a negative value can be valid, maybe not now but in future this can be the case, it should be a strict ```!= 0```

Emistry: ok done

Issue #1569, opened 2017-02-08
Chat Room on nochat cells


bWolfie:
In a map with no chat cells, there is a way to put chats on them.

1. First player creates chat room on a chat-allowed cell.
2. Second player, who is currently standing on a nochat cell, enters chat room.
3. First player leaves chat room.

Chat room is now on nochat cell.

psychonaut04:
If i remember it's possible to(don't know if it was fixed):
1. Open chat room creation window, put the name but don't create it yet.
2. Start walking to a nochat cell.
3. Create the chat room while walking to the nochat cell but while still in the chat-allowed cell.
Character will move to the nochat cell.

PR #2443, opened 2019-04-14
Disable chatroom creation if cell_chknochat

Emistry:
### Pull Request Prelude

### Changes Proposed
- doesn't allow chatroom creation/switch if the coordinated cell are set with `cell_nochat`.
- chatroom will be auto deleted.

<details>

```
-	script	sample	-1,{

	OnInit:
		.@map$ = "prontera";
		
		.@max_x = getmapinfo(MAPINFO_SIZE_X, .@map$);
		.@max_y = getmapinfo(MAPINFO_SIZE_Y, .@map$);
		setcell(.@map$, 0, 0, .@max_x, .@max_y, cell_nochat, true);
		
		.@range = 10;
		.@center_x = 155;
		.@center_y = 181;
		setcell(.@map$, (.@center_x - .@range), (.@center_y - .@range), (.@center_x + .@range), (.@center_y + .@range), cell_nochat, false);
		end;
}
```
player 1 - create a normal chatroom at normal cells that dont have `cell_nochat`.
player 2 - move to `cell_nochat` area, and join the chatroom
player 1 - leave the chatroom
** at this point, chatroom shall be auto deleted since the new owner are located in the `cell_nochat` area.
</details>

**Issues addressed:** 
https://github.com/HerculesWS/Hercules/issues/1569

### DIFF
--- a/src/map/chat.c
+++ b/src/map/chat.c
@@ -255,6 +255,14 @@ static int chat_leavechat(struct map_session_data *sd, bool kicked)
 	}
 
 	if( leavechar == 0 && cd->owner->type == BL_PC ) {
+
+		// check if new location are CELL_CHKNOCHAT
+		if (map->getcell(cd->usersd[0]->bl.m, NULL, cd->usersd[0]->bl.x, cd->usersd[0]->bl.y, CELL_CHKNOCHAT)) {
+			for (i = (cd->users - 1); i >= 0; i--)
+				chat->leave(cd->usersd[i], false);
+			return 2;
+		}
+
 		// Set and announce new owner
 		cd->owner = &cd->usersd[0]->bl;
 		clif->changechatowner(cd, cd->usersd[0]);


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2443)

bWolfie:
Nice one! Much better solution than what I was using where player on nochat cell just can't join the chat.

src/map/chat.c
```diff
@@ -255,6 +255,14 @@ static int chat_leavechat(struct map_session_data *sd, bool kicked)
 	}
 
 	if( leavechar == 0 && cd->owner->type == BL_PC ) {
+
+		// check if new location are CELL_CHKNOCHAT
+		if (map->getcell(cd->usersd[0]->bl.m, NULL, cd->usersd[0]->bl.x, cd->usersd[0]->bl.y, CELL_CHKNOCHAT)) {
+			for (i = (cd->users - 1); i >= 0; i--)
```
hemagx: Is there a reason the loop have to be descending instead of ascending?

Emistry: doesn't have to shift the array.
just my own preference when delete data from array.

previously i tried with `chat->npckickall(...)` but no idea, it crashed for unknown reason.
btw...  `chat->npckickall(...)` also did go through the loop descending when kicking member, maybe the same reason as stated above.

hemagx: Ah, this makes perfect sense then.

PR #2447, opened 2019-04-18
Restore the ability for HPM Hooks to hook to private interfaces

hemagx:
### Pull Request Prelude

### Changes Proposed

- Starting from 5db7c799055c6ae9c4463f6cf4c88a35597d5d31 the ability to hook to private interfaces was removed, this commit restores this ability by introducing two new macros for specifically.
* addHookPrePriv to add a pre-hook for private interface member
* addHookPostPriv to add a post-hook for private interface member

**Issues addressed:** 

### DIFF
--- a/src/plugins/HPMHooking.h
+++ b/src/plugins/HPMHooking.h
@@ -50,9 +50,19 @@ HPExport struct HPMHooking_interface HPMHooking_s;
 		HPMi->hooking->AddHook(HOOK_TYPE_PRE, #ifname "->" #funcname, (hook), HPMi->pid) \
 		)
 
+#define addHookPrePriv(ifname, funcname, hook) ( \
+		(void)((HPMHOOK_pre_PRIV__ ## ifname ## _ ## funcname)0 == (hook)), \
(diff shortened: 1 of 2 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2447)

hemagx:
I will fix the build issue after the weekend.

bWolfie:
what is use of this private hook i dont understand

hemagx:
@bWolfie Please check lclif.c lclif.h and lclif.p.h it's the only example in hercules at the moment that uses private (header) interface.
you would notice in lclif that there's functions defined in the private header and not the public one, the plugins was not able to hook to those functions until this PR.
later on we will have more files separated this way, starting with my refinery PR

bWolfie:
i guess its cause i don't know what private interface means

hemagx:
Think of it this:
lclif is an object.
lclif->foo is a public member of the object, can be called inside of it and outside of it.
lclif->p->bar is a private member of the object, can be only called inside the object and from plugins.

so, public interfaces are the public members of an object and the private interfaces are the private members of an object.

PR #2462, opened 2019-05-07
Dehardcode Rental Expire/Renting item checks

hemagx:
### Pull Request Prelude

### Changes Proposed
Implemented new script fields in item db executes when item is rented and when item rent expires
Implemented a new script command getfont() gets the current player font id
Fixed hardcoded checks for rental items in source

**Issues addressed:** 
#140

### DIFF
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -565,72 +565,8 @@ static int pc_inventory_rental_clear(struct map_session_data *sd)
 /* assumes i is valid (from default areas where it is called, it is) */
 static void pc_rental_expire(struct map_session_data *sd, int i)
 {
-	int nameid;
-
 	nullpo_retv(sd);
 	Assert_retv(i >= 0 && i < sd->status.inventorySize);
-	nameid = sd->status.inventory[i].nameid;
-
-	/* Soon to be dropped, we got plans to integrate it with item db */
-	switch( nameid ) {
-		case ITEMID_BOARDING_HALTER:
-			status_change_end(&sd->bl,SC_ALL_RIDING,INVALID_TIMER);
-			break;
-		case ITEMID_LOVE_ANGEL:
-			if( sd->status.font == 1 ) {
-				sd->status.font = 0;
-				clif->font(sd);
-			}
-			break;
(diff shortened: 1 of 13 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2462)

4144:
what will happened if other script already running and rental script will be run?

src/map/pc.c
```diff
@@ -4820,12 +4758,21 @@ static int pc_delitem(struct map_session_data *sd, int n, int amount, int type,
 
 	sd->status.inventory[n].amount -= amount;
 	sd->weight -= sd->inventory_data[n]->weight*amount ;
+
+	// It's here because the data would most likely get zeroed in following if [Hemagx]
+	struct item_data *itd = sd->inventory_data[n];
+	bool is_rentral = (sd->status.inventory[n].expire_time > 0) ? true : false;
```
dastgirp:
```suggestion
	bool is_rental = (sd->status.inventory[n].expire_time > 0) ? true : false;
```

src/map/script.c
```diff
@@ -25454,6 +25467,36 @@ static void script_run_item_unequip_script(struct map_session_data *sd, struct i
 	script->current_item_id = 0;
 }
 
+static void script_run_item_rental_start_script(struct map_session_data *sd, struct item_data *data, int oid) __attribute__((nonnull(1, 2)));
```
dastgirp: Why is there declaration of function here?

hemagx:
@4144 Hmm, good point. we can probably set an ```sd->npc_id```(Not sure if the variable name correct), check, since rentals on regular basis checked a few seconds extra of renting is not a problem.

src/map/script.c
hemagx: it has a nonull attribute

4144:
i mean if you do attach to player, it will terminate previous script. but i not sure is you attaching script to player or run in deattached

hemagx:
@4144 Item scripts works in attached mode, yet it can still be used while talking/executing NPC perfectly fine.

PR #2487, opened 2019-06-02
Add identify/identifyidx script commands and identifyall atcommand

guilherme-gm:
### Pull Request Prelude

### Changes Proposed

This PR implements script commands to identify items, although lots of scripts nowadays kind of fake that behavior by using a combination of `delitem2` + `getitem2`, this leads to wrong behavior when items have more information, like random options (as shown in this [forum post][post], where doing this leads to losing dropped options). New script commands:
- `identify(<Item ID>)`: Identifies the first unidentified item with ID = "Item ID"
- `identifyidx(<idx>)`: Identifies item at inventory index `idx`.

I took the chance to also add `@identifyall` atcommand, because it is quite boring to identify item per item when you are testing lots of drops.

**Issues addressed:** 

[post]: https://herc.ws/board/topic/17079-random-options-on-monster-drops/?do=findComment&comment=91951

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -6769,23 +6769,34 @@ ACMD(refreshall)
 }
 
 /*==========================================
- * @identify
+ * @identify / @identifyall
  * => GM's magnifier.
  *------------------------------------------*/
 ACMD(identify)
 {
 	int num = 0;
+	bool identifyall = (strcmpi(info->command, "identifyall") == 0);
 
-	for (int i = 0; i < sd->status.inventorySize; i++) {
-		if(sd->status.inventory[i].nameid > 0 && sd->status.inventory[i].identify!=1){
-			num++;
+	if (!identifyall) {
+		for (int i = 0; i < sd->status.inventorySize; i++) {
+			if (sd->status.inventory[i].nameid > 0 && sd->status.inventory[i].identify != 1) {
+				num++;
+			}
+		}
+	} else {
+		for (int i = 0; i < sd->status.inventorySize; i++) {
+			if (sd->status.inventory[i].nameid > 0 && sd->status.inventory[i].identify != 1) {
+				skill->identify(sd, i);
+				num++;
+			}
 		}
 	}
-	if (num > 0) {
-		clif->item_identify_list(sd);
-	} else {
+	
+	if (num == 0)
 		clif->message(fd,msg_fd(fd,1238)); // There are no items to appraise.
-	}
+	else if (!identifyall)
+		clif->item_identify_list(sd);
+
 	return true;
 }
 
@@ -10061,6 +10072,7 @@ static void atcommand_basecommands(void)
 		ACMD_DEF(refresh),
 		ACMD_DEF(refreshall),
 		ACMD_DEF(identify),
+		ACMD_DEF2("identifyall", identify),
 		ACMD_DEF(misceffect),
 		ACMD_DEF(mobsearch),
 		ACMD_DEF(cleanmap),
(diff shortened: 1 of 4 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2487)

src/map/atcommand.c
```diff
@@ -6727,23 +6727,27 @@ ACMD(refreshall)
...
  *------------------------------------------*/
 ACMD(identify)
 {
 	int num = 0;
+	bool identifyall = (strcmpi(info->command, "identifyall") == 0);
 
 	for (int i = 0; i < sd->status.inventorySize; i++) {
-		if(sd->status.inventory[i].nameid > 0 && sd->status.inventory[i].identify!=1){
+		if (sd->status.inventory[i].nameid > 0 && sd->status.inventory[i].identify != 1) {
+			if (identifyall)
```
Emistry: perhaps  move this line to before the `for-loop`? so that it won't need to go through the loop everytime even if we are just using `@identify` to identify one item.

src/map/script.c
```diff
@@ -25305,6 +25305,75 @@ static BUILDIN(openrefineryui)
...
+{
+	struct map_session_data *sd = script_rid2sd(st);
+
+	if (sd == NULL) {
+		script_pushint(st, -2);
+		return true;
+	}
+
+	int itemid = script_getnum(st, 2);
+	if (!itemdb->exists(itemid)) {
```
Asheraf: `itemdb->exists` returns a pointer, please check for `== NULL`

src/map/script.c
```diff
@@ -25305,6 +25305,75 @@ static BUILDIN(openrefineryui)
...
+		ShowError("script_identify: Invalid item ID (%d)\n", itemid);
```
Asheraf: please use `builtin_` for function name, just to be consistence with the reset of the commands.

src/map/script.c
```diff
@@ -25305,6 +25305,75 @@ static BUILDIN(openrefineryui)
...
+		script_pushint(st, -2);
+		return true;
+	}
+
+	int idx = -1;
+	ARR_FIND(0, sd->status.inventorySize, idx, (sd->status.inventory[idx].nameid == itemid && !sd->status.inventory[idx].identify));
```
Asheraf: `identify` is not a bool so please check for 0

src/map/atcommand.c
guilherme-gm: I was wondering that too, but it would duplicate most of the loop code so I wasn't sure if I should

dastgirp: @Emistry it needs to loop even if you use `@identify`, to know how many item needs to be identified.

guilherme-gm: Yes, it is as dastgir said. Can I mark as resolved and leave it as is?

4144: for faster code ``if (identifyall)`` better move outside of ``for``
this mean for faster code need two ``for`` one inside if block and second inside else block

guilherme-gm: Done. I'm wondering, would make more sense to make a separate command block instead of using the same block + strcmp?

MishimaHaruna:
@guilherme-gm I force-pushed on your branch a minute after you did (without realizing) - if you had made any changes (or just to restore your gpg signature, if any), please force-push again, otherwise I'll merge this as is

guilherme-gm:
@MishimaHaruna no problem :) you can merge it as is

PR #2494, opened 2019-06-16
Update minimum value of *input script command.

Emistry:
### Pull Request Prelude

### Changes Proposed
All `input()` shall have default values, especially old scripts. 
Maybe consider to set all parameter for `input()` as mandatory?

**How to reproduce:** 
https://github.com/HerculesWS/Hercules/blob/stable/conf/map/script.conf
```
 'input_min_value:  <any negative value> 
```
talk to any NPC that use `input()` to exchange item/zeny/points and enter negative values to cause the exploit. (Ex:  `points = (points - (-amount)); ` became  `points = (points + amount); `

**Issues addressed:** 
- avoid potential hacks for old scripts that use `input()` script commands. 
- One of the example:  _there are alot more scripts to list, I will just pick one_
https://github.com/HerculesWS/Hercules/blob/stable/npc/other/arena/arena_room.txt#L1008-L1035
- It prompt user to input value, but the value could be negative (due to https://github.com/HerculesWS/Hercules/pull/2375) and create exploits where they could farm the `arena_points` or `Zeny` in other scripts.
- some of the script are safe due to it has negative value validation, but a lot scripts didn't have the validation.

**Other Issue:** 
- Previously, `input()` doesn;t allow negative, so I am not sure why some NPC does/doesn't check for negative value. Hence it might affect minor NPC that has the if-else checking for negative value.

### DIFF
--- a/npc/airports/airships.txt
+++ b/npc/airports/airships.txt
@@ -491,9 +491,8 @@ airplane,50,66,5	script	Apple Merchant#airplane	4_M_04,{
 			" '0' to cancel your order.");
 		next;
 		while (1) {
-			input .@input;
-			.@pay = .@input * 15;
-			if (.@input == 0) {
+			input(.@input);
+			if (.@input <= 0) {
 				mes("[Fruitz]");
 				mes("Thanks for stopping\r"
 					"by my shop. Farewell!");
@@ -502,7 +501,7 @@ airplane,50,66,5	script	Apple Merchant#airplane	4_M_04,{
 					"Apple to snack on~");
 				close;
 			}
-			else if (.@input < 1 || .@input > 500) {
+			if (.@input > 500) {
 				mes("[Fruitz]");
 				mes("You've entered a number\r"
 					"higher than the maximum\r"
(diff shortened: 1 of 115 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2494)

dastgirp:
Didnt we had restriction for this before? 
I think it would be better to not allow negative value Unless specified in input

Emistry:
But this issue only happen if `input_min_value` are set to negative. 
so if anyone changed it into negative value, and use `input(.@value);` without define the `min` and `max`, then it might cause issue.
although this could be said as "user fault" but the old script that use `input()` are the culprit, those script mostly doesn't define the value of `min` and `max`.

I think a safety way to use `input()` would be change the script command into `input(min, max)` where `min max` are mandatory field.

4144:
I think it was merged incomplete negative value implimentation.
Correct way is control negatives in ``input``
As simple half fix probably need add optional parameters to input with allowed limits. And by default without negative values

MishimaHaruna:
I edited the pull request and removed the minimum bound where it's not necessary (wherever a negative value is already explicitly or implicitly handled by the script).
In many cases I also changed the `== 0` checks to be `<= 0`.

PR #2523, opened 2019-08-25
Name partial match

MishimaHaruna:
### Pull Request Prelude

### Changes Proposed

When the `partial_name_scan` setting is enabled, it currently affects various name lookups done in source and script, even where it is not safe to do so and a complete match would be necessary.

This PR limits the partial match to atcommands and player/client commands, where it's safe to do so.


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2523)

src/map/map.c
```diff
@@ -2268,46 +2268,45 @@ static struct map_session_data *map_charid2sd(int charid)
...
 	nicklen = strlen(nick);
 	iter = mapit_getallusers();
 
 	found_sd = NULL;
 	for (sd = BL_UCAST(BL_PC, mapit->first(iter)); mapit->exists(iter); sd = BL_UCAST(BL_PC, mapit->next(iter))) {
-		if( battle_config.partial_name_scan )
-		{// partial name search
-			if( strnicmp(sd->status.name, nick, nicklen) == 0 )
-			{
+		if (battle_config.partial_name_scan && allow_partial) {
```
4144: i not sure, but probably if better move outside of ``for`` for better performance

MishimaHaruna: Done, thank you

PR #2526, opened 2019-08-26
Convert sc_config to use libconfig

Asheraf:
### Pull Request Prelude

### Changes Proposed

just simple conversion because the old format makes me sad..

**Issues addressed:** 


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2526)

src/map/status.c
```diff
@@ -13397,25 +13397,127 @@ static bool status_readdb_sizefix(char *fields[], int columns, int current)
...
+	nullpo_retr(false, source);
+	Assert_retr(false, type >= SC_NONE && type < SC_MAX);
+
+	int i = 0;
+	struct config_setting_t *t = NULL;
+	while ((t = libconfig->setting_get_elem(it, i++)) != NULL) {
+		const char *flag = config_setting_name(t);
+		bool on = libconfig->setting_get_bool_real(t);
+
+		if (strcmpi(flag, "NoDeathReset") == 0) {
```
hemagx: you can change this into an array of structure holding the config name and bitmask value, then loop for each value found to match it with one of those array members, this will simplify the code and make it simpler to add more values in the future, also please do a check twice one with strcmp and one with strcmpi, if strcmpi success and strcmp fail give a warning, case insensitivity shall not tolerated.

src/map/status.c
```diff
@@ -13397,25 +13397,127 @@ static bool status_readdb_sizefix(char *fields[], int columns, int current)
...
+				status->dbs->DisplayType[type] = false;
+			}
+		} else {
+			status->read_scdb_libconfig_sub_flag_additional(it, type, source);
+		}
+	}
 	return true;
 }
+
+static void status_read_scdb_libconfig_sub_flag_additional(struct config_setting_t *it, int type, const char *source)
```
hemagx: change this structure to be bool and default return of false, plugins shall change this to true, otherwise main function returns a warning.

src/map/status.c
```diff
@@ -13397,25 +13397,127 @@ static bool status_readdb_sizefix(char *fields[], int columns, int current)
 	return true;
 }
 
-static bool status_readdb_scconfig(char *fields[], int columns, int current)
+static bool status_read_scdb_libconfig(void)
 {
-	int val = 0;
-	char* type = fields[0];
+	struct config_t status_conf;
+	const char *config_filename = "db/sc_config.conf"; // FIXME hardcoded name
```
Emistry:
```suggestion
	const char *config_filename = map->db_path"/sc_config.conf"; // FIXME hardcoded name
```

src/map/status.c
```diff
@@ -13397,25 +13397,127 @@ static bool status_readdb_sizefix(char *fields[], int columns, int current)
...
+
+static bool status_read_scdb_libconfig_sub(struct config_setting_t *it, int idx, const char *source)
+{
+	nullpo_retr(false, it);
+	nullpo_retr(false, source);
+
+	int i32;
+	const char *name = config_setting_name(it);
+
+	if (!script->get_constant(name, &i32) || i32 < SC_NONE || i32 >= SC_MAX) {
```
Emistry:
```suggestion
	if (!script->get_constant(name, &i32) || i32 <= SC_NONE || i32 >= SC_MAX) {
```
I think `SC_NONE` also considered as invalid status

src/map/status.c
```diff
@@ -13397,25 +13397,127 @@ static bool status_readdb_sizefix(char *fields[], int columns, int current)
...
+	status->read_scdb_libconfig_sub_flag(fg, i32, source);
+
+	return true;
+}
+
+static bool status_read_scdb_libconfig_sub_flag(struct config_setting_t *it, int type, const char *source)
+{
+	nullpo_retr(false, it);
+	nullpo_retr(false, source);
+	Assert_retr(false, type >= SC_NONE && type < SC_MAX);
```
Emistry:
```suggestion
	Assert_retr(false, type > SC_NONE && type < SC_MAX);
```
I think `SC_NONE` also considered as invalid status

hemagx:
> 
> 
> Might as well consider to include enhancement for this too?
> 
> https://github.com/HerculesWS/Hercules/blob/ce07ce1a7c5bcc118e8f08342f43a580c1f4bfe8/src/map/script.c#L12414-L12423

I think he should keep it this way for now, if we add this it will break the convertor and people will have to do a lot more to ensure their sc_config.txt is converted correctly, perhaps this shall be done in later PR, i will merge this PR as soon as it's approved to master and move as much hardcoded values to the new sc_config.conf

src/map/status.c
4144: if do this fix, fixme comment can be removed

tools/scconfigconverter.py
```diff
@@ -0,0 +1,76 @@
+import csv
```
4144: add python header to allow run correct python without adding binary before script
for python without version can be: ``#! /usr/bin/env python``
or for python2: ``#! /usr/bin/env python2``
And also script should be with executable flag if it not set

hemagx:
> 
> 
> Do you think about adding official options that are missing ?
> 
>     * **Send**
> 
>     * **IgnorePretendDead**
> 
>     * **ActorAppearance**
> 
>     * **SendMultiCast**
> 
>     * **["backward compatibility send"]**

in future all hard-coded settings shall be moved, for now to make it simple and easy on people to migrate we're sticking to a simple conversion without adding feautres

db/sc_config.conf
```diff
@@ -0,0 +1,2810 @@
...
+//= along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//================= Description ===========================================
+// Configurations file for status effects
+//=========================================================================
+
+/**************************************************************************
+ ************* Entry structure ********************************************
+ **************************************************************************
+SC_TYPE: {
+	Visible:           (bool) SC can be visible for all players
```
hemagx: It doesn't state it's optional

src/map/status.c
```diff
@@ -13397,25 +13397,102 @@ static bool status_readdb_sizefix(char *fields[], int columns, int current)
...
+	const char *name = config_setting_name(it);
+
+	if (!script->get_constant(name, &i32) || i32 <= SC_NONE || i32 >= SC_MAX) {
+		ShowWarning("status_read_scdb_libconfig_sub: Invalid status type (%s) in \"%s\" entry #%d, skipping.\n", name, source, idx);
+		return false;
 	}
 
+	libconfig->setting_lookup_bool_real(it, "Visible", &status->dbs->DisplayType[i32]);
+
+	struct config_setting_t *fg = libconfig->setting_get_member(it, "Flags");
```
hemagx: the old format allowed flags to be optional, aka value of 0, perhaps we should retrain this in the new feature, right bow it will assert.

src/map/status.c
```diff
@@ -13397,25 +13397,102 @@ static bool status_readdb_sizefix(char *fields[], int columns, int current)
...
+			if (strcmp(flag, flags[j].name) != 0) {
+				ShowWarning("status_read_scdb_libconfig_sub_flag: flag (%s) for status effect (%d) is casesensitive, correct it to (%s).", flag, type, flags[i].name);
+			}
+			if (on) {
+				status->dbs->sc_conf[type] |= flags[j].value;
+			} else {
+				status->dbs->sc_conf[type] &= ~flags[j].value;
+			}
+		} else {
+			status->read_scdb_libconfig_sub_flag_additional(it, type, source);
```
hemagx: the value of this still not checked, it should warn you if it return false.

tools/scconfigconverter.py
```diff
@@ -0,0 +1,99 @@
...
+			print('\t\tNoClearanceReset: true')
+		if int(sc[1]) & 16:
+			print('\t\tBuff: true')
+		if int(sc[1]) & 32:
+			print('\t\tDebuff: true')
+		if int(sc[1]) & 64:
+			print('\t\tNoMadoReset: true')
+		if int(sc[1]) & 128:
+			print('\t\tNoAllReset: true')
+		print('\t}')
```
hemagx: I think we should add sanity check here, after each flag is set remove it's bit mask, then at the end check if the value doesn't equal 0, it will be useful in future if someone forgot to update it and/or if a person have custom flags.

src/map/status.c
```diff
@@ -13397,25 +13397,104 @@ static bool status_readdb_sizefix(char *fields[], int columns, int current)
...
+
+	if (!script->get_constant(name, &i32) || i32 <= SC_NONE || i32 >= SC_MAX) {
+		ShowWarning("status_read_scdb_libconfig_sub: Invalid status type (%s) in \"%s\" entry #%d, skipping.\n", name, source, idx);
+		return false;
 	}
 
+	libconfig->setting_lookup_bool_real(it, "Visible", &status->dbs->DisplayType[i32]);
+
+	struct config_setting_t *fg = libconfig->setting_get_member(it, "Flags");
+	if (fg != NULL)
```
hemagx: change the scope to the if only? eg ```if ((struct config_setting_t *fg = libconfig->setting_get_member(it, "Flags")) != NULL)

PR #2547, opened 2019-10-04
Dehardcode DB files path

guilherme-gm:
### Pull Request Prelude

### Changes Proposed

This pull request aims at dehardcoding all `db/*` file paths, allowing users to change which folder to load the database files from. It applies to all files used by char and map servers.

For char-server, the `db_path` global variable in `char.c` got moved to the `chr` interface in `char.h` so it is possible to use it in other files.

**Issues addressed:** 
None, as far as I know

Special thanks to @4144 for giving me some ideas

### DIFF
--- a/conf/common/map-index.conf
+++ b/conf/common/map-index.conf
@@ -1,0 +1,34 @@
+//================= Hercules Configuration ================================
+//=       _   _                     _
+//=      | | | |                   | |
+//=      | |_| | ___ _ __ ___ _   _| | ___  ___
+//=      |  _  |/ _ \ '__/ __| | | | |/ _ \/ __|
+//=      | | | |  __/ | | (__| |_| | |  __/\__ \
+//=      \_| |_/\___|_|  \___|\__,_|_|\___||___/
+//================= License ===============================================
+//= This file is part of Hercules.
+//= http://herc.ws - http://github.com/HerculesWS/Hercules
+//=
+//= Copyright (C) 2014-2019  Hercules Dev Team
+//=
+//= Hercules is free software: you can redistribute it and/or modify
+//= it under the terms of the GNU General Public License as published by
+//= the Free Software Foundation, either version 3 of the License, or
+//= (at your option) any later version.
+//=
+//= This program is distributed in the hope that it will be useful,
+//= but WITHOUT ANY WARRANTY; without even the implied warranty of
+//= MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//= GNU General Public License for more details.
+//=
+//= You should have received a copy of the GNU General Public License
+//= along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//=========================================================================
+//= Map Index configuration file.
+//=========================================================================
+
+mapindex_configuration: {
+    // Full path to the map_index.txt file
+    // Default: db/map_index.txt
+    file_path: "db/map_index.txt"
+}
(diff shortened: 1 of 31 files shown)


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2547)

4144:
look like in mapindex.h missing include

guilherme-gm:
Yes, fixed. Thanks!

src/char/geoip.c
```diff
@@ -141,7 +142,9 @@ static void geoip_init(void)
 
 	geoip->data->active = true;
 
-	db = fopen("./db/GeoIP.dat","rb");
+	char file_path[256];
+	libconfig->format_db_path("GeoIP.dat", file_path, sizeof(file_path));
+	db = fopen(file_path,"rb");
```
4144: while changed this line, better add space after coma

src/common/conf.c
```diff
@@ -32,6 +33,31 @@
...
+ * @param filename File name to format (e.g. re/item_db.conf)
+ * @param path_buf Where to save the path to
+ * @param buffer_len Maximun length of path_buf
+ */
+static void config_format_db_path(const char *filename, char *path_buf, int buffer_len)
+{
+	nullpo_retv(filename);
+	nullpo_retv(path_buf);
+	
+	snprintf(path_buf, buffer_len, "%s/%s", libconfig->db_path, filename);
```
4144: better use safesnprintf

src/map/status.c
```diff
@@ -13339,10 +13339,12 @@ static void status_read_job_db(void)
 	int i = 0;
 	struct config_t job_db_conf;
 	struct config_setting_t *jdb = NULL;
+	char config_filename[256];
+	
 #ifdef RENEWAL_ASPD
-	const char *config_filename = "db/re/job_db.conf";
+	libconfig->format_db_path("re/job_db.conf", config_filename, sizeof(config_filename));
```
4144: probably here need use defines similar to ``DBPATH``?
For example ``DBPATHRE`` and ``DBPATHPRERE``

conf/common/map-index.conf
```diff
@@ -0,0 +1,34 @@
...
+//= GNU General Public License for more details.
+//=
+//= You should have received a copy of the GNU General Public License
+//= along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//=========================================================================
+//= Map Index configuration file.
+//=========================================================================
+
+mapindex_configuration: {
+    // Full path to the map_index.txt file from the repository root.
```
MishimaHaruna: This doesn't strictly have to be a relative path from the repository root, does it? (but it can rather be an absolute path if the user wishes so)

guilherme-gm: Ah, I haven't thought about this possiblity, but I think so. I will check/update the message later today. Thank you!

guilherme-gm: Yes, an absolute path does work! Comment updated. Thanks Haru!

src/common/mapindex.c
```diff
@@ -159,8 +160,58 @@ static const char *mapindex_id2name_sub(uint16 id, const char *file, int line, c
...
+static bool mapindex_config_read(void)
+{
+	struct config_t config;
+	const char *filename = "conf/common/map-index.conf";
+	
+	if (!libconfig->load_file(&config, filename))
+		return false; // Error message is already shown by libconfig->load_file
+
+	if (!mapindex_config_read_dbpath(filename, &config))
+		return false;
```
MishimaHaruna: Memory leak, missing a `libconfig->destroy(&config);` (both here and in the successful case)

PR #2551, opened 2019-10-05
Fixes status changes effect timer when relogging

### LINKED ISSUE #2018: Effects does not reset the counter when relogin character

### Issue Prelude

### Description

Today I was testing some things in the kRO when I noticed that when the relogin character the time counter of the effects was not reset.

the bar continued from where it stopped, in hercules when you relogin character or close emulator open again the time loading bar of effects starts from the beginning again.

The bar I say is that of the effect time the one of white color and turns red when 1 min is missing.

sem titulo

[//]: # (Description of the problem or issue at length.)
[//]: # (Please specify any battle configuration related to the components of this issue that have been changed from the default values. This will allow quicker determination of the cause of the problem.)

### Expected Behavior

[//]: # (Tell us what should happen instead.)

**Branch(es):**
- [x] master
- [ ] other

**Hercules rev. hash/commit:** 

Git revision src: 7b097da8b433ea3b768b6c306c12801f7ddef0a5

[//]: # (Copy the first 3 lines of the login-server, char-server or map-server startup.)
[//]: # ( [Info]: Hercules 64-bit for Mac OS X )
[//]: # ( [Info]: Git revision src: 'a5918b329ca0826b04dca32ede783586403f58db' )
[//]: # ( [Info]: Git revision scripts: 'a5918b329ca0826b04dca32ede783586403f58db' )

[//]: # (Mac OS X 10.12.3 16D32 [x86_64])
[//]: # (Thank you for adhering to this process! It ensures your issue is resolved quickly and that neither your nor our time is needlessly wasted.)
[//]: # (This template is for problem reports. For other types of report, edit it accordingly.)
[//]: # (For fixes please create a Pull Request.)

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
This Pull Request fixes the effect timer of status changes when user relogs. Currently we send the client the remaining time as both the total time and the remaining time, thus making the timer go back to "full". The changes in this PR adds a new column to `sc_data` with the total time so we can restore the timer as it was before logging out.

I chose to make those "wrapper" functions around `status_change_start` and `clif_status_change` because they already have lots of parameters and are used in many places, simply adding another parameter would require me to update lots of places with repeated information.

**Issues addressed:** 
Fixes #2018 


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2551)

sql-files/upgrades/2019-10-05--19-01.sql
```diff
@@ -0,0 +1,28 @@
...
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+-- Adds new total_tick column
+ALTER TABLE `sc_data` ADD COLUMN `total_tick` INT(11) NOT NULL DEFAULT '0' AFTER `tick`;
```
guilherme-gm: Is it possible to not use `DEFAULT` here? as the other columns doesn't have default values, it feels wrong to keep it in this one.

sql-files/main.sql
```diff
@@ -685,6 +685,7 @@ CREATE TABLE IF NOT EXISTS `sc_data` (
   `char_id` INT(11) UNSIGNED NOT NULL,
   `type` SMALLINT(11) UNSIGNED NOT NULL,
   `tick` INT(11) NOT NULL,
+  `total_tick` INT(11) NOT NULL DEFAULT '0',
```
dastgirp: Sql upgrade file is not having default value.
So either have it on none or both

PR #2555, opened 2019-10-09
Fixes close merchant vending

### LINKED ISSUE #2554: Failed in closing vending.

### Issue Prelude

### Description
When trying to close a vending, the character freezes and can no longer move. It looks like the store is not closing properly, because when you use @refresh the store reappears even though it has been closed.

PACKETVER used: 20180404 or 20181212

### Steps To Reproduce The Issue
1. Create a store using the MC_VENDING skill (ID: 41).
2. Then close the vending.
3. At this stage your character will no longer be able to walk, he will be "frozen".
4. Use the @refresh command and notice that the Vending is still open.

**Branch(es):**
- [x] master
- [x] other

### Hercules rev. hash/commit:
Git revision src: https://github.com/HerculesWS/Hercules/commit/d026e28fd78f60dfa21da381f17f2a8cc9ae3d11

guilherme-gm:
### Pull Request Prelude

### Changes Proposed

This PR removes some checks that were preventing players to close their vending (from merchant skill). The issue was basically because we had a check that prevented a player in any trading state (including vending) to close it.

I also removed the check for a dead player, because the client itself allows that, and the same problem would happen.

**Issues addressed:** 
Fixes #2554 

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -14298,7 +14298,7 @@ static void clif_parse_CloseVending(int fd, struct map_session_data *sd) __attri
 /// 012e
 static void clif_parse_CloseVending(int fd, struct map_session_data *sd)
 {
-	if (pc_istrading(sd) || pc_isdead(sd))
+	if (sd->npc_id || sd->state.buyingstore || sd->state.trading)
 		return;
 
 	vending->close(sd);


HerculesWSAPI:
This change is [Reviewable](https://reviewable.io/reviews/herculesws/hercules/2555)

src/map/clif.c
```diff
@@ -14298,7 +14298,7 @@ static void clif_parse_CloseVending(int fd, struct map_session_data *sd) __attri
 /// 012e
 static void clif_parse_CloseVending(int fd, struct map_session_data *sd)
 {
-	if (pc_istrading(sd) || pc_isdead(sd))
+	if (sd->npc_id || sd->state.buyingstore || sd->state.trading)
```
4144: probably better leave pc_isdead? Or you think dead players should close vending too?

guilherme-gm: The original client allows you to close it while dead (tested on 2018-10), so if we keep the pc_isdead check, we will prevent them from closing while dead (although the client would still close it, as it doesn't wait for an ACK), a player could close it while dead and when they respawn the shop would be open again (client-side wise), which is a wrong behaviour.

I haven't tried, but I think adding a check that prevents them from closing while dead would also cause issues with a priest using ressurrect, example (not tested, just an idea):
- merchant dies
- merchant clicks on close (client shows it as closed / server doesn't close it)
- priest ressurrects player
- merchant still has the vending state, thus couldn't walk, but they doesn't have an open shop in their UI -> player gets stuck = bug

Issue #2349, opened 2019-01-04
mysql 8 throw error on varchar(255) table creation on myisam engine


AnnieRuru:
**Describe the bug**
mysql 8 doesn't allow table creation with varchar(255) on myisam engine
https://stackoverflow.com/questions/1814532/1071-specified-key-was-too-long-max-key-length-is-767-bytes
https://stackoverflow.com/questions/20014106/how-to-create-this-table-1071-specified-key-was-too-long-max-key-length-is/42497309

**To Reproduce**
1. install mysql 8
2. run sql-files/main.sql

**Expected behavior**
.........

**Screenshots**
![mysql8](https://raw.githubusercontent.com/AnnieRuru/customs/master/screenshots/mysql8.png)

**System specs (please complete the following information):**
 - OS: [e.g. windows 7, 64-bit]
 - Hercules Version [e.g. v2019.01.01]
 - Mode: [renewal]
 - Packet version: _still setting up test server_
 - Client type: _still setting up test server_

**Plugins used or source modifications**
_still setting up test server_

**Additional context**
https://github.com/rathena/rathena/issues/3093

dastgirp:
@4144 @Helianthella

PR #2583, opened 2019-11-22
Fix for https://github.com/HerculesWS/Hercules/issues/2349.

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

ipban_log() only inserts the first 3 octets of an IP and an asterisk for last one, so we can set the columns length of `list` in `ipbanlist` table to 13.

**Issues addressed:** #2349


sql-files/upgrades/2019-11-22--23-58.sql
```diff
@@ -0,0 +1,23 @@
...
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+ALTER TABLE `ipbanlist` MODIFY `list` VARCHAR(13) NOT NULL DEFAULT '';
+INSERT INTO `sql_updates` (`timestamp`) VALUES (1574463539);
```
dastgirp: add this in main.sql too

Kenpachi2k13: My mistake, sorry. Done!

4144:
in this pr also useless merge commit

pazkero:
Question, some custom plugins which I've saw on the forums add a command called "@ipban", for example, which is a non-automatic way of adding IP Bans.

An IP official length is 15 or 16 (xxx.xxx.xxx.xxx). Wouldn't this cause issues when IP bans are not done automatically? (Also, in some rather specific cases, it might be a desired behavior to ban only the specific IP. Yes, I am aware of how dynamic IP works and why we usually use a star.)

TL;DR Doesn't this causes issues or SQL errors if there are IP Bans of length 15?

dastgirp:
@pazkero 
If you see the source, we always add only 3 octet and wildcard

**ipban.c:**
```c
if (SQL_ERROR == SQL->Query(ipban->sql_handle, "INSERT INTO `%s`(`list`,`btime`,`rtime`,`reason`) VALUES ('%u.%u.%u.*', NOW() , NOW() +  INTERVAL %u MINUTE ,'Password error ban')",
			ipban->dbs->table, p[3], p[2], p[1], login->config->dynamic_pass_failure_ban_duration))
```

As you can see here, the value inserted is `'%u.%u.%u.*'`.

I don't know why this is done. Also for custom source, I guess you would need to manually change sql for now. I will open the issue on your behalf as you have mentioned it.

PR #2585, opened 2019-11-23
Update packets up to 2019-12-11

4144:
### Pull Request Prelude

### Changes Proposed
Update packets table up to 2019-12-11.
Update shuffle packets up to 2019-12-11.
Update packets keys up to 2019-12-11.
Update packet ZC_NOTIFY_EFFECT3 to new version.
Update packet CZ_SE_CASHSHOP_OPEN to new version.
Add script function specialeffectnum.


src/map/script.c
```diff
@@ -15815,6 +15815,56 @@ static BUILDIN(specialeffect)
 	return true;
 }
 
+/*==========================================
+ * Special effects with num [4144]
+ *------------------------------------------*/
+static BUILDIN(specialeffectnum)
+{
```
Kenpachi2k13: Maybe add something like this to prevent misuse of the command.
```
#if PACKETVER < 20191127
	ShowError("buildin_specialeffectnum: %s needs at least PACKETVER 20191127. You are using %d.", script->getfuncname(st), PACKETVER);
	st->state = END;
	return false;
#endif
```

doc/script_commands.txt
```diff
@@ -6231,6 +6231,15 @@ Example usage:
 
 ---------------------------------------
 
+*specialeffectnum(<effect number>, <num1>, <num2>{, <send_target>{, <unit id>{, <account id>}}})
+*specialeffectnum(<effect number>, <num1>, <num2>{, <send_target>{, "<NPC Name>"{, <account id>}}})
+
```
Kenpachi2k13: Please explain how `<num1>` and `<num2>` should be used.

src/map/script.c
4144: i think terminate script if something not supported is bad idea. because this is only effect, better ignore it as now

doc/script_commands.txt
4144: it always effect depend. num1 and num2 combined into int64 and then sent to effect function

Kenpachi2k13: Okay, I see...

src/map/script.c
Kenpachi2k13: At least print a warning, to notify the user that something is wrong.

4144: about error. here exists clients older than 2006?
because this packet not supported only in 2006 versions older than 2006

Kenpachi2k13: I just noticed your PACKETVER switch in `clif_specialeffect_value`. Just nevermind my change request...

doc/script_commands.txt
```diff
@@ -6231,6 +6231,15 @@ Example usage:
...
+Works same as specialeffect but also send effect numbers to client.
+For PACKETVER >= 20191127 support two numbers (num1, num2).
+For older packet versions only num1 supported.
```
dastgirp: Any info on what num2 affects?

4144: num2 is high dword of num. Script engine not support int64, because this here num and num2. Inside packet its one int32 or one int64 depend on packet version

PR #2586, opened 2019-11-24
Implement support for switching madogear type

Asheraf:
### Pull Request Prelude

### Changes Proposed
Support for the new mado gear outfit.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -4378,7 +4378,7 @@ falcon and false if they don't.
 
 ---------------------------------------
 
-*setmount({<flag>})
+*setmount({<flag>{, <type>}})
 *checkmount()
 
 If <flag> is MOUNT_NONE this command will remove the mount from the
@@ -4413,6 +4413,12 @@ The following flag values are accepted:
 
 Unlike 'setfalcon' and 'setcart' this will not work at all if they aren't of a
 class which can ride a mount.
+
+For clients PACKETVER_MAIN_NUM >= 20191120 || PACKETVER_RE_NUM >= 20191106
+you can pass a <type> flag which specifies the madogear wanted.
+The Available types:
+	MADO_ROBOT
+	MADO_SUITE
 
 The accompanying function will return MOUNT_NONE if the invoking
 character is not on a mount, and a non-zero value (according to the
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -4212,12 +4212,19 @@ ACMD(mount_peco)
 		return true;
 	}
 	if ((sd->job & MAPID_THIRDMASK) == MAPID_MECHANIC) {
+		int mtype = MADO_ROBOT;
+		if (!*message)
+			sscanf(message, "%d", &mtype);
+		if (mtype < MADO_ROBOT || mtype >= MADO_MAX) {
+			clif->message(fd, msg_fd(fd, 173)); // Please enter a valid madogear type.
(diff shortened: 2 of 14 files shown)


src/map/skill.c
```diff
@@ -9433,7 +9433,7 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 		case NC_SELFDESTRUCTION:
 			if (sd) {
 				if (pc_ismadogear(sd))
-					 pc->setmadogear(sd, false);
+					 pc->setmadogear(sd, false, 0);
```
4144: probably here need replace 0 to enum value?

src/map/script.h
```diff
@@ -551,6 +551,16 @@ enum siege_type {
...
 };
 
+/**
+ * Types of MadoGear
+ */
+enum mado_type {
+	MADO_ROBOT = 0x00,
+	// unused  = 0x01,
+	MADO_SUITE = 0x02,
+	MADO_MAX
```
4144: here probably better add ``#ifndef MADO_MAX ...``

doc/script_commands.txt
```diff
@@ -4414,6 +4414,9 @@ The following flag values are accepted:
 Unlike 'setfalcon' and 'setcart' this will not work at all if they aren't of a
 class which can ride a mount.
 
+For clients PACKETVER_MAIN_NUM >= 20191120 || PACKETVER_RE_NUM >= 20191106
+you can pass a type flag which specifies the madogear wanted.
```
Emistry:
```suggestion
you can pass a <type> which specifies the madogear wanted.
```

Emistry: perhaps list the madogear type too? 

sometime I find that its troublesome / less user friendly because had to browse `script.c` or the `doc/constants` just to find the constants which user may or may not know what are the constants name to begin with.

src/map/script.c
```diff
@@ -10688,6 +10691,8 @@ static BUILDIN(setmount)
 	if (script_hasdata(st,2))
 		flag = script_getnum(st,2);
 
+	enum mado_type mtype = script_hasdata(st, 3) ? script_getnum(st, 3) : MADO_ROBOT;
```
Emistry: what if they pass any number other than `MADO_ROBOT` and `MADO_SUITE`? 
which the madogear types doesn't exists or undefined.

src/map/atcommand.c
```diff
@@ -4212,12 +4212,19 @@ ACMD(mount_peco)
 		return true;
 	}
 	if ((sd->job & MAPID_THIRDMASK) == MAPID_MECHANIC) {
+		int mtype = MADO_ROBOT;
+		if (!*message)
+			sscanf(message, "%d", &mtype);
+		if (mtype < MADO_ROBOT || mtype >= MADO_MAX) {
```
Emistry: what if they pass any number other than MADO_ROBOT and MADO_SUITE?
which the madogear types doesn't exists or undefined.

src/map/script.c
Asheraf: We check for valid type in `mtype < MADO_ROBOT || mtype >= MADO_MAX` and show an error, anything else is handled by the client, for Gravity's clients they default anything that is not valid to 0 which is robo mado.

src/map/atcommand.c
Asheraf: same as previous comment.

PR #2588, opened 2019-11-25
Fix racial crit bonuses not being affected by katar crit bonus

skyleo:
### Pull Request Prelude

### Changes Proposed

Make racial crit bonuses affected by the katar crit bonus.
(Such as Goblin Steamrider Card for example)

### Proof/Test
First: We're assuming that this part of katar crit bonus didn't change in Renewal.
Test was done on 11.3 Ep Aegis Server.
Assassin had 42 LUK.
Equipment was Specialty Jur[4] with 4x Goblin Steamrider Cards and 2x Critical Rings.
Crit Shield of Sleeper is 5.6
Crit without Goblin Steamrider Cards is 50 - 5.6 = 44.4
IF katar crit bonus doesn't affect goblin steamrider cards then that would be 72,4 Crit against Sleepers.
Since it does affect them it's 100,4 as the test showed.

**Issues addressed:** None?

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -4661,12 +4661,16 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 	{
 		short cri = sstatus->cri;
 		if (sd != NULL) {
+			// Racial crit bonuses are affected by katar's crit bonus.
+			if (battle_config.show_katar_crit_bonus && sd->weapontype == W_KATAR)
+				cri += sd->critaddrace[tstatus->race] * 2;
+			else
+				cri += sd->critaddrace[tstatus->race];
+
 			// if show_katar_crit_bonus is enabled, it already done the calculation in status.c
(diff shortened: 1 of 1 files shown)


skyleo:
Reviews are welcome now.

src/map/battle.c
```diff
@@ -4661,13 +4661,17 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 	{
 		short cri = sstatus->cri;
 		if (sd != NULL) {
+			// Racial crit boni are affected by katar's crit bonus.
```
Emistry:
```suggestion
			// Racial crit bonus are affected by katar's crit bonus.
```

skyleo: Im willing to settle for bonuses as english is unfortunately a joke of a language that does'nt use the only true plural "boni" of bonus

skyleo: Applied "bonuses"

PR #2597, opened 2019-12-14
Deprecate set() script command.

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed
In my opinion it's pointless to have two script commands, which do the same thing in different ways.
That's why I'd like to deprecate `set()` completely, as already mentioned in doc/script_commands.txt, and where needed use `setd()` instead.

Modified `BUILDIN(setd)` to accept references to variables, too.
Marked `set()` as deprecated.
Updated doc/script_commands.txt.
Updated non-custom scripts to **not** use `set()` anymore. (Custom scripts will follow.)

**Issues addressed:** None.

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -18318,32 +18318,65 @@ static BUILDIN(swap)
 
 // [zBuffer] List of dynamic var commands --->
 
+/**
+ * Sets the value of a variable or an array.
+ * The variable's name can be passed as a string and thus can dynamically be constructed.
+ *
+ * [Kenpachi]
+ * Passing a reference to a variable is supported, too.
+ * Useful when using getvariableofnpc(), getvariableofpc(), getd() or getarg().
+ *
+ * @code{.herc}
+ *	setd("<variable name>", <value>);
+ *	setd(<reference to variable>, <value>);
+ * @endcode
+ *
+ */
 static BUILDIN(setd)
 {
-	struct map_session_data *sd = NULL;
+	int idx = 0;
+	int length = 0;
 	char varname[100];
 	const char *buffer;
-	int elem;
-	buffer = script_getstr(st, 2);
+	const void *value;
+	struct map_session_data *sd = NULL;
+	struct script_data *data = script_getdata(st, 2);
+	struct reg_db *ref = reference_getref(data);
 
-	if(sscanf(buffer, "%99[^[][%d]", varname, &elem) < 2)
-		elem = 0;
+	if (data_isreference(data)) {
+		buffer = reference_getname(data);
+		idx = reference_getindex(data);
 
-	if( not_server_variable(*varname) )
-	{
-		sd = script->rid2sd(st);
-		if( sd == NULL )
-		{
-			ShowError("script:setd: no player attached for player variable '%s'\n", buffer);
-			return true;
-		}
-	}
-
-	if (is_string_variable(varname)) {
-		script->setd_sub(st, sd, varname, elem, script_getstr(st, 3), NULL);
+		if (not_server_variable(*buffer))
+			ref = NULL;
 	} else {
-		script->setd_sub(st, sd, varname, elem, (const void *)h64BPTRSIZE(script_getnum(st, 3)), NULL);
+		buffer = script_getstr(st, 2);
 	}
+
+	/// Special case: Variable of type string was passed directly or by getd().
+	if (data_isreference(data) && reference_tovariable(data) && script_isstringtype(st, 2)) {
+		if (data->type == C_STR) /// Type has changed now, bacause script->get_val() was called by script_isstringtype().
+			buffer = data->u.str; /// Get the value of the passed variable instead of its name.
+	}
+
+	if (sscanf(buffer, "%99[^[][%d]", varname, &idx) < 2 && idx == 0)
+		idx = 0;
+
+	for (; varname[++length] != '\0';);
+
+	if (length > SCRIPT_VARNAME_LENGTH) {
+		ShowError("script:%s: Variable name %s is too long (%d)! Maximum allowed length is %d.\n",
+			  script->getfuncname(st), varname, length, SCRIPT_VARNAME_LENGTH);
+		return false;
+	}
+
+	if (not_server_variable(*varname) && (sd = script->rid2sd(st)) == NULL) {
+		ShowError("script:%s: No character attached for character variable %s!\n", script->getfuncname(st), varname);
+		return false;
+	}
+
+	value = is_string_variable(varname) ? script_getstr(st, 3) : (const void *)h64BPTRSIZE(script_getnum(st, 3));
+	script->setd_sub(st, sd, varname, idx, value, ref);
 
 	return true;
 }
@@ -26150,7 +26183,7 @@ static void script_parse_builtin(void)
 		BUILDIN_DEF(warpguild,"siii?"), // [Fredzilla]
 		BUILDIN_DEF(setlook,"ii"),
 		BUILDIN_DEF(changelook,"ii"), // Simulates but don't Store it
-		BUILDIN_DEF2(__setr,"set","rv"),
+		BUILDIN_DEF2_DEPRECATED(__setr,"set","rv"),
 		BUILDIN_DEF(setarray,"rv*"),
 		BUILDIN_DEF(cleararray,"rvi"),
 		BUILDIN_DEF(copyarray,"rri"),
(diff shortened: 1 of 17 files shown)


4144:
deprecating ``set`` will provokes errors in code. because ``setd`` a bit complex, and cant be checked by script engine in any way.

if really need deprecate ``set`` for normal assignments, probably better try to check value and if value is not reference, show warning? I not totally sure is this possible.

Kenpachi2k13:
I tested my modified version of `setd()` with (I guess) every command that returns a reference. I tested varibles and arrays. Of course I tested dynamically build strings, too. Everything works.
The only difference to `set()` is, that you can't assign variables/arrays directly with `setd()`, which isn't wanted anyway.

src/map/script.c
```diff
@@ -18318,33 +18318,66 @@ static BUILDIN(swap)
...
-		sd = script->rid2sd(st);
-		if( sd == NULL )
-		{
-			ShowError("script:setd: no player attached for player variable '%s'\n", buffer);
-			return true;
-		}
+		if (not_server_variable(*buffer))
+			ref = NULL;
+	} else {
+		buffer = script_getstr(st, 2);
```
4144: this look like allow for example ``setd(10, 20)``. probably here need check for is ``st,2`` really string?

src/map/script.c
```diff
@@ -18318,33 +18318,66 @@ static BUILDIN(swap)
...
-		script->setd_sub(st, sd, varname, elem, script_getstr(st, 3), NULL);
-	} else {
-		script->setd_sub(st, sd, varname, elem, (const void *)h64BPTRSIZE(script_getnum(st, 3)), NULL);
+	/// Special case: Variable of type string was passed directly or by getd().
+	if (data_isreference(data) && reference_tovariable(data) && script_isstringtype(st, 2)) {
+		if (data->type == C_STR) /// Type has changed now, bacause script->get_val() was called by script_isstringtype().
+			buffer = data->u.str; /// Get the value of the passed variable instead of its name.
 	}
 
+	if (sscanf(buffer, "%99[^[][%d]", varname, &idx) < 2 && idx == 0)
```
4144: even this is old code, probably better add size limit for ``%d`` because it can be too big and overflow
for example replace to this ``%30d``

src/map/script.c
Kenpachi2k13: Changed and tested.

src/map/script.c
Kenpachi2k13: Added and tested.

MishimaHaruna:
I don't think setd can be used to replace set at all. I'm looking at this example, which seems ambiguous to me:

```
setd(getd(".@var$"), "Hercules!");
```

Does that mean that it's setting the variable `.@var$` (unexpected)? Or is it setting the variable whose name is in `.@var$` (expected)?

Right now, always dereferences the variable passed to it, since it requires a string (so you can safely pass a string reference to it, and it'll evaluate the string as expected). With your suggested change, it will no longer do so, making it even more unsafe to use, since you'd always need to be careful of dereferencing the string you pass to it. An example:

```
set(getvariableofnpc(.foo$, "NPCName"), "Value"); // Takes the .foo$ variable of NPCName as reference, and sets it to "Value"
setd(getvariableofnpc(.foo$, "NPCName"), "Value"); // Takes the .foo$ variable of NPCName, dereferences it and uses its current value to build the name of the variable to set to "Value"
```

With your suggested change, the above `setd()` call would become equivalent to the deprecated `set()` call (silently breaking any scripts that relied on the existing documented value).

Kenpachi2k13:
> Does that mean that it's setting the variable `.@var$` (unexpected)? Or is it setting the variable whose name is in `.@var$` (expected)?

As expected it sets the variable whose name is in `.@var$`. That's verfied through this part of the code:
```
/// Special case: Variable of type string was passed directly or by getd().
if (data_isreference(data) && reference_tovariable(data) && script_isstringtype(st, 2)) {
	if (data->type == C_STR) /// Type has changed now, bacause script->get_val() was called by script_isstringtype().
		buffer = data->u.str; /// Get the value of the passed variable instead of its name.
}
```

> Right now, always dereferences the variable passed to it, since it requires a string (so you can safely pass a string reference to it, and it'll evaluate the string as expected). With your suggested change, it will no longer do so, making it even more unsafe to use, since you'd always need to be careful of dereferencing the string you pass to it. An example:
> 
> ```
> set(getvariableofnpc(.foo$, "NPCName"), "Value"); // Takes the .foo$ variable of NPCName as reference, and sets it to "Value"
> setd(getvariableofnpc(.foo$, "NPCName"), "Value"); // Takes the .foo$ variable of NPCName, dereferences it and uses its current value to build the name of the variable to set to "Value"
> ```
That's not completely true. `setd_sub()` looks up the variables key by its name by using `script->add_variable()`. So the reference stays valid in any case.

> With your suggested change, the above `setd()` call would become equivalent to the deprecated `set()` call (silently breaking any scripts that relied on the existing documented value).

Sure it would be equivalent, but that's supposed. The only use of `set()` is to be able to work with reference to variables. Now this can be done by using `setd()`, which makes `set()` useless. 
And it would not break anything, since `setd()` still works the way it did before, but has additional features.

Kenpachi2k13:
@MishimaHaruna Since you added `codereview:needsedits`, it would be nice if you could tell me, what exactly needs to be edited in your opinion.
I tested my changes to `setd()` in every case I could think of and everything works as expected.
Just tell me, what I should do, to prove that this works correctly and I'll try to do it.

If deprecating `set()` generally isn't wanted, just tell me, so I can stop putting effort in this. :wink:

MishimaHaruna:
Deprecating `set` is really wanted, but I'm absolutely against having `setd` take the role of set (if anything, the use of `setd` should ALSO be discouraged - but we can't do that, since the language isn't expressive enough, and several things would become impossible to do).

But the reason why I marked this as in need of edits is that it breaks existing behavior in the way it dereferences variables (I insist it does, see the following test case). And we should never break existing behavior silently.

Test case:
```
-	script	test_npc	FAKE_NPC,{
OnInit:
	.foo$ = ".foo$";
	.bar$ = "hello1";
	.foo_d$ = ".bar_d$";
	.bar_d$ = "world1";

	sleep(1000);
	// The following line should set the variable .foo$ of other_npc to "EDITED1" (set expects variable references and treats them as such)
	set(getvariableofnpc(.foo$, "other_npc"), "EDITED1");
	// The following line reads the variable .foo_d$ of other_npc as a string, and then sets the variable of test_npc whose name is equal to the value of that string, to "EDITED2" (setd expects strings, and treats them as such)
	setd(getvariableofnpc(.foo_d$, "other_npc"), "EDITED2");

	consolemes(CONSOLEMES_DEBUG, "variables in test_npc:");
	consolemes(CONSOLEMES_DEBUG, "set foo: %s; bar: %s", .foo$, .bar$);
	consolemes(CONSOLEMES_DEBUG, "setd foo: %s; bar: %s", .foo_d$, .bar_d$);
}

-	script	other_npc	FAKE_NPC,{
OnInit:
	.foo$ = ".foo$";
	.bar$ = "hello2";
	.foo_d$ = ".bar_d$";
	.bar_d$ = "world2";

	sleep(2000);
	consolemes(CONSOLEMES_DEBUG, "variables in other_npc:");
	consolemes(CONSOLEMES_DEBUG, "set foo: %s; bar: %s", .foo$, .bar$);
	consolemes(CONSOLEMES_DEBUG, "setd foo: %s; bar: %s", .foo_d$, .bar_d$);
}
```

Documented existing behavior:

```
[Debug]: consolemes: variables in test_npc:
[Debug]: consolemes: set foo: .foo$; bar: hello1
[Debug]: consolemes: setd foo: .bar_d$; bar: EDITED2
[Debug]: consolemes: variables in other_npc:
[Debug]: consolemes: set foo: EDITED1; bar: hello2
[Debug]: consolemes: setd foo: .bar_d$; bar: world2
```

New unexpected behavior:

```
[Debug]: consolemes: variables in test_npc:
[Debug]: consolemes: set foo: .foo$; bar: hello1
[Debug]: consolemes: setd foo: .bar_d$; bar: world1
[Debug]: consolemes: variables in other_npc:
[Debug]: consolemes: set foo: EDITED1; bar: hello2
[Debug]: consolemes: setd foo: .bar_d$; bar: EDITED2
```

As you can see, after this patch, the behavior of setd changes completely, and it no longer treats its argument as a string. In order to restore the old behavior, people would need to get creative and dereference their string explicitly (i.e. `setd(sprintf("%s", getvariableofnpc(.foo_d$, "other_npc")), "value");`. What's worse, any script that uses the currently documented behavior would be broken.

This is why I'm against moving the set functionality to setd, since they're two different commands with different goals.

Kenpachi2k13:
I had a look at the issue you pointed out and you're absolutely right.
Since it's almost impossible to decide whether to treat the data as reference or as string, I decided to abort this task.
`close();`

PR #2546, opened 2019-10-03
Refactor unit.c - Part I

skyleo:
### Pull Request Prelude

- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

This PR is about refactoring unit.c, the main goal is to make this file very readable and thus improve finding bugs and working on it, as well maintaining it in the long run.

The following refactoring is made per function:  
- Coding Style is updated according to our standard  
- nullpo_ret and other macros are used where approriate  
- misusing ints in a boolean way is fixed
- doxygen documentation is added  
- enums are added when appropriate  
- functions that do nothing or are never called are deleted or marked.  
- functions with misleading names are renamed. 
- reused variables with unexplanative names are changed to multiple variables with understandable names
- Code repetition is removed by adding new functions or using existing structures/macros etc
- anything I missed...
  
Note:  
This PR covers the first 700~ lines of unit.c, there might still be slight changes for oversights or some changes might be saved for when every function has been refactored.
Things such as creating a private interface for private functions will be done at the end of refactoring every function of unit.c  
  
Special Note:  
Unfortunately there is 1 commit that isn't much atomic (namely: `Refactor unit_walktoxy_timer`), I apologize for that, I've tried splitting it up with rebasing and did create some more atomic commits, but the way the hunks are I can't split them further easily and this was the best I could fix it to.  
Every commit after that is extremely atomic and should be very intuitive and easy to read. 
I promise to continue working on this as seen in my last 10~20 commits and I'll keep them as atomic as possible to make reviewing easier and enjoyable.
  
Thank you very much for reviewing my PR!

**Issues addressed:** None

[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style

Issue #2434, opened 2019-04-12
pets after evolve by pet evolution are acting like monsters


AnnieRuru:
**Describe the bug**
pets after evolve by pet evolution are acting like monsters

**To Reproduce**
1. `@makeegg 1002` create a poring egg
2. `@hatch` hatch the poring egg
3. `@petfriendly 1000` intimacy to max
4. `@item Leaf_Of_Yggdrasil 10` `@item Unripe_Apple 3` and select pet evolution to Mastering
5. `@hatch` the mastering egg
-> now your mastering is uncontrollable

**Expected behavior**
this mastering pet should behave like poring

**Screenshots**
none

**System specs (please complete the following information):**
 - OS: Windows 7 Service Pack 1 Ultimate
 - Hercules Version: 2019-04-07
 - Packet version: 20180620
 - Mode: Renewal
 - Client type: ~[2018-06-20eRagexeRE](https://rathena.org/board/topic/117168-packetver-20180620-client-release-2018-06-20eragexere/)~ [2019-02-20aRagexeRE](http://nemo.herc.ws/clients/2019-02-20aRagexeRE/) tested both

**Plugins used or source modifications**
none

**Additional context**
_Originally posted by @AnnieRuru in https://github.com/HerculesWS/Hercules/pull/2428#issuecomment-482582880_

Jedzkie:
Additional information, when you return the pet to egg state, the egg turns broken. And you can't see the egg in you inventory. You can see the broken egg only by using `Pet Incubator` or `@hatch` command

![image](https://user-images.githubusercontent.com/4826387/56075001-12c4d580-5dee-11e9-8398-bb3bc1993002.png)

Asheraf:
Re-posting my comment from pr#2428

> the bug you're reporting is related to missing pet info in the database, basically mastering have a HungerDelay value of 0 which means once you obtain it it loses all the hunger and loyalty and its gone. due to missing infos about these new pets nothing was added to their values.

Zarbony:
Got this issue too.
If i use @petfriendly 1000 after the per acting like mob, it snapping from another place to near of the master like macro using body relocation

Zarbony:
Found out what this Issue was called.
one Look at the pet_db.conf and i see that some Evolve Pets dont have any infos, like this:
`{
	Id: 1090
	SpriteName: "MASTERING"
	Name: "Mastering"
	EggItem: "Mastering_Egg"
	AutoFeed: true
	Evolve: {
[...]

Issue #303, opened 2014-07-04
Increase MAX_MOB_DB to support newest mobs w/ID 3,000+


jTynne:
Because the last thousand entries to MAX_MOB_DB are reserved for clones, requesting src/map/mob.h to be increased at least to 5,000 rather than the current:
# define MAX_MOB_DB 4000

... to support mobs with ID range of 3,000-4,000 (Jitterbug, Faceworm Queen, etc).

dastgir:
@MishimaHaruna

jTynne:
I do wonder about this. Any progress?

MishimaHaruna:
Queued in my list of things to do :x

dastgir:
@jTynne 
This was done in https://github.com/HerculesWS/Hercules/commit/228a826dfdde3581f8e5d757ec458119d6079524

However, I keep this issue open as that causes serious problem of clone's being conflicted with JobID as warning says..

@hemagx @MishimaHaruna

PR #2600, opened 2019-12-26
Preparation for upcoming pet DB update.

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

* Added enumerations for pet hunger/intimacy levels
* Added value capping to `pet_set_intimate()` function.
* Adjusted pet catch rate calculation. (Thanks to @hemagx.)
* Adjusted pet intimacy calculation when feeding. (Thanks to @hemagx.)
* Made `pet_hungry_timer` only be added if a pet's `HungerDelay` is greater than 0.
* Added a `HungerDelay` validation to `pet_hungry()` function.
* Added a base rate validation to `pet_target_check()` function.
* Added a monster ID validation to `pet_read_db_sub()` function.
* Removed `SpriteName` field from pet DB.
* Changed `EggItem` field in pet DB to be mandatory.
* Added default values and limits to pet DB fields.
* Added new field `HungerDecrement` to pet DB.
* Added new field `Intimacy.StarvingDelay` to pet DB.
* Added new field `Intimacy.StarvingDecrement` to pet DB.
* Removed `pet_hungry_friendly_decrease` config setting.
* Increased `MAX_MOB_DB` to 22000.
* Added some layout changes to the pet DB file headers. (entry structure)
* Added pet DB documentation file. (doc/pet_db.txt)
* Removed fields from pet DB where default values can be used.
* Added intimacy validation to pet DB `EquipScript` fields.
* Removed `pet_equip_min_friendly` config setting.
* Adjusted `inter_pet_tosql()` and `inter_pet_fromsql()` functions to use prepared statements.
* Added `pet_set_hunger()` function.
* Applied code style to touched functions in src/char/int_pet.c.
* Applied code style to touched functions in src/map/atcommand.c.
* Applied code style to touched functions in src/map/clif.c.
* Applied code style to touched functions in src/map/pc.c.
* Applied code style to touched functions in src/map/pet.c.
* Applied code style to touched functions in src/map/script.c.
* Applied code style to pet related part of `status_calc_pc_()` function.
* Won't apply code style to touched functions in src/map/unit.c, because of https://github.com/HerculesWS/Hercules/pull/2546.

**Issues addressed:** #2434. #303


src/map/status.c
```diff
@@ -2572,12 +2572,27 @@ static int status_calc_pc_(struct map_session_data *sd, enum e_status_calc_opt o
...
+
+#ifndef RENEWAL
+		if (pd != NULL && pd->petDB != NULL && pd->petDB->equip_script != NULL &&
+		    pd->pet.intimate >= battle_config.pet_equip_min_friendly)
+			script->run(pd->petDB->equip_script, 0, sd->bl.id, 0);
+#else
+		/**
+		 * [Kenpachi]
+		 * The intimacy will be removed, when pet DB gets updated, since the intimacy level verification will be done in the pet's 'EquipScript' field then.
+		 *
```
dastgirp: Why it will be removed?

Kenpachi2k13: That's a type. Not the intimacy will be removed, but the check for its level (`pd->pet.intimate >= PET_INTIMACY_CORDIAL`).

dastgirp:
Other than iRO, is it confirmed to have pet intimacy between 100-399?

src/map/atcommand.c
```diff
@@ -2744,39 +2744,42 @@ ACMD(guildlevelup)
...
+	else if ((id = mob->db_searchname(message)) != 0) /// For monster name.
+		;
 	else
-		if ((id = mob->db_searchname(message)) != 0) // for monster name
-			;
-		else
-			id = atoi(message);
+		id = atoi(message);
 
-	pet_id = pet->search_petDB_index(id, PET_CLASS);
```
dastgirp: I feel having assignment in new line is better and is having better readability . Is there any standard for this?

Asheraf:
> TODO: Remove initial intimacy from pet DB and make it random between 100 and 399? (https://irowiki.org/wiki/Cute_Pet_System#Mechanics)

This is an iRO only thing, other aegis servers have a clear
```c++
CPC::GetPetEgg()
{
	....
	property.relationship = 250;
	....
}
```
But feel free to add randomness support by passing an array as the value `Initial: [250, 300]`  (just don't make it default).

Kenpachi2k13:
If only iRO is using random values, I won't change it. Thanks for the info.

conf/map/battle/pet.conf
```diff
@@ -57,12 +57,19 @@ pet_attack_support: false
 // When the master receives damage from the monster, whether or not the pet attacks back. (Note 1)
 pet_damage_support: false
 
-// Minimum intimacy necessary for a pet to support their master. Default is 900
-// (intimacy goes from 0 to 1000). At this minimum, support rate is 50% of pet's normal value.
+// Minimum intimacy necessary for a pet to support their master. Default is 900 for pre-RE and 910 for RE.
+// (Intimacy goes from 0 to 1000.) At this minimum, support rate is 50% of pet's normal value.
 // At max (1000) support rate is 150%.
+// Min value: 0
+// Max value: 950
```
dastgirp: 2&3 lines above, max is written as 1000, so why this is limited till 950?

src/map/atcommand.c
Kenpachi2k13: I don't know about a standard for this, but since assignments inside an if-condition are used all over the code, and I like that syntax, I'd prefer this. Just look a few lines above:
`if ((item_data = itemdb->search_name(message)) != NULL)`
`if ((id = mob->db_searchname(message)) != 0)`

conf/map/battle/pet.conf
Kenpachi2k13: The max value for intimacy is 1000. The max value for these two settings is 950.
I actually don't know, why it's capped to 950, but that's nothing I've changed. I just wrote that note to the setting descriptions.

Kenpachi2k13: Okay, I did some research. These settings' max value needs to be less than 1000 because the calculation of `pd->rate_fix` would end up in a devided by 0 error otherwise.
`status_calc_pet_()`:
```pd->rate_fix = 1000*(pd->pet.intimate - battle_config.pet_support_min_friendly)/(1000- battle_config.pet_support_min_friendly) +500;```

File: src/map/pet.c [hunk 4]
```diff
@@ -162,48 +203,55 @@ static int pet_attackskill(struct pet_data *pd, int target_id)
...
+ *
+ */
 static int pet_target_check(struct map_session_data *sd, struct block_list *bl, int type)
 {
-	struct pet_data *pd;
 	int rate;
+	struct pet_data *pd;
 
 	nullpo_ret(sd);
+	nullpo_ret(bl);
```
hemagx: Fix variables scope here. move nullpos to top.

src/map/pet.c
```diff
@@ -226,60 +274,66 @@ static int pet_sc_check(struct map_session_data *sd, int type)
...
+ * @param data Unused.
+ * @return 1 on failure, 0 on success.
+ *
+ */
 static int pet_hungry(int tid, int64 tick, int id, intptr_t data)
 {
+	int interval;
 	struct map_session_data *sd;
 	struct pet_data *pd;
-	int interval;
```
hemagx: Fix variables scope here.

src/map/pet.c
```diff
@@ -226,60 +274,66 @@ static int pet_sc_check(struct map_session_data *sd, int type)
...
+
+	if (battle_config.pet_hungry_delay_rate != 100)
+		interval = pd->petDB->hungry_delay * battle_config.pet_hungry_delay_rate / 100;
+	else
+		interval = pd->petDB->hungry_delay;
 
 	pd->pet.hungry--;
-	/* Pet Autofeed */
+
+	/** Pet Autofeed */
```
hemagx: if it starts with a single * let it end up with single *, if two then let it end up with two etc..

src/map/pet.c
```diff
@@ -226,60 +274,66 @@ static int pet_sc_check(struct map_session_data *sd, int type)
...
-		}
+
 		status_calc_pet(pd, SCO_NONE);
-		clif->send_petdata(sd,pd,1,pd->pet.intimate);
+		clif->send_petdata(sd, pd, 1, pd->pet.intimate);
+		interval = 20000; /// While starving, it's every 20 seconds.
 	}
-	clif->send_petdata(sd,pd,2,pd->pet.hungry);
-
-	if(battle_config.pet_hungry_delay_rate != 100)
```
hemagx: with this code change the rate would not longer affect the special case of starving, please move it back to where it was.

src/map/pet.c
```diff
@@ -315,21 +369,30 @@ static int pet_hungry_timer_delete(struct pet_data *pd)
...
 {
 	int val;
 
 	nullpo_retr(1, pd);
-	if (pd->pet.intimate > 900)
-		val = (pd->petDB->s_perfor > 0)? 4:3;
-	else if(pd->pet.intimate > 750) //TODO: this is way too high
+
+	if (pd->pet.intimate >= PET_INTIMACY_LOYAL)
+		val = pd->petDB->s_perfor > 0 ? 4 : 3;
```
hemagx: Parentheses around expression please

src/map/pet.c
```diff
@@ -315,21 +369,30 @@ static int pet_hungry_timer_delete(struct pet_data *pd)
...
+ * @param sd Unused.
+ * @param pd The pet.
+ * @return Always 1.
+ *
+ */
 static int pet_performance(struct map_session_data *sd, struct pet_data *pd)
 {
 	int val;
 
 	nullpo_retr(1, pd);
```
hemagx: Fix variables scope here. move nullpos to top.

src/map/pet.c
```diff
@@ -563,57 +626,63 @@ static int pet_catch_process1(struct map_session_data *sd, int target_class)
...
+ */
 static int pet_catch_process2(struct map_session_data *sd, int target_id)
 {
+	int i = 0;
+	int pet_catch_rate = 0;
 	struct mob_data *md = NULL;
 	struct block_list *bl = NULL;
-	int i = 0, pet_catch_rate = 0;
 
 	nullpo_retr(1, sd);
```
hemagx: Fix variables scope here. move nullpos to top.

src/map/pet.c
```diff
@@ -563,57 +626,63 @@ static int pet_catch_process1(struct map_session_data *sd, int target_class)
...
+
+	if (md == NULL || md->bl.prev == NULL) { /// Invalid inputs/state, abort capture.
+		clif->pet_roulette(sd, 0);
 		sd->catch_target_class = -1;
 		sd->itemid = -1;
 		sd->itemindex = -1;
 		return 1;
 	}
 
-	//FIXME: delete taming item here, if this was an item-invoked capture and the item was flagged as delay-consume [ultramage]
```
hemagx: FIXME comments prefix ```//FIXME``` shall stay as it is, and also TODO etc

src/map/pet.c
```diff
@@ -670,42 +739,49 @@ static bool pet_get_egg(int account_id, int pet_class, int pet_id)
...
 	struct item_data *egg_id;
-	nullpo_ret(sd);
-	if (sd->pd == NULL)
-		return 1;
 
-	//You lost the pet already.
-	if(!sd->status.pet_id || sd->pd->pet.intimate <= 0 || sd->pd->pet.incubate)
-		return 1;
+	nullpo_retr(1, sd);
+	nullpo_retr(1, sd->pd);
```
hemagx: Fix variables scope here. move nullpos to top.

src/map/pet.c
```diff
@@ -830,50 +906,59 @@ static int pet_unequipitem(struct map_session_data *sd, struct pet_data *pd)
...
+ */
 static int pet_food(struct map_session_data *sd, struct pet_data *pd)
 {
-	int i, food_id;
+	int food_id;
+	int i;
 
+	nullpo_retr(1, sd);
 	nullpo_retr(1, pd);
+
```
hemagx: Fix variables scope here. move nullpos to top.

src/map/pet.c
```diff
@@ -918,117 +1003,128 @@ static int pet_randomwalk(struct pet_data *pd, int64 tick)
...
+ * @param pd The pet.
+ * @param sd The pet's master.
+ * @param tick Timestamp of last support.
+ * @return Always 0.
+ *
+ */
 static int pet_ai_sub_hard(struct pet_data *pd, struct map_session_data *sd, int64 tick)
 {
 	struct block_list *target = NULL;
+
```
hemagx: Fix variables scope here. move nullpos to top.

src/map/pet.c
```diff
@@ -918,117 +1003,128 @@ static int pet_randomwalk(struct pet_data *pd, int64 tick)
...
 		if (DIFF_TICK(tick, pd->ud.canmove_tick) < 0)
-			return 0; //Can't move yet.
-		pd->status.speed = (sd->battle_status.speed>>1);
-		if(pd->status.speed <= 0)
-			pd->status.speed = 1;
-		if (!unit->walktobl(&pd->bl, &sd->bl, 3, 0))
-			pet->randomwalk(pd,tick);
+			return 0; /// Can't move yet.
+
+		pd->status.speed = sd->battle_status.speed >> 1;
```
hemagx: Change to / 2

src/map/pet.c
```diff
@@ -918,117 +1003,128 @@ static int pet_randomwalk(struct pet_data *pd, int64 tick)
...
-			return 0; //Wait until the pet finishes walking back to master.
+			return 0; /// Wait until the pet finishes walking back to master.
+
 		pd->status.speed = pd->petDB->speed;
-		pd->ud.state.change_walk_target = pd->ud.state.speed_changed = 1;
+		pd->ud.state.speed_changed = 1;
+		pd->ud.state.change_walk_target = 1;
 	}
 
 	if (pd->target_id) {
```
hemagx: ``` != 0 ```

src/map/pet.c
```diff
@@ -1142,45 +1238,53 @@ static int pet_lootitem_drop(struct pet_data *pd, struct map_session_data *sd)
...
+ *
+ */
 static int pet_skill_bonus_timer(int tid, int64 tick, int id, intptr_t data)
 {
-	struct map_session_data *sd=map->id2sd(id);
-	struct pet_data *pd;
 	int bonus;
 	int duration = 0;
+	struct map_session_data *sd;
+	struct pet_data *pd;
```
hemagx: Fix variables scope here.

src/map/script.c
```diff
@@ -10995,29 +10995,32 @@ static BUILDIN(cooking)
...
+ *	makepet(<pet id>);
+ * @endcode
+ *
+ */
 static BUILDIN(makepet)
 {
+	int id = script_getnum(st, 2);
+	int pet_id;
 	struct map_session_data *sd;
-	int id,pet_id;
```
hemagx: Fix variables scope here.

src/map/script.c
```diff
@@ -27164,6 +27167,23 @@ static void script_hardcoded_constants(void)
...
 
+	script->constdb_comment("pet hunger levels");
+	script->set_constant("PET_HUNGER_NONE", PET_HUNGER_NONE, false, false);
+	script->set_constant("PET_HUNGER_VERY_HUNGRY", PET_HUNGER_VERY_HUNGRY, false, false);
+	script->set_constant("PET_HUNGER_HUNGRY", PET_HUNGER_HUNGRY, false, false);
+	script->set_constant("PET_HUNGER_NEUTRAL", PET_HUNGER_NEUTRAL, false, false);
+	script->set_constant("PET_HUNGER_SATISFIED", PET_HUNGER_SATISFIED, false, false);
+	script->set_constant("PET_HUNGER_STUFFED", PET_HUNGER_STUFFED, false, false);
+
+	script->constdb_comment("pet intimacy levels");
```
hemagx: Pet*

src/map/script.c
```diff
@@ -27164,6 +27167,23 @@ static void script_hardcoded_constants(void)
 	script->set_constant("PETINFO_EVO_EGGID", PETINFO_EVO_EGGID, false, false);
 	script->set_constant("PETINFO_AUTOFEED", PETINFO_AUTOFEED, false, false);
 
+	script->constdb_comment("pet hunger levels");
```
hemagx: Pet*

src/map/battle.c
```diff
@@ -7092,7 +7092,11 @@ static const struct battle_data {
 	{ "pet_status_support",                 &battle_config.pet_status_support,              0,      0,      1,              },
 	{ "pet_attack_support",                 &battle_config.pet_attack_support,              0,      0,      1,              },
 	{ "pet_damage_support",                 &battle_config.pet_damage_support,              0,      0,      1,              },
+#ifdef RENEWAL
+	{ "pet_support_min_friendly_re",        &battle_config.pet_support_min_friendly,        910,    0,      950,            },
```
Emistry: just curios, why a new config? since both just using different value? 
could have just remain one config, its not possible to trigger/use both in server.
this config seem redundant.

src/map/pet.c
```diff
@@ -563,57 +626,63 @@ static int pet_catch_process1(struct map_session_data *sd, int target_class)
...
+	if (i < 0 || sd->catch_target_class != md->class_) {
+		clif->emotion(&md->bl, E_AG); /// Mob will do /ag if wrong lure is used on it.
+		clif->pet_roulette(sd, 0);
 		sd->catch_target_class = -1;
 		return 1;
 	}
 
-	pet_catch_rate = (pet->db[i].capture + (sd->status.base_level - md->level)*30 + sd->battle_status.luk*20)*(200 - get_percentage(md->status.hp, md->status.max_hp))/100;
+	pet_catch_rate = (pet->db[i].capture + (sd->status.base_level - md->level) * 30 + sd->battle_status.luk * 20) * (200 - get_percentage(md->status.hp, md->status.max_hp)) / 100;
+	pet_catch_rate = max(pet_catch_rate, 1);
```
Emistry:
```suggestion
	pet_catch_rate = cap_value(pet_catch_rate, 1, 10000);
```
maybe `cap_value` are more appropriate? and maybe move to lower part (or add another too)? since the `pet_catch_rate` value could be updated again later.

src/map/battle.c
Emistry: just curios, why a new config? this config seem redundant.
could have just remain one config, and assign different value based on `RE/PRE-RE`?

src/map/pet.c
```diff
@@ -563,57 +626,63 @@ static int pet_catch_process1(struct map_session_data *sd, int target_class)
...
 
-	if(pet_catch_rate < 1) pet_catch_rate = 1;
-	if(battle->bc->pet_catch_rate != 100)
-		pet_catch_rate = (pet_catch_rate*battle->bc->pet_catch_rate)/100;
+	if (battle->bc->pet_catch_rate != 100)
```
Emistry: what if they put catch rate as `0`, server crash maybe? since the config seem like allow `0` rate. or maybe limit the config minimum value to `1`.

src/map/pet.c
hemagx: actually, this sounds so reasonable.

src/map/pet.c
Kenpachi2k13: Nothing would happen. If `battle->bc->pet_catch_rate` is 0 the calculation would return 0:
```pet_catch_rate = pet_catch_rate * 0 / 100;``` ->
```pet_catch_rate = 0 / 100;``` ->
```pet_catch_rate = 0;```

File: src/map/battle.c (hunk 21 above)
Comment (Kenpachi2k13) [reply src/map/battle.c]: Sure it's redundant. I just added a separate config for user friendliness, so that `pet_support_min_friendly` don't have to be adjusted if server is compiled in pre-RE mode.
This may gets removed, depending on `PET_INTIMACY_LOYAL` has changed to 910 or not. (See @hemagx' requested changes.)

Comment (Kenpachi2k13) [reply src/map/battle.c]: It hasn't changed. New config gets removed.

File: src/map/atcommand.c [hunk 24]
```diff
@@ -2798,73 +2802,89 @@ ACMD(hatch)
...
+		clif->message(fd, msg_fd(fd, 1016)); /// Please enter a valid value (usage: @petfriendly <0-1000>).
 		return false;
 	}
 
-	pd = sd->pd;
-	if (!pd) {
-		clif->message(fd, msg_fd(fd,184)); // Sorry, but you have no pet.
+	int friendly = atoi(message);
+
+	if (friendly < PET_INTIMACY_NONE || friendly > PET_INTIMACY_MAX) {
```
Emistry: instead of checking for min/max value, how about just cap the value and let it proceed?
```
friendly = cap_value(friendly, PET_INTIMACY_NONE, PET_INTIMACY_MAX);
```

src/map/atcommand.c
```diff
@@ -2798,73 +2802,89 @@ ACMD(hatch)
...
+		clif->message(fd, msg_fd(fd, 1017)); /// Please enter a valid number (usage: @pethungry <0-100>).
 		return false;
 	}
 
-	pd = sd->pd;
-	if (!sd->status.pet_id || !pd) {
-		clif->message(fd, msg_fd(fd,184)); // Sorry, but you have no pet.
+	int hungry = atoi(message);
+
+	if (hungry < PET_HUNGER_NONE || hungry > PET_HUNGER_STUFFED) {
```
Emistry: instead of checking for min/max value, how about just cap the value and let it proceed?
```suggestion
	hungry = cap_value(hungry, PET_HUNGER_NONE, PET_HUNGER_STUFFED);
```

src/map/clif.c
```diff
@@ -10727,7 +10727,7 @@ static void clif_parse_LoadEndAck(int fd, struct map_session_data *sd)
 			}
 		}
 
-		if(sd->pd && sd->pd->pet.intimate > 900)
+		if(sd->pd != NULL && sd->pd->pet.intimate >= PET_INTIMACY_LOYAL)
 			clif->pet_emotion(sd->pd,(sd->pd->pet.class_ - 100)*100 + 50 + pet->hungry_val(sd->pd));
```
Emistry:
```suggestion
			clif->pet_emotion(sd->pd, (sd->pd->pet.class_ - 100) * 100 + 50 + pet->hungry_val(sd->pd));
```

src/map/pet.c
```diff
@@ -60,62 +60,102 @@ struct pet_interface *pet;
...
 static int pet_create_egg(struct map_session_data *sd, int item_id)
 {
-	int pet_id = pet->search_petDB_index(item_id, PET_EGG);
 	nullpo_ret(sd);
-	if (pet_id < 0) return 0; //No pet egg here.
-	if (!pc->inventoryblank(sd)) return 0; // Inventory full
+
+	int pet_id = pet->search_petDB_index(item_id, PET_EGG);
+
+	if (pet_id < 0)
```
Emistry:
```suggestion
	if (pet_id == INDEX_NOT_FOUND)
```

src/map/pet.c
```diff
@@ -563,57 +623,57 @@ static int pet_catch_process1(struct map_session_data *sd, int target_class)
...
+	/// catch_target_class == 0 is used for universal lures (except bosses for now). [Skotlex]
+	if (sd->catch_target_class == 0 && (md->status.mode & MD_BOSS) == 0)
 		sd->catch_target_class = md->class_;
-	if(i < 0 || sd->catch_target_class != md->class_) {
-		clif->emotion(&md->bl, E_AG); //mob will do /ag if wrong lure is used on them.
-		clif->pet_roulette(sd,0);
+
+	int i = pet->search_petDB_index(md->class_, PET_CLASS);
+
+	if (i < 0 || sd->catch_target_class != md->class_) {
```
Emistry:
```suggestion
	if (i == INDEX_NOT_FOUND || sd->catch_target_class != md->class_) {
```

src/map/pet.c
```diff
@@ -563,57 +623,57 @@ static int pet_catch_process1(struct map_session_data *sd, int target_class)
...
+		clif->emotion(&md->bl, E_AG); /// Mob will do /ag if wrong lure is used on it.
+		clif->pet_roulette(sd, 0);
 		sd->catch_target_class = -1;
 		return 1;
 	}
 
-	pet_catch_rate = (pet->db[i].capture + (sd->status.base_level - md->level)*30 + sd->battle_status.luk*20)*(200 - get_percentage(md->status.hp, md->status.max_hp))/100;
+	int pet_catch_rate = pet->db[i].capture * (100 - get_percentage(md->status.hp, md->status.max_hp)) / 100 + pet->db[i].capture;
+	pet_catch_rate = cap_value(pet_catch_rate, 1, 10000);
```
Emistry:
```suggestion
```
move to lower line, due to value still get update after this.

src/map/pet.c
```diff
@@ -563,57 +623,57 @@ static int pet_catch_process1(struct map_session_data *sd, int target_class)
...
 
-	if(pet_catch_rate < 1) pet_catch_rate = 1;
-	if(battle->bc->pet_catch_rate != 100)
-		pet_catch_rate = (pet_catch_rate*battle->bc->pet_catch_rate)/100;
+	if (battle->bc->pet_catch_rate != 100)
+		pet_catch_rate *= battle->bc->pet_catch_rate / 100;
```
Emistry:
```suggestion
	pet_catch_rate = cap_value(pet_catch_rate, 1, 10000);
```
cap the value after done all calculations.

src/map/script.c
```diff
@@ -10995,29 +10995,33 @@ static BUILDIN(cooking)
...
-	sd = script->rid2sd(st);
-	if( sd == NULL )
+	if ((sd = script->rid2sd(st)) == NULL)
 		return true;
 
-	pet_id = pet->search_petDB_index(id, PET_CLASS);
+	int id = script_getnum(st, 2);
+	int pet_id = pet->search_petDB_index(id, PET_CLASS);
 
 	if (pet_id < 0)
```
Emistry:
```suggestion
	if (pet_id == INDEX_NOT_FOUND)
```

src/map/script.c
```diff
@@ -10995,29 +10995,33 @@ static BUILDIN(cooking)
...
 		pet_id = pet->search_petDB_index(id, PET_EGG);
-	if (pet_id >= 0 && sd) {
+
+	if (pet_id >= 0) {
```
Emistry:
```suggestion
	if (pet_id != INDEX_NOT_FOUND) {
```

src/map/atcommand.c
Kenpachi2k13: I think throwing an error message here, to notifiy the user, is the better way,

src/map/atcommand.c
Kenpachi2k13: Same as above. I think throwing an error message here, to notifiy the user, is the better way,

src/map/clif.c
Kenpachi2k13: Actually you're right, but when I start applying code style here, I'd have to do it for the whole 350+ lines of that function.
This seems to be a little bit too much effort just for changing a value in an if-condition. Don't you agree?

src/map/pet.c
Kenpachi2k13: Nope. It should be capped before apllying `battle->bc->pet_catch_rate`.  This ensures that the calculation works properly.
An example case. The base calculation returns (for whatever reason) 11000 and `battle->bc->pet_catch_rate` is set to 50.
Let's see what happens when applying `battle->bc->pet_catch_rate`  before capping:
```
int pet_catch_rate = 11000;
pet_catch_rate *= battle->bc->pet_catch_rate / 100; // 11000 * 50 / 100 = 550000 / 100 = 5500
// Capping would do anything.
```

And now let's see what prior capping does:
```
int pet_catch_rate = 11000;
// Capping to 10000
pet_catch_rate *= battle->bc->pet_catch_rate / 100; // 10000 * 50 / 100 = 50000 / 100 = 5000
```

See? This would end up with the expected value of 5000, while your execution order would not.

src/map/atcommand.c
Emistry: Imo, I find this this way are "more friendly". 
For example: current commands like `@zeny 123` give 100 Zeny or  `@stats 100` would give +100 stat, and `@stats 129837` would give me a max stats (+32767), and wont block me and throw a error message like `Please type number between 1~32767`. 
However it does have some cons that give negative result due to overflowed numbers like `@zeny 102938120983019283` may result in negative amount.

Kenpachi2k13: @Emistry: I agree with you in most cases, but an example: The user's intention was to set intimacy to 900 but accidentally wrote 9000. When capping the value, the intimacy would be set to max, even though this was not what user wanted to do. So when whrowing an error message, there's at least a chance to correct this.
BTW; You don't have to instantly resolve a conservation just because I have another opinion. You're welcome to throw in your suggestion and discuss it. I'm always open for suggestions and advices. :-)

Emistry: > The user's intention was to set intimacy to 900 but accidentally wrote 9000. When capping the value, the intimacy would be set to max, even though this was not what user wanted to do. So when throwing an error message, there's at least a chance to correct this.

ah i see. 
about the resolve conversation, its ok, I just randomly throwing question/opinion sometime. no worry :D everything good.

db/pre-re/pet_db.conf
```diff
@@ -33,193 +33,41 @@ pet_db:(
...
+	Name: "Pet Name"                     (string)
+	EggItem: Egg Id                      (string)
 	// ================ Optional fields ===============================
-	TamingItem: "Taming_Item_Name"                             (string)
-	FoodItem: "Food_Item_Name"                                 (string)
-	AccessoryItem: "Accessory_Item_Name"                       (string)
-	FoodEffectiveness: HungerPoints                               (int)
-	HungerDelay: IntervalDuration                                 (int)
-	HungerDecrement: HungerPoints                                 (int)
+	TamingItem: Taming Item              (string, defaults to 0)
```
dastgirp: Same for some fields like TamingItem, FoodId, AccessoryItem, its better to quote them so no mistakes are made by others.

db/pre-re/pet_db.conf
```diff
@@ -33,193 +33,41 @@ pet_db:(
  **************************************************************************
 {
 	// ================ Mandatory fields ==============================
-	Id: MonsterID						      (int)
-	Name: "Pet Name"                                           (string)
-	EggItem: "Egg_Item_Name"                                   (string)
+	Id: ID                               (int)
+	Name: "Pet Name"                     (string)
+	EggItem: Egg Id                      (string)
```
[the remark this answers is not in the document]
Kenpachi2k13: You're absolutely right... Changed!

db/pre-re/pet_db.conf
Kenpachi2k13: Changed!

MishimaHaruna:
Splitting this into atomic commits before it can be reviewed and merged...

Kenpachi2k13:
Changed commits to be atomic.
Code style will be applied later.

Kenpachi2k13:
Code style is applied, too, now.
Ready for re-review.

src/map/pet.c
```diff
@@ -592,7 +592,7 @@ static int pet_catch_process2(struct map_session_data *sd, int target_id)
 		return 1;
 	}
 
-	pet_catch_rate = (pet->db[i].capture + (sd->status.base_level - md->level)*30 + sd->battle_status.luk*20)*(200 - get_percentage(md->status.hp, md->status.max_hp))/100;
+	pet_catch_rate = pet->db[i].capture * (100 - get_percentage(md->status.hp, md->status.max_hp)) / 100 + pet->db[i].capture;
```
MishimaHaruna: Since this is quite a big reduction of the success rate, I can see how controversial this change might be on an already launched server. What do you think about offering a battle flag to toggle between the custom eAthena formula and the official formula, like we do for various other things (and defaulting to the official formula)?

src/char/int_pet.c
```diff
@@ -139,36 +143,47 @@ static int inter_pet_fromsql(int pet_id, struct s_pet *p)
...
-	{
-		Sql_ShowDebug(inter->sql_handle);
+	const char *query = "SELECT "
+		"`class`, `name`, `account_id`, `char_id`, `level`, `egg_id`, `equip`, `intimate`, `hungry`, `rename_flag`, `incubate`, `autofeed` "
+		"FROM `%s` WHERE `pet_id`=?";
+
+	if (SQL_ERROR == SQL->StmtPrepare(stmt, query, pet_db) ||
+	    SQL_ERROR == SQL->StmtBindParam(stmt, 0, SQLDT_UINT32, &pet_id, sizeof(pet_id)) ||
+	    SQL_ERROR == SQL->StmtExecute(stmt) ||
+	    SQL_ERROR == SQL->StmtBindColumn(stmt, 0, SQLDT_UINT32, &p->class_, sizeof(p->class_), NULL, NULL) ||
```
MishimaHaruna: Can you double-check the types of these fields? They're pretty much all signed and have platform-dependent sizes, so most of these should be changed to `SQLDT_INT`, `SQLDT_SHORT`, etc. (matching the actual data type in the struct)

src/map/pet.c
Kenpachi2k13: Sounds reasonable. I'll implement this.

src/char/int_pet.c
MishimaHaruna: Yes, our sql code is a bit messy and not very well documented, unfortunately.

The `SqlStmt_BindColumn()` and `SqlStmt_BindParam()` functions (through `Sql_P_BindSqlDataType()` set the `buffer_type` and `is_unsigned` fields of the `MYSQL_BIND` struct that is later passed to `mysql_stmt_fetch_column()` / `mysql_stmt_bind_param()`.

The [MySQL C API documentation](https://dev.mysql.com/doc/refman/8.0/en/c-api-prepared-statement-type-codes.html) for those `MYSQL_TYPE_` constants used mentions:

> The buffer_type member of MYSQL_BIND structures indicates the data type of the C language variable bound to a statement parameter or result set column. For input, buffer_type indicates the type of the variable containing the value to be sent to the server. For output, it indicates the type of the variable into which a value received from the server should be stored.

(and it also shows which types are allowed for the different SQL column types.

We should probably update our documentation to mention this...

Kenpachi2k13: Okay, so for input/output data always use the SQL data type corresponding to the C data type. That's described precisely.

But what about params used for comparison, like `pet_id` in this statement?
That's not mentioned in the linked API documentation. (Or I just didn't notice it.)

src/map/pet.c
Kenpachi2k13: Done.

src/char/int_pet.c
Kenpachi2k13: Done.

src/map/pet.c
```diff
@@ -433,68 +433,73 @@ static int pet_return_egg(struct map_session_data *sd, struct pet_data *pd)
...
 	map->addiddb(&pd->bl);
-	status_calc_pet(pd,SCO_FIRST);
-
+	status_calc_pet(pd, SCO_FIRST);
 	pd->last_thinktime = timer->gettick();
 	pd->state.skillbonus = 0;
 
-	if( battle_config.pet_status_support )
-		script->run_pet(pet->db[i].pet_script,0,sd->bl.id,0);
+	if (pd->petDB != NULL) {
```
MishimaHaruna, line 497: This is initialized to a value that should never be NULL (`&pet->db[i]`) just a bit above, so it's probably fine to remove the NULL check here (not sure why it existed to begin with)

src/map/pet.c
```diff
@@ -437,15 +437,12 @@ static int pet_data_init(struct map_session_data *sd, struct s_pet *petinfo)
 		if( pd->petDB->equip_script )
 			status_calc_pc(sd,SCO_NONE);
 
-		if( battle_config.pet_hungry_delay_rate != 100 )
-			interval = (pd->petDB->hungry_delay*battle_config.pet_hungry_delay_rate)/100;
-		else
-			interval = pd->petDB->hungry_delay;
+		if (pd->petDB->hungry_delay > 0) {
+			int interval = pd->petDB->hungry_delay * battle_config.pet_hungry_delay_rate / 100;
+			pd->pet_hungry_timer = timer->add(timer->gettick() + max(interval, 1), pet->hungry, sd->bl.id, 0);
```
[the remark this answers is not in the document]
Kenpachi2k13: Done.

src/map/pet.c
Kenpachi2k13: Done.

Issue #1861, opened 2017-10-17
warpparty() and warpguild() are checking for nowarp mapflag


bWolfie:
### Issue Prelude

[//]: # (Please complete these mandatory steps and check the following boxes by putting an `x` inside the brackets _before_ filing your issue)

-  [x] I have not modified the source prior to reproducing this issue.
-  [x] I am using the latest version of Hercules.
-  [x] I am aware that this report will be closed or deleted if it becomes obvious that I am stating the false.

### Description

[//]: # (Description of the problem or issue at length.)
[//]: # (Please specify any battle configuration related to the components of this issue that have been changed from the default values. This will allow quicker determination of the cause of the problem.)

`BUILDIN(warpparty)` and `BUILDIN(warpguild)` are checking for the `nowarp` and `noreturn` mapflags.

### Current Behavior

[//]: # (Describe at length what you noticed during your analysis.)
[//]: # (If this is a crash, post the core/stack-dump or crash-log to https://gist.github.com/)
[//]: # (If you are referencing from sources such as iROwiki or ratemyserver.net, please quote specific information rather than providing the links alone.)
These script commands cannot be used to warp guilds/parties around on nowarp mapflags.

### Expected Behavior

[//]: # (Tell us what should happen instead.)
Just like the regular `warp()` command, you should be able to warp regardless.

**Branch(es):**
- [x] master
- [ ] other

mekolat:
I believe it would be a better idea to just add an extra parameter to `warp()` so it can take a GID, and then deprecate `warpparty()` and `warpguild()` since `getunits()` makes them obsolete. It will also help avoid having yet more warp commands when we merge the Clan system

psychonaut04:
I did something just to test.

https://github.com/HerculesWS/Hercules/compare/stable...psychonaut04:patch-2

Deprecated warpparty and warpguild, just with a warning(it still works).
Warp using switch, a char change it to warpguild, can change it somehow still to add more.

Just a code testing(i did for fun only) to give .

PR #2604, opened 2020-01-05
Add flag to ignore mapflag restrictions in warpparty() and warpguild() script commands.

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed
* Added a flag to ignore mapflag restrictions in `warpparty()` and `warpguild()` script commands.
* Applied code style to `warpparty()` and `warpguild()` script commands.
* Updated doc/script_command.txt.

**Issues addressed:** #1861

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -6828,21 +6828,17 @@ static BUILDIN(warpparty)
 
 		switch( type ) {
 		case 0: // Random
-			if (!map->list[pl_sd->bl.m].flag.nowarp)
-				pc->randomwarp(pl_sd, CLR_TELEPORT);
+			pc->randomwarp(pl_sd, CLR_TELEPORT);
 			break;
 		case 1: // SavePointAll
-			if (!map->list[pl_sd->bl.m].flag.noreturn)
-				pc->setpos(pl_sd, pl_sd->status.save_point.map, pl_sd->status.save_point.x, pl_sd->status.save_point.y, CLR_TELEPORT);
+			pc->setpos(pl_sd, pl_sd->status.save_point.map, pl_sd->status.save_point.x, pl_sd->status.save_point.y, CLR_TELEPORT);
 			break;
 		case 2: // SavePoint
-			if (!map->list[pl_sd->bl.m].flag.noreturn)
-				pc->setpos(pl_sd, sd->status.save_point.map, sd->status.save_point.x, sd->status.save_point.y, CLR_TELEPORT);
+			pc->setpos(pl_sd, sd->status.save_point.map, sd->status.save_point.x, sd->status.save_point.y, CLR_TELEPORT);
 			break;
 		case 3: // Leader
 		case 4: // m,x,y
-			if (!map->list[pl_sd->bl.m].flag.noreturn && !map->list[pl_sd->bl.m].flag.nowarp)
-				pc->setpos(pl_sd, map_index, x, y, CLR_TELEPORT);
+			pc->setpos(pl_sd, map_index, x, y, CLR_TELEPORT);
 			break;
 		}
 	}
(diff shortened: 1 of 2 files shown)


Kenpachi2k13:
There is no script that relies on that restrictions. `warpguild()` isn't used in any script and `warpparty()`is only used in `F_CashPartyCall()`.
Iin my opinion merging all warp commands into one seems to be a very bad idea. It would cause a parameter overload. And it would also cause that the meaning of one parameter depends on the value of another parameter (flag for passed ID is a guild or party ID) which isn't good programming practice.

Asheraf:
I am not talking about the scripts included with the repository (those can always be fixed), but the tens of thousands of scripts written by the community since these commands were introduced a long long time ago, unfortunately opinions on how it should work does not matter here, the scripting engine MUST ALWAYS keep backward compatibility, unless you stop the script from loading you cannot change the behavior in a manner that breaks that compatibility.

Kenpachi2k13:
Ah, okay. You're right. Thanks for pointing that out.
Requested changes are committed, please re-review.

As a side note:
The documentation for `warpparty()` implies that `<include leader>` can be passed without passing `<from map name>`, which is wrong but makes sense.
I'm indecisive, if I should repair this within this PR or in a new one when this one was merged.
Feel free to give an advice.

src/map/script.c
```diff
@@ -6752,169 +6752,218 @@ static BUILDIN(warpchar)
...
-			case 2: // SavePoint
-				if (!map->list[pl_sd->bl.m].flag.noreturn)
-					pc->setpos(pl_sd, sd->status.save_point.map, sd->status.save_point.x, sd->status.save_point.y, CLR_TELEPORT);
-				break;
-			case 3: // m,x,y
-				if (!map->list[pl_sd->bl.m].flag.noreturn && !map->list[pl_sd->bl.m].flag.nowarp)
-					pc->setpos(pl_sd, script->mapindexname2id(st, str), x, y, CLR_TELEPORT);
-				break;
-			}
+		const char *m_name_from = script_hasdata(st, 6 + offset) ? script_getstr(st, 6 + offset) : NULL;
```
dastgirp: Just a thought, would it be nice to assign these variables outside the loop instead of reassigning it everytime in loop?

Kenpachi2k13: Nope, I discussed this with @hemagx and it should be inside the loop.

4144: yes, look like calculation what not depend on ``i`` can be moved outside of ``for``. if need for safety can be added ``const`` to vars.

why move outside? because normally this code will be executed again and again for each ``for`` iteration, and this is can be slow.

Asheraf:
Hmm im not sure if i like inserting the new argument in the middle and that offset changing even though it works, anyways you missed updating the building definition with an extra optional argument.

Kenpachi2k13:
I did not put the flag at he end of the argument lists, because this would force the user to pass a `<from map name>` (and for `warpparty()` also `<include leader>`), too, which makes no sense.

src/map/script.c
Kenpachi2k13: Unless you have a good reason to act different, variables should be declared in the smallest possible scope, which, in this case, is the `for` loop. This standard will be added to our style, when I'm ready to update it. (See topic in staff forum.)

<details><summary>Quote from my chat with @hemagx if you're interested:</summary>
 
    **Kenpachi**
    Assuming you need a variable only inside a for loop and it will have the same value for every iteration.
    Should it still be declared inside the loop?
 
    **Kenpachi**
    https://softwareengineering.stackexchange.com/a/296724
    This guy for example has good arguments to declare it outside the loop.
    https://stackoverflow.com/a/7959658
    But this guy has good arguments to declare it inside the loop.
    That's fucking confusing... :scream:
 
    **Ema**
    do you understand the Big O notation?
 
    **Kenpachi**
    I know the basics...
 
    **Ema**
    so think of it
    is it worth it or not
    99% it's not.
    and you should keep it inside loop no matter what the costs are
 
    **Kenpachi**
    I know the costs are low in my case, since the iteration count is very low.
    But it seems to be so badly unlogical/unneecessary to me, to set foo to n for x times. You know?
 
    **Ema**
    Safety comes before efficiency.
    show me the case tho?
 
    **Kenpachi**
    My "problem" is more like rating the benefit of keep the variable in the smallest possible scope.
 
    **Ema**
    the benefit?
    we have 10 years old bugs that no one was able to fix until we started fix scopes.
 
    **Kenpachi**
    Really? Well, that's a benefit of course. :rofl:
 
    **Ema**
    variables with no hard defined scopes are also prune to memory violations of all kinds << happens all the time.
 
    **Kenpachi**
    Okay, so like you said. "Safety comes before efficiency."
    That's an argument I can deal with. :slight_smile:
 
    **Ema**
    sometimes it can be the other way around
    you just need to have a really good argument.
    but as i said 99% it will never be good argument
   </details>

4144: this talk not about variables but about code what can be run only once.
inside for or while this code will be run N times. this is huge performance difference.

Example 1
```
int k;  // wrong
for (int i = 0; i < 10; i++)
{
  k = 10;
  // do something here with i and k
}
```
fixed version will be:

```
for (int i = 0; i < 10; i++)
{
  int k = 10;
  // do something here with k
}
```

Example 2

```
for (int i = 0; i < 10; i++)
{
  int k = call_some_slow_function();  // wrong
  // do something here with i and k
}
```
fixed version will be:


```
int k = call_some_slow_function();
for (int i = 0; i < 10; i++)
{
  // do something here with i and k
}
```

And in this example function ``call_some_slow_function`` not depend on ``i`` and this mean it can be moved outside of ``for``

hemagx: I actually do agree with 4144, it's slow function also const adds enough safety level.

Kenpachi2k13: Changed...

PR #2615, opened 2020-01-15
Add exploration achievements

Asheraf:
### Pull Request Prelude

### Changes Proposed

This PR adds the exploration achievements npcs and script commands needed for them.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -10601,6 +10601,14 @@ returns progress on success and false on failure
 
 ---------------------------------------
 
+*achievement_iscompleted(<ach_id>{, <account_id>})
+
+Checks whether the attached player have completed all the given achievement objectives.
+
+returns true if yes and false if no.
+
+---------------------------------------
+
 *itempreview(<index>)
 
 Update already opened preview window with item from
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -7039,6 +7039,18 @@ NPCs talking while hidden then revealing... you can wonder around =P).
 
 ---------------------------------------
 
+*cloakonnpc("<NPC object name>"{, <account_id>})
+*cloakoffnpc("<NPC object name>"{, <account_id>})
+
+These commands are used to apply the cloaking effect on npcs
+this is a visual effect only and it does NOT stop the player
+from interacting with the npc.
+
+If an account_id is specified then the effect will only display to the given id,
+otherwise it's displayed to the entire npc area.
+
+---------------------------------------
+
 *doevent("<NPC object name>::<event label>")
 
 This command will start a new execution thread in a specified NPC object
--- a/npc/re/other/achievement_treasures.txt
+++ b/npc/re/other/achievement_treasures.txt
@@ -1,0 +1,342 @@
+//================= Hercules Script =======================================
+//=       _   _                     _
+//=      | | | |                   | |
+//=      | |_| | ___ _ __ ___ _   _| | ___  ___
+//=      |  _  |/ _ \ '__/ __| | | | |/ _ \/ __|
+//=      | | | |  __/ | | (__| |_| | |  __/\__ \
+//=      \_| |_/\___|_|  \___|\__,_|_|\___||___/
+//================= License ===============================================
+//= This file is part of Hercules.
+//= http://herc.ws - http://github.com/HerculesWS/Hercules
+//=
+//= Copyright (C) 2020 Hercules Dev Team
+//=
+//= Hercules is free software: you can redistribute it and/or modify
+//= it under the terms of the GNU General Public License as published by
+//= the Free Software Foundation, either version 3 of the License, or
+//= (at your option) any later version.
+//=
+//= This program is distributed in the hope that it will be useful,
+//= but WITHOUT ANY WARRANTY; without even the implied warranty of
+//= MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//= GNU General Public License for more details.
+//=
+//= You should have received a copy of the GNU General Public License
+//= along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//=========================================================================
+// Exploration Achievements
+//=========================================================================
+
+-	script	achievement_tr	FAKE_NPC,{
+	sscanf(strnpcinfo(NPC_NAME_HIDDEN), "tr%d", .@id);
+	if (!achievement_iscompleted(.@id))
+		achievement_progress(.@id, 1, 1, true);
+	end;
+OnInit:
+	cloakonnpc(strnpcinfo(NPC_NAME));
+	end;
+}
+
+-	script	achievement_wp	FAKE_NPC,{
+OnTouch:
+	sscanf(strnpcinfo(NPC_NAME_HIDDEN), "wp%d", .@id);
+	if (!achievement_iscompleted(.@id))
+		cloakoffnpc(sprintf("#tr%d", .@id), playerattached());
+	end;
+}
+
+// Treasure Chests
+prt_fild01,146,126,4	duplicate(achievement_tr)	#tr120001	4_TREASURE_BOX
+prt_fild02,142,221,4	duplicate(achievement_tr)	#tr120002	4_TREASURE_BOX
+prt_fild03,172,139,4	duplicate(achievement_tr)	#tr120003	4_TREASURE_BOX
+prt_fild04,118,57,4	duplicate(achievement_tr)	#tr120004	4_TREASURE_BOX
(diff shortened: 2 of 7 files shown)


doc/script_commands.txt
```diff
@@ -10589,6 +10601,14 @@ returns progress on success and false on failure
 
 ---------------------------------------
 
+*achievement_iscompleted(<ach_id>)
```
Kenpachi2k13: I'd like to have the optional parameter `<account ID>`, like `achievement_progress()` has.

src/map/clif.c
```diff
@@ -4032,51 +4032,58 @@ static void clif_misceffect(struct block_list *bl, int type)
...
 static void clif_changeoption(struct block_list *bl)
 {
-	unsigned char buf[32];
-	struct status_change *sc;
-	struct map_session_data* sd;
-
 	nullpo_retv(bl);
 
-	if ( !(sc = status->get_sc(bl)) && bl->type != BL_NPC ) return; //How can an option change if there's no sc?
+	struct map_session_data *sd = BL_CAST(BL_PC, bl);
```
Kenpachi2k13: You only need `sd` if `if (sc == NULL && bl->type != BL_NPC)` is `false`, so the declaration should be moved below this check.

src/map/clif.c
```diff
@@ -4032,51 +4032,58 @@ static void clif_misceffect(struct block_list *bl, int type)
...
-	WBUFW(buf,6) = (sc) ? sc->opt1 : 0;
-	WBUFW(buf,8) = (sc) ? sc->opt2 : 0;
-	WBUFL(buf,10) = (sc != NULL) ? sc->option : ((bl->type == BL_NPC) ? BL_UCCAST(BL_NPC, bl)->option : 0);
-	WBUFB(buf,14) = (sd)? sd->status.karma : 0;
+	struct PACKET_ZC_STATE_CHANGE p;
+	p.packetType = HEADER_ZC_STATE_CHANGE;
+	p.AID = bl->id;
+	p.bodyState = (sc != NULL) ? sc->opt1 : 0;
+	p.healthState = (sc != NULL) ? sc->opt2 : 0;
+	p.effectState = (sc != NULL) ? sc->option : ((bl->type == BL_NPC) ? BL_UCCAST(BL_NPC, bl)->option : 0);
```
Kenpachi2k13: When reaching this part of code and `sc` is `NULL`, the only possible `bl->type` is `BL_NPC`, since you checked `if (sc == NULL && bl->type != BL_NPC)` a few lines above and `return;` if this is `true`.
So `(bl->type == BL_NPC) ?` is superfluous.

src/map/clif.c
```diff
@@ -4032,51 +4032,58 @@ static void clif_misceffect(struct block_list *bl, int type)
...
-	WBUFL(buf,10) = (sc != NULL) ? sc->option : ((bl->type == BL_NPC) ? BL_UCCAST(BL_NPC, bl)->option : 0);
-	WBUFB(buf,12) = (sd)? sd->status.karma : 0;
+}
+
+static void clif_changeoption_target(struct block_list *bl, struct block_list *target_bl, enum send_target target)
+{
+	nullpo_retv(bl);
+	nullpo_retv(target_bl);
+
+	struct map_session_data *sd = BL_CAST(BL_PC, bl);
```
Kenpachi2k13: Same as above. You only need `sd` if `if (sc == NULL && bl->type != BL_NPC)` is `false`, so the declaration should be moved below this check.

src/map/clif.c
```diff
@@ -4032,51 +4032,58 @@ static void clif_misceffect(struct block_list *bl, int type)
...
+
+	if (sc == NULL && bl->type != BL_NPC) // How can an option change if there's no sc?
+		return;
+
+	struct PACKET_ZC_STATE_CHANGE p;
+	p.packetType = HEADER_ZC_STATE_CHANGE;
+	p.AID = bl->id;
+	p.bodyState = (sc != NULL) ? sc->opt1 : 0;
+	p.healthState = (sc != NULL) ? sc->opt2 : 0;
+	p.effectState = (sc != NULL) ? sc->option : ((bl->type == BL_NPC) ? BL_UCCAST(BL_NPC, bl)->option : 0);
```
Kenpachi2k13: Same as above. When reaching this part of code and `sc` is `NULL`, the only possible `bl->type` is `BL_NPC`, since you checked `if (sc == NULL && bl->type != BL_NPC)` a few lines above and `return;` if this is `true`.
So `(bl->type == BL_NPC) ?` is superfluous.

src/map/script.c
```diff
@@ -12489,6 +12489,53 @@ static BUILDIN(hideonnpc)
 	npc->enable(str,4);
 	return true;
 }
+/*==========================================
+ *------------------------------------------*/
+static BUILDIN(cloakonnpc)
+{
+	struct npc_data *nd = npc->name2id(script_getstr(st, 2));
+	if (nd == NULL)
+		return false;
```
Kenpachi2k13: I would show an error message if the NPC wasn't found, but that's up to you.

Asheraf: returning false already prints an error, no need for another one.

Asheraf:
> `clif_changeoption_target()` isn't added to plugin interface yet. Is this intended?

it's added to the clif interface just fine.. maybe your browser didn't load that part.

src/map/script.c
Kenpachi2k13: Well, return `false` just triggers `script_reportsrc()`.
And, in my opinion, a debug message like this isn't really helpful in that case:

> [Debug]: Source (NPC): Guard#DEV1 at prontera (147,169)

Kenpachi2k13:
> > `clif_changeoption_target()` isn't added to plugin interface yet. Is this intended?
> 
> it's added to the clif interface just fine.. maybe your browser didn't load that part.

That was kind of misleading, I guess... I was talking about the _HPMHooking_.

Asheraf:
> > > `clif_changeoption_target()` isn't added to plugin interface yet. Is this intended?
> > 
> > 
> > it's added to the clif interface just fine.. maybe your browser didn't load that part.
> 
> That was kind of misleading, I guess... I was talking about the _HPMHooking_.

HPM Hooks get generated by HerculesWSAPI once commits are merged to master, we're only required to add it to pull requests when changes break CI build.

doc/script_commands.txt
```diff
@@ -7039,6 +7039,18 @@ NPCs talking while hidden then revealing... you can wonder around =P).
 
 ---------------------------------------
 
+*cloakonnpc("<NPC object name>"{, <AID>})
+*cloakoffnpc("<NPC object name>"{, <AID>})
```
Emistry:
```suggestion
*cloakoffnpc("<NPC object name>"{, <account_id>})
```
instead of `AID` , perhaps `account_id` a better choice, just like other? easier to newbie to understand what it is?

npc/re/other/achievement_treasures.txt
```diff
@@ -0,0 +1,342 @@
...
+//= GNU General Public License for more details.
+//=
+//= You should have received a copy of the GNU General Public License
+//= along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//=========================================================================
+// Exploration Achievements
+//=========================================================================
+
+-	script	achievement_tr	FAKE_NPC,{
+	.@id = atoi(substr(strnpcinfo(NPC_NAME_HIDDEN), 2, getstrlen(strnpcinfo(NPC_NAME_HIDDEN)) - 1));
```
Emistry:
```suggestion
	sscanf("tr%d", .@id);
```
maybe an alternative ? imo this way seem easier to read.

npc/re/other/achievement_treasures.txt
```diff
@@ -0,0 +1,342 @@
...
+	if (!achievement_iscompleted(.@id))
+		achievement_progress(.@id, 1, 1, true);
+	end;
+OnInit:
+	cloakonnpc(strnpcinfo(NPC_NAME));
+	end;
+}
+
+-	script	achievement_wp	FAKE_NPC,{
```
Emistry: why need 2 npc? cant merge the `OnTouch` into the `achievement_tr` ?  2 npc to maintain for each achievement npc if players ever decide to add a new one? and extra npcs mean less npc slot to use for other custom npc per map (although the limit around 512 per maps)

default npc with hidden sprite , then `clockoffnpc + changeclass` would do the trick.
i believe it would still trigger the event if player get near...

Asheraf: That's because this is a conversion from the official script and I want to keep the same set of npcs to be able to reference it with that, in addition with this format the trigger area can be different from the chest area.

MishimaHaruna:
Merged from the command line due to conflicts

Issue #819, opened 2015-10-25
error in bonus items


xBeret:
In hercules some bonus don't work like the aegis, seem a solution to work like.

Hercules is using a script entirely wrong, because when you use the item you are using an effect and not the skill as it is in the aegis,

for example the item:

```
    Inc_Agi_10_Scroll
```

script hercules:

```
Script: <"
    if(Hp>15) {
        skilleffect AL_INCAGI,0;
        sc_start SC_INC_AGI,240000,10;
        heal -15,0;
    }
">
```

script aegis:

```
item Inc_Agi_10_Scroll
    event OnConsume:
        SkillToMe AL_INCAGI 10
    return
```

Hercules is working totally wrong, needing extra code to function as the aegis.

I suggest create SkillToMe command, this command applies the skill the character who used the item, noting that if the skill is target it applies without needing to click on target to receive the skill.

dastgir:
I agree, we need those commands.

Ridley8819:
Bumping this :)

iRagno:
What about `unitskilluseid(getcharid(CHAR_ID_ACCOUNT), AL_INCAGI, 10);`?

Makes the trick.

Murilo-BiO:
@iRagno When using this item and `Guarana Candy` it shouldn't have cast time, it must be instantaneous.

PR #2616, opened 2020-01-24
Modified itemskill() script command to be usable like AEGIS' SkillToMe command.

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

 * Fixed the `flag` parameter of `itemskill()` script command which was broken for more than 6 years.
 * Added new option flag to be able to cast the skill without cast time.
 * Added new option flag to be able to forcefully cast the skill on the invoking character.
 * Updated the documentation of `itemskill()` script command.
 * Updated item databases.
 * Implemented the consumption of 10 SP when using Earth Spike Scroll while SC_EARTHSCROLL is active. (TK_SPTIME skill.)

**Issues addressed:** #819


db/pre-re/item_db.conf
```diff
@@ -1159,7 +1159,7 @@ item_db: (
 	Id: 601
 	AegisName: "Wing_Of_Fly"
 	Name: "Fly Wing"
-	Type: "IT_DELAYCONSUME"
```
dastgirp: Any reason to change this?

Kenpachi2k13: Skill casting items should not be of type `IT_DELAYCONSUME`.
In case of Fly Wing it wouldn't matter, but for the sake of consistency I changed it for all skill casting items.

src/map/skill.c
```diff
@@ -13976,6 +13976,10 @@ static int skill_check_condition_castbegin(struct map_session_data *sd, uint16 s
 
 	nullpo_ret(sd);
 
+	if ((sd->state.itemskill_conditions_checked == 1 || sd->state.itemskill_no_conditions == 1)
```
4144: this will return allow flag even if player in chat and not checking any other conditions, this is correct?

also because ``if`` is more than one line, better add ``{`` and ``}``

src/map/pc.c
```diff
@@ -5302,6 +5309,11 @@ static int pc_useitem(struct map_session_data *sd, int n)
 	script->run_use_script(sd, sd->inventory_data[n], npc->fake_nd->bl.id);
 	script->potion_flag = 0;
 
+	/// If Earth Spike Scroll is used while SC_EARTHSCROLL is active, there is a chance to don't consume the scroll. [Kenpachi]
+	if ((nameid == ITEMID_EARTH_SCROLL_1_3 || nameid == ITEMID_EARTH_SCROLL_1_5) && sd->sc.count > 0
```
4144: here because ``if`` is multi line better add ``{`` and ``}`` too

src/map/skill.c
Kenpachi2k13: > this will return allow flag even if player in chat and not checking any other conditions, this is correct?

Since you can't use items, while a chat window is open, this check can be done prior to `sd->chat_id != 0`.

4144: you mean you cant do this is in default client.
but some one can use bot or modified or alternative clients. because this never trust to client.

Kenpachi2k13: This is checked server side. `clif_parse_UseItem()` will return without calling `pc->useitem()` if a chat, trading, vending or NPC windows is open or the character is dead:
```
	if (pc_isvending(sd))
		return;

	if (pc_isdead(sd)) {
		clif->clearunit_area(&sd->bl, CLR_DEAD);
		return;
	}

	if ((!sd->npc_id && pc_istrading(sd)) || sd->chat_id != 0)
		return;
```

4144: skill_check_condition_castbegin called not only from use item packet but from different unit and other functions.

Kenpachi2k13: Sure, but if `skill_check_condition_castbegin()` is **not** called by a skill casting item, `sd->itemskill_id == sd->skillitem` and `sd->itemskill_lv == sd->skillitemlv` will both be `false` and thus `return 1;` will not be reached...

src/map/pc.c
```diff
@@ -5257,6 +5257,14 @@ static int pc_useitem(struct map_session_data *sd, int n)
 #endif
 			if( battle_config.item_restricted_consumption_type && sd->status.inventory[n].expire_time == 0 ) {
 				clif->useitemack(sd,n,sd->status.inventory[n].amount-1,true);
+
+				/// If Earth Spike Scroll is used while SC_EARTHSCROLL is active, there is a chance to don't consume the scroll. [Kenpachi]
+				if ((nameid == ITEMID_EARTH_SCROLL_1_3 || nameid == ITEMID_EARTH_SCROLL_1_5)
```
Asheraf: I am not a fan of this, we already have so many hardcoded statement like this that we're trying to get rid of, adding more is not..fun, this should be converted to some item db flag.
also the message of the related commit is misleading 
>Implemented the consumption of 10 SP when using Earth Spike Scroll while SC_EARTHSCROLL is active.

while the change is for NOT consuming sp if i understand correctly.

src/map/script.c
```diff
@@ -10976,33 +10976,63 @@ static BUILDIN(guildopenstorage)
...
-	clif->item_skill(sd,id,lv);
+
+	sd->state.itemskill_no_casttime = ((flag & ISF_INSTANTCAST) == ISF_INSTANTCAST) ? 1 : 0; /// Unset in unit_skilluse_id() if ISF_CASTONSELF is not set and skill was not aborted while target selection.
+
+	/// itemskill_conditions_checked/itemskill_no_conditions/itemskill_no_casttime abuse prevention.
+	/// Unset in unit_skilluse_id() if ISF_CASTONSELF is not set and skill was not aborted while target selection.
+	sd->itemskill_id = sd->skillitem;
+	sd->itemskill_lv = sd->skillitemlv;
+
+	if ((flag & ISF_CASTONSELF) == ISF_CASTONSELF) {
```
Asheraf: In aegis this is done by changing the skill type in 0x147.. but this works too, you can probably ignore this comment.

src/map/unit.c
```diff
@@ -1041,11 +1041,20 @@ static int unit_stop_walking(struct block_list *bl, int flag)
...
-		skill->cast_fix(src, skill_id, skill_lv),
-		skill->get_castcancel(skill_id)
-	);
+	int casttime = skill->cast_fix(src, skill_id, skill_lv);
+	int castcancel = skill->get_castcancel(skill_id);
+	int ret = unit->skilluse_id2(src, target_id, skill_id, skill_lv, casttime, castcancel);
+	struct map_session_data *sd = BL_CAST(BL_PC, src);
+
+	if (sd != NULL) {
+		sd->itemskill_id = 0;
```
Asheraf: you have these called multiple times, a `pc_itemskill_clear` might help.

src/map/pc.c
Kenpachi2k13: > I am not a fan of this, we already have so many hardcoded statement like this that we're trying to get rid of, adding more is not..fun, this should be converted to some item db flag.

I agree with you. But in my opinion this should be done in a separate PR for all/most of these hardcoded statements, because you can create a general solution then, instead of fixing just this one spot.

> also the message of the related commit is misleading
> 
> > Implemented the consumption of 10 SP when using Earth Spike Scroll while SC_EARTHSCROLL is active.
> 
> while the change is for NOT consuming sp if i understand correctly.

You probably mixed up some commits here. This part of the code is for not consuming the Earth Spike Scroll by chance and is not related to SP consumption.
 * Commit for not consuming the Earth Spike Scroll by chance while SC_EARTHSCROLL is active: https://github.com/HerculesWS/Hercules/pull/2616/commits/ff082a27bc3a6e6e86edbd8e66ff563244f6ba33
 * Commit for consuming 10 SP when using Earth Spike Scroll while SC_EARTHSCROLL is active: https://github.com/HerculesWS/Hercules/pull/2616/commits/803b1fae79d3a387791d9cb18c55cd580940c733

src/map/script.c
Kenpachi2k13: Thanks for the hint. I'll look into it and maybe do it like AEGIS, depending on the effort/benefit.

src/map/unit.c
Kenpachi2k13: Good point. I'll outsource this to a separate function.

src/map/script.c
Kenpachi2k13: Changed this to work like official.

src/map/unit.c
Kenpachi2k13: Done.

src/map/unit.c
```diff
@@ -1041,11 +1041,15 @@ static int unit_stop_walking(struct block_list *bl, int flag)
...
-		skill->cast_fix(src, skill_id, skill_lv),
-		skill->get_castcancel(skill_id)
-	);
+	int casttime = skill->cast_fix(src, skill_id, skill_lv);
+	int castcancel = skill->get_castcancel(skill_id);
+	int ret = unit->skilluse_id2(src, target_id, skill_id, skill_lv, casttime, castcancel);
+	struct map_session_data *sd = BL_CAST(BL_PC, src);
+
+	if (sd != NULL)
+		pc->itemskill_clear(sd);
```
4144: for safety better call ``pc->itemskill_clear`` from all skill related packets from client for avoid any bypass exploits
and may be from all functions ``unit_skilluse_*``

src/map/skill.c
4144: still better for avoid exploits add as many as possible checks before bypass all other checks.

src/map/unit.c
Kenpachi2k13: If doing so, `sd->itemskill_id` and `sd->itemskill_lv` would be unset even if the skill casting was aborted before selecting a target and thus makes the whole thing abusable. So this is a bad idea.

Kenpachi2k13: To avoid abusing the `ISF_` flags you have the `sd->itemskill_id` and `sd->itemskill_lv` variables and the `skill_is_item_skill()` function. So it's not necessary to call `pc_itemskill_clear` somewhere else.

src/map/skill.c
Kenpachi2k13: Done.

src/map/unit.c
4144: some bypass for checks still probably possible. If use ``itemskill``, and then manually use same skill.
It not mean it really possible but can be possible because code too complex and missing cleanup code.
Better in all places where new skill invoke happens clean all variables what may set by previous skill invokes.

Kenpachi2k13: Done.

src/map/clif.c
```diff
@@ -12922,6 +12922,8 @@ static void clif_parse_UseSkillMap(int fd, struct map_session_data *sd)
 
 	pc->delinvincibletimer(sd);
 	skill->castend_map(sd,skill_id,map_name);
+
+	pc->itemskill_clear(sd);
```
4144: probably clear need call before invoking skill?

Kenpachi2k13: Nope. `skill_castend_map()` calls `skill_check_condition_castend()` and `skill_consume_requirement()`, which both use itemskill flags. If you'd clear those flags before calling `skill_castend_map()`, those flags won't work.

src/map/clif.c
```diff
@@ -6747,21 +6747,24 @@ static void clif_item_refine_list(struct map_session_data *sd)
...
-	safestrncpy(WFIFOP(fd,14),skill->get_name(skill_id),NAME_LENGTH);
-	WFIFOB(fd,38)=0;
-	WFIFOSET(fd,packet_len(0x147));
+	int fd = sd->fd;
+	int type = skill->get_inf(skill_id);
+
+	if (sd->state.itemskill_castonself == 1 && skill->is_item_skill(sd, skill_id, skill_lv))
+		type = INF_SELF_SKILL;
+
+	WFIFOHEAD(fd, packet_len(0x147));
```
4144: if you know how, better convert this packet into struct.

Kenpachi2k13: I never worked with the packet related code, but I'll try to get into it at the weekend.
If I get stuck, I'll talk to you at Discord. 😄

Kenpachi2k13: Commited. Thanks for your help. :+1:

db/pre-re/item_db.conf
MishimaHaruna: By saying "should not be", do you mean that they are officially consumed immediately, or that one must never set them as delay consumed, even if they're creating a custom item?

Kenpachi2k13: In AEGIS they are consumed immediately after using them.
And see my note in doc/script_commands.txt:

> Important: Items which use itemskill() should be of type IT_USABLE.
>            If the item type is IT_DELAYCONSUME and ISF_IGNORECONDITIONS is set,
>            the item won't be consumed when using the item!

Using `IT_DELAYCONSUME` brings some risk.

Issue #2558, opened 2019-10-11
incorrect interaction between status reco and evil druid armor


Xelliekins:
**Describe the bug**
Using status recovery on an evil druid armor user in pvp/woe succeeds and also applies the blind status 

**To Reproduce**
1. Go to pvp/woe
2.  Apply sleep status to player
3. Use status recovery on sleeping ally with evil druid equipped

**Expected behavior**
Status recovery should simply fail. No recovery of status or application of blind should happen

**Screenshots**
Here is a video demonstrating the official behaviour on iRO Classic (aegis) 
https://youtu.be/LCQZUIF_naU

**System specs (please complete the following information):**
 - OS: Debian GNU/Linux 8.11 (jessie) [x86_64]
 - Hercules Version [v2019.09.22]
 - Mode: pre-renewal
 - Packet version: [20130807]
 - Client type: [RE]

Kenpachi2k13:
Partially confirmed.
The _Sleep_ status won't be cured when _Evil Druid Card_ is equipped, but the application of _Blind_ status is possible.
https://www.youtube.com/watch?v=7xBtw29_kho

PR #2618, opened 2020-01-25
Updated PR_STRECOVERY to only cure status effects if defence element is not Undead.

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

`PR_STRECOVERY` should cure status effects only if the target's defence/armor element is not `ELE_UNDEAD`.

**Issues addressed:** #2558

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -7306,8 +7306,8 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 			break;
 
 		case PR_STRECOVERY:
-			if(status->isimmune(bl)) {
-				clif->skill_nodamage(src,bl,skill_id,skill_lv,0);
+			if (status->isimmune(bl) != 0) {
+				clif->skill_nodamage(src, bl, skill_id, skill_lv, 0);
 				break;
 			}
 
@@ -7321,16 +7321,19 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 				}
 
 				status_change_end(bl, SC_NETHERWORLD, INVALID_TIMER);
+			} else {
+				int rate = 100 * (100 - (tstatus->int_ / 2 + tstatus->vit / 3 + tstatus->luk / 10));
+				int duration = skill->get_time2(skill_id, skill_lv);
+
+				duration *= (100 - (tstatus->int_ + tstatus->vit) / 2) / 100;
+				status->change_start(src, bl, SC_BLIND, rate, 1, 0, 0, 0, duration, SCFLAG_NONE);
 			}
-			//Is this equation really right? It looks so... special.
-			if( battle->check_undead(tstatus->race,tstatus->def_ele) ) {
-				status->change_start(src, bl, SC_BLIND,
-				                     100*(100-(tstatus->int_/2+tstatus->vit/3+tstatus->luk/10)), 1,0,0,0,
-				                     skill->get_time2(skill_id, skill_lv) * (100-(tstatus->int_+tstatus->vit)/2)/100,SCFLAG_NONE);
-			}
-			clif->skill_nodamage(src,bl,skill_id,skill_lv,1);
-			if(dstmd)
-				mob->unlocktarget(dstmd,tick);
+
+			clif->skill_nodamage(src, bl, skill_id, skill_lv, 1);
+
+			if (dstmd != NULL)
+				mob->unlocktarget(dstmd, tick);
+
 			break;
 
 		// Mercenary Supportive Skills
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -7310,14 +7310,18 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 				clif->skill_nodamage(src,bl,skill_id,skill_lv,0);
 				break;
 			}
-			if (tsc && tsc->opt1) {
-				status_change_end(bl, SC_FREEZE, INVALID_TIMER);
-				status_change_end(bl, SC_STONE, INVALID_TIMER);
-				status_change_end(bl, SC_SLEEP, INVALID_TIMER);
-				status_change_end(bl, SC_STUN, INVALID_TIMER);
-				status_change_end(bl, SC_WHITEIMPRISON, INVALID_TIMER);
+
+			if (!battle->check_undead(tstatus->race, tstatus->def_ele)) {
+				if (tsc != NULL && tsc->opt1 != 0) {
+					status_change_end(bl, SC_FREEZE, INVALID_TIMER);
+					status_change_end(bl, SC_STONE, INVALID_TIMER);
+					status_change_end(bl, SC_SLEEP, INVALID_TIMER);
+					status_change_end(bl, SC_STUN, INVALID_TIMER);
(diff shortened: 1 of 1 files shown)


src/map/skill.c
```diff
@@ -7310,14 +7310,18 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
...
 				break;
 			}
-			if (tsc && tsc->opt1) {
-				status_change_end(bl, SC_FREEZE, INVALID_TIMER);
-				status_change_end(bl, SC_STONE, INVALID_TIMER);
-				status_change_end(bl, SC_SLEEP, INVALID_TIMER);
-				status_change_end(bl, SC_STUN, INVALID_TIMER);
-				status_change_end(bl, SC_WHITEIMPRISON, INVALID_TIMER);
+
+			if (bl->type == BL_PC && tstatus->def_ele != ELE_UNDEAD) {
```
hemagx: you just broke the whole thing.
1. you should only check undead using battle->check_undead otherwise it ignores server configuration
2. by adding the bl->type == BL_PC you broke the skill behavior, it should be able to status recovery non-pc but not on undead property objects

src/map/skill.c
```diff
@@ -7310,14 +7310,18 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
...
+					status_change_end(bl, SC_SLEEP, INVALID_TIMER);
+					status_change_end(bl, SC_STUN, INVALID_TIMER);
+					status_change_end(bl, SC_WHITEIMPRISON, INVALID_TIMER);
+				}
+
+				status_change_end(bl, SC_NETHERWORLD, INVALID_TIMER);
 			}
-			status_change_end(bl, SC_NETHERWORLD, INVALID_TIMER);
 			//Is this equation really right? It looks so... special.
 			if( battle->check_undead(tstatus->race,tstatus->def_ele) ) {
```
hemagx: This part is of the original report, the skill should indeed not apply blind on PC objects, this was not fixed.

src/map/skill.c
```diff
@@ -7306,27 +7306,37 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
...
 				break;
 			}
-			if (tsc && tsc->opt1) {
-				status_change_end(bl, SC_FREEZE, INVALID_TIMER);
-				status_change_end(bl, SC_STONE, INVALID_TIMER);
-				status_change_end(bl, SC_SLEEP, INVALID_TIMER);
-				status_change_end(bl, SC_STUN, INVALID_TIMER);
-				status_change_end(bl, SC_WHITEIMPRISON, INVALID_TIMER);
+
+			if (bl->type == BL_PC && tstatus->def_ele != ELE_UNDEAD) {
```
hemagx: you just broke the whole thing.
1. you should only check undead using battle->check_undead otherwise it ignores server configuration
2. by adding the bl->type == BL_PC you broke the skill behavior, it should be able to status recovery non-pc but not on undead property objects
3. the skill shall fail on undead property players, you should instead add a check for PC object their element and do `clif->skill_nodamage(src, bl, skill_id, skill_lv, 0);` similar to the isimmune check above. (or simply merge it there), while on non-PC objects that are undead it shall only apply the blind formula down there.

src/map/skill.c
```diff
@@ -7306,27 +7306,37 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
...
+				if (tsc != NULL && tsc->opt1 > 0) {
```
hemagx: ```tsc->opt1 != 0```
please keep translate the other behavior as it was written before, also it's a bitmask variable which makes `> 0` incorrect.

src/map/skill.c
```diff
@@ -7306,27 +7306,37 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
...
 			}
-			status_change_end(bl, SC_NETHERWORLD, INVALID_TIMER);
-			//Is this equation really right? It looks so... special.
-			if( battle->check_undead(tstatus->race,tstatus->def_ele) ) {
-				status->change_start(src, bl, SC_BLIND,
-				                     100*(100-(tstatus->int_/2+tstatus->vit/3+tstatus->luk/10)), 1,0,0,0,
-				                     skill->get_time2(skill_id, skill_lv) * (100-(tstatus->int_+tstatus->vit)/2)/100,SCFLAG_NONE);
+
+			/// Is this equation really right? It looks so... special.
+			if (battle->check_undead(tstatus->race, tstatus->def_ele)) {
```
hemagx: the report clearly states that blind status should not be applied on PC (which is the correct/expected behavior)

src/map/skill.c
Kenpachi2k13: > 1. you should only check undead using battle->check_undead otherwise it ignores server configuration

Yeah, my fault....

> 2. by adding the bl->type == BL_PC you broke the skill behavior, it should be able to status recovery non-pc but not on undead property objects

Does it cure status effects on Mercenaries/Homunculi? I just tested monsters and they are not cured, so I assumed it's only for players. In any case, monsters should be excluded.

src/map/skill.c
Kenpachi2k13: Why resolved? :open_mouth:

src/map/skill.c
Kenpachi2k13: Since it's an unsigned integer, `> 0` doesn't make more or less sense than `!= 0` to me, even  if it's a bitmask variable. Correct me if I'm wrong.

> please keep translate the old behavior as it was written before

I should check implicitly even though it's not a boolean? This is how it was written before I changed it: `if (tsc && tsc->opt1) {`
Maybe I just misunderstood you.

src/map/skill.c
hemagx: The bug is reporting in pre-re and you're testing on renewal.

src/map/skill.c
hemagx: `>` `<` `+` `-` operations are invalid operations on bitmask fields, therefore your `> 0` give wrong impression.
and by translating old behavior it means
``` if (tsc->opt1) ```
is true whenever opt1 is not zero, therefore when you fix the coding style you should preserve the old behavior `!= 0` unless you're explicitly fixing a behavior. which is not the case here anyway.

Kenpachi2k13: Ah, okay. Thanks for the explanation. :+1:

src/map/skill.c
Kenpachi2k13: I'm curently unable to test, neither on pre-RE, nor on RE. Would you be so nice and give me some details, since you seem to know the correct mechanism for both?

Kenpachi2k13: Removed the BL_PC restriction and replaced `tstatus->def_ele != ELE_UNDEAD` with a call of `battle_check_undead()`.

src/map/skill.c
Kenpachi2k13: Replaced `tsc->opt1 > 0` with `tsc->opt1 != 0`.

src/map/skill.c
hemagx: Actually you can ignore this, the report was invalid after a second check.

PR #2619, opened 2020-01-25
Crash fixes

dastgirp:
### Pull Request Prelude

### Changes Proposed

There were 2 cases causing map-server to crash.
1) When getunits was called with invalid mapname, mapname2mapid would be called, and it will return -1, which would then call map_forcountinmap, which does not perform any checks. So, a error message is added to getunits, to know exactly where something went wrong.
2) Source can use map_forcountinmap without validating the inputs, or even passing wrong value of mapid, which would have caused the map-server crash.

**Issues addressed:** 

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -11775,6 +11775,13 @@ static BUILDIN(getunits)
 		const char *mapname = script_getstr(st, 5);
 		int16 m = map->mapname2mapid(mapname);
 
+		if (m == -1) {
+			ShowError("script:getunits: Invalid map(%s) provided.\n", mapname);
+			script->reportdata(data);
+			st->state = END;
+			return false;
+		}
+
 		if (script_hasdata(st, 9)) {
 			int16 x1 = script_getnum(st, 6);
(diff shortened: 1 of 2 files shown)


src/map/script.c
```diff
@@ -11775,6 +11775,13 @@ static BUILDIN(getunits)
 		const char *mapname = script_getstr(st, 5);
 		int16 m = map->mapname2mapid(mapname);
 
+		if (m == -1) {
```
Kenpachi2k13: I'd prefer `INDEX_NOT_FOUND` instead of `-1`, but that's up to you.

dastgirp: You are correct. will change map_mapname2mapid and map_mapindex2mapid functions too, so that it will be in sync

dastgirp: @Kenpachi2k13 I will do it in new PR, as the purpose of that would be different, and it would be better to change at all places where mapname2mapid is called.

Kenpachi2k13: :+1:

PR #2622, opened 2020-02-01
Make mob DB field ViewData->HairStyleId defaulting to 1 instead of 0.

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

When using the `ViewData` block in mob_db.conf, `HairStyleId` currently defaults to 0.
This causes client errors, because there is no head sprite with ID 0.
So `HairStyleId` should default to 1 instead.

**Issues addressed:** None.

### DIFF
--- a/doc/mob_db.txt
+++ b/doc/mob_db.txt
@@ -82,7 +82,7 @@ mob_db: (
     HeadTopId: top headgear id       (int, defaults to 0)
     HeadMidId: middle headgear id    (int, defaults to 0)
     HeadLowId: lower headgear id     (int, defaults to 0)
-    HairStyleId: hair style id       (int, defaults to 0)
+    HairStyleId: hair style id       (int, defaults to 1)
     BodyStyleId: clothes id          (int, defaults to 0)
     HairColorId: hair color id       (int, defaults to 0)
     BodyColorId: clothes color id    (int, defaults to 0)
@@ -193,7 +193,7 @@ Element: Monster's element. Sets element type and level.
 Mode: Monster AI behaviour. If this block is omitted, monster doesn't react to anything.
       All the settings in this group are boolean values,
       Default value is false (mode not set) for any missing setting.
-      See /doc/sample/mob_db_mode_list.txt for more information about monsters Mode types.
+      See /doc/mob_db_mode_list.md for more information about monsters Mode types.
 
 MoveSpeed: Monster's speed. Sets speed (cells/sec).
            MoveSpeed is calculated to Hercules with this formula: 1000 / SPEED (CELLS/SEC)
@@ -278,3 +278,4 @@ DamageTakenRate:
 ViewData:
   Overrides the default view data sent to the client with the given values for:
     Sprite, Weapon, Shield, Robe, HeadTop, HeadMid, HeadLow, HairStyle, BodyStyle, HairColor, BodyColor, Gender, Options
+  Note: HairStyleId will only default to 1, if the ViewData block is defined.
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -4206,8 +4206,12 @@ static void mob_read_db_viewdata_sub(struct mob_db *entry, struct config_setting
 		entry->vd.head_mid = libconfig->setting_get_int(it);
 	if ((it = libconfig->setting_get_member(t, "HeadLowId")) != NULL)
 		entry->vd.head_bottom = libconfig->setting_get_int(it);
+
 	if ((it = libconfig->setting_get_member(t, "HairStyleId")) != NULL)
 		entry->vd.hair_style = libconfig->setting_get_int(it);
+	else
+		entry->vd.hair_style = 1;
+
 	if ((it = libconfig->setting_get_member(t, "BodyStyleId")) != NULL)
 		entry->vd.body_style = libconfig->setting_get_int(it);
 	if ((it = libconfig->setting_get_member(t, "HairColorId")) != NULL)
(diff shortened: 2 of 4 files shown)


src/map/mob.c
```diff
@@ -4208,6 +4208,8 @@ static void mob_read_db_viewdata_sub(struct mob_db *entry, struct config_setting
 		entry->vd.head_bottom = libconfig->setting_get_int(it);
 	if ((it = libconfig->setting_get_member(t, "HairStyleId")) != NULL)
 		entry->vd.hair_style = libconfig->setting_get_int(it);
+	if (entry->vd.hair_style == 0)
```
dastgirp: So What if I explicitly want hair style to be 0? It will still change to 1

doc/mob_db.txt
```diff
@@ -82,7 +82,7 @@ mob_db: (
     HeadTopId: top headgear id       (int, defaults to 0)
     HeadMidId: middle headgear id    (int, defaults to 0)
     HeadLowId: lower headgear id     (int, defaults to 0)
-    HairStyleId: hair style id       (int, defaults to 0)
+    HairStyleId: hair style id       (int, defaults to 1)
```
dastgirp: Maybe add documentation somewhere, that if viewData is defined, then hairstyle will be default to 1

Kenpachi2k13: In my opinion this isn't necessary. It should be obvious that the fields inside a block will only be assigned if that block is defined. If I'd start documenting this explicitly we'd have to do it for more fields, too.
For example:
 * item_db->Trade->override
 * item_db->Nouse->override
 * job_db->BaseASPD->all fields except *Shield*
 * mob_db->ViewData->SpriteId
 * mob_db->ViewData->Gender
 * refine_db->Rates->Level->all fields except *Bonus*

And why only if it's a field inside a block? Actually we'd have to document that behaviour for every field which doesn't default to 0, then. I don't think that's useful...

src/map/mob.c
Kenpachi2k13: Don't know if the client supports hair style 0. But I'll change it to only default to 1 if `HairStyleId` was not defined.

dastgirp: other games too use hercules, so it's better if we allow value

doc/mob_db.txt
dastgirp: actually not,
Those override of item_db defaults to 100, even if block is not defined. (for item, itemdb_validate_entry).
For mob_db->ViewData->SpriteID, it is defaulted to mob_read_db_sub just after "Id" is read.
mob_db->ViewData->Gender defaults to SEX_FEMALE (Value: 0)

For RefineDB, even before reading all the fields, default value is set.
```
for (int i = 0; i < REFINE_CHANCE_TYPE_MAX; i++)
			for (int j = 0; j < MAX_REFINE; j++)
				chance[i][j] = 100; // default value for all rates.
```

Kenpachi2k13: I don't want to argue, but do you really think that a user wants to know that `HairStyleId` won't default to 1, if he doesn't even use the `ViewData` block?
Well, to don't push this topic too far, I add a note in _doc/mob_db.txt_.

Kenpachi2k13:
> can you please squash the commits together.

Done. :thumbsup:

doc/mob_db.txt
```diff
@@ -278,3 +278,4 @@ DamageTakenRate:
 ViewData:
   Overrides the default view data sent to the client with the given values for:
     Sprite, Weapon, Shield, Robe, HeadTop, HeadMid, HeadLow, HairStyle, BodyStyle, HairColor, BodyColor, Gender, Options
+  Note: HairStyleId will only default to 1, if the ViewData block is defined.
```
Emistry:
```suggestion
  Note: HairStyleId will only default to 1, if the ViewData block is not defined.
```
perhaps you meant this?
or alternatives
> When not specified, becomes 1.

just like how others field's default values are mentioned in the doc.

Kenpachi2k13: The sentence is correct. Please read the [request](https://github.com/HerculesWS/Hercules/pull/2622#discussion_r373753837) made by @dastgirp.

dastgirp: @Emistry Currently, it's coded like, it will be 0 if ViewData block is not defined, but if defined, default value is 1

Emistry: hmm ... perhaps i understand it in a wrong way then.
these are what I thought it meant if:

**if ViewData block is defined, a random values**
```
{
   HairStyleId: 123
}
```
then the result will be override as `1` regardless of what value i defined there?
because the doc said it default to 1 if defined. Does it meant the value can always be `1` only, and not using the value provided in the doc? 
> Note: HairStyleId will only default to 1, if the ViewData block is defined.

Doc sound confusing to me, perhaps just my english aren't good to read it.
currently its sound like ... if I defined a values there, then the value will always be `1`, regardless what I put there.


**if ViewData block is defined, but 0 value**
```
{
   HairStyleId: 0
}
```
then result will override `0`. 

**if ViewData block is not defined**
```
{
   not defined
}
```
then result will override as `1`  (default).


but aren't document should be mentioning this scenario?
** if ViewData block is not defined, value default to 1**
```
{
   not defined
}
```
then result will override as `1`  (default).

Emistry: or does libconfig actually allow us to define config in such a way?
```
{
   HairStyleId: 
}
```
without a value? (only define a field without value)
I thought it always need a value, just like normal JSON kind of format.

dastgirp:
```c
{
 ViewData: {
  // Other fields except Hairstyle
 }
}
```
 ^ In this case, Hairstyle 1

```c
{
 // no view Data block
}
```
 ^ In this case, Hairstyle 0


```c
{
 ViewData: {
  // Other fields
  HairStyle: 123
 }
}
```
 ^ In this case, Hairstyle 123

Emistry: oic, thx for the info

Issue #1311, opened 2016-06-12
Duplicate 'type' in pc_readparam


Emistry:
```
#ifndef RENEWAL_CAST
        case SP_VARCASTRATE:
        ....
        ....
#endif
```

and

```
#ifdef RENEWAL_CAST
        case SP_VARCASTRATE:     val = sd->bonus.varcastrate; break;
        ....
        ....
#endif
```

hemagx:
it's not duplicate, tho i think it's bad code styling (and i believe i did it myself)
as you can notice one is under if renewal enabled and the other if not :)

Emistry:
o-o .. lol .. i actually didnt noticed the `n` within the `ifndef` and `ifdef`

PR #2624, opened 2020-02-06
Merge pre-RE/RE case SP_VARCASTRATE in pc_readparam() function

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

In `pc_readparam()` function, `case SP_VARCASTRATE` is defined twice.
On is defined in `#ifndef RENEWAL_CAST` and the other one in `#ifdef RENEWAL_CAST`.
I merged both and made the assignment of `val` depending on `RENEWAL_CAST`.

**Issues addressed:** #1311

### DIFF
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -8417,8 +8417,12 @@ static int64 pc_readparam(const struct map_session_data *sd, int type)
 		case SP_FLEE1:           val = sd->battle_status.flee; break;
 		case SP_FLEE2:           val = sd->battle_status.flee2; break;
 		case SP_DEFELE:          val = sd->battle_status.def_ele; break;
-#ifndef RENEWAL_CAST
 		case SP_VARCASTRATE:
+#ifdef RENEWAL_CAST
+			val = sd->bonus.varcastrate;
+			break;
+#else
+			FALLTHROUGH
 #endif
 		case SP_CASTRATE:
 				val = sd->castrate;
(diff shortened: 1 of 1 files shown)


src/map/pc.c
```diff
@@ -8417,8 +8417,12 @@ static int64 pc_readparam(const struct map_session_data *sd, int type)
 		case SP_FLEE1:           val = sd->battle_status.flee; break;
 		case SP_FLEE2:           val = sd->battle_status.flee2; break;
 		case SP_DEFELE:          val = sd->battle_status.def_ele; break;
-#ifndef RENEWAL_CAST
 		case SP_VARCASTRATE:
+#ifdef RENEWAL_CAST
+			val = sd->bonus.varcastrate;
+			break;
+#else
+			FALLTHROUGH
```
4144: Is this correct what SP_VARCASTRATE on pre renewal will works same as SP_CASTRATE?
Before SP_VARCASTRATE on pre renewal was ignored.

Kenpachi2k13: It's the same as before. The old code:
```
#ifndef RENEWAL_CAST
		case SP_VARCASTRATE:
#endif
		case SP_CASTRATE:
				val = sd->castrate;
			break;
```
If `RENEWAL_CAST` is not defined, `SP_VARCASTRATE` falls through to `SP_CASTRATE`.
I just explicitly stated the `FALLTHROUGH` for better readability,

4144: FALLTHROUGH here need for mark what missing break is not at issue. This need for compiler

PR #2628, opened 2020-02-09
Add unitiswalking() script command

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

Regarding [this topic](https://herc.ws/board/topic/17846-possible-to-move-characters-player), I'd like to add `unitiswalking()` script command to improve the usability of `unitwalk()` script command.

**Issues addressed:** None.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -6957,6 +6957,37 @@ Examples:
 
 ---------------------------------------
 
+*unitiswalking({<GID>})
+
+This command checks, if a unit is walking or not.
+If <GID> is omitted, the currently attached character is used.
+Returns 1 if the unit is walking, 0 if the unit is not walking and -1 on error.
+
+Note: There's no differentiation between script and client initiated walking.
+
+Example:
+
+prontera,155,185,5	script	Check Walking	1_F_MARIA,{
+	mes("Enter character name.");
+	mes("");
+	input(.@name$);
+	.@GID = getcharid(CHAR_ID_ACCOUNT, .@name$);
+	if (.@GID != 0) {
+		.@iswalking = unitiswalking(.@GID);
+		if (.@iswalking == 1)
+			mesf("%s is walking.", .@name$);
+		else if (.@iswalking == 0)
+			mesf("%s is not walking.", .@name$);
+		else
+			mesf("Can't get %s's walking state.", .@name$);
+	} else {
+		mesf("%s not found!", .@name$);
+	}
+	close();
+}
+
+---------------------------------------
+
 *unitkill(<GID>)
 *unitwarp(<GID>, <Mapname>, <x>, <y>)
 *unitattack(<GID>, <Target ID>)
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -21025,6 +21025,38 @@ static BUILDIN(unitwalk)
 	return true;
 }
 
+/**
+ * Checks if a unit is walking.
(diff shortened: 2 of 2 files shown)


dastgirp:
Although, this will still return 1 when player has manually initiated the walk (by client), adding that note to script command should avoid confusion

Kenpachi2k13:
> Although, this will still return 1 when player has manually initiated the walk (by client), adding that note to script command should avoid confusion

Done.

Emistry:
was this an attempt to create a script that required player to walk to somewhere and then trigger the next event? some sort like that?
if yes, I would said rAthena's approach is way better ... which only trigger the next event label upon reach destination.

having `while + sleep` in the script doesn't seem like what Herc would want promote such practice. 
I believe @MishimaHaruna would like to elimate the `sleep()` in the future, as its a bad practice for scripts. (discussion exist in previous issue, not sure which)

btw, what if player are currently lag or having super high latency? does it still return 1 while walking halfway, and serious lag occur like every 3 sec for 1 cells o-o

doc/script_commands.txt
```diff
@@ -6957,6 +6957,34 @@ Examples:
...
+prontera,155,185,5	script	Walky	1_F_MARIA,{
+	// Get the character's GID.
+	.@GID = getcharid(CHAR_ID_ACCOUNT);
+
+	// Make the character walk.
+	unitwalk(.@GID, 155, 175);
+
+	// Pausing the script while the character is walking.
+	while (unitiswalking(.@GID) == 1)
+		sleep2(50);
```
4144: sleep is bad idea. better use npc timer

src/map/script.c
```diff
@@ -21025,6 +21025,37 @@ static BUILDIN(unitwalk)
...
+ * Returns 1 if unit is walking, 0 if unit is not walking and -1 on error.
+ *
+ * @code{.herc}
+ *	unitiswalking(<GID>);
+ * @endcode
+ *
+ **/
+static BUILDIN(unitiswalking)
+{
+	struct block_list *bl = map->id2bl(script_getnum(st, 2));
```
4144: parameter probably can be optional.
if parameter missing, you may use ``st->oid``

doc/script_commands.txt
Kenpachi2k13: Using `*npctimer()` commands would cause an infinity loop error and `freeloop()` freezes the script.
So `sleep2()` is the only working solution here.

src/map/script.c
Kenpachi2k13: Done.

Kenpachi2k13:
@Emistry 
1. Please don't nail me down on the referenced topic, it just inspired me. rA's solution is fine, and may be added, too, but maybe someone really wants to know if a unit is currently walking or not.

2. Using `*npctimer()` commands would cause an infinity loop error and `freeloop()` freezes the script.
So `sleep2()` is the only working solution here. And well, it's just an example to show what the command does. Nobody is forced, to use it this way.

3. I can't test the case you described, but since it uses the internal `unit_is_walking()` function, it should be safe.

4144:
hm, why npctimer will create loop?
it's simple timer. it call label, code did what it want, then exit, timer triggered again etc.

With sleep code left executed in npc or player, this is not good to do this without reason.

4144:
this way may works for npc or player or anything else

PR #2168, opened 2018-08-06
Adds options to @fakename so guild and party names can be displayed.

bWolfie:
[//]: # (**********************************)
[//]: # (** Fill in the following fields **)
[//]: # (**********************************)

[//]: # (Note: Lines beginning with syntax such as this one, are comments and will not be visible in your report!)

### Pull Request Prelude

[//]: # (Thank you for working on improving Hercules!)

[//]: # (Please complete these steps and check the following boxes by putting an `x` inside the brackets _before_ filing your Pull Request.)

- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR will be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

[//]: # (Describe at length, the changes that this pull request makes.)
Adds a parameter to the `@fakename` command, allowing the display of guild/party names. I previously made this PR, but deleted due to resetting my fork.

E.g.
`@fakename 0 Great Name` - would display the fake name 'Great Name' and hide party, guild and position names.
`@fakename 1 Great Name` - would display the fake name 'Great Name' and hide party name, but not guild and position.

**Affected Branches:** 

[//]: # (Master? Slave?)
- [x] master
- [x] stable

**Issues addressed:**

[//]: # (Issue Tracker Number if any.)

### Known Issues and TODO List

[//]: # (Insert checklist here)
[//]: # (Syntax: - [ ] Checkbox)

[//]: # (**NOTE** Enable the setting "[√] Allow edits from maintainers." when creating your pull request if you have not already enabled it.)

[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style

PR #2637, opened 2020-02-22
Fix/extend @fakename

### LINKED ISSUE #1966: fakename stops guild/party names from showing

### Issue Prelude

### Description

[//]: # (Description of the problem or issue at length.)
[//]: # (Please specify any battle configuration related to the components of this issue that have been changed from the default values. This will allow quicker determination of the cause of the problem.)

When you have a fakename from using **@fakename**, your guild and party names and guild positions won't be shown.

In the function `clif_charnameack`, I was able to resolve this issue. I'm unsure if it has side effects, but seems to work okay so far!

Find:
```
			if (ssd->fakename[0] != '\0') {
				WBUFW(buf, 0) = cmd = 0x195;
				memcpy(WBUFP(buf,6), ssd->fakename, NAME_LENGTH);
				WBUFB(buf,30) = WBUFB(buf,54) = WBUFB(buf,78) = 0;
- 				break;
			}
+			else
- 			memcpy(WBUFP(buf,6), ssd->status.name, NAME_LENGTH);
+ 				memcpy(WBUFP(buf,6), ssd->status.name, NAME_LENGTH);
```

Also, this is not a bug, rather an undesired feature. I will try to figure out how PR works to see if I can do it myself.

**Branch(es):**
- [x] master
- [ ] other

**Hercules rev. hash/commit:** 

Git revision src:

[//]: # (Copy the first 3 lines of the login-server, char-server or map-server startup.)
[//]: # ( [Info]: Hercules 64-bit for Mac OS X )
[//]: # ( [Info]: Git revision src: 'a5918b329ca0826b04dca32ede783586403f58db' )
[//]: # ( [Info]: Git revision scripts: 'a5918b329ca0826b04dca32ede783586403f58db' )

### Operating System

[//]: # (Mac OS X 10.12.3 16D32 [x86_64])
[//]: # (Thank you for adhering to this process! It ensures your issue is resolved quickly and that neither your nor our time is needlessly wasted.)
[//]: # (This template is for problem reports. For other types of report, edit it accordingly.)
[//]: # (For fixes please create a Pull Request.)

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

 * Fix the issue, where not even the fake name shows up, when using `@fakename` in RE clients.
 * Add parameter `options` to `@fakename`, to be able to select which names will be displayed.

Closes #2168
Closes #1966

### DIFF
--- a/doc/atcommands.txt
+++ b/doc/atcommands.txt
@@ -968,10 +968,19 @@ If no position is given, the command defaults to headgear.
 
 ---------------------------------------
 
-@fakename {<text string>}
+@fakename {<options>} {<fake_name>}
 
-Temporarily changes name to the specified string.
-If no string is given, the character's real name will be re-applied.
+Temporarily changes the character's name to <fake_name>.
+If no parameter is passed, the character's real name will be re-applied.
+The <options> parameter is used, to modify which names will be displayed.
+The fake name will always be displayed, regardless of <options>.
+Valid <options> flags (combinable):
+  0 - Only the fake name is displayed. (Default value.)
+  1 - Display party name.
+  2 - Display guild name.
+  4 - Display guild position.
+  8 - Display clan position.
+ 16 - Display title.
 
 ---------------------------------------
 
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -7856,6 +7856,7 @@ ACMD(fakename)
 		if (sd->fakename[0])
 		{
 			sd->fakename[0] = '\0';
+			sd->fakename_options = FAKENAME_OPTION_NONE;
 			clif->blname_ack(0, &sd->bl);
 			if( sd->disguise )
 				clif->blname_ack(sd->fd, &sd->bl);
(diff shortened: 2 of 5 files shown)


src/map/atcommand.c
```diff
@@ -7845,39 +7845,53 @@ ACMD(monsterignore)
...
 ACMD(fakename)
 {
-	if (!*message)
-	{
-		if (sd->fakename[0])
-		{
+	char fake_name[NAME_LENGTH] = {'\0'};
+	int options = FAKENAME_OPTION_NONE;
+
+	if (*message == '\0' || sscanf(message, "%23s %3d", fake_name, &options) < 1) {
```
MishimaHaruna: This will fail to match strings containing spaces, and incorrectly parse strings containing spaces and numbers.

For example, the following will have unexpected results:

- `@fakename Hello World` -> `fake_name`: "Hello", `options`: unmodified, `sscanf()` result: 1 ("World" discarded)
- `@fakename Hello World 1` -> `fake_name`: "Hello", `options`: unmodified, `sscanf()` result: 1 ("World 1" discarded)
- `@fakename Hello 1 World` -> `fake_name`: "Hello", `options`: 1, `sscanf()` result: 2 ("World" discarded)
- `@fakename Hello 123 1` ->  `fake_name`: "Hello", `options`: 123, `sscanf()` result: 2 ("1" discarded)

It might be easier to move the options flag as the first argument (with `sscanf(message, "%d %[^\n]", &options, fake_name)`), and handle the case where `*message != '\0'` but `sscanf() < 2`, by using `message` as is (so that the flag is optional). It would still create ambiguity for those that want to use a fakename such as "123 foo", but those could just use `@fakename 0 123 foo`.

Kenpachi2k13: This won't work, since `sscanf()` would fail if only the fake name is passed because it tries to parse a number as first argument and thus returns 0.
Guess we have to work with quotation marks here...

MishimaHaruna: I was thinking of something along the lines of:

```
	char buf[NAME_LENGTH] = "";
	char *fake_name = NULL;
	int options = FAKENAME_OPTION_NONE

	if (*message == '\0') {
		/* disable fakename */
	} else {
		if (sscanf(message, "%d %23[^\n]", &options, buf, &options) == 2) // new syntax
			fake_name = buf;
		else // old syntax
			fake_name = message;

		/* enable fakename */
	}
```

Kenpachi2k13: Yeah, this looks good. :+1:

Kenpachi2k13: Done.

doc/atcommands.txt
```diff
@@ -968,10 +968,19 @@ If no position is given, the command defaults to headgear.
...
+If no parameter is passed, the character's real name will be re-applied.
+The <options> parameter is used, to modify which names will be displayed.
+The fake name will always be displayed, regardless of <options>.
+Valid <options> flags (combinable):
+  0 - Only the fake name is displayed. (Default value.)
+  1 - Display party name.
+  2 - Display guild name.
+  4 - Display guild position.
+  8 - Display clan position.
+ 16 - Display title.
```
Helianthella, line 983: the enum in atcommand.h should be exposed to the script engine as constants rather than having magic numbers in the documentation

Kenpachi2k13: Since you can't use those constants when executing @fakename, I don't see any benefit in this.

Issue #1211, opened 2016-03-21
Improvised Song not working


Shyne:
Improvised Song not release any magic skill.
I tried use @addperm skill_unconditional to spam but nothing.

Kenpachi2k13:
This happens because `sd->state.abra_flag` is set to 0 in `skill_check_condition_castbegin()`.
For Hocus-Pocus skills, this is correct, because they require catalyst items.
Since skills, casted by Improvised Song, do **NOT** require catalyst items, `sd->state.abra_flag` should only be set to 0 in `skill_check_condition_castbegin()`, if it's value is 1 (Hocus-Pocus).
If `sd->state.abra_flag` is 2 (Improvised Song), it should be kept for usage in `skill_check_condition_castend()`.

PR #2641, opened 2020-02-22
Make skills, casted by Improvised Song, ignore all requirements

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

Skills, casted by Improvised Song (`WM_RANDOMIZESPELL`), should ignore all requirements.
Therefore `skill_check_condition_castbegin()` should return 1, if `abra_flag` is 2. The same applies to `skill_check_condition_castend()`, but here `abra_flag` should be unset before returning 1.

**Issues addressed:** #1211 

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -14029,6 +14029,9 @@ static int skill_check_condition_castbegin(struct map_session_data *sd, uint16 s
 		return 1;
 	}
 
+	if (sd->state.abra_flag == 2) /// Skills, casted by Improvised Song, ignore all requirements.
+		return 1;
+
 	if (pc_has_permission(sd, PC_PERM_SKILL_UNCONDITIONAL) && sd->skillitem != skill_id) {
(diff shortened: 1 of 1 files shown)


Kenpachi2k13:
> i hope abra flag reset in all cases after skill complete or failed?
> i mean for avoid exploiting persistent enabled abra flag

I had a closer look at the `abra_flag` handling and came to the conclusion, that it really is abusable in some cases and actually is a huge mess.
 * The `abra_flag` state itself is hardly used in favour of the checking `skillitem` all over the code.
 * A lot of stuff is superfluous.
 * Some important checks are missing.
 * The helper variables I added in #2616 can be replaced by using `abra_flag` state 3.

I'll close this PR and open a new one, where the whole `abra_flag` related stuff gets cleaned up and secured as far as possible. (Since the skill casts don't have unique identifiers, the only way to validate automatically cast skills, is by validating the skill's ID and level, which is not 100% safe.)

PR #2207, opened 2018-09-02
Fixed Issue #2180

dastgirp:
[//]: # (**********************************)
[//]: # (** Fill in the following fields **)
[//]: # (**********************************)

[//]: # (Note: Lines beginning with syntax such as this one, are comments and will not be visible in your report!)

### Pull Request Prelude

[//]: # (Thank you for working on improving Hercules!)

[//]: # (Please complete these steps and check the following boxes by putting an `x` inside the brackets _before_ filing your Pull Request.)

- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR will be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

[//]: # (Describe at length, the changes that this pull request makes.)
OnEquipScript and OnUnequipScript is now not called if equipment/card is nullified

**Affected Branches:** 

[//]: # (Master? Slave?)

**Issues addressed:**

#2180 

### Known Issues and TODO List

[//]: # (Insert checklist here)
[//]: # (Syntax: - [ ] Checkbox)

[//]: # (**NOTE** Enable the setting "[√] Allow edits from maintainers." when creating your pull request if you have not already enabled it.)

[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style

Issue #2180, opened 2018-08-14
 map_zone_db (disabled_items)


kyeme:
[//]: # (**********************************)
[//]: # (** Fill in the following fields **)
[//]: # (**********************************)

[//]: # (Note: Lines beginning with syntax such as this one, are comments and will not be visible in your report!)

### Issue Prelude

[//]: # (Please complete these mandatory steps and check the following boxes by putting an `x` inside the brackets _before_ filing your issue)

-  [x] I have not modified the source prior to reproducing this issue.
-  [x] I am using the latest version of Hercules.
-  [x] I am aware that this report will be closed or deleted if it becomes obvious that I am stating the false.

Even the item is disabled on map_zone_db OnEquipScript: <" "> and OnUnequipScipt: <" "> still works.
I think @AnnieRuru fixes was broken.

My old report: http://herc.ws/oldboard/tracker/issue-8271-map-zone-db-disabled-items/?gopid=23175#entry23175

@AnnieRuru fixes: https://github.com/HerculesWS/Hercules/commit/30d748f4ab332b501b6db5ce59421ea42387a8b8

PR #2642, opened 2020-02-23
Fix (un)equip script execution for map zone restrcited items

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

 * Prevent the execution of an item's (un)equip script, regardless of the `unequip_restricted_equipment` configuration, if the item is restricted in that zone.
 * Force the execution of an item's unequip script when entering a zone where the item is restricted, even if the item won't be unequipped.

This replaces #2207, since there seems to be no more progress.

@MishimaHaruna
This may be merged manually after #2600, since that PR applies code style to `clif_parse_LoadEndAck()`.

**Issues addressed:** #2180

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -10585,6 +10585,65 @@ static void clif_parse_LoadEndAck(int fd, struct map_session_data *sd)
 		clif->updatestatus(sd, SP_CARTINFO);
 	}
 
+	/**
+	 * In official servers, an item's unequip script is executed when entering a zone where the item is restricted,
+	 * even if the item won't be unequipped.
+	 *
+	 **/
+	if (map->list[sd->bl.m].zone != NULL && map->list[sd->bl.m].zone->disabled_items_count != 0) {
+		struct map_zone_data *zone = map->list[sd->bl.m].zone;
+		int dis_items_cnt = zone->disabled_items_count;
+		int handled_equip = 0x00000000;
+
+		for (int i = 0; i < EQI_MAX; i++) {
+			if (sd->equip_index[i] == INDEX_NOT_FOUND)
+				continue;
+
+			int inv_idx = sd->equip_index[i];
+			struct item_data *equip_data = sd->inventory_data[inv_idx];
+
+			if (equip_data == NULL)
+				continue;
+
+			if ((handled_equip & equip_data->equip) != 0)
+				continue; // Equipment takes multiple slots and was already handled.
+
+			handled_equip |= equip_data->equip;
+
+			if (equip_data->unequip_script != NULL) {
+				int idx;
+
+				ARR_FIND(0, dis_items_cnt, idx, zone->disabled_items[idx] == equip_data->nameid);
+
+				if (idx < dis_items_cnt)
+					script->run_item_unequip_script(sd, equip_data, npc->fake_nd->bl.id);
+			}
+
+			if (inv_idx != sd->equip_index[i])
+				continue; // Unequip script execution corrupted the inventory index.
+
+			struct item *equip = &sd->status.inventory[inv_idx];
+
+			if (equip != NULL && !itemdb_isspecial(equip->card[0])) {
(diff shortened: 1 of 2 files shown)


src/map/clif.c
```diff
@@ -10585,6 +10585,45 @@ static void clif_parse_LoadEndAck(int fd, struct map_session_data *sd)
...
 	}
 
+	/**
+	 * In official servers, an item's unequip script is executed when entering a zone where the item is restricted,
+	 * even if the item won't be unequipped.
+	 *
+	 **/
+	for (int i = 0; i < EQI_MAX; i++) {
+		if (sd->equip_index[i] >= 0) {
+			struct item_data *equip_data = sd->inventory_data[sd->equip_index[i]];
```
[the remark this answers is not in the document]
Kenpachi2k13: Actually this should be impossible, since the item is equipped, but I added a `NULL` validation.
Additionally I added a validation for item which take multiple equipment slots, do prevent running the same unequip script multiple times.

src/map/clif.c
```diff
@@ -10585,6 +10585,45 @@ static void clif_parse_LoadEndAck(int fd, struct map_session_data *sd)
...
+			struct map_zone_data *zone = map->list[sd->bl.m].zone;
```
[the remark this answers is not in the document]
Kenpachi2k13: The "All" zone is attached to every map, but I added a `NULL` validation, too.

src/map/clif.c
```diff
@@ -10585,6 +10585,58 @@ static void clif_parse_LoadEndAck(int fd, struct map_session_data *sd)
...
+	 * even if the item won't be unequipped.
+	 *
+	 **/
+	if (map->list[sd->bl.m].zone != NULL && map->list[sd->bl.m].zone->disabled_items_count != 0) {
+		struct map_zone_data *zone = map->list[sd->bl.m].zone;
+		int dis_items_cnt = zone->disabled_items_count;
+		int handled_equip = 0x00000000;
+
+		for (int i = 0; i < EQI_MAX; i++) {
+			struct item_data *equip_data = sd->inventory_data[sd->equip_index[i]];
```
dastgirp: sd->equip_index can contain -1, so it's better to check if it's >= 0, and then use it as index

Kenpachi2k13: Good catch, thanks. 👍

src/map/clif.c
```diff
@@ -10585,6 +10585,61 @@ static void clif_parse_LoadEndAck(int fd, struct map_session_data *sd)
...
+			if (equip_data->unequip_script != NULL) {
+				int idx;
+
+				ARR_FIND(0, dis_items_cnt, idx, zone->disabled_items[idx] == equip_data->nameid);
+
+				if (idx < dis_items_cnt)
+					script->run_item_unequip_script(sd, equip_data, npc->fake_nd->bl.id);
+			}
+
+			struct item *equip = &sd->status.inventory[sd->equip_index[i]];
```
4144: need validate ``sd->equip_index[i]`` again after unequip script was executed

Kenpachi2k13: Done.

src/map/clif.c
```diff
@@ -10585,6 +10585,65 @@ static void clif_parse_LoadEndAck(int fd, struct map_session_data *sd)
...
+
+					struct item_data *card_data = itemdb->exists(equip->card[slot]);
+
+					if (card_data != NULL && card_data->unequip_script != NULL) {
+						int idx;
+
+						ARR_FIND(0, dis_items_cnt, idx, zone->disabled_items[idx] == card_data->nameid);
+
+						if (idx < dis_items_cnt)
+							script->run_item_unequip_script(sd, card_data, npc->fake_nd->bl.id);
```
4144: this is probably for other pr, but if unequip script called from different places, like real unequip and item restrictions, probably need set script variable with flag about why it called

Kenpachi2k13: Sounds reasonable. I'll work on that for the April release.

PR #2647, opened 2020-03-01
Fixed Experience Gain message

dastgirp:
### Pull Request Prelude

### Changes Proposed

Since macro is used in message, I believe we can't move it to messages.conf yet, and it will display the format specifier instead of value

**Issues addressed:** 
None

### DIFF
--- a/conf/messages.conf
+++ b/conf/messages.conf
@@ -694,7 +694,7 @@
 886: Gained %dz.
 887: %s stole an Unknown Item (id: %i).
 888: %s stole %s.
-889: Experience Gained Base:%"PRIu64" (%.2f%%) Job:%"PRIu64" (%.2f%%)
+889: Experience Gained Base:%llu (%.2f%%) Job:%llu (%.2f%%)
 890: [KS Warning!! - Owner : %s]
 891: [Watch out! %s is trying to KS you!]
 892: Growth: hp:%d sp:%d str(%.2f) agi(%.2f) vit(%.2f) int(%.2f) dex(%.2f) luk(%.2f)
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -7120,8 +7120,8 @@ static bool pc_gainexp(struct map_session_data *sd, struct block_list *src, uint
 	if(sd->state.showexp) {
 		char output[256];
 		sprintf(output,
-			msg_sd(sd, 889), // Experience Gained Base:%"PRIu64" (%.2f%%) Job:%"PRIu64" (%.2f%%)
-			base_exp, nextbp * (float)100, job_exp, nextjp * (float)100);
+			msg_sd(sd, 889), // Experience Gained Base:%llu (%.2f%%) Job:%llu (%.2f%%)
+			(unsigned long long)base_exp, nextbp * 100.0f, (unsigned long long)job_exp, nextjp * 100.0f);
 		clif_disp_onlyself(sd, output);
 	}
 


conf/messages.conf
```diff
@@ -694,7 +694,7 @@
 886: Gained %dz.
 887: %s stole an Unknown Item (id: %i).
 888: %s stole %s.
-889: Experience Gained Base:%"PRIu64" (%.2f%%) Job:%"PRIu64" (%.2f%%)
```
4144: probably better left format string as is. but in C code at first replace ``%"PRIu64"`` to actual format parameters. then use fixed format string in sprintf?

dastgirp: oh, I didn't knew. I always though PRIu64 corresponds to different identifier based on OS type (without checking it). 
It seems to be same identifier for every OS. changed it.

MishimaHaruna: This seems to be a Microsoft extension, not compatible with other compiles (at least not compatible with gcc-6 on debian, gcc-9 on gentoo and gcc-9 on macos)

```
#include <stdint.h>
#include <stdio.h>

int main(void)
{
	int64_t foo = 1234567890123456789ull;
	printf("%I64u\n", foo);
}
```



```
debian$ gcc-6 -std=c11 -pedantic -Wall -Wextra test_I64u.c && ./a.out
test_I64u.c: In function ‘main’:
test_I64u.c:7:9: warning: ISO C does not support the 'I' printf flag [-Wformat=]
  printf("%I64u\n", foo);
         ^~~~~~~~~
test_I64u.c:7:14: warning: format ‘%u’ expects argument of type ‘unsigned int’, but argument 2 has type ‘int64_t {aka long int}’ [-Wformat=]
  printf("%I64u\n", foo);
              ^
                                                      2112454933
```

```
gentoo % gcc-9.2.0 -std=c11 -pedantic -Wall -Wextra test_I64u.c && ./a.out
test_I64u.c: In function ‘main’:
test_I64u.c:7:9: warning: ISO C does not support the 'I' printf flag [-Wformat=]
    7 |  printf("%I64u\n", foo);
      |         ^~~~~~~~~
test_I64u.c:7:14: warning: format ‘%u’ expects argument of type ‘unsigned int’, but argument 2 has type ‘int64_t’ {aka ‘long int’} [-Wformat=]
    7 |  printf("%I64u\n", foo);
      |          ~~~~^     ~~~
      |              |     |
      |              |     int64_t {aka long int}
      |              unsigned int
      |          %I64lu
                                                      2112454933
```

```
macos% gcc-9 -std=c11 -pedantic -Wall -Wextra test_I64u.c && ./a.out
test_I64u.c: In function 'main':
test_I64u.c:7:9: warning: ISO C does not support the 'I' printf flag [-Wformat=]
    7 |  printf("%I64u\n", foo);
      |         ^~~~~~~~~
test_I64u.c:7:14: warning: format '%u' expects argument of type 'unsigned int', but argument 2 has type 'int64_t' {aka 'long long int'} [-Wformat=]
    7 |  printf("%I64u\n", foo);
      |          ~~~~^     ~~~
      |              |     |
      |              |     int64_t {aka long long int}
      |              unsigned int
      |          %I64llu
test_I64u.c:4:14: warning: unused parameter 'argc' [-Wunused-parameter]
    4 | int main(int argc, char *argv[])
      |          ~~~~^~~~
test_I64u.c:4:26: warning: unused parameter 'argv' [-Wunused-parameter]
    4 | int main(int argc, char *argv[])
      |                    ~~~~~~^~~~~~
I64u
```

```
macos% clang -std=c11 -pedantic -Wall -Wextra test_I64u.c && ./a.out
test_I64u.c:7:11: warning: length modifier 'I64' results in undefined behavior or no effect with 'u' conversion specifier [-Wformat]
        printf("%I64u\n", foo);
                ~^~~~
test_I64u.c:4:14: warning: unused parameter 'argc' [-Wunused-parameter]
int main(int argc, char *argv[])
             ^
test_I64u.c:4:26: warning: unused parameter 'argv' [-Wunused-parameter]
int main(int argc, char *argv[])
                         ^
3 warnings generated.
I64u
```

I suggest to use `%llu`, and do an explicit cast to `(unsigned long long)` in the printf function that references this messages.

bWolfie:
bump amigos

conf/messages.conf
MishimaHaruna: I applied the change I suggested and force-pushed

PR #2649, opened 2020-03-07
Validate duplicated item_db entries

Emistry:
### Pull Request Prelude

### Changes Proposed

- Enable user to detect "used" item_db (happen when server has custom item_db that clashed with latest update from Herc if any)
- able to detect duplicated items where inherit field aren't used.
- Enable user to detect "duplicated" item_db (happen when user aren't aware of multiple duplicated entries in the past, only happen if `inherit: true` aren't used.)
- Encourage user to use `inherit: true` field for their custom changes to existing item_db.

**Issues addressed:** 
Detect whichever item_db ID are currently clashed with each other (if any)
Lesser duplicated entries of item_db, especially server who has alot custom items.

### DIFF
--- a/src/map/itemdb.c
+++ b/src/map/itemdb.c
@@ -2033,6 +2033,10 @@ static int itemdb_readdb_libconfig_sub(struct config_setting_t *it, int n, const
 		}
 	}
 
+	if (itemdb->exists(id.nameid) && inherit == false) {
+		ShowWarning("itemdb_readdb_libconfig_sub: Duplicated item %d found in \"%s\".\n", id.nameid, source);
+	}
+
 	if( !libconfig->setting_lookup_string(it, "AegisName", &str) || !*str ) {
 		if( !inherit ) {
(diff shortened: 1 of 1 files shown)


MishimaHaruna:
We already have a duplication check (within the same file), to avoid that the same item id is defined more than once in item_db or item_db2, so I think this is redundant.

There are use cases for non-inheriting entries in the item_db2 (when you want to completely override an item, without inheriting any fields from the main entry). If we don't want to allow that, then we should remove the `inherit` option, and make the item_db2 always inherit instead of overriding.

The thing I'd really do, instead of forcing people to inherit instead of overriding, is to update the duplicate check in itemdb_readdb_libconfig, to work for item IDs > MAX_ITEMDB (i.e. it could use a dbmap instead of an array)

Emistry:
hmm ... if i recall correctly, currently the duplicated items are detected only if they exists within same item_db file.
if i have multiple item_db or i create same item in both item_db and item_db2, it doesn't warn me, it will just simply overwrite the existing item.

but if they want to totally overwrite an existing item, they could still do it, but overwrite all values, but just required the extra inherit field.

4144:
may be better if need really duplicate, add some special flag into duplicated db entry?

PR #2657, opened 2020-03-13
Clean up auto-cast related code

### LINKED ISSUE #1211: Improvised Song not working

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

Currently the auto-cast related code is a mess.
 * We have `map_session_data->state.autocast` which is used to temporarily flag a skill as auto-cast for skills cast by `bAutoSpell*` bonuses or by Auto Shadow Spell. It's also (mis)used to prevent damage reflection from infinite looping.
 * We have `map_session_data->state.abra_flag` which isn't actually used. It's just set when casting Abracadabra or Improvised Song, but then never used for validation.
 * We have `map_session_data->skillitem` and `map_session_data->skillitemlv` which hold the auto-cast skill's ID and level. `map_session_data->skillitem` is misused to validate auto-casting all over the code.
 * As of late we have the `itemskill()` script command related states and helper variables.
 * Unsetting/validating the auto-cast data in the right place currently is more likely gambling.

\------------------------------------------------------------------------------------------------------------------

I spent some days in getting into the subject and cleaned it up as far as possible.
 * I added an enumeration for auto-cast types to be able to handle each one separately and use those types for validation instead of the auto-cast skill's ID.
 * I added a structure to unify all auto-cast related data in one place.
 * I added a function which validates the plausibility of auto-cast related data.
 * I revised where and when `pc_autocast_clear()` (former `pc_itemskill_clear()`) gets called.
 * I removed obsolete code parts.
 * I made skills cast by Improvised Song ignore all requirements.

Closes #1211


src/map/pc.h
```diff
@@ -173,6 +173,17 @@ struct pc_combos {
 	int id; /* this combo id */
 };
 
+/** Auto-cast related data. **/
+struct autocast_data {
+	enum autocast_type type; // The auto-cast type.
+	int skill_id; // The auto-cast skill ID.
+	int skill_lv; // The auto-cast skill level.
+	bool itemskill_conditions_checked; // Used by itemskill() script command, to prevent second check of conditions after target was selected.
```
4144: probably here better use bit fields for bools as it was for old fields?
this will reduce memory usage a bit

src/map/skill.c
```diff
@@ -2191,6 +2195,9 @@ static int skill_onskillusage(struct map_session_data *sd, struct block_list *bl
 	if( sd == NULL || !skill_id )
 		return 0;
 
+	// Preserve auto-cast tpye if bAutoSpellOnSkill was triggered by a skill which was cast by Abracadabra, Improvised Song or an item.
```
4144: typo in some comments ``tpye`` -> ``type``

src/map/pc.h
Kenpachi2k13: Wouldn't make a difference in this case.
 * `bool` has a size of 1 byte which makes a total of 4 bytes.
 * A struct containing 4 `unsigned int` fields with a width of 1 (that's how we use bit fields) has a size of 4 bytes.
```
struct {
	unsigned int bit_field1 : 1;
	unsigned int bit_field2 : 1;
	unsigned int bit_field3 : 1;
	unsigned int bit_field4 : 1;
} bit_fields;
ShowError("sizeof(bit_fields): %d\n", sizeof(bit_fields));
struct {
	bool boolean1;
	bool boolean2;
	bool boolean3;
	bool boolean4;
} booleans;
ShowError("sizeof(booleans): %d\n", sizeof(booleans));
```
```
[Error]: sizeof(bit_fields): 4
[Error]: sizeof(booleans): 4
```

There may be compiler depending differences in fields sizes, but as far as i know, this is the most common variation.

src/map/skill.c
Kenpachi2k13: Done.

src/map/pc.h
4144: in this case yes, but if add one new flag, it may use 8 bytes for bools (because padding) and still 4 bytes for bit fields.
But yes probably left it as is for now, if add new flags, then may be convert to bit fields

PR #2691, opened 2020-04-11
Extend and clean up monster skill DB code

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

- Add missing constants to enable global skills assignment.
- Harmonize behaviour of cast condition constants.
- Add some error messages.
- Add proper defaulting and capping when reading monster skill DBs.
- Apply code style.
- Add Windows support for predefined identifier \_\_func__. (Not related to the topic but useful.)

**Issues addressed:** None.

### DIFF
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -5564,51 +5564,70 @@ static bool mob_skill_db_libconfig_sub(struct config_setting_t *it, int n)
 	return true;
 }
 
+/**
+ * Reads a single monster skill from DB.
+ *
+ * @param it The libconfig settings block, which contains the skill data.
+ * @param n The skill data block's index within the parent monster block.
+ * @param mob_id The monster's ID.
+ * @return true on success, false on failure.
+ *
+ **/
 static bool mob_skill_db_libconfig_sub_skill(struct config_setting_t *it, int n, int mob_id)
 {
-	int i, j, idx = 0;
-	int skill_id = 0;
-	int skill_idx = 0;
-	bool clearskills = false;
-	const char *name = config_setting_name(it);
-	struct mob_skill *ms, gms;
-
 	nullpo_retr(false, it);
 	Assert_retr(false, mob_id <= 0 || mob->db(mob_id) != mob->dummy);
 
-	if (!(skill_id = skill->name2id(name))) {
-		ShowWarning("mob_skill_db_libconfig_sub_skill: Non existant skill id %d in monster %d, skipping.\n", skill_id, mob_id);
+	int skill_id = 0;
+	const char *name = config_setting_name(it);
+	const char *mob_str = (mob_id < 0) ? "global ID" : "monster";
+
+	if ((skill_id = skill->name2id(name)) == 0) {
+		ShowWarning("%s: Non existant skill %d in %s %d, skipping.\n", __func__, skill_id, mob_str, mob_id);
 		return false;
 	}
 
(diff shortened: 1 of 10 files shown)


src/map/mob.c
```diff
@@ -3515,8 +3515,6 @@ static int mobskill_use(struct mob_data *md, int64 tick, int event)
 					flag = ((fmd = mob->getfriendstatus(md, ms[i].cond1, ms[i].cond2)) != NULL); break;
 				case MSC_SLAVELT: // slave < num
 					flag = (mob->countslave(&md->bl) < c2 ); break;
-				case MSC_ATTACKPCGT: // attack pc > num
-					flag = (unit->counttargeted(&md->bl) > c2); break;
```
4144, line 3519: it's not same. comparison ``>`` and ``>=``
probably if this things going from aegis, better left all existing constants and code?

Kenpachi2k13: You're right. It's not the same. But `MSC_ATTACKPCGT` is not used and `MSC_ATTACKPCGE` (which is used) can do the same, you just have to increase the value for `ConditionData` by 1.

4144: yes, but i mean if it was added because exists in aegis, probably better left it as is?

Asheraf: @Kenpachi2k13 it's not used in our default database, people might be using it in theirs you can't just remove it without a reason..

src/map/mob.c
```diff
@@ -3513,8 +3513,6 @@ static int mobskill_use(struct mob_data *md, int64 tick, int event)
 				case MSC_FRIENDSTATUSON: // friend status[num] on
 				case MSC_FRIENDSTATUSOFF: // friend status[num] off
 					flag = ((fmd = mob->getfriendstatus(md, ms[i].cond1, ms[i].cond2)) != NULL); break;
-				case MSC_SLAVELT: // slave < num
```
Asheraf, line 3516: same for this one, why is it being removed.

src/map/mob.c
Kenpachi2k13: @Asheraf 
I understand your point of view, but nonetheless I think it's okay to remove those two constants.

- They are not documented so nobody should know about them.
- The feature itself isn't removed. One just need to use another constant and increase the value for `ConditionData` by 1.
- Following your argumentation it would be "forbidden" to rename constants, too, even though their names are misleading.
- Having two constants of same purpose is pointless.

src/map/mob.c
Kenpachi2k13: Removed related commits.

src/map/mob.c
```diff
@@ -6059,7 +6155,7 @@ void mob_defaults(void)
 	mob->getmasterhpltmaxrate = mob_getmasterhpltmaxrate;
 	mob->getfriendstatus_sub = mob_getfriendstatus_sub;
 	mob->getfriendstatus = mob_getfriendstatus;
-	mob->skill_use = mobskill_use;
+	mob->skill_use = mob_skill_use;
```
skyleo, line 6158: Rename the function pointer mob->skill_use as well, so that compilation errors are thrown for all custom code that isn't aware that the return value behavior changed.

skyleo (commented):
@Kenpachi2k13 Please address these suggestions in possibly a new pull request.

Kenpachi2k13:
@skyleo Done in #2887.

Issue #1026, opened 2015-12-26
Using item scroll while casting skill


kyeme:
On Official, even if you're still casting a skill, (example, Storm Gust) and you use an Item Scroll like Heal Scroll, after the said skill has finished its casting, the effect of the said Item Scroll will work automatically. 

On Hercules, this doesn't happen. Please watch my video from iRO.
https://www.youtube.com/watch?v=LimA7o2Umo8

Issue #816, opened 2015-10-24
Item Skill Scroll not working properly


Jedzkie:
Follow up! XD

http://herc.ws/board/tracker/issue-7739-item-skill-scroll-not-working-properly/

Kenpachi2k13:
I can confirm that using items doesn't abort the skill.
But casting another skill - even a self skill - aborts it, so this part can't be confirmed.

Issue #2412, opened 2019-03-25
Unable to "spam" skill


kyeme:
**Describe the bug**
Herc: The next skill will only be activated AFTER the previous skill has finished its casting time.

**Expected behavior**
Official: When your character is about to finish casting it's recent skill, (like the picture shown below) and you prompted to use a new skill by pressing its corresponding hotkey even before it finished it's casting, it will automatically be casted right after your recent skill.

Official Video: https://youtu.be/MR-RLcTnfqA

Screenshot:
![1](https://user-images.githubusercontent.com/4656578/54901989-12ce6580-4f13-11e9-9014-b6831cbd5eea.jpg)

Notes:
	- Also, the Skill being casted is about 95%~ complete and you prompted your character to WALK, its going to WALK RIGHT AWAY. (Maybe because it only reads a command AFTER you fully used/casted a skill)
	- Ground skills are good. You can compare the Ground Skills to Self-Skills (with no delay skill like Energy Coat, CH_SOULCOLLECT etc.) for you guys to reproduce the bug easily.
	- The "95%" casting is only based on my observation wherein you can follow up a next skill after the cast time of your recent skill. (Again, by clicking the desired skill to be used/casted before it finished its casting)Notes:
	- Also, the Skill being casted is about 95%~ complete and you prompted your character to WALK, its going to WALK RIGHT AWAY. (Maybe because it only reads a command AFTER you fully used/casted a skill)
	- Ground skills are good. You can compare the Ground Skills to Self-Skills (with no delay skill like Energy Coat, CH_SOULCOLLECT etc.) for you guys to reproduce the bug easily.
	- The "95%" casting is only based on my observation wherein you can follow up a next skill after the cast time of your recent skill. (Again, by clicking the desired skill to be used/casted before it finished its casting)

**Screenshots**

**System specs (please complete the following information):**
 - OS: [e.g. windows 10 version 1809, macOS 10.14]
 - Hercules Version [e.g. v2018.09.22]
 - Mode: [renewal or pre-renewal?]
 - Packet version: [e.g. 20101130]
 - Client type: [e.g. main, RE, zero, sak, ad]

**Plugins used or source modifications**
If you are using any plugins besides the ones that come bundled with Hercules,
please enumerate them here.

**Additional context**
Add any other context about the problem here. If the bug report is about a
crash, please attach the core/stack-dump or crash-log, if any.

AnnieRuru:
[...]

PR #2699, opened 2020-04-21
Enable multiple itemskill() calls per item

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

The current auto-cast code doesn't allow multiple `itemskill()` calls per item.
I changed the code to be vector based to be able to hold multiple auto-cast datasets.
Additionally the auto-cast cleaning related code was adjusted accordingly.
Other side effects are:
 * Using an item while the target cursor is active won't abort the skill anymore.
 * Auto-cast skills which are cast while casting another skill will now be executed when the current casting is finished.

**Issues addressed:** #816, #1026, #2750


src/map/skill.c
```diff
@@ -1194,14 +1194,16 @@ static void skill_validate_autocast_data(struct map_session_data *sd, int skill_
 	// Determine if called by clif_parse_UseSkillMap().
 	bool use_skill_map = (skill_lv == 0 && (skill_id == AL_WARP || skill_id == AL_TELEPORT));
 
-	if (sd->auto_cast_current.type == AUTOCAST_NONE)
+	struct autocast_data auto_cast = sd->auto_cast_current;
```
4144, line 1197: here will be copy of struct autocast_data from sd->auto_cast_current. This can be bit slow and allocate memory for copy. Is it really need here?

If dont need, you can get address of sd->auto_cast_current
``struct autocast_data *auto_cast = &sd->auto_cast_current;`` and use it like this ``auto_cast->type == ... ``

Kenpachi2k13: Thanks for the hint. Fixed!

kyeme:
The autocast of item skill is sometimes not working.

Kenpachi2k13:
> The autocast of item skill is sometimes not working.

Since you posted here I assume you added this changes to your emulator for test purposes.
Could tell me a little bit more about the circumstances? Can you reproduce it? Does it happen for specific items/skills? Any other info?

kyeme:
I used heal scroll. I don't know why it's not working.
https://www.youtube.com/watch?v=RVvW-TPeKGc&feature=youtu.be

Kenpachi2k13:
Okay, I know how to reproduce this issue.
If the item's skill is used within the last period of casting the item's skill won't be cast because of `ud->skilltimer` is set. The critical point is close to the middle of casting but I could figured out the exact time yet.
It's odd that using the item's skill at the beginning isn't affected by `ud->skilltimer`, since it's set then, too.

As far as I can tell at the moment, the actual problem isn't the auto-cast related code (this PR) but the issue you reported in #2412.

MishimaHaruna:
I'm moving this to the next release so that I have time to give it a proper review

Issue #1037, opened 2015-12-28
Validate string variable (mapreg, charreg, etc) length


MishimaHaruna:
This is similar to #1003, but with the string variable content, rather than its name.

We currently limit the length to 255 (should we increase it to something larger maybe?), but in some places that length isn't enforced or checked correctly. More info in the discussion of #1003.

PR #2705, opened 2020-05-01
Cap string variable value length and unify corresponding SQL columns size

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

Currently the maximum length of string variable values is inconsistent.
 * If a variable is not stored in a SQL table its value's length is not restricted.
 * If a variable is stored in `acc_reg_str_db`, `char_reg_str_db` or `global_acc_reg_str_db` table its value's length is restricted to 254.
 * If a variable is stored in `mapreg` table its value's length is restricted to 255.

Furthermore some fields are too small to use the maximum supported value length.

To have a consistent size I capped the length of all string variable values to 255 and unified the corresponding SQL columns size accordingly.

**Issues addressed:** #1037


sql-files/upgrades/2020-05-01--04-44.sql
```diff
@@ -0,0 +1,25 @@
...
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+ALTER TABLE `acc_reg_str_db` MODIFY `value` VARCHAR(255) NOT NULL DEFAULT '';
```
4144, line 21: default value look like different with defaults from main.sql

src/map/atcommand.c
```diff
@@ -8945,13 +8945,17 @@ ACMD(accinfo)
...
+	char reg[SCRIPT_VARNAME_LENGTH + 1];
+	char val[SCRIPT_STRING_VAR_LENGTH + 1];
 	struct script_data* data;
 	int toset = 0;
 	bool is_str = false;
 	size_t len;
 
-	if (!*message || (toset = sscanf(message, "%32s %253[^\n]", reg, val)) < 1) {
+	char format[20];
+	snprintf(format, sizeof(format), "%%%ds %%%d[^\n]", SCRIPT_VARNAME_LENGTH, SCRIPT_STRING_VAR_LENGTH);
```
4144, line 8956: better use safesnprintf

4144: also if SCRIPT_VARNAME_LENGTH will be 2^32, will be buffer overflow

Kenpachi2k13: Why should someone set `SCRIPT_VARNAME_LENGTH` to 4294967296? o.O

sql-files/upgrades/2020-05-01--04-44.sql
Kenpachi2k13: Fixed.

src/map/atcommand.c
Kenpachi2k13: > better use safesnprintf
Fixed.

> also if SCRIPT_VARNAME_LENGTH will be 2^32, will be buffer overflow
Added static assert check for SCRIPT_VARNAME_LENGTH.

src/map/script.c
```diff
@@ -3439,43 +3439,53 @@ static int set_reg(struct script_state *st, struct map_session_data *sd, int64 n
 	if (is_string_variable(name)) {// string variable
 		const char *str = (const char*)value;
 
+		if (strlen(str) > SCRIPT_STRING_VAR_LENGTH) {
```
MishimaHaruna, line 3442: How about only limiting the variables that are persisted to the database? Their use in scripts is (or, should) be far less common than the non-persisted variables, and unless there's a compelling reason to limit the maximum string length I'd avoid it at all costs.

Kenpachi2k13: Sure, easy to do... But I thought you're a friend of having consistency. That's why I made no exception for temporary variables.

MishimaHaruna: I agree with the consistency part, but it's also true that most use cases for string variables only care about scope variables, and are never saved to the database.

I'm mostly thinking of long menu sequences constructed by concatenating options into a string (which the client limits to 127 options if I remember correctly, and we surely want each of them to be able to be more than 1 character long, not including the `:`), or of long text to print through mes (I believe the client can support up to 4k characters in a single message dialog, with some NEMO patches applied?), so it makes sense not to add an artificial restriction on those strings unless it's necessary for technical reasons

Issue #2670, opened 2020-03-28
Skill index/tree calculation broken


Kenpachi2k13:
**Describe the bug**
A user reported via Discord that the calculation for skill index/tree is broken:

> hey. i just want to ask you, do you know about problem with new skill indexes & bard/dancer linked skills?
> Is server side problem
> Check pc_calc_skilltree
> There must be i+8, cause indexes are reversed now
> actually somehow it affects all this function
> we just revert this PR (https://github.com/HerculesWS/Hercules/pull/2596) on our server until you will fix all these indexes, it's not only affects bard/dancer linked skill, it also affects skill reset, tree calculation and etc.
![unknown](https://user-images.githubusercontent.com/3476227/77814947-18c2cd80-70b6-11ea-9ccf-b567e650138d.png)

MishimaHaruna:
I have a draft to fix this part in an index-independent way (doing that kind of arithmetic on an index is just plain wrong anyway...)

PR #2710, opened 2020-05-03
Fix skill idx

MishimaHaruna:
### Pull Request Prelude

### Changes Proposed

This PR adds some robustness to the code related to `skill_get_index()` and fixes some wrong assumptions and dangerous index arithmetic index in the Bard/Dancer Soul Link temporarily granted skills, reported in #2670 .

Warnings have been added, to help detecting incorrectly defined ranges for `skill_get_index()`.

Finally, the use of the `SKILL_FLAG_REPLACED_LV_0` value (and in general, values from `enum e_skill_flag`) have been sanitized, since suspicion on them was raised in a private report about attempts to save negative skill indexes to the database (which may or may not have been a secondary effect of the Bard/Dancer granted skills issue).

**Issues addressed:** #2670 

### DIFF
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -1627,31 +1627,32 @@ static int pc_calc_skilltree(struct map_session_data *sd)
 			sd->status.skill[i].flag = SKILL_FLAG_PERMANENT;
 		}
 
-		if( sd->sc.count && sd->sc.data[SC_SOULLINK] && sd->sc.data[SC_SOULLINK]->val2 == SL_BARDDANCER && skill->dbs->db[i].nameid >= DC_HUMMING && skill->dbs->db[i].nameid <= DC_SERVICEFORYOU )
-		{ //Enable Bard/Dancer spirit linked skills.
-			if (sd->status.sex) {
-				// Link dancer skills to bard.
-				if (i < 8) {
-					Assert_report(i >= 8);
-					continue;
-				}
-				if (sd->status.skill[i-8].lv < 10)
-					continue;
-				sd->status.skill[i].id = skill->dbs->db[i].nameid;
-				sd->status.skill[i].lv = sd->status.skill[i-8].lv; // Set the level to the same as the linking skill
-				sd->status.skill[i].flag = SKILL_FLAG_TEMPORARY; // Tag it as a non-savable, non-uppable, bonus skill
-			} else {
-				// Link bard skills to dancer.
-				if (i < 8) {
-					Assert_report(i >= 8);
-					continue;
-				}
-				if (sd->status.skill[i].lv < 10)
-					continue;
-				sd->status.skill[i-8].id = skill->dbs->db[i-8].nameid;
-				sd->status.skill[i-8].lv = sd->status.skill[i].lv; // Set the level to the same as the linking skill
-				sd->status.skill[i-8].flag = SKILL_FLAG_TEMPORARY; // Tag it as a non-savable, non-uppable, bonus skill
+		if (sd->sc.count && sd->sc.data[SC_SOULLINK] && sd->sc.data[SC_SOULLINK]->val2 == SL_BARDDANCER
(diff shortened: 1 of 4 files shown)


src/map/pc.c
```diff
@@ -1627,31 +1627,32 @@ static int pc_calc_skilltree(struct map_session_data *sd)
...
-				sd->status.skill[i-8].id = skill->dbs->db[i-8].nameid;
-				sd->status.skill[i-8].lv = sd->status.skill[i].lv; // Set the level to the same as the linking skill
-				sd->status.skill[i-8].flag = SKILL_FLAG_TEMPORARY; // Tag it as a non-savable, non-uppable, bonus skill
+		if (sd->sc.count && sd->sc.data[SC_SOULLINK] && sd->sc.data[SC_SOULLINK]->val2 == SL_BARDDANCER
+		 && ((skill->dbs->db[i].nameid >= BA_WHISTLE && skill->dbs->db[i].nameid <= BA_APPLEIDUN)
+		  || (skill->dbs->db[i].nameid >= DC_HUMMING && skill->dbs->db[i].nameid <= DC_SERVICEFORYOU))
+		) {
+			//Enable Bard/Dancer spirit linked skills.
+			int linked_nameid = skill->get_linked_song_dance_id(skill->dbs->db[i].nameid);
+			if (linked_nameid) {
```
4144, line 1633: probably need check ``linked_nameid == 0``?

MishimaHaruna: You're right, thank you

src/map/skill.c
```diff
@@ -133,17 +133,23 @@ static int skill_get_index(int skill_id)
...
 			skill_idx += (skill_idx_ranges[i].end - skill_id);
+			found = true;
 			break;
 		}
 		// Add the difference of current range
 		skill_idx += (skill_idx_ranges[i].end - skill_idx_ranges[i].start + 1);
 	}
 
+	if (!found) {
+		ShowWarning("skill_get_index: skill id '%d' (idx: %d) is not handled as it lies outside the defined ranges!\n", skill_id, skill_idx);
```
hemagx, line 150: maybe assert it too?

src/map/skill.c
```diff
@@ -10938,6 +10938,32 @@ static int skill_count_wos(struct block_list *bl, va_list ap)
...
+			return DC_SERVICEFORYOU;
+		case DC_HUMMING:
+			return BA_WHISTLE;
+		case DC_DONTFORGETME:
+			return BA_ASSASSINCROSS;
+		case DC_FORTUNEKISS:
+			return BA_POEMBRAGI;
+		case DC_SERVICEFORYOU:
+			return BA_APPLEIDUN;
+	}
```
hemagx, line 10963: in here it silently fail, should not assert instead?

MishimaHaruna: I'll document that it returns 0 if there's no linked ID, so that it can be used in a place that doesn't need any knowledge of the skill IDs if we need to

src/map/skill.c
```diff
@@ -96,47 +124,24 @@ static int skill_name2id(const char *name)
...
-			CUSTOM_SKILL_RANGES};
-	int length = sizeof(skillRange) / sizeof(int);
-	STATIC_ASSERT(sizeof(skillRange) / sizeof(int) % 2 == 0, "skill_get_index: skillRange should be multiple of 2");
-
-
-	if (skill_id < skillRange[0] || skill_id > skillRange[length - 1]) {
+	int length = ARRAYLENGTH(skill_idx_ranges);
+
+
+	if (skill_id < skill_idx_ranges[0].start || skill_id > skill_idx_ranges[length - 1].end) {
```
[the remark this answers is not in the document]
MishimaHaruna: Good point. I added an Assert_report() to every failure branch condition of skill_get index, so that it'll generate a backtrace

src/map/skill.c
MishimaHaruna: Handled, see above

src/map/skill.c
MishimaHaruna: Done

Issue #768, opened 2015-10-04
error occur for unitskillusepos script command.


Emistry:
Sample Script : http://pastebin.com/raw.php?i=cjmDVb5A
## GIT Version

```
[Info]: Git revision (src): '4db986b93566d673c226971fa045f1dc99645242'
[Info]: Git revision (scripts): '4db986b93566d673c226971fa045f1dc99645242'
```
## Step to Reproduce :

Load the Sample Script given using by restart server / @reloadscript (First Error Shown) 
![Image](http://i.imgur.com/0CvStyT.png)

After that, execute @reloadscript ( second error below shown ) , you may try several times, it will show same result.
![Image](http://i.imgur.com/4ZThFxR.png)
Skill ID 85 = Lord of Vermillion.
## In-game Screenshot :

![Image](http://i.imgur.com/kx04Jwq.png)
The skill will continue cast by the NPC, but the error will not shown in map-server.
## Side Note :

Beside, somehow when I first try it, the map-server spam some other error from the core, I unable to reproduce the errors at the moment, but if I recall correctly, it's something related to nullpo reference I guess.
## Off-Topic :

It seem like NPC casting area skill is kinda useless... it's only 1 damage.. :-1:

Ancyker:
In regard to the off topic bit, this command can be used with a monsters GID as well as a players GID (account ID/RID).

emistry:
ya, but it become kinda useless if cast by NPC since NPC has no status build. only deal 1 damage.

Kenpachi2k13:
This happens because the NPC's skill units can't be deleted when the NPC is unloaded.
The reason for this is the condition for calling `skill_clear_unitgroup()` in `npc_unload()`:
```
	if (nd->ud != NULL && nd->ud != &npc->base_ud)
		skill->clear_unitgroup(&nd->bl);
```
It's almost impossible to fulfil this condition since `nd->ud` always is `NULL` or `&npc->base_ud` if `unit_bl2ud2()` was never called for the NPC's `block_list` (for example `npcspeed()`).

PR #2712, opened 2020-05-06
Fix condition for deleting skill units when unloading NPC

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

Currently it's almost impossible to fulfil the condition for calling `skill_clear_unitgroup()` in `npc_unload()` since `nd->ud` always is `NULL` or `&npc->base_ud` if `unit_bl2ud2()` was never called for the NPC's block_list (for example by `npcspeed()`).
I changed `skill_initunitgroup()` to call `unit_bl2ud2()` instead of `unit_bl2ud()` if `src` is of type `BL_NPC`.

**Issues addressed:** #768 

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -17766,13 +17766,20 @@ static int skill_get_new_group_id(void)
 
 static struct skill_unit_group *skill_initunitgroup(struct block_list *src, int count, uint16 skill_id, uint16 skill_lv, int unit_id, int limit, int interval)
 {
-	struct unit_data* ud = unit->bl2ud( src );
 	struct skill_unit_group* group;
 	int i;
 
 	if(!(skill_id && skill_lv)) return 0;
 
 	nullpo_retr(NULL, src);
+
+	struct unit_data *ud;
+
+	if (src->type == BL_NPC)
+		ud = unit->bl2ud2(src);
+	else
+		ud = unit->bl2ud(src);
+
 	nullpo_retr(NULL, ud);
 
 	// find a free spot to store the new unit group


src/map/npc.c
```diff
@@ -3026,7 +3026,7 @@ static int npc_unload(struct npc_data *nd, bool single, bool unload_mobs)
 {
 	nullpo_ret(nd);
 
-	if (nd->ud != NULL && nd->ud != &npc->base_ud)
+	if (nd->ud != NULL)
```
4144, line 3029: this mean if unload one npc, it will remove units from all npcs?

4144: probably need call unit->bl2ud2 if adding any unit group (call in skill_initunitgroup)?

Kenpachi2k13: >this mean if unload one npc, it will remove units from all npcs?
No, only from the one that was passed to `npc_unload()`.

> probably need call unit->bl2ud2 if adding any unit group (call in skill_initunitgroup)?
This would fix the bug, too.

4144: if ud == npc->base_ud, this mean unit_data struct is shared between all default looks npcs. and unit_groups look like part or unit_data. this mean if change in base ud, it will affect any npc with same ud. npc id here completely unused

Kenpachi2k13: In `skill_initunitgroup()` the NPC's ID is assigned like this:
`group->src_id      = src->id;` (Where `src` is the NPC's `block_list`.)

MishimaHaruna: While adding a call to `unit->bl2ud2()` in `skill_initunitgroup()` might be worse for performance, it is probably the cleanest approach, since it'd also solve a problem that I can expect to happen right now: if more than `MAX_SKILLUNITGROUP` NPCs in different locations use `unitskillusepos()` in rapid succession, and they end up deleting each others' unit groups because the `ud->skillunit[]` is full.

Kenpachi2k13: Changed. :smile:

Issue #862, opened 2015-11-13
Pre-Renewal: When Talking to NPC


kyeme:
On Ofifcial pre-renewal you can use usable items and self skills (Energy Coat, Sight etc..) while talking to NPC.

dastgir:
Kinda buggy behavior?
I can't think any good reason to allow it...

kyeme:
For example, in a situation that you accidentally clicked a npc and there are mobs, in pre-renewal official you can still use pots

MishimaHaruna:
I agree with kyeme, that's a more than valid reason to allow it (and I don't believe it'd break anything).

AnisotropicDefixation:
On a side note, would it be possible to block interaction with any other NPC as long as you are already in a dialogue or have a buying/selling window/cash shop open? Maybe have it as a .conf option.

AnnieRuru:
https://github.com/HerculesWS/Hercules/blob/master/doc/script_commands.txt#L5042
we already allow to change equipment during npc dialog (allow by default)
so just need to add 'allow to use pots and self buff skills'
or ... maybe a bitmask

Aeromesi:
+1 for this

kyeme:
@Kenpachi2k13

Kenpachi2k13:
Are all usable item types affected by this or only specific ones? (For example only items of type `IT_HEALING`.)

kyeme:
All items including IT_HEALING, IT_USABLE, IT_DELAYCONSUME, IT_CASH are allowed to use.
https://github.com/rathena/rathena/commit/be2428b83200756316adc54bf58b208268ba35c6

Kenpachi2k13:
Okay, I'll have a look at it.

PR #2718, opened 2020-05-10
Add/extend battle flags to enable item/skill usage while interacting with NPC

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

To enable consuming usable items and using skills while interacting with a NPC I did the following:
 * Extend `item_enabled_npc` battle flag with option for usable items.
 * Add `skill_enabled_npc` battle flag.

**Issues addressed:** #862

### DIFF
--- a/conf/map/battle/items.conf
+++ b/conf/map/battle/items.conf
@@ -100,10 +100,14 @@ autospell_stacking: false
 // Default: true (official)
 item_restricted_consumption_type: true
 
-// Enable all NPC to allow changing of equipments while interacting? (Note 1)
-// Script commands 'enable_items/disable_items' will not be override. (see doc/script_commands.txt)
-// Default: true (official)
-item_enabled_npc: true
+// Which item actions are allowed while interacting with NPC? (Note 3)
+// Script commands 'enable_items/disable_items' will not be overridden. (See doc/script_commands.txt.)
+// 0x0 (ITEMENABLEDNPC_NONE)    - Don't allow any item actions.
+// 0x1 (ITEMENABLEDNPC_EQUIP)   - Allow changing equipment.
+// 0x2 (ITEMENABLEDNPC_CONSUME) - Allow consuming usable items.
+// Official RE: 0x1 (Default value.)
+// Official Pre-RE: 0x3
+item_enabled_npc: 0x1
 
 // Unequip the equipments that has disabled by map_zone_db.conf ?
 // 0 : disabled equipments and cards are nullify (official)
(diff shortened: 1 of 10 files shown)


src/map/clif.c
```diff
@@ -12832,7 +12832,12 @@ static void clif_useSkillToIdReal(int fd, struct map_session_data *sd, int skill
 	// Whether skill fails or not is irrelevant, the char ain't idle. [Skotlex]
 	pc->update_idle_time(sd, BCIDLE_USESKILLTOID);
 
-	if (sd->npc_id || sd->state.workinprogress & 1) {
+#ifdef RENEWAL
+	if (sd->npc_id != 0 || (sd->state.workinprogress & 1) != 0) {
+#else
+	if ((sd->npc_id != 0 && ((tmp & INF_SELF_SKILL) == 0 || (skill->get_nk(skill_id) & NK_NO_DAMAGE) == 0))
```
4144, line 12838: what if skill remove some item from inventory? or add new item?

Kenpachi2k13: Ask Gravity? :grin:

I don't know. There may be some rare cases where adding/removing items while interacting with a NPC could cause an error but it's official behaviour for pre-RE and using items is allowed, too...

kyeme:
Yes please make configuration to disable/enable :D

src/map/clif.c
Nurkz: You can add there additional check for item usage script state. Which already exists in some scripts (buildin_disableitemuse)

conf/map/battle/items.conf
```diff
@@ -100,10 +100,14 @@ autospell_stacking: false
...
 item_restricted_consumption_type: true
 
-// Enable all NPC to allow changing of equipments while interacting? (Note 1)
-// Script commands 'enable_items/disable_items' will not be override. (see doc/script_commands.txt)
-// Default: true (official)
-item_enabled_npc: true
+// Which item actions are allowed while interacting with NPC? (Note 3)
+// Script commands 'enable_items/disable_items' will not be overridden. (See doc/script_commands.txt.)
+// 0x0 (ITEMENABLEDNPC_NONE)    - Don't allow any item actions.
+// 0x1 (ITEMENABLEDNPC_EQUIP)   - Allow changing equipment.
```
4144, line 106: probably need add comments about possible exploits for mode 0x1 and 0x2

4144: if item usage/equip in npc scripts need rare, may be better add allow use/equip flags into each item?

Kenpachi2k13: > probably need add comments about possible exploits for mode 0x1 and 0x2

`0x1` is official in pre-RE as well as in RE and is enabled in Hercules for many years by default.
For `0x2` I actually I can't think of any real exploits and aside from that it's official in pre-RE.

> if item usage/equip in npc scripts need rare, may be better add allow use/equip flags into each item?

It's not rarely needed. In official it's enabled for all NPCs.

Issue #718, opened 2015-09-17
Warning


kyeme:
When you're going to PM someone using the PM Box whose name is compose of 24 characters or more, theres a warning msg that will appear on the console.
![error_pm](https://cloud.githubusercontent.com/assets/4656578/9923834/16574d5c-5d2c-11e5-9813-c5f71d42a74b.jpg)
Im using 2015-05-13client

hemagx:
that pretty normal ? the name should never be longer than 23 character

dastgir:
Char name length 23 and a termination char,
Making it 24.
Also I wonder if >24 char name exist, since client would limit off the name.

kyeme:
Opps, I mean any non existing name.
![1](https://cloud.githubusercontent.com/assets/4656578/9924067/5990d442-5d2e-11e5-863b-b93d2e48203e.jpg)

dastgir:
Hmm.. This seems frustrating..
Maybe NULL character is not been added.

kyeme:
~ Up ~

MishimaHaruna:
If the client allows sending names longer than 23 characters (or in other words, can send names without a NULL terminator, then we can remove the warning message.

Kenpachi2k13:
I did a some testing and Gravity's developers really are strange people.

The name field in the character creation windows as well as the name field in the chat box can hold a maximum of 24 characters.
When trying to create a character name with 24 characters you'll get an error message which tells you that you're limited to 23 characters.
But when trying to PM a character name with 24 characters you'll get the "There is no such character name or the user is offline." message.

In my opinion we should simply add a `NULL` terminator ourselves and remove the warning message.

4144:
probably 24 chars works some times, because packets placed in buffer, and in buffer often exists 0 code, this mean almost always 24 chars will not crash server or client, but may crash randomly if here not allowed 24 chars.
need check with disassembler each packet for check is it works with 24 or 23 chars

Kenpachi2k13:
Hmm, we're not supporting names with more than 23 characters in any case.
I don't think we should start making exceptions now just to mimic an obvious bug of Gravity.

MishimaHaruna:
Ok, since the character creation is limited to 23 (and we limit it to 23 nearly everywhere anyway), there's no point in supporting 24 in the whisper packet. Let's enforce a `'\0'` on the 24th character as a safety measure

Kenpachi2k13:
[...]

PR #2721, opened 2020-05-13
Enable receiving not NULL terminated whisper names with a length of 24 characters

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

This adds the benefit of gaining an additional character for NPC names when whispering NPCs and does not break anything...

**Issues addressed:** #718

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -10389,7 +10389,8 @@ static const char *clif_process_chat_message(struct map_session_data *sd, const 
  * @param[in]  sd             The source character.
  * @param[in]  packet         The packet data.
  * @param[out] out_name       The parsed target name buffer (must be a valid
- *                            buffer of size NAME_LENGTH).
+ *                            buffer of size NAME_LENGTH + 1 because the client
+ *                            can send 24 characters without NULL terminator).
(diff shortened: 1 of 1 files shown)


dastgirp:
If I see #718, it was discussed to send NUL at 24th char, but I see that you increased the name length to 25 characters.
There might be some side effects of this, as they won't receive null terminated strings (only 24 characters in packet), and they are known to be bad at this, (like some clients send nul terminated chats, some don't). Might need to check this with old clients as well.

Kenpachi2k13:
I checked this and there are no side effects.
The name received in in `clif_process_whisper_message()` is only used in `clif_parse_WisMessage()` and nowhere else.
Within `clif_parse_WisMessage()` are three different functions called with that name:

- `npc->name2id()`
- `channel->search()`
- `map->nick2sd()`

None of these function will run into problems since the name is `NULL` terminated in any case.
`safestrncpy(out_name, packet->name, NAME_LENGTH + 1);`

The name just might be too long (24 characters), but this is handled everywhere.

4144:
issue with name and without null termination can be if send packet from server to client, but here look like all only inside server

PR #2745, opened 2020-05-27
Prevent unequipping items on login before calculating the character's status data

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

`pc_setpos()` checks for invalid equipment and unequips it if required. On login, the character's status data was not yet calculated at this point and thus the server could crash due to missing memory allocation.
To fix this, the equipment check should be moved to `clif_parse_LoadEndAck()` because it's called after the character's status data was initially calculated in `pc_reg_received()`.

**Issues addressed:** None. Private report via Discord.

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -10850,6 +10850,12 @@ static void clif_parse_LoadEndAck(int fd, struct map_session_data *sd)
 		clif->updatestatus(sd, SP_SKILLPOINT);
 		clif->initialstatus(sd);
 
+		// Unequip items which can't be equipped by the character.
+		for (int i = 0; i < EQI_MAX; i++) {
+			if (sd->equip_index[i] >= 0 && pc->isequip(sd , sd->equip_index[i]) == 0)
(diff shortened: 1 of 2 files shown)


src/map/clif.c
```diff
@@ -10657,6 +10657,12 @@ static void clif_parse_LoadEndAck(int fd, struct map_session_data *sd)
 		clif->updatestatus(sd, SP_CARTINFO);
 	}
 
+	// Unequip items which can't be equipped by the character.
+	for (int i = 0; i < EQI_MAX; i++) {
```
MishimaHaruna, line 10661: Maybe it could be worth running this only on login, rather than on every map change (i.e. `if (sd->state.connect_new != 0)`)

Kenpachi2k13: Woops. My bad. It was intended to add this to the `if (sd->state.connect_new != 0)` branch.
*fixed*

hemagx: Uh no it's intended to happen in every map change, otherwise zone item restrictions won't be checked.

Issue #1762, opened 2017-06-04
Storage and teleporting


Asheraf:
[//]: # (**********************************)
[//]: # (** Fill in the following fields **)
[//]: # (**********************************)

[//]: # (Note: Lines beginning with syntax such as this one, are comments and will not be visible in your report!)

### Issue Prelude

[//]: # (Please complete these mandatory steps and check the following boxes by putting an `x` inside the brackets _before_ filing your issue)

-  [x] I have not modified the source prior to reproducing this issue.
-  [x] I am using the latest version of Hercules.
-  [x] I am aware that this report will be closed or deleted if it becomes obvious that I am stating the false.

### Current Behavior
Using teleportation skill while storage is open wont close it.

[//]: # (Describe at length what you noticed during your analysis.)
[//]: # (If this is a crash, post the core/stack-dump or crash-log to https://gist.github.com/)
[//]: # (If you are referencing from sources such as iROwiki or ratemyserver.net, please quote specific information rather than providing the links alone.)

### Expected Behavior
Test on kRO shows that the storage got closed when the player uses teleportation skill

[//]: # (Tell us what should happen instead.)

### Steps To Reproduce The Issue

1. Open storage.
2. use teleport skill.

**Branch(es):**
- [x] master

```
[Info]: Hercules 64-bit for Linux
[Info]: Git revision (src): '10cb83a30cc50d757e05924fb29c11fc68abda49'
[Info]: Git revision (scripts): '10cb83a30cc50d757e05924fb29c11fc68abda49'
[Info]: OS version: 'Ubuntu 14.04.5 LTS [x86_64]'
[Info]: CPU: 'Intel(R) Core(TM) i5-4200H CPU @ 2.80GHz [4]'
[Info]: Compiled with GCC v4.8.4
[...]

PR #2756, opened 2020-05-31
Storage now closes when using teleport skill.

Jedzkie:
### Pull Request Prelude

### Changes Proposed
Storage now closes when using teleport skill.

**Issues addressed:** 
https://github.com/HerculesWS/Hercules/issues/1762

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -47,6 +47,7 @@
 #include "map/refine.h"
 #include "map/script.h"
 #include "map/status.h"
+#include "map/storage.h"
 #include "map/unit.h"
 #include "common/cbasetypes.h"
 #include "common/ers.h"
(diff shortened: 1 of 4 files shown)


Emistry:
previously they are able to do other action when using storage? 
anyway, maybe apply it for guild storage too ?

Jedzkie:
> previously they are able to do other action when using storage?
> anyway, maybe apply it for guild storage too ?

You can only use self skills, and yeah I forgot on the Guild Storage.

src/map/skill.c
```diff
@@ -11027,17 +11028,18 @@ static int skill_castend_map(struct map_session_data *sd, uint16 skill_id, const
...
-				pc->setpos(sd,sd->status.save_point.map,sd->status.save_point.x,sd->status.save_point.y,CLR_TELEPORT);
-
-			clif->refresh_storagewindow(sd);
+			if (strcmp(mapname, "Random") == 0)
+				pc->randomwarp(sd, CLR_TELEPORT);
+			else if (sd->menuskill_val > 1) // Need lv2 to be able to warp here.
+				pc->setpos(sd, sd->status.save_point.map, sd->status.save_point.x, sd->status.save_point.y, CLR_TELEPORT);
+
+			if (battle_config.teleport_close_storage == 1 && sd->state.storage_flag != STORAGE_FLAG_CLOSED) {
+				clif->storageclose(sd);
```
Kenpachi2k13, line 11039: Use `storage->close(sd);` instead. This will preventively save the storage data, call `clif->storageclose(sd);` and unset `sd->state.storage_flag`.

src/map/skill.c
```diff
@@ -11027,17 +11028,18 @@ static int skill_castend_map(struct map_session_data *sd, uint16 skill_id, const
...
+				sd->state.storage_flag = STORAGE_FLAG_CLOSED;
+			} else
+				clif->refresh_storagewindow(sd);
```
Kenpachi2k13, line 11042: From [style guide](https://github.com/HerculesWS/Hercules/wiki/Coding-Style#chapter-3-placing-braces-and-spaces):

> Do not unnecessarily use braces where a single statement will do.
> This does not apply if only one branch of a conditional statement is a single statement; in the latter case use braces in both branches.

Kenpachi2k13:
> When I looked on the code `gstorage->close` uses `clif->storageclose(sd)` too.

The client does not distinguish between account storage and guild storage. Client-side it's the same thing. That's why `gstorage->close()` uses `clif->storageclose(sd)`, too.

4144:
> > When I looked on the code `gstorage->close` uses `clif->storageclose(sd)` too.
> 
> The client does not distinguish between account storage and guild storage. Client-side it's the same thing. That's why `gstorage->close()` uses `clif->storageclose(sd)`, too.

actually storage and guild storage should use different packets, but for now some packets hercules send same. because this probably better use separate close functions depend on storage type. or one generic function for close any opened storage.

Issue #2751, opened 2020-05-30
Megaphone Behaviors


Jedzkie:
**Describe the bug**
Original issue reported by kyeme - https://herc.ws/oldboard/tracker/issue-7065-megaphone-behavior-minor-bug/?fbclid=IwAR2DLAJB0rjd3dNn9vsJFAMe3ausg3ybY-Zs0B3DjX0-lvYKkuVWQ980wEE

**To Reproduce**
1. When someone shouts and i double clicked his name in General Message it won't automatically register his name in my chat.

**Expected behavior**
1. in Official, if you double click his name in General Message, it will automatically register the name in the chat.

Additional Info:
if the message input is open.
 - You can walk
 - You can use skill
 - You can use item
 - The message should still be broadcast even thought you died. (Here in hercules when you're going to broadcast a message using a megaphone and suddenly you die the message you were going to send is cancelled.)

**Screenshots**
https://youtu.be/5WVNc_haDqo

**System specs (please complete the following information):**
 - OS: Windows 10 18363
 - Hercules Version: v2020.05.03
 - Mode: Pre-Renewal
 - Packet version: 2016-03-16
 - Client type: RE

**Plugins used or source modifications**

**Additional context**

Kenpachi2k13:
@MishimaHaruna 
#2758 fixed only the part where double clicking the shouters name had no effect.

*reopen*

PR #2758, opened 2020-05-31
Add loudhailer() script command

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

Aegis uses the `LoudSpeaker` command for broadcasting megaphone messages.
Namely this is a special use case of packet `0x009a` where the message's first 34 bytes are reserved for string `micc` (4B) which identifies the broadcast as megaphone shout, the character's name (24B) and the text color (6B).
Side note: I had to use "loudhailer" because "megaphone" is already taken by item 7040. :disappointed_relieved:

**Issues addressed:** The "double clickable name" part of #2751

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -12401,6 +12401,50 @@ static BUILDIN(mobattached)
 	return true;
 }
 
+/**
+ * Announces a colored text in '<char_name> Shouts : <message>' format.
+ * Default color is white ("FFFFFF").
+ *
+ * This is a special use case of packet 0x009a where the message's first 34 bytes
+ * are reserved for string "micc" (4B) which identifies the broadcast as megaphone shout,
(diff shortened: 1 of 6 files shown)


src/map/script.c
```diff
@@ -12401,6 +12401,50 @@ static BUILDIN(mobattached)
...
+	const char *color = script_hasdata(st, 3) ? script_getstr(st, 3) : "FFFFFF";
+
+	Assert_retr(false, strlen(color) == 6);
+
+	struct map_session_data *sd = script->rid2sd(st);
+
+	if (sd == NULL)
+		return false;
+
+	char mes_formatted[CHAT_SIZE_MAX + 30] = "";
```
4144, line 12435: here probably better memset whole array to 0

PR #2774, opened 2020-06-13
Fix required item index selection for various skills

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

A user reported via Discord that Slim Potion Pitcher does not work as intended because of an incorrect index selection for required items.
> on skill level 10 item_idx is 9, which cause 100% fail, because max index of requirement item array is 2, and actually skill level 3 works as level 10, with heal part, but without consuming any requirement items

To fix this issue I added `skill_get_item_index()` function.

**Issues addressed:** None. Reported via Discord.

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -7972,7 +7972,14 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 				}
 				if( sd ) {
 					int bonus = 100, potion = min(500+skill_lv,505);
-					int item_idx = (skill_lv - 1) % MAX_SKILL_ITEM_REQUIRE;
+					int item_idx = skill->get_item_index(skill_id, skill_lv);
+
+					if (item_idx == INDEX_NOT_FOUND) {
+						clif->skill_fail(sd, skill_id, USESKILL_FAIL_LEVEL, 0, 0);
+						map->freeblock_unlock();
+						return 1;
+					}
+
 					int item_id = skill->get_itemid(skill_id, item_idx);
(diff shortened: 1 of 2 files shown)


src/map/skill.c
```diff
@@ -369,6 +369,40 @@ static int skill_get_spiritball(int skill_id, int skill_lv)
...
+
+	Assert_retr(INDEX_NOT_FOUND, idx != 0);
+
+	int item_index = INDEX_NOT_FOUND;
+
+	for (int i = 0; i < MAX_SKILL_ITEM_REQUIRE; i++) {
+		if (skill->dbs->db[idx].req_items.item[i].id == 0)
+			continue;
+
+		if (skill->dbs->db[idx].req_items.item[i].amount[skill_get_lvl_idx(skill_lv)] != -1) {
```
4144, line 397: probably call to ``skill_get_lvl_idx`` better put outside of ``for`` and save value to variable. inside ``for`` only read saved variable with idx

Kenpachi2k13: Yeah, good point. Changed...

CordlessThesaurus:
Hi, i think your referring the report i made on discord about the SPP Bug.

CordlessThesaurus:
Thesaurus06/13/2020
Hi anyone here willing to help about Slim Potion Pitcher bug?
I got a strange bug on my server, it can cast a slim potion pitcher but it doesn't heal anything.

So yeah how can i fix the problem. How can i input and start to fix things on the server.

PR #2778, opened 2020-06-15
Extend ci with more checks and add different fixes

### LINKED ISSUE #2776: Skill DB loading errors in script-checker/minimal mode
**Describe the bug**

When running the server in minimal mode (for example through the script-checker script), several errors are printed similar to the following:

```
[Warning]: skill_validate_sp_cost: Invalid SP cost 8 specified in level 1 for skill ID 5 in db/pre-re/skill_db.conf! Minimum is 0, maximum is 0. Defaulting to 0... 
[Warning]: skill_validate_sp_cost: Invalid SP cost 8 specified in level 2 for skill ID 5 in db/pre-re/skill_db.conf! Minimum is 0, maximum is 0. Defaulting to 0... 
[Warning]: skill_validate_sp_cost: Invalid SP cost 8 specified in level 3 for skill ID 5 in db/pre-re/skill_db.conf! Minimum is 0, maximum is 0. Defaulting to 0... 
[Warning]: skill_validate_sp_cost: Invalid SP cost 8 specified in level 4 for skill ID 5 in db/pre-re/skill_db.conf! Minimum is 0, maximum is 0. Defaulting to 0... 
[Warning]: skill_validate_sp_cost: Invalid SP cost 8 specified in level 5 for skill ID 5 in db/pre-re/skill_db.conf! Minimum is 0, maximum is 0. Defaulting to 0... 
```

This is happening because in minimal mode, the battle config isn't loaded, but the skill DB validation code requires it (see battle.c:21918):

```
 if (sp_cost >= 0 && sp_cost <= battle_config.max_sp)
```

A possible fix is to load the battle config in minimal mode (by moving the call to `battle->config_read()` outside the `if (!minimal)` block in `do_init()` of map.c. This needs to be tested as to whether there are other unexpected interactions with the minimal mode.

**To Reproduce**
1. Compile Hercules
2. Run the script `./script-checker npc/dev/test.txt` (or the equivalent command `./map-server --script-checker --load-script npc/dev/test.txt`)
3. See several warnings printed to the output

**Expected behavior**

No errors or warnings are printed

 - OS: Debian GNU/Linux 9.11 (stretch) [x86_64]
 - Hercules Version v2020.05.31+1
 - Mode: both renewal and pre-renewal
 - Packet version: any (default)
 - Client type: n/a, the error is reproducible in the server running in minimal mode

**Plugins used or source modifications**
None

**Additional context**
The error can also be seen on https://haru.ws/scriptchecker/

4144:
### Pull Request Prelude

### Changes Proposed

**Different**
Force use signed chars in platforms with unsigned chars.
Use python2 in validateinterfaces.py for avoid issues with default python3.
Read battle config if map server running in minimal mode.
Remove not existing maps from map index.
Fixes #2776

**Scripts**
Remove not existing maps from scripts.
Fix conflicting npc names in different custom scripts.
Remove duplicated function time2str from bg_common.txt.
Add missing mobs and items for custom or disabled scripts in pre-re.

**Mysql**
Update ci script for adding sql user for support any version of mysql/maridb or broken travis vms.
Add github actions for testing supported external mysql/mariadb servers.

**CI**
Improve travis ci clone speed.
Add arm64 and ppc64le travis builds.
Reduce travis jobs count because travis too slow.
Add into travis.sh check with all custom scripts uncommented.
Add github actions for testing different flags and compilers.

**Plugins**
Fix datacheck array and variable conflicts between server binary and plugins.
Fix double fclose in map cache plugin.
Add all plugins into make file.
Run servers in ci with and without all existing plugins.
Call all plugins events even if server runs in minimal mode.


db/map_index.txt
```diff
@@ -642,8 +642,8 @@ ve_fild04
 ve_fild05
 ve_fild06
 ve_fild07
-poring_c01
-poring_c02
+//poring_c01
+//poring_c02
 que_ng
```
MishimaHaruna, line 647: In order to avoid issues with saved data, indexes of existing maps shouldn't ever change - can you force `que_ng` to keep its old index? (I believe it should be `que_ng 630`, but please double-check on your side)

Kenpachi2k13:
Maybe fix the import string for our script engine test script in this PR, too.

In npc/scripts_dev.conf change this:
`//"npc: npc/dev/test.txt",`
To:
`//"npc/dev/test.txt",`

src/common/atomic.h
```diff
@@ -104,7 +104,7 @@ forceinline volatile int64 InterlockedExchange64(volatile int64 *target, int64 v
 // The __sync functions are available on x86 or ARMv6+
 #if !defined(__x86_64__) && !defined(__i386__) \
 	&& ( !defined(__ARM_ARCH_VERSION__) || __ARM_ARCH_VERSION__ < 6 )
-#error Your Target Platfrom is not supported
+//#error Your Target Platfrom is not supported
```
Kenpachi2k13, line 107: Why keep the check if nothing happens on failure?

4144: technically here should be error, but i not sure conditions what to really check.
current condition show error on new arm arches, and this is wrong.

Kenpachi2k13: Okay, I see.

MishimaHaruna: This is because we don't currently detect ARMv8 - see cbasetypes.h:

```
// Standardize the ARM platform version, if available (the only values we're interested in right now are >= ARMv6)
#if defined(__ARMV6__) || defined(__ARM_ARCH_6__) || defined(__ARM_ARCH_6J__) || defined(__ARM_ARCH_6K__) \
	|| defined(__ARM_ARCH_6Z__) || defined(__ARM_ARCH_6ZK__) || defined(__ARM_ARCH_6T2__) // gcc ARMv6
#define __ARM_ARCH_VERSION__ 6
#elif defined(__ARM_ARCH_7A__) || defined(__ARM_ARCH_7R__) || defined(__ARM_ARCH_7S__) // gcc ARMv7
#define __ARM_ARCH_VERSION__ 7
#elif defined(_M_ARM) // MSVC
#define __ARM_ARCH_VERSION__ _M_ARM
#else
#define __ARM_ARCH_VERSION__ 0
#endif
```

In order to detect it, and if https://opensource.apple.com/source/WTF/WTF-7601.4.2/wtf/Platform.h.auto.html is accurate (it probably is, considering Apple's heavy use of ARM processors) we can probably simply add the blocks equivalent to:

```
#elif defined(__ARM_ARCH_8__)
#define WTF_ARM_ARCH_VERSION 8
```

and

```
/* RVCT sets _TARGET_ARCH_ARM */
#elif defined(__TARGET_ARCH_ARM)
#define WTF_ARM_ARCH_VERSION __TARGET_ARCH_ARM
```

(the last one being optional since we don't officially support RVCT, and I have no ARM machines to test things on)

MishimaHaruna:
I added it myself since I found some more info (including one ARMv7 we were missing - not sure what processors it's actually enabled on though)

PR #2783, opened 2020-06-26
Refactor unit.c - Part II

skyleo:
### Pull Request Prelude

### Changes Proposed

All commits are guaranteed to compile *on their own*. If not, please mention this in the PR.
All commits are guaranteed to be as atomic as possible to make the reviewing as enjoyable and easy as possible.

The following functions have been refactored:
- unit->walktobl
- unit->run_hit
- unit->run
- unit->escape
- unit->movepos
- return value fix in unit->walk_toxy_timer
- unit->blown
- path->blownpos (related to unit->blown change)
- added npc->handle_touch_events function

Changes that were made can be summarized as follows:
- Add proper documentation in doxygen-format.
- Move variables declaration closer to their first use
- Correct mistreatment on checks of non-boolean variables
- Rename some variables to clarify their use
- Save re-used *calculations* in variables or create local macros for them
- Simplify logical checks / conditions when possible
- Change returning error-code in functions to obey code-style guidelines (**Functions that are affected by this are renamed so that custom code fails to compile as to point at that change of behavior.**)
- Split too long lines according to code-style guidelines
- Make functions use enums for directions (enum unit_dir), when dealing with a direction context
- Reduce code-repetition by separating code-chunks into new functions
- Fix remaining code-style after applying all these changes if necessary

Thanks to all the reviewers in advance for looking into my Pull Requests! 

*This is the continuation of https://github.com/HerculesWS/Hercules/pull/2546*

**Issues addressed:** None?

### DIFF
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -625,7 +625,7 @@ static int mob_once_spawn(struct map_session_data *sd, int16 m, int16 x, int16 y
 
 			if (gc != NULL) {
 				struct guild *g = guild->search(gc->guild_id);
-				
+
 				md->guardian_data = (struct guardian_data*)aCalloc(1, sizeof(struct guardian_data));
 				md->guardian_data->castle = gc;
 				md->guardian_data->number = MAX_GUARDIANS;
@@ -1626,7 +1626,7 @@ static int mob_warpchase(struct mob_data *md, struct block_list *target)
 	map->foreachinrange(mob->warpchase_sub, &md->bl,
 	                    md->db->range2, BL_NPC, target, &warp, &distance);
 
-	if (warp && unit->walktobl(&md->bl, &warp->bl, 1, 1))
+	if (warp != NULL && unit->walktobl(&md->bl, &warp->bl, 1, 1) == 0)
 		return 1;
 	return 0;
 }
@@ -1796,7 +1796,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 				return true;/* we are already moving */
 			map->foreachinrange (mob->ai_sub_hard_bg_ally, &md->bl, view_range, BL_PC, md, &tbl, mode);
 			if( tbl ) {
-				if( distance_blxy(&md->bl, tbl->x, tbl->y) <= 3 || unit->walktobl(&md->bl, tbl, 1, 1) )
+				if (distance_blxy(&md->bl, tbl->x, tbl->y) <= 3 || unit->walktobl(&md->bl, tbl, 1, 1) == 0)
 					return true;/* we're moving or close enough don't unlock the target. */
 			}
 		}
@@ -1827,7 +1827,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 			if (!can_move) //Stuck. Wait before walking.
 				return true;
 			md->state.skillstate = MSS_LOOT;
-			if (!unit->walktobl(&md->bl, tbl, 1, 1))
+			if (unit->walktobl(&md->bl, tbl, 1, 1) != 0)
 				mob->unlocktarget(md, tick); //Can't loot...
 			return true;
 		}
@@ -1908,8 +1908,8 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 
 	//Follow up if possible.
 	//Hint: Chase skills are handled in the walktobl routine
-	if(!mob->can_reach(md, tbl, md->min_chase, MSS_RUSH) ||
-		!unit->walktobl(&md->bl, tbl, md->status.rhw.range, 2))
+	if (mob->can_reach(md, tbl, md->min_chase, MSS_RUSH) == 0
+	    || unit->walktobl(&md->bl, tbl, md->status.rhw.range, 2) != 0)
 		mob->unlocktarget(md,tick);
 
 	return true;
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -1704,7 +1704,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 			    )
 			 && md->state.attacked_count++ >= RUDE_ATTACKED_COUNT
 			 && mob->use_skill(md, tick, MSC_RUDEATTACKED) != 0 // If can't rude Attack
-			 && can_move && unit->escape(&md->bl, tbl, rnd()%10 +1) // Attempt escape
+			 && can_move != 0 && unit->escape(&md->bl, tbl, rnd() % 10 + 1) == 0 // Attempt escape
 			) {
 				//Escaped
 				md->attacked_id = 0;
@@ -1732,7 +1732,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 				// Rude attacked
 				if (md->state.attacked_count++ >= RUDE_ATTACKED_COUNT
 				 && mob->use_skill(md, tick, MSC_RUDEATTACKED) != 0 && can_move != 0
-				 && !tbl && unit->escape(&md->bl, abl, rnd()%10 +1)
+				 && tbl == NULL && unit->escape(&md->bl, abl, rnd() % 10 + 1) == 0
 				) {
 					//Escaped.
 					//TODO: Maybe it shouldn't attempt to run if it has another, valid target?
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -1626,7 +1626,7 @@ static int mob_warpchase(struct mob_data *md, struct block_list *target)
 	map->foreachinrange(mob->warpchase_sub, &md->bl,
 	                    md->db->range2, BL_NPC, target, &warp, &distance);
 
-	if (warp != NULL && unit->walktobl(&md->bl, &warp->bl, 1, 1) == 0)
+	if (warp != NULL && unit->walk_tobl(&md->bl, &warp->bl, 1, 1) == 0)
 		return 1;
 	return 0;
 }
@@ -1796,7 +1796,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 				return true;/* we are already moving */
 			map->foreachinrange (mob->ai_sub_hard_bg_ally, &md->bl, view_range, BL_PC, md, &tbl, mode);
 			if( tbl ) {
-				if (distance_blxy(&md->bl, tbl->x, tbl->y) <= 3 || unit->walktobl(&md->bl, tbl, 1, 1) == 0)
+				if (distance_blxy(&md->bl, tbl->x, tbl->y) <= 3 || unit->walk_tobl(&md->bl, tbl, 1, 1) == 0)
 					return true;/* we're moving or close enough don't unlock the target. */
 			}
 		}
@@ -1827,7 +1827,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 			if (!can_move) //Stuck. Wait before walking.
 				return true;
 			md->state.skillstate = MSS_LOOT;
-			if (unit->walktobl(&md->bl, tbl, 1, 1) != 0)
+			if (unit->walk_tobl(&md->bl, tbl, 1, 1) != 0)
 				mob->unlocktarget(md, tick); //Can't loot...
 			return true;
 		}
@@ -1909,7 +1909,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 	//Follow up if possible.
 	//Hint: Chase skills are handled in the walktobl routine
 	if (mob->can_reach(md, tbl, md->min_chase, MSS_RUSH) == 0
-	    || unit->walktobl(&md->bl, tbl, md->status.rhw.range, 2) != 0)
+	    || unit->walk_tobl(&md->bl, tbl, md->status.rhw.range, 2) != 0)
 		mob->unlocktarget(md,tick);
 
 	return true;
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -1704,7 +1704,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 			    )
 			 && md->state.attacked_count++ >= RUDE_ATTACKED_COUNT
 			 && mob->use_skill(md, tick, MSC_RUDEATTACKED) != 0 // If can't rude Attack
-			 && can_move != 0 && unit->escape(&md->bl, tbl, rnd() % 10 + 1) == 0 // Attempt escape
+			 && can_move != 0 && unit->attempt_escape(&md->bl, tbl, rnd() % 10 + 1) == 0 // Attempt escape
 			) {
 				//Escaped
 				md->attacked_id = 0;
@@ -1732,7 +1732,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 				// Rude attacked
 				if (md->state.attacked_count++ >= RUDE_ATTACKED_COUNT
 				 && mob->use_skill(md, tick, MSC_RUDEATTACKED) != 0 && can_move != 0
-				 && tbl == NULL && unit->escape(&md->bl, abl, rnd() % 10 + 1) == 0
+				 && tbl == NULL && unit->attempt_escape(&md->bl, abl, rnd() % 10 + 1) == 0
 				) {
 					//Escaped.
 					//TODO: Maybe it shouldn't attempt to run if it has another, valid target?
--- a/src/map/npc.c
+++ b/src/map/npc.c
@@ -907,6 +907,37 @@ static int npc_event(struct map_session_data *sd, const char *eventname, int ont
 	return npc->event_sub(sd,ev,eventname);
 }
 
+/**
+ * Executes OnTouch, OnUnTouch events when necessary. Also unsets `touching_id` of @p sd if not touching anymore.
+ * @remark @p sd and @p bl is expected to be not `NULL`
+ * @param sd map_session_data of BL_PC to check on
+ * @param x int of x-coordinate to check
+ * @param y int of y-coordinate to check
+ * @param check_if_warped set to true if you want to check if BL_PC has been warped by script
+ * @retval 0 Success
+ * @retval 1 only if check check_if_warped == true and OnTouch might have warped @p bl
+ * @retval -1 @p sd or @p bl is `NULL`
+ */
+static int npc_handle_touch_events(struct map_session_data *sd, int x, int y, bool check_if_warped)
+{
+	nullpo_retr(-1, sd);
+	struct block_list *bl = &sd->bl;
+	nullpo_retr(-1, bl);
+
+	if (sd->touching_id != 0)
+		npc->touchnext_areanpc(sd, false);
+
+	if (map->getcell(bl->m, bl, x, y, CELL_CHKNPC) != 0) {
+		npc->touch_areanpc(sd, bl->m, x, y);
+		if (check_if_warped && bl->prev == NULL) // Script could have warped char
+			return 1;
+	} else {
+		npc->untouch_areanpc(sd, bl->m, x, y);
+	}
+
+	return 0;
+}
+
 /*==========================================
  * Sub chk then execute area event type
  *------------------------------------------*/
@@ -5969,6 +6000,7 @@ void npc_defaults(void)
 	npc->gettimerevent_tick = npc_gettimerevent_tick;
 	npc->settimerevent_tick = npc_settimerevent_tick;
 	npc->event = npc_event;
+	npc->handle_touch_events = npc_handle_touch_events;
 	npc->touch_areanpc_sub = npc_touch_areanpc_sub;
 	npc->touchnext_areanpc = npc_touchnext_areanpc;
 	npc->touch_areanpc = npc_touch_areanpc;
--- a/src/map/path.c
+++ b/src/map/path.c
@@ -24,6 +24,7 @@
 #include "path.h"
 
 #include "map/map.h"
+#include "map/unit.h"
 #include "common/cbasetypes.h"
 #include "common/db.h"
 #include "common/memmgr.h"
@@ -82,10 +83,10 @@ static const unsigned char walk_choices [3][3] =
 };
 
 /*==========================================
- * Find the closest reachable cell, 'count' cells away from (x0,y0) in direction (dx,dy).
+ * Find the closest reachable cell, 'count' cells away from (x0,y0) in direction dir.
  * Income after the coordinates of the blow
  *------------------------------------------*/
-static int path_blownpos(struct block_list *bl, int16 m, int16 x0, int16 y0, int16 dx, int16 dy, int count)
+static int path_blownpos(struct block_list *bl, int16 m, int16 x0, int16 y0, enum unit_dir dir, int count)
 {
 	struct map_data *md;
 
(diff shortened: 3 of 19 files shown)


src/map/unit.c
```diff
@@ -699,7 +699,7 @@ static int unit_walk_tobl(struct block_list *bl, struct block_list *tbl, int ran
 	nullpo_retr(1, bl);
 	nullpo_retr(1, tbl);
 
-	if (!(status_get_mode(bl)&MD_CANMOVE))
+	if ((status_get_mode(bl) & MD_CANMOVE) != 0x0)
```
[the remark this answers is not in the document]
skyleo: 😨good catch

src/map/unit.c
```diff
@@ -827,13 +827,7 @@ static bool unit_run(struct block_list *bl, struct map_session_data *sd, enum sc
 		to_y += dir_y;
 	}
 
-	// Can't run forward
-	if( (to_x == bl->x && to_y == bl->y ) || (to_x == (bl->x+1) || to_y == (bl->y+1)) || (to_x == (bl->x-1) || to_y == (bl->y-1))) {
```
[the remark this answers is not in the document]
skyleo: No. What case do you think is different? I was pretty sure it was going to be equivalent.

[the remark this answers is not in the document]
4144: i not see in other functions same check. may be similar but not same checks present,
probably better left fast fail as is here

src/map/unit.c
```diff
@@ -959,71 +964,60 @@ static enum unit_dir unit_getdir(const struct block_list *bl)
...
+/**
+ * Pushes a unit by given amount of cells into given direction.
+ * @remark Only map cell restrictions are respected.
+ * @param bl: unit to push
+ * @param dir: direction of the push motion
+ * @param count: amount of cells to push
+ * @param no_update: don't send position update packets when true
+ * @return amount of pushed cells.
+ */
+static int unit_blown(struct block_list *bl, enum unit_dir dir, int count, bool no_update)
```
Kenpachi2k13, line 976: Don't use negative names for booleans, That's bad practice.
Just have a look at this one:
`if (!no_update)`

This is read as "If not no update do ..." which is a double negative and doesn't make much sense.

src/map/unit.c
```diff
@@ -959,71 +964,60 @@ static enum unit_dir unit_getdir(const struct block_list *bl)
...
-		sd = BL_CAST(BL_PC, bl);
-		su = BL_CAST(BL_SKILL, bl);
-
-		result = path->blownpos(bl, bl->m, bl->x, bl->y, dx, dy, count);
-
-		nx = result>>16;
-		ny = result&0xffff;
+	if (count == 0)
+		return 0;
+	nullpo_ret(bl);
```
Kenpachi2k13, line 980: Do the `NULL` pointer check first, please.

src/map/unit.c
```diff
@@ -959,71 +964,60 @@ static enum unit_dir unit_getdir(const struct block_list *bl)
...
+	int result = path->blownpos(bl, bl->m, bl->x, bl->y, dir, count);
+	int pushto_x = result >> 16;
+	int pushto_y = result & 0xffff;
 
-		if( sd ) {
-			unit->stop_stepaction(bl); //Stop stepaction when knocked back
-			sd->ud.to_x = nx;
-			sd->ud.to_y = ny;
-		}
+	struct skill_unit* su = BL_CAST(BL_SKILL, bl);
```
Kenpachi2k13, line 986: The asterisk should be adjacent to the field name.

src/map/unit.c
```diff
@@ -959,71 +964,60 @@ static enum unit_dir unit_getdir(const struct block_list *bl)
...
+	if (su == NULL)
+		unit->stop_walking(bl, STOPWALKING_FLAG_NONE);
 
-		dx = nx-bl->x;
-		dy = ny-bl->y;
+	struct map_session_data* sd = BL_CAST(BL_PC, bl);
```
Kenpachi2k13, line 990: The asterisk should be adjacent to the field name.

src/map/unit.c
```diff
@@ -959,71 +964,60 @@ static enum unit_dir unit_getdir(const struct block_list *bl)
...
+	int delta_x = pushto_x - bl->x;
+	int delta_y = pushto_y - bl->y;
 
-			if(su) {
-				skill->unit_move_unit_group(su->group, bl->m, dx, dy);
-			} else {
-				map->moveblock(bl, nx, ny, timer->gettick());
-			}
+	if (delta_x != 0 || delta_y != 0) {
+		enum bl_type receiver_type = bl->type == BL_PC ? BL_ALL : BL_PC;
```
Kenpachi2k13, line 1001: Please enclose the condition block (`bl->type == BL_PC`) in brackets.

src/map/unit.c
```diff
@@ -959,71 +964,60 @@ static enum unit_dir unit_getdir(const struct block_list *bl)
...
-					npc->untouch_areanpc(sd, bl->m, bl->x, bl->y);;
-				}
-			}
-		}
+		if (!no_update)
+			clif->blown(bl);
 
-		count = path->distance(dx, dy);
+		if (sd != NULL)
+			npc->handle_touch_events(sd, bl, bl->x, bl->y);
```
Kenpachi2k13, line 1015: The original code doesn't check for `bl->prev == NULL`. Don't know if simply calling `npc->handle_touch_events()` here will cause unexpected behaviour in case of `bl->prev` being `NULL`.

src/map/unit.c
```diff
@@ -959,71 +964,60 @@ static enum unit_dir unit_getdir(const struct block_list *bl)
...
 	}
 
-	return count;  // return amount of knocked back cells
+	count = path->distance(delta_x, delta_y);
+
+	return count;
```
Kenpachi2k13, line 1020: You could simply do `return path->distance(delta_x, delta_y);` here.

src/map/unit.c
```diff
@@ -959,71 +964,60 @@ static enum unit_dir unit_getdir(const struct block_list *bl)
...
+	if (count == 0)
+		return 0;
+	nullpo_ret(bl);
 
-		if(!su) {
-			unit->stop_walking(bl, STOPWALKING_FLAG_NONE);
-		}
+	int result = path->blownpos(bl, bl->m, bl->x, bl->y, dir, count);
+	int pushto_x = result >> 16;
+	int pushto_y = result & 0xffff;
```
Kenpachi2k13, line 984: These variables are first used inside the `if (sd != NULL)` block, so please move them down there.

src/map/script.c
```diff
@@ -23880,11 +23886,7 @@ static BUILDIN(pushpc)
 		cells = -cells;
 	}
 
-	Assert_retr(false, dir >= UNIT_DIR_FIRST && dir < UNIT_DIR_MAX);
-	dx = dirx[dir];
-	dy = diry[dir];
-
-	unit->blown(&sd->bl, dx, dy, cells, 0);
+	unit->blown(&sd->bl, dir, cells, 0);
```
Kenpachi2k13, line 23889: You've changed the data type of the last parameter of `unit_blown()` to `bool`, so passing 0 is invalid.

src/map/npc.c
```diff
@@ -907,6 +907,34 @@ static int npc_event(struct map_session_data *sd, const char *eventname, int ont
...
+/**
+ * Executes OnTouch, OnUnTouch events when necessary. Also unsets `touching_id` of @p sd if not touching anymore.
+ * @remark @p sd and @p bl is expected to be not `NULL`
+ * @param sd: map_session_data of BL_PC to check on
+ * @param bl: block_list of BL_PC to check on
+ * @param x: short of x-coordinate to check
+ * @param y: short of y-coordinate to check
+ * @retval 0 Success
+ * @retval 1 OnTouch might have warped @p bl
+ * @retval -1 @p sd or @p bl is `NULL`
```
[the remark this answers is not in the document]
skyleo: It's to differentiate between a "tolerated" error and an unwanted error. I guess I should have used nullpo macros in that case though. Will think about it.

src/map/npc.c
```diff
@@ -907,6 +907,34 @@ static int npc_event(struct map_session_data *sd, const char *eventname, int ont
 	return npc->event_sub(sd,ev,eventname);
 }
 
+/**
+ * Executes OnTouch, OnUnTouch events when necessary. Also unsets `touching_id` of @p sd if not touching anymore.
+ * @remark @p sd and @p bl is expected to be not `NULL`
+ * @param sd: map_session_data of BL_PC to check on
+ * @param bl: block_list of BL_PC to check on
+ * @param x: short of x-coordinate to check
+ * @param y: short of y-coordinate to check
```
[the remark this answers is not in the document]
skyleo: I was wondering about that! Thanks for clarifying.

src/map/skill.c
```diff
@@ -10828,11 +10823,11 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 					break;
 				}
 
-				if (unit->movepos(src, bl->x, bl->y, 0, 0)) {
+				if (unit->move_pos(src, bl->x, bl->y, 0, false) == 0) {
 					clif->skill_nodamage(src, src, skill_id, skill_lv, 1);
 					clif->slide(src, bl->x, bl->y) ;
 					sc_start(src, src, SC_CONFUSION, 25, skill_lv, skill->get_time(skill_id, skill_lv));
-					if ( !is_boss(bl) && unit->stop_walking(&sd->bl, STOPWALKING_FLAG_FIXPOS) && unit->movepos(bl, x, y, 0, 0) ) {
+					if (is_boss(bl) == 0x0 && unit->stop_walking(&sd->bl, STOPWALKING_FLAG_FIXPOS) != 0 && unit->move_pos(bl, x, y, 0, false) == 0) {
```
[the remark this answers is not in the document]
skyleo: Will do so

src/map/unit.c
skyleo: Will change to update instead

skyleo: I might have to rename the function in that case to have custom code not compile instead of having unexpected behavior.

src/map/unit.c
skyleo: Woops.

src/map/script.c
skyleo: Good catch!

src/map/mob.c
```diff
@@ -1704,7 +1704,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 			    )
 			 && md->state.attacked_count++ >= RUDE_ATTACKED_COUNT
 			 && mob->skill_use(md, tick, MSC_RUDEATTACKED) == 0 // If can't rude Attack
-			 && can_move && unit->escape(&md->bl, tbl, rnd()%10 +1) // Attempt escape
+			 && can_move != 0 && unit->attempt_escape(&md->bl, tbl, rnd() % 10 + 1) == 0 // Attempt escape
```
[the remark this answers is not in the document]
skyleo: I'd prefer to not do so. 👀

src/map/unit.c
```diff
@@ -682,97 +675,106 @@ static int unit_walktobl_timer(int tid, int64 tick, int id, intptr_t data)
...
+ * @param bl: block_list of unit to move
+ * @param tbl: block_list of target unit to move to
+ * @param range: stop walking once within a distance of @p range with @p tbl
+ * @param flag: flag parameter with following options:
+ *  - `& 1` -> 1/0 = easy / hard
+ *  - `& 2` -> 1 = chase to attack upon arrival within range.
+ * .
+ * @return 0: success, 1: failure
+ */
+static int unit_walk_tobl(struct block_list *bl, struct block_list *tbl, int range, int flag)
```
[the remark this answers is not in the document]
skyleo: Kind of true, but I'm kind of very lazy ._.

src/map/unit.c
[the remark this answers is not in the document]
skyleo: @4144 

Explaining this based on the following commit: https://github.com/skyleo/Hercules/commit/0964a8869a361f24800fec664591eb43bc252d83
Alright so:
**First:**
If `to_x == bl->x && to_y == bl->y` then`i == 0`
so we don't return on `if (i > 1 ...)`

*)
```
	do {
		to_x -= dir_x;
		to_y -= dir_y;
	} while (--i > 0 && unit->walk_toxy(bl, to_x, to_y, 1) != 0);
```
is executed but to_x and to_y are just local variables so we don't care about any changes since `--i > 0` is false in the case of `i == 0`
so
```
	if (i <= 0) {
		unit->run_hit(bl, sc, sd, type);
		return false;
	}
```
is executed fully, which is the old behavior, same order of execution with 2 unnecessary subtractions but that's neglible.

**Other case**
if `(to_x == (bl->x+1) || to_y == (bl->y+1)) || (to_x == (bl->x-1) || to_y == (bl->y-1))`
which is equivalent to `to_x == bl->x + 1 || to_y == bl->y + 1 || to_x == bl->x - 1 || to_y == bl->y - 1`
which then means that since dir_x and dir_y are in {-1, 1, 0}
must mean that to_x and to_y (which initially are set to bl->x and bl->y)
have had exactly 1 addition of dir_x and dir_y
which then means that `i == 1`

so `i > 1` is false and the rest is analog to *)

So as we can see the logic did not change. *q.e.d* :)

I will only not do this if someone can disprove me with a counter-example.

src/map/unit.c
```diff
@@ -763,21 +763,19 @@ static int unit_walk_tobl(struct block_list *bl, struct block_list *tbl, int ran
  */
 static void unit_run_hit(struct block_list *bl, struct status_change *sc, struct map_session_data *sd, enum sc_type type)
 {
-	int lv;
-	struct unit_data *ud;
-
 	Assert_retv(type >= 0 && type < SC_MAX);
-	lv = sc->data[type]->val1;
+
```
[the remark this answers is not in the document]
skyleo: Done in unit->run. But will add for extra safety if people break stuff or misuse this function..

src/map/unit.c
```diff
@@ -796,13 +796,9 @@ static void unit_run_hit(struct block_list *bl, struct status_change *sc, struct
  **/
 static bool unit_run(struct block_list *bl, struct map_session_data *sd, enum sc_type type)
 {
-	struct status_change *sc;
-	short to_x,to_y,dir_x,dir_y;
-	int i;
-
 	nullpo_retr(false, bl);
```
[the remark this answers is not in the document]
skyleo: Added.

src/map/path.c
```diff
@@ -98,11 +99,9 @@ static int path_blownpos(struct block_list *bl, int16 m, int16 x0, int16 y0, int
 		ShowWarning("path_blownpos: count too many %d !\n",count);
 		count=25;
 	}
-	if( dx > 1 || dx < -1 || dy > 1 || dy < -1 ){
-		ShowError("path_blownpos: illegal dx=%d or dy=%d !\n",dx,dy);
-		dx=(dx>0)?1:((dx<0)?-1:0);
-		dy=(dy>0)?1:((dy<0)?-1:0);
-	}
+	Assert_retr((x0 << 16) | y0, dir >= UNIT_DIR_FIRST && dir < UNIT_DIR_MAX);
```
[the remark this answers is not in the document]
skyleo: Not true. If you'd have dx == 0 && dy == 0 it would still return the same, if you choose to use values of |dx| > 1 and |dy| > 1 you can't, it's ensured by using the enum unit_dir that you use a valid direction. The only existing call of path->blownpos still behaves the same, any custom code will need to obey, there's no advantage in using non-standard dx and dy values.

src/map/skill.c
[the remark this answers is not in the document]
skyleo: Fixed

src/map/npc.c
```diff
@@ -907,6 +907,34 @@ static int npc_event(struct map_session_data *sd, const char *eventname, int ont
...
+ * @retval 1 OnTouch might have warped @p bl
+ * @retval -1 @p sd or @p bl is `NULL`
+ */
+static int npc_handle_touch_events(struct map_session_data *sd, struct block_list *bl, short x, short y)
+{
+	nullpo_retr(-1, sd);
+	nullpo_retr(-1, bl);
+	if (sd->touching_id != 0)
+		npc->touchnext_areanpc(sd, false);
+	if (map->getcell(bl->m, bl, x, y, CELL_CHKNPC) != 0) {
```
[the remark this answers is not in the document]
skyleo: Done.

src/map/npc.c
```diff
@@ -907,6 +907,34 @@ static int npc_event(struct map_session_data *sd, const char *eventname, int ont
...
+ * @remark @p sd and @p bl is expected to be not `NULL`
+ * @param sd: map_session_data of BL_PC to check on
+ * @param bl: block_list of BL_PC to check on
+ * @param x: short of x-coordinate to check
+ * @param y: short of y-coordinate to check
+ * @retval 0 Success
+ * @retval 1 OnTouch might have warped @p bl
+ * @retval -1 @p sd or @p bl is `NULL`
+ */
+static int npc_handle_touch_events(struct map_session_data *sd, struct block_list *bl, short x, short y)
```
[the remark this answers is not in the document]
skyleo: Done.

src/map/npc.c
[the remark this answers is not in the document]
skyleo: -1 is for nullpos. Just a nice addition. Especially if one wants to know if the BL_PC has been warped potentially if bl->prev was NULL.

src/map/unit.c
```diff
@@ -407,15 +407,8 @@ static int unit_walk_toxy_timer(int tid, int64 tick, int id, intptr_t data)
...
-			npc->touchnext_areanpc(sd, false);
-		if (map->getcell(bl->m, bl, x, y, CELL_CHKNPC)) {
-			npc->touch_areanpc(sd, bl->m, x, y);
-			if (bl->prev == NULL) //Script could have warped char, abort remaining of the function.
-				return 0;
-		} else {
-			npc->untouch_areanpc(sd, bl->m, x, y);
-		}
+		if (npc->handle_touch_events(sd, bl, x, y) != 0)
+			return 1;
```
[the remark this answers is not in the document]
skyleo: Fixed that.

src/map/unit.c
skyleo: Added flag for npc->handle_touch_events to opt in to that check.

skyleo:
All essential requests have been addressed.

skyleo:
> 
> 
> A general hint:
> You don't have to add newly created functions to the HPMHooking files manually. This will be done automatically when the PR gets merged.
> 
> And a general request regarding your unit code refactoring:
> In unit.h you defined `extern const short dirx[8];` and `extern const short diry[8];`.
> These names were not well chosen because they look like local variables everywhere they are used.
> Furthermore those variables are not part of the unit interface which makes them not usable in plugins.
> Please change this.

I haven't defined extern const short dirx and diry. It was like this before. This may be changed in a future PR.

skyleo:
All essential requests for changes have been addressed.

src/map/unit.c
```diff
@@ -682,97 +675,108 @@ static int unit_walktobl_timer(int tid, int64 tick, int id, intptr_t data)
...
-static int unit_walktobl(struct block_list *bl, struct block_list *tbl, int range, int flag)
+/**
+ * Makes the unit of @p bl walk up to the target @p tbl.
+ * See @p flag for additional actions or options.
+ * @param bl block_list of unit to move
+ * @param tbl block_list of target unit to move to
+ * @param range stop walking once within a distance of @p range with @p tbl
+ * @param flag *flag* parameter with following options:
+ *  - `& 1` -> 1/0 = easy / hard
+ *  - `& 2` -> 1 = chase to attack upon arrival within range.
```
Kenpachi2k13, line 686: Please add `@n` to force line breaks in the created documentation.

```suggestion
 * @param flag *flag* parameter with following options: @n
 *  - `& 1` -> 1/0 = easy / hard @n
 *  - `& 2` -> 1 = chase to attack upon arrival within range.
```

src/map/unit.c
```diff
@@ -682,97 +675,108 @@ static int unit_walktobl_timer(int tid, int64 tick, int id, intptr_t data)
...
+ * .
```
Kenpachi2k13, line 687: Is that dot intended?

src/map/unit.c
```diff
@@ -682,97 +675,108 @@ static int unit_walktobl_timer(int tid, int64 tick, int id, intptr_t data)
 	return 0;
 }
 
-// Chases a tbl. If the flag&1, use hard-path seek,
-// if flag&2, start attacking upon arrival within range, otherwise just walk to that character.
-static int unit_walktobl(struct block_list *bl, struct block_list *tbl, int range, int flag)
+/**
+ * Makes the unit of @p bl walk up to the target @p tbl.
+ * See @p flag for additional actions or options.
+ * @param bl block_list of unit to move
```
Kenpachi2k13, line 681: Please separate the description and the parameter list with an empty line.

src/map/unit.c
```diff
@@ -682,97 +675,108 @@ static int unit_walktobl_timer(int tid, int64 tick, int id, intptr_t data)
...
-	return 0;
+	return 1;
 }
 
 
 /**
- * Called by unit_run when an object was hit
- * @param sd Required only when using SC_WUGDASH
- **/
+ * Called by unit->run when an object was hit.
```
Kenpachi2k13, line 751: Please separate the description and the parameter list with an empty line.

src/map/unit.c
```diff
@@ -781,146 +785,152 @@ static void unit_run_hit(struct block_list *bl, struct status_change *sc, struct
 
 /**
  * Makes character run, used for SC_RUN and SC_WUGDASH
- * @param sd Required only when using SC_WUGDASH
+ * @param bl block_list of the unit that is running
```
Kenpachi2k13, line 788: Please separate the description and the parameter list with an empty line.

src/map/unit.c
```diff
@@ -781,146 +785,152 @@ static void unit_run_hit(struct block_list *bl, struct status_change *sc, struct
...
 
-	return 1;
+	return true;
 }
 
-//Makes bl attempt to run dist cells away from target. Uses hard-paths.
-static int unit_escape(struct block_list *bl, struct block_list *target, short dist)
+/**
+ * Makes @p bl attempt to run at most @p dist cells away from @p target. Uses hard-paths.
+ * @param bl block_list of the unit that is escaping
```
Kenpachi2k13, line 851: Please separate the description and the parameter list with an empty line.

src/map/unit.c
```diff
@@ -781,146 +785,152 @@ static void unit_run_hit(struct block_list *bl, struct status_change *sc, struct
...
+		return 1;
+#undef x_dist
+#undef y_dist
 }
 
-//Instant warp function.
-static int unit_movepos(struct block_list *bl, short dst_x, short dst_y, int easy, bool checkpath)
+/**
+ * _Instantly_ moves / warps @p bl to the coordinates @p dst_x and @p dst_y .
+ * @param bl block_list of unit to move
```
Kenpachi2k13, line 875: Please separate the description and the parameter list with an empty line.

src/map/unit.c
```diff
@@ -959,71 +969,58 @@ static enum unit_dir unit_getdir(const struct block_list *bl)
...
 }
 
-// Pushes a unit by given amount of cells into given direction. Only
-// map cell restrictions are respected.
-// flag:
-//  &1  Do not send position update packets.
-static int unit_blown(struct block_list *bl, int dx, int dy, int count, int flag)
+/**
+ * Pushes a unit by given amount of cells into given direction.
+ * @remark Only map cell restrictions are respected.
```
Kenpachi2k13, line 974: Please separate the description and the parameter list with an empty line.

Kenpachi2k13 (changes requested):

src/map/unit.c
skyleo:
```
You can end a list by starting a new paragraph or by putting a dot (.) on an empty line at the same indentation level as the list you would like to end.
```
- doxygen doc

Yes.

skyleo: Indentation screwed up though.

Kenpachi2k13 (approved):

src/map/unit.c
```diff
@@ -420,10 +420,12 @@ static int unit_walk_toxy_timer(int tid, int64 tick, int id, intptr_t data)
 		// Movement was successful, reset walktoxy_fail_count
 		md->walktoxy_fail_count = 0;
 
-		if (map->getcell(bl->m, bl, x, y, CELL_CHKNPC) != 0 && npc->touch_areanpc2(md))
-			return 0; // Warped
-		else
+		if (map->getcell(bl->m, bl, x, y, CELL_CHKNPC) != 0) {
+			if (npc->touch_areanpc2(md))
```
Kenpachi2k13, line 424:
```suggestion
			if (npc->touch_areanpc2(md) != 0)
```

Kenpachi2k13 (changes requested):

Kenpachi2k13 (approved):

skyleo:
All conflicts with the base branch have been resolved.

MishimaHaruna (approved):

Issue #548, opened 2015-06-10
New item script for reducement based on Monster's element (not elemental attack)


zackdreaver:
Based on : 
http://herc.ws/board/tracker/issue-7769-tatacho-card-effect/

Hercules need new script for this reducement.
List items that affected : 
2684 Ring_Of_Wind
4442 Tatacho Card
4443 Aqua Elemental Card
4444 Draco Card
4445 Luciola Vespa Card
4447 Centipede Card
4448 Cornus Card
4449 Dark Shadow Card
5508 Shark_Hat
5509 Sting_Hat
5593 K_Rabbit_Bonnet
15008 Flame_Sprits_Armor__

AnisotropicDefixation:
Also need AutoSpellOnSkill for offensive heals.

zackdreaver:
@AnisotropicDefixation
You should create new thread for your issue

zackdreaver:
*bumps
@MishimaHaruna @4144 @malufett 
anyone?

iRagno:
This has been already introduced on rAthena:

https://github.com/rathena/rathena/commit/5a4cd7426ee67bc87a845651bedc55bce2ff4487

Would be possible to solve this?

PR #2790, opened 2020-07-02
Add bonuses to reduce received damage based on attacker's defense element

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

- Add `bonus3(bSubDefEle, <element>, <rate>, <flag>)` item bonus to emulate `SubDamage_Property User <element> <rate>` from Aegis.
- Add `bonus3(bMagicSubDefEle, <element>, <rate>, <flag>)` item bonus to emulate `SubMDamage_Property User <element> <rate>` from Aegis.
- Update/add several items and item combos which use those bonuses in RE database. (Pre-RE doesn't use them.)

**Note**:
I tested several skills processed in `battle_calc_misc_attack()` on Aegis and none of them was affected by `SubDamage_Property` or `SubMDamage_Property`. Thus I decided to not add code to the `BF_MISC` case in `battle_calc_cardfix()`.
If someone knows about any "misc" type damage which should be affected by those bonuses please let me know.

**Issues addressed:** #548

### DIFF
--- a/db/re/item_db.conf
+++ b/db/re/item_db.conf
@@ -152580,12 +152580,14 @@ item_db: (
 	AegisName: "S_Basis_Shield"
 	Name: "Basis Shadow Shield"
 	Type: "IT_ARMOR"
-	Buy: 10
+	Buy: 20
 	Loc: "EQP_SHADOW_SHIELD"
 	Script: <"
-		/* Reduces physical and magical damage receive from Neutral property monsters by 1%. */
-		/* Refined to +7, reduces physical and magical damage receive from Neutral property monsters by 1%. */
-		/* Refined to +9, reduces physical and magical damage receive from Neutral property monsters by 1%. */
+		.@refine = getrefine();
+		bonus(bMaxHP, .@refine * 10);
+		.@rate = ((.@refine > 8) ? 3 : ((.@refine > 6) ? 2 : 1));
+		bonus3(bSubDefEle, Ele_Neutral, .@rate, 3);
+		bonus3(bMagicSubDefEle, Ele_Neutral, .@rate, 3);
 	">
 },
 {
@@ -152593,12 +152595,14 @@ item_db: (
 	AegisName: "S_Hallowed_Shield"
 	Name: "Hallowed Shadow Shield"
 	Type: "IT_ARMOR"
-	Buy: 10
+	Buy: 20
 	Loc: "EQP_SHADOW_SHIELD"
 	Script: <"
-		/* Reduces physical and magical damage receive from Shadow property monsters by 1%. */
-		/* Refined to +7, reduces physical and magical damage receive from Shadow property monsters by 1%. */
-		/* Refined to +9, reduces physical and magical damage receive from Shadow property monsters by 1%. */
+		.@refine = getrefine();
+		bonus(bMaxHP, .@refine * 10);
+		.@rate = ((.@refine > 8) ? 3 : ((.@refine > 6) ? 2 : 1));
+		bonus3(bSubDefEle, Ele_Dark, .@rate, 3);
+		bonus3(bMagicSubDefEle, Ele_Dark, .@rate, 3);
 	">
 },
 {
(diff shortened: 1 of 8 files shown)


db/re/item_db.conf
```diff
@@ -135583,8 +135691,8 @@ item_db: (
 	Refine: false
 	ViewSprite: 844
 	Script: <"
-		bonus2 bSubRace,RC_Fish,15;
-		bonus3 bAddMonsterDropItem,579,RC_Fish,5;
+		bonus3(bSubDefEle, Ele_Water, 15, 3);
+		bonus3(bAddMonsterDropItem, 579, RC_Fish, 50);
```
Jedzkie, line 135695: `bonus3(bAddMonsterDropItem, Delicious_Fish, RC_Fish, 50);`

Kenpachi2k13: Good catch, thanks. 👍 
_fixed_

src/map/battle.c
```diff
@@ -1234,6 +1234,20 @@ static int64 battle_calc_cardfix(int attack_type, struct block_list *src, struct
 							}
 							cardfix = cardfix * (100 - ele_fix_lh) / 100;
 						}
+
+						// Apply bSubDefEle damage reduction.
+						switch (src->type) {
+						case BL_MOB:
+							ele_fix = tsd->sub_def_ele[status_get_element(src)].rate_mob;
```
4144, line 1254: better check return value of status_get_element. probably with assert

Kenpachi2k13: Fixed.

src/map/battle.c
```diff
@@ -1070,15 +1070,20 @@ static int64 battle_calc_cardfix(int attack_type, struct block_list *src, struct
...
-					switch (src->type) {
-					case BL_MOB:
-						ele_fix += tsd->magic_sub_def_ele[status_get_element(src)].rate_mob;
-						break;
-					case BL_PC:
-						ele_fix += tsd->magic_sub_def_ele[status_get_element(src)].rate_pc;
-						break;
-					default:
-						break;
+					enum elements def_ele = status_get_element(src);
```
4144, line 1073: if type is not mob or pc calling ``status_get_element`` is useless
probably move ``enum elements def_ele = status_get_element(src);`` under if?

Kenpachi2k13: Fixed.
I'm using `sstatus->def_ele` instead of `status_get_element(src)` now.

4144:
see ci errors
look like old gcc not like comparison. probably cast them to int

db/re/item_db.conf
```diff
@@ -135583,8 +135580,8 @@ item_db: (
 	Refine: false
 	ViewSprite: 844
 	Script: <"
-		bonus2 bSubRace,RC_Fish,15;
-		bonus3 bAddMonsterDropItem,579,RC_Fish,5;
+		bonus3(bSubDefEle, Ele_Water, 15, 3);
+		bonus3(bAddMonsterDropItem, Delicious_Fish, RC_Fish, 50);
```
MishimaHaruna: Just to make sure this isn't an error - is this change from 5 to 50 intentional?

db/re/item_db.conf
```diff
@@ -34324,8 +34322,9 @@ item_db: (
 	Loc: "EQP_GARMENT"
 	EquipLv: 65
 	Script: <"
-		bonus2 bSubEle,Ele_Fire,5;
-		bonus2 bAddRaceTolerance,RC_Demon,5;
+		bonus3(bSubDefEle, Ele_Fire, 5, 3);
+		bonus3(bMagicSubDefEle, Ele_Fire, 5, 3);
+		bonus2(bAddRaceTolerance, RC_Demon, -5);
```
MishimaHaruna, line 34327: Is this change from 5 to -5 intentional?

MishimaHaruna: Ok, this seems correct according to the official item description

db/re/item_db.conf
MishimaHaruna: Ok, this also seems to be correct at 0.5% chance

Issue #2779, opened 2020-06-16
gettimetick(0) returns negative value


bWolfie:
**Describe the bug**
gettimetick(0) should only return positive integers.

**To Reproduce**
Create a script to test values, then run the script.
```
.@tick = gettimetick(0);
consolemes(CONSOLEMES_DEBUG, "tick = %u, tick = %u", gettimetick(0), .@tick);
```

In my example, it produced
```[Jun/16 10:09:42][Debug]: consolemes: tick = 3775957166, tick = 3775950043```

But since these values exceed INT_MAX, they become negative according to the script engine.

**Expected behavior**
We shouldn't allow the value to exceed INT_MAX since script engine doesn't support it.

Emistry:
What should be the value if it capped at `INT_MAX`, always return `INT_MAX`? seem not appropriate.

Perhaps we can consider to trim the millisecond? since these millisecond value are hardly used (never really seen anyone using it). 
That way we could have more buffer to use?

PR #2791, opened 2020-07-03
Prevent overflow when using gettimetick(0)

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

Prevent overflow of script variables (`signed integer`) when using `gettimetick(0)` by doing a bitwise `AND` operation on `timer->gettick()`' and `INT_MAX`.
This decreases the tick loop interval from ~50 to ~25 days, but I think this is the best approach to fix this issue.

**Issues addressed:** #2779

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -3626,7 +3626,7 @@ If the player is not found, returns -1.
 *gettimetick(<type>)
 
 Valid types are :
-	0 - server's tick (milleseconds), unsigned int, loops every ~50 days
+	0 - server's tick (milleseconds), unsigned int, loops every ~25 days
 	1 - time since the start of the current day in seconds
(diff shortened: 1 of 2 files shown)


Lemongrass3110:
This will still leave to wrong condition checks for many user scripts that do checks based on ticks for time difference.

Kenpachi2k13:
@Lemongrass3110 
> This will still lead to wrong condition checks for many user scripts that do checks based on ticks for time difference.

`gettimetick(0)` is supposed to be used where one needs to precisely validate a **short** lapse of time.
So the only problem I can imagine is accidentally having the loop interval being reset between setting the start tick and setting the end tick.
If there's something I don't consider, please tell me or provide an example case. :blush:

Issue #2789, opened 2020-07-01
Improve changesex/changecharsex implementation


MrKeiKun:
**Describe the bug**
Regular changesex thru `@changesex` using 2018 client gives similar result like #985 .

**To Reproduce**
1. Character is currently Job: 4020, which is Clown.
2. Enter `@changesex`
3. Login again

**System specs (please complete the following information):**
 - OS: Centos 8
 - Hercules Version: v2020.05.31+1
 - Mode: Pre-Renewal
 - Packet version: 20180418
 - Client type: RE

MrKeiKun:
If someone might tell me because i used the wrong ACMD, but in the first place whythe existence of two ACMD which does same thing but handle two different methods when emulator can only handles one at the time. Either old one which handle sex via `login` table and the newer one which is under `char` table which requires client >= 20141022

Kenpachi2k13:
Confirmed.
This happens because the SQL query in `char_change_sex_sub()` does not update the gender. (Resetting the robe view sprite ID is missing, too.):
```c
	if (SQL_ERROR == SQL->Query(inter->sql_handle, "UPDATE `%s` SET `class`='%d', `weapon`='0', `shield`='0', "
				"`head_top`='0', `head_mid`='0', `head_bottom`='0' WHERE `char_id`='%d'",
				char_db, class, char_id))
```

`@changecharsex` works because the gender is updated in `char_changecharsex()` prior to calling `char_change_sex_sub()`.
(`@changesex` calls `char_change_sex_sub()` via `char_parse_fromlogin_changesex_reply()`.)

MrKeiKun:
> 
> 
> Confirmed.
> This happens because the SQL query in `char_change_sex_sub()` does not update the gender. (Resetting the robe view sprite ID is missing, too.):
> 
> ```c
> 	if (SQL_ERROR == SQL->Query(inter->sql_handle, "UPDATE `%s` SET `class`='%d', `weapon`='0', `shield`='0', "
> 				"`head_top`='0', `head_mid`='0', `head_bottom`='0' WHERE `char_id`='%d'",
> 				char_db, class, char_id))
> ```
> 
> `@changecharsex` works because the gender is updated in `char_changecharsex()` prior to calling `char_change_sex_sub()`.
> (`@changesex` calls `char_change_sex_sub()` via `char_parse_fromlogin_changesex_reply()`.)

[...]

PR #2796, opened 2020-07-06
Fix @changesex not saving new gender to SQL DB

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

- Adjust the SQL query in `char_change_sex_sub()` to update the character's gender and robe view ID, too.
- Remove unnecessary gender update SQL query from `char_changecharsex()`. Gender gets updated in `char_change_sex_sub()` now, so we can safely skip it in `char_changecharsex()`.
- Adjust `char_change_sex_sub()` to use prepared statements.
- Adjust `char_changecharsex()` to use prepared statement.
- Apply code style to `char_change_sex_sub()`.
- Apply code style to `char_changecharsex()`.

**Issues addressed:** #2789

### DIFF
--- a/src/char/char.c
+++ b/src/char/char.c
@@ -3560,9 +3560,9 @@ static int char_changecharsex(int char_id, int sex)
 	}
 
 	const char *query = "SELECT `account_id`, `class`, `guild_id` FROM `%s` WHERE `char_id`=?";
-	int account_id;
-	int class;
-	int guild_id;
+	int account_id = 0;
+	int class = 0;
+	int guild_id = 0;
 
 	/** Abort changing gender if there was an error while loading the data. **/
 	if (SQL_ERROR == SQL->StmtPrepare(stmt, query, char_db)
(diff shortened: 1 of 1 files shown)


src/char/char.c
```diff
@@ -3562,6 +3560,9 @@ static int char_changecharsex(int char_id, int sex)
 	}
 
 	const char *query = "SELECT `account_id`, `class`, `guild_id` FROM `%s` WHERE `char_id`=?";
+	int account_id;
```
4144, line 3563: probably better left 0 for initial value of vars. it will prevent errors if code bellow will use this vars without checking.

Kenpachi2k13: That's actually impossible. If there's an error while getting the data from the SQL server the function returns.
But I'll re-apply the initial values regardless...

Kenpachi2k13:
> you checked is all works fine for clients with char sex support and without char sex support?

I don't have such an old kRO folder so I can't test it. But I'm pretty sure that defaulting to `U` is right:
```c
#if PACKETVER >= 20141016
	char gender = (sex == SEX_MALE) ? 'M' : ((sex == SEX_FEMALE) ? 'F' : 'U');
#else
	char gender = 'U';
#endif
```

At least @MishimaHaruna defaults to `99` which is `U` in `char_mmo_gender()` , too.(https://github.com/HerculesWS/Hercules/blame/stable/src/char/char.c#L1029)

MishimaHaruna:
Yes, that's correct, `U` (undefined) is the correct gender to use when the client doesn't support per-character gender

PR #2806, opened 2020-07-31
Added item bonus effects of Runes Boots.

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

Based on DP
```
When pure STR is 90 or higher, ATK + 10. When pure STR is 120 or higher, ATK + 10 additionally.
When pure INT is 90 or higher, MATK + 3%. When pure INT is 120 or higher, MATK + 2% additionally.
```

Also weight should be `10` not `100`

**Issues addressed:** 

### DIFF
--- a/db/re/item_db.conf
+++ b/db/re/item_db.conf
@@ -31812,6 +31812,12 @@ item_db: (
 	Def: 5
 	Slots: 1
 	Loc: "EQP_SHOES"
+	Script: <"
+		if (readparam(bStr) >= 90) bonus(bAtk, 10);
+		if (readparam(bStr) >= 120) bonus(bAtk, 20);
+		if (readparam(bInt) >= 90) bonus(bMatkRate, 3);
+		if (readparam(bInt) >= 120) bonus(bMatkRate, 5);
+	">
 },
 {
 	Id: 2482


db/re/item_db.conf
```diff
@@ -31808,10 +31808,16 @@ item_db: (
...
-	Weight: 100
+	Weight: 10
 	Def: 5
 	Slots: 1
 	Loc: "EQP_SHOES"
+	Script: <"
+		if(readparam(bStr) >= 90 && readparam(bStr) <= 119) bonus bAtk, 10;
+		if(readparam(bStr) >= 120) bonus bAtk, 20;
+		if(readparam(bInt) >= 90 && readparam(bInt) <= 119) bonus bMatkRate, 3;
+		if(readparam(bInt) >= 120) bonus bMatkRate, 5;
```
Emistry, line 31819:
```suggestion
		if (readparam(bStr) >= 90) bonus bAtk, 10;
		if (readparam(bStr) >= 120) bonus bAtk, 20;
		if (readparam(bInt) >= 90) bonus bMatkRate, 3;
		if (readparam(bInt) >= 120) bonus bMatkRate, 5;
```
effect are stackable when stats are higher...

db/re/item_db.conf
```diff
@@ -31808,10 +31808,16 @@ item_db: (
 	Name: "Rune Boots"
 	Type: "IT_ARMOR"
 	Buy: 20
-	Weight: 100
+	Weight: 10
```
Emistry, line 31811:
```suggestion
	Weight: 100
```
I believe the weight is correct at `100` which is `10.0`. Your changes was wrong perhaps?

db/re/item_db.conf
MrKeiKun: oh yeah. i forgot that part.

db/re/item_db.conf
MrKeiKun: oh weight is `int`

MishimaHaruna (approved):

Issue #2704, opened 2020-04-27
Attendance System


PixelBr:
The Attendance System is not working as it is on official servers.

The reward is per character, and not on account as it should be, and when the player is eligible to receive the reward he must open the window automatically, which is not happening either.

I performed the test on Kro and iRO.

GilbertoBitt:
the reward system i think can be easily changed to happen on the account and on the windows automatically open I think it is fixable using a simple NPC timer.

PR #2812, opened 2020-08-15
Attendance System

Samuel23:
### Pull Request Prelude

- [√] I have followed [proper Hercules code styling][code].
- [√] I have read and understood the [contribution guidelines][cont] before making this PR.
- [√] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

- Changed attendance system from being character bound to be account …
- Makes the attendance system pop up on login when there's pending reward to be claimed

**Issues addressed:** 

https://github.com/HerculesWS/Hercules/issues/2101
https://github.com/HerculesWS/Hercules/issues/2704


sql-files/upgrades/2020-08-15--13-00.sql
```diff
@@ -0,0 +1,30 @@
...
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+-- Add attendance data to accounta_data table
+ALTER TABLE `account_data` ADD COLUMN `attendance_count` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0' AFTER `base_death`;
+ALTER TABLE `account_data` ADD COLUMN `attendance_timer` BIGINT(20) NULL DEFAULT '0' AFTER `attendance_count`;
+
+-- Delete column for attendance at char table
+ALTER TABLE `char` DROP COLUMN `attendance_count`;
```
MishimaHaruna, line 26: Would it make sense to update the new `account_data` fields with the nonzero values from the `char` table, so that existing data isn't lost?

src/map/clif.c
```diff
@@ -11098,6 +11098,10 @@ static void clif_parse_LoadEndAck(int fd, struct map_session_data *sd)
 		if (battle_config.show_tip_window != 0)
 			clif->open_ui_send(sd, ZC_TIPBOX_UI);
 #endif
+#if PACKETVER_RE_NUM >= 20180307 || PACKETVER_MAIN_NUM >= 20180404 || PACKETVER_ZERO_NUM >= 20180411
+		if ((battle_config.feature_enable_attendance_system != 0) && (clif->attendance_timediff(sd)))
```
Emistry, line 11102:
```suggestion
		if (battle_config.feature_enable_attendance_system != 0 && clif->attendance_timediff(sd) == true)
```
btw is it possible to make it optional to popup on every login if the reward aren't redeemed?
Maybe the user would like to redeem it at a later time and already acknowledged the time is completed?

sql-files/upgrades/2020-08-15--13-00.sql
Samuel23: @MishimaHaruna You mean getting the current values from the char table then insert into the account data table? Hmm, i tried this before making this PR, but I'm getting problem as account_data table not always have values saved in it unless bank or the mod exp was given value, then in char table wherein I should just get the highest value for attendance count and attendance timer should just be copied only to be sent into the account_data table

src/map/clif.c
Samuel23: @Emistry Hmm it just works like the tipbox UI and will just load upon character login, when it popup and you didn't redeem, it will not popup again unless your switch character or relogin the said character.

Emistry: tipsbox are configurable...
https://github.com/HerculesWS/Hercules/commit/ddf679b3808e56309a7352eead2cd1f93c9f99c9#diff-ff11d33dedaf6be65dc557a335aa2d55

Samuel23: @Emistry 
that is configured by the server owner, that could be added, will commit it in a while
Maybe the user would like to redeem it at a later time and already acknowledged the time is completed? - I'm referring to this one as I thought you want it to be a player option

4144 (approved):

sql-files/upgrades/2020-08-15--13-00.sql
MishimaHaruna: Have you tried with a `REPLACE ... SELECT` query?  Something similar to:

```
REPLACE INTO `account_data` (`account_id`, `attendance_count`, `attendance_timer`)
SELECT `account_id`, MAX(`attendance_count`), MAX(`attendance_timer`)
FROM `char`
WHERE `attendance_count` <> 0 AND `attendance_timer` <> 0
GROUP BY `account_id`
```

MishimaHaruna: I updated the migration to include this

MishimaHaruna (approved):

PR #2827, opened 2020-09-02
A few updates to conf folder files

EPuncker:
### Pull Request Prelude

### Changes Proposed
I've updated most of the files from the conf folder, fixed a few typos and enhanced a few setting descriptions


4144 (approved):

conf/common/socket.conf
```diff
@@ -39,7 +39,7 @@ socket_configuration: {
 	//	(Maximum Supported Connections)/2
 	// NOTE: this controls the maximum collected socket-events per-cycle (call to epoll_wait())
 	//       for example settings this to 32 will allow up to 32 events (incoming data/new connections
-	//       per server-cycle.
+	//       per server-cycle).
```
Asheraf, line 42: this bracket should be moved to the line above so it's correct in both
`32 events (incoming data/new connections) per server-cycle.` and `32 events per server-cycle.`

conf/global/console.conf
```diff
@@ -38,7 +38,8 @@ console: {
 	//  %I:%M:%S %p (hour:minute:second 12 hour, AM/PM format)
 	//  %H:%M:%S (hour:minute:second, 24 hour format)
 	//  %d/%b/%Y (day/Month/year)
-	// For full format information, consult the strftime() manual.
+	// For full format information, consult the strftime() manual on
+	// <https://www.php.net/manual/en/function.strftime.php>
```
Asheraf, line 42: `strftime` is not a php function so you should not link to the php version of it, you can use https://man7.org/linux/man-pages/man3/strftime.3.html

Asheraf (changes requested):

conf/map/script.conf
```diff
@@ -51,13 +51,13 @@ script_configuration: {
...
-	// Defaults to 0.
+	// Defaults to 0
 	input_min_value: 0
 
 	// Default value of the 'max' argument of the script command 'input'.
 	// When the 'max' argument isn't provided, this value is used instead.
-	// Defaults to INT_MAX.
-	//input_max_value: 2147483647
+	// Max allowed value that can be used is INT_MAX (2147483647).
+	// Defaults to 10000000
```
MishimaHaruna, line 60: Here it was actually true that the default was `INT_MAX` and not `10000000`. There's a discrepancy between the value we set in the default config file and the default value set in source (which takes effect when you comment it out from the config file).

Please either mention it in the comment here, or change the default in script.c (and in that case, I'll mention it in the release notes)

MishimaHaruna (changes requested):

conf/map/script.conf
EPuncker: done

EPuncker:
I'd be very glad if any of you guys could give me a better description for these two settings:

> dynamic_mobs at conf/map/battle/monster.conf
explain what it does in a few words (right now it doesn't explain anything at all)

> resurrection_exp at conf/map/battle/exp.conf
same as above because right now it is very confusing:

> // The rate of exp. that is gained by the process of resurrection, a unit is 0.01%.
> // Experience calculations for the experience value * level difference of the person revived / 100 * resurrection_exp/10000 which the revived player has can be got.

MishimaHaruna:
Sure!

For dynamic mobs, it could be something along these lines (feel free to reword):

> Whether to spawn and de-spwan mobs dynamically when players are present on the map.
> When enabled, naturally spawning mobs mobs aren't spawned until a player enters the map, and are de-spawned after the last player leaves, according to the mob_remove_damaged and mob_remove_delay settings.
> When disabled, all naturally spawning mobs are spawned and kept in memory for as long as the server is running.

For resurrection exp (which is a setting that likely no one will miss if we remove it, as abusable as it is):

> The rate of exp gained when revived by another player (Resurrection, Redemptio, Yggdrasil Leaf), in 0.01% increments - 100is 1%, 10000 is 100%.
> Base and job experience are only gained if the resurrected player's base and job level, respectively, are higher than the resurrecting player.
> The amount of gained experience is calculated as current experience * level difference / 100 * resurrection_exp / 10000

MishimaHaruna (approved):

PR #2837, opened 2020-09-18
Add configure option for set max connections count

4144:

### Changes Proposed
Use configure flag --with-maxconn=VALUE for set maximum of supported connections.


configure
```diff
@@ -1407,7 +1407,7 @@ Optional Packages:
   --with-key3[=ARG]       Set the third obfuscation key (ignored unless the
                           other two are also specified)
   --with-maxconn[=ARG]    optionally set the maximum connections the core can
-                          handle (default: 16384) NOT USED YET - EXPERIMENTAL
+                          handle (default: 1024)
```
MishimaHaruna, line 1410: Shouldn't this say that the default is 3072 if epoll is enabled and FD_SETSIZE (usually 1024) otherwise?

MishimaHaruna (changes requested):

configure
4144: changed comment, and rebased

MishimaHaruna (approved):

Issue #2848, opened 2020-09-24
battleground type 2 score board has gone missing


AnnieRuru:
**Describe the bug**
where the hell did the score board on battleground go ?

**To Reproduce**
`prontera	mapflag	battleground	2`

**Expected behavior**
I thought my memory wrong, or somebody play a mind trick on me
I load **rathena test server** to check my memory is still correct
and its there

**Screenshots**
![](https://raw.githubusercontent.com/AnnieRuru/customs/master/screenshots/screen2020rAthena001.jpg)

**System specs (please complete the following information):**
 - OS: Windows 10 pro 64-bit
 - Hercules Version: latest (always pull latest git)
 - Packet version: 20180620
 - Mode: Renewal
 - Client type: [2018-06-20eRagexeRE](https://rathena.org/board/topic/117168-packetver-20180620-client-release-2018-06-20eragexere/)
 - Branch: Main

**Plugins used or source modifications**
doesn't matter

**Additional context**
1. it seems battleground script isn't popular in hercules, otherwise why nobody even report this issue ?
2. another side effect of hercules don't accept custom scripts, rathena has my bg_pvp/bg_emp script template and people can report if its broken or something ...

AnnieRuru:
confirm now

```
guild_vs2	mapflag	battleground	2

prontera,155,185,5	script	ksdhfksjd	1_F_MARIA,{
	bg_updatescore "guild_vs2", 99, 99;
	if (!.asdf)
		.asdf = bg_create_team("guild_vs2", 50,50);
	if (!getcharid(CHAR_ID_BG))
		bg_join_team .asdf, getcharid(CHAR_ID_ACCOUNT);
	bg_warp .asdf, "guild_vs2", 50,50;
	end;
OnInit:
	.@c = getunits(BL_PC, .@u, 0);
	for (.@i = 0; .@i < .@c; ++.@i) {
		attachrid .@u[.@i];
		bg_leave;
	}
}
```
1. login 2 characters into prontera
2. use player A to click this npc, this player will warp to guild_vs2 without seeing the score board
3. use player B to click the npc from prontera, players B doesn't see the score board, but the score board suddenly appear on player A

I repeat the process and there ALWAYS ONE PLAYER couldn't see the score board, and its always the last player warp into the map couldn't see it

PR #2849, opened 2020-09-25
fix missing scoreboard for battleground type 2

AnnieRuru:
### Pull Request Prelude

### Issues addressed
https://github.com/HerculesWS/Hercules/issues/2848

### Changes Proposed
cap the value as 1 or 2 only
if it is 0, it becomes 1(default), and if it is over 2, becomes 2

```
guild_vs2	mapflag	battleground
```
tested is 1, because the sscanf return 0 didn't found w4

### Affected Branches
* Master

blame https://github.com/HerculesWS/Hercules/pull/2654 ... yeah its @Emistry again

### DIFF
--- a/src/map/npc.c
+++ b/src/map/npc.c
@@ -5028,7 +5028,7 @@ static const char *npc_parse_mapflag(const char *w1, const char *w2, const char 
 		struct map_zone_data *zone;
 		if (state != 0) {
 			if (w4 != NULL && sscanf(w4, "%d", &state) == 1)
-				map->list[m].flag.battleground = (state != 0) ? 1 : 0;
+				map->list[m].flag.battleground = cap_value(state, 0, 2);
 			else
 				map->list[m].flag.battleground = 1; // Default value
 		} else {


src/map/npc.c
```diff
@@ -5028,7 +5028,7 @@ static const char *npc_parse_mapflag(const char *w1, const char *w2, const char
 		struct map_zone_data *zone;
 		if (state != 0) {
 			if (w4 != NULL && sscanf(w4, "%d", &state) == 1)
-				map->list[m].flag.battleground = (state != 0) ? 1 : 0;
+				map->list[m].flag.battleground = cap_value(state, 1, 2);
```
4144, line 5031: may be better cap from 0? ``cap_value(state, 0, 2)``
because with your code if put parameter as 0, it will be used as 1, i think this is wrong.

4144 (changes requested):

src/map/npc.c
AnnieRuru:
```
	else if (strcmpi(w3, "battleground") == 0) {
		struct map_zone_data *zone;
		if (state != 0) {
			if (w4 != NULL && sscanf(w4, "%d", &state) == 1)
				map->list[m].flag.battleground = (state != 0) ? 1 : 0;
			else
				map->list[m].flag.battleground = 1; // Default value
		} else {
			map->list[m].flag.battleground = 0;
		}
```
wait, you mean we can use
```
prontera	mapflag	battleground	0
--same as--
//prontera	mapflag	battleground // not declare it
```
to turn off the mapflag ?? then what's the point of adding a mapflag line then ?

4144: yes i mean this, but i not saw check before. it blocks ``0`` already.
this can be need if don't want remove map flag but want disable it.

hm, but may be add value validation check? and if it not valid, show error message?
because old and new code silently accept any values...

AnnieRuru: I really don't see the point of using value 0 
there are already *removemapflag script command that can remove a mapflag

just changing the value 1 into 0 doesn't mean anything, but doing it doesn't cost me my arm or leg, so I'll do it

AnnieRuru:
just remember after doing this change, now will create inconsistent
```
prontera	mapflag	battleground	0
```
which doesn't do anything ...... !!!

while the *setmapflag script command can only set value 1 or 2

Kenpachi2k13 (approved):

MishimaHaruna (approved):

Issue #2815, opened 2020-08-19
Soul Link Dancer


Samuel23:
**Describe the bug**
Dancer with soul link doesn't get the bard skills counterpart

**To Reproduce**
1. Make a Dancer with all skills (tried even not using @skillall, manual leveling and skill adding)
2. Soul Link Buff for Bard/Dancer
3. Check Skill tree
4. No additional skills added

**Expected behavior**
Dancer with soul link should get the bard skills counterpart

**Screenshots**
https://imgur.com/v6RzuUF

https://imgur.com/OYKbaWB

https://imgur.com/pIz7A4Q

https://imgur.com/Z0gqqpB

**System specs (please complete the following information):**
 - OS: [windows 10]
 - Hercules Version [Release v2020.07.26]
 - Mode: [renewal or pre-renewal?] - tried both
 - Packet version: [e.g. 20190322]
 - Client type: [Main]

**Plugins used or source modifications**
No source modifications
Tried with some of my made plugins
Tried as well to disable all plugins

**Additional context**
No crash error, no console error

AnnieRuru:
confirm, someone on discord also tell me this issue

AnnieRuru:
blame https://github.com/HerculesWS/Hercules/commit/bfa423fea55aa5659506cf85e6c418346804ac89
@Samuel23 that's wrong fix, in fact haru also wrong, why bard copy the correct skill index but dancer did the other way round ?

```
	pc->calc_skilltree_clear(sd);
ShowDebug ("SkillActualID : %d %d %d %d\n", BA_WHISTLE, BA_APPLEIDUN, DC_HUMMING, DC_SERVICEFORYOU);
ShowDebug ("Skillindex : %d %d %d %d\n", skill->get_index(BA_WHISTLE), skill->get_index(BA_APPLEIDUN), skill->get_index(DC_HUMMING), skill->get_index(DC_SERVICEFORYOU));
	for (int i = 0; i < MAX_SKILL_DB; i++) {
		if (sd->status.skill[i].flag == SKILL_FLAG_TEMPORARY || sd->status.skill[i].flag >= SKILL_FLAG_REPLACED_LV_0) {
			// Restore original level of skills after deleting earned skills.
			sd->status.skill[i].lv = (sd->status.skill[i].flag == SKILL_FLAG_TEMPORARY) ? 0 : sd->status.skill[i].flag - SKILL_FLAG_REPLACED_LV_0;
[...]

PR #2852, opened 2020-09-29
fix dancer with soul link didn't get the bard skills counterpart

AnnieRuru:
### Pull Request Prelude

### Issues addressed
https://github.com/HerculesWS/Hercules/issues/2815

### Changes Proposed
the actual reason why dancer didn't get the skill is because the dancer's skill index is lower than bard's
when it continue on the loop, it hit the condition above and set skill level into 0

so move the dancer/bard soul link check out of the loop

### Affected Branches
* Master

### DIFF
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -1624,35 +1624,10 @@ static int pc_calc_skilltree(struct map_session_data *sd)
 			sd->status.skill[i].lv = (sd->status.skill[i].flag == SKILL_FLAG_TEMPORARY) ? 0 : sd->status.skill[i].flag - SKILL_FLAG_REPLACED_LV_0;
 			sd->status.skill[i].flag = SKILL_FLAG_PERMANENT;
 		}
-
-		if (sd->sc.count && sd->sc.data[SC_SOULLINK] && sd->sc.data[SC_SOULLINK]->val2 == SL_BARDDANCER
-		 && ((skill->dbs->db[i].nameid >= BA_WHISTLE && skill->dbs->db[i].nameid <= BA_APPLEIDUN)
-		  || (skill->dbs->db[i].nameid >= DC_HUMMING && skill->dbs->db[i].nameid <= DC_SERVICEFORYOU))
-		) {
-			//Enable Bard/Dancer spirit linked skills.
-			int linked_nameid = skill->get_linked_song_dance_id(skill->dbs->db[i].nameid);
-			if (linked_nameid == 0) {
-				Assert_report("Linked bard/dance skill not found");
-				continue;
-			}
-			int copy_from_index;
-			int copy_to_index;
-			if (sd->status.sex == SEX_MALE && skill->dbs->db[i].nameid >= BA_WHISTLE && skill->dbs->db[i].nameid <= BA_APPLEIDUN) {
-				copy_from_index = i;
-				copy_to_index = skill->get_index(linked_nameid);
-			} else {
-				copy_from_index = skill->get_index(linked_nameid);
-				copy_to_index = i;
-			}
-			if (copy_from_index < copy_to_index)
-				continue; // Copy only after the source skill has been filled into the tree
-			if (sd->status.skill[copy_from_index].lv < 10)
-				continue; // Copy only if the linked skill has been mastered
-			sd->status.skill[copy_to_index].id = skill->dbs->db[copy_to_index].nameid;
-			sd->status.skill[copy_to_index].lv = sd->status.skill[copy_from_index].lv; // Set the level to the same as the linking skill
-			sd->status.skill[copy_to_index].flag = SKILL_FLAG_TEMPORARY; // Tag it as a non-savable, non-uppable, bonus skill
-		}
 	}
+
+	//Enable Bard/Dancer spirit linked skills.
+	skill->add_bard_dancer_soullink_songs(sd);
 
 	if( pc_has_permission(sd, PC_PERM_ALL_SKILL) ) {
 		for (int i = 0; i < MAX_SKILL_DB; i++) {
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -20613,6 +20613,48 @@ static bool skill_parse_row_changematerialdb(char *split[], int columns, int cur
 }
 
 /**
+ * Enable Bard/Dancer to learn songs from their counterpart when spirit link
+ * * @param *sd    the actual player gain the skills
+ */
+static void skill_add_bard_dancer_soullink_songs(struct map_session_data *sd)
+{
+	nullpo_retv(sd);
+	
+	if (sd->sc.count == 0 || sd->sc.data[SC_SOULLINK] == NULL || sd->sc.data[SC_SOULLINK]->val2 != SL_BARDDANCER)
+		return;
+	
+	const int bard_song_skillid[4] = {BA_WHISTLE, BA_ASSASSINCROSS, BA_POEMBRAGI, BA_APPLEIDUN};
+	const int dancer_song_skillid[4] = {DC_HUMMING, DC_DONTFORGETME, DC_FORTUNEKISS, DC_SERVICEFORYOU};
+	
+	STATIC_ASSERT(ARRAYLENGTH(bard_song_skillid) == ARRAYLENGTH(dancer_song_skillid), "bard_song_skillid and dancer_song_skillid must be the same size");
+
+	int copy_from_index;
+	int copy_to_index;
+	for (int i = 0; i < ARRAYLENGTH(bard_song_skillid); i++) {
+		if (sd->status.sex == SEX_MALE) {
+			copy_from_index = skill->get_index(bard_song_skillid[i]);
+			copy_to_index = skill->get_index(dancer_song_skillid[i]);
+		} else {
+			copy_from_index = skill->get_index(dancer_song_skillid[i]);
+			copy_to_index = skill->get_index(bard_song_skillid[i]);
+		}
+
+		if (copy_from_index == 0 || copy_to_index == 0) {
+			Assert_report("Linked bard/dancer skill not found");
+			continue;
+		}
+
+		if (sd->status.skill[copy_from_index].lv < 10) // Copy only if the linked skill has been mastered
+			continue;
+
+		sd->status.skill[copy_to_index].id = skill->dbs->db[copy_to_index].nameid;
+		sd->status.skill[copy_to_index].lv = sd->status.skill[copy_from_index].lv; // Set the level to the same as the linking skill
+		sd->status.skill[copy_to_index].flag = SKILL_FLAG_TEMPORARY; // Tag it as a non-savable, non-uppable, bonus skill
+	}
+	return;
+}
+
+/**
  * Sets Level based configuration for skill groups from skill_db.conf [ Smokexyz/Hercules ]
  * @param *conf    pointer to config setting.
  * @param *arr     pointer to array to be set.
@@ -24285,4 +24327,5 @@ void skill_defaults(void)
 	skill->check_npc_chaospanic = skill_check_npc_chaospanic;
 	skill->count_wos = skill_count_wos;
 	skill->get_linked_song_dance_id = skill_get_linked_song_dance_id;
+	skill->add_bard_dancer_soullink_songs = skill_add_bard_dancer_soullink_songs;
 }
--- a/src/map/skill.h
+++ b/src/map/skill.h
@@ -2292,6 +2292,7 @@ struct skill_interface {
 	int (*check_npc_chaospanic) (struct block_list *bl, va_list args);
 	int (*count_wos) (struct block_list *bl, va_list ap);
 	int (*get_linked_song_dance_id) (int skill_id);
+	void (*add_bard_dancer_soullink_songs) (struct map_session_data *sd);
 };
 
 #ifdef HERCULES_CORE


skyleo:
Reword commit message to `Fix Dancer with Soul Link not getting the Bard skills counterpart`

src/map/pc.c
```diff
@@ -1624,33 +1624,30 @@ static int pc_calc_skilltree(struct map_session_data *sd)
...
-		  || (skill->dbs->db[i].nameid >= DC_HUMMING && skill->dbs->db[i].nameid <= DC_SERVICEFORYOU))
-		) {
-			//Enable Bard/Dancer spirit linked skills.
-			int linked_nameid = skill->get_linked_song_dance_id(skill->dbs->db[i].nameid);
-			if (linked_nameid == 0) {
-				Assert_report("Linked bard/dance skill not found");
-				continue;
+	//Enable Bard/Dancer spirit linked skills.
+	if (sd->sc.count && sd->sc.data[SC_SOULLINK] && sd->sc.data[SC_SOULLINK]->val2 == SL_BARDDANCER) {
+		for (int i = 0; i < MAX_SKILL_DB; i++) {
```
skyleo, line 1631: You definitely shouldn't be looping over all skills again. Why was it necessary to move this out of the upper loop? Instead, can't a change of order help here if the loops first if statement is the initial problem?

skyleo (changes requested):

src/map/pc.c
AnnieRuru: >You definitely shouldn't be looping over all skills again

that's because hercules has compress the skill ID into skill index, which Ind made to save server memory
https://herc.ws/board/topic/512-skill-id-processing-overhaul/
if let's say for example, I add a custom skill with ID ...
```
static const struct {
	int start;
	int end;
} skill_idx_ranges[] = {
	{ NV_BASIC, NPC_LEX_AETERNA },
// somewhere here
	{ KN_CHARGEATK, SA_ELEMENTWIND },
	{ RK_ENCHANTBLADE, AB_SILENTIUM },
	{ WL_WHITEIMPRISON, SC_FEINTBOMB },
	{ LG_CANNONSPEAR, SR_GENTLETOUCH_REVITALIZE },
	{ WA_SWING_DANCE, WA_MOONLIT_SERENADE },
	{ MI_RUSH_WINDMILL, MI_HARMONIZE },
	{ WM_LESSON, WM_UNLIMITED_HUMMING_VOICE },
	{ SO_FIREWALK, SO_EARTH_INSIGNIA },
	{ GN_TRAINING_SWORD, GN_SLINGITEM_RANGEMELEEATK },
	{ AB_SECRAMENT, LG_OVERBRAND_PLUSATK },
	{ ALL_ODINS_RECALL, ALL_LIGHTGUARD },
	{ RL_GLITTERING_GREED, RL_GLITTERING_GREED_ATK },
	{ KO_YAMIKUMO, OB_AKAITSUKI },
	{ ECL_SNOWFLIP, ALL_THANATOS_RECALL },
	{ GC_DARKCROW, NC_MAGMA_ERUPTION_DOTDAMAGE },
	{ SU_BASIC_SKILL, SU_SPIRITOFSEA },
	{ HLIF_HEAL, MH_VOLCANIC_ASH },
	{ MS_BASH, MER_INVINCIBLEOFF2 },
	{ EL_CIRCLE_OF_FIRE, EL_STONE_RAIN },
	{ GD_APPROVAL, GD_DEVELOPMENT },
```
so please **DO NOT assume the skill index will stay the same for all server**
some server has already made their own custom skill ID

I think that's the reason why Haru made that commit in the first place https://github.com/HerculesWS/Hercules/commit/bfa423fea55aa5659506cf85e6c418346804ac89

------------------------------------------------

> Instead, can't a change of order help here if the loops first if statement is the initial problem?

with clean hercules with no custom skills, bard skill index is higher than dancer

do you have any better idea how to do it then ?

AnnieRuru: EDIT -> I think I get it now
use ARRAY
ok I make another one

AnnieRuru:
no more ugly loop,
with this method, we don't even need the function `skill->get_linked_song_dance_id`

src/map/pc.c
```diff
@@ -1624,33 +1624,30 @@ static int pc_calc_skilltree(struct map_session_data *sd)
...
+		if (sd->status.sex == SEX_MALE) {
+			for (int i = 0; i < 4; i++) {
+				if (sd->status.skill[bard_song_skillidx[i]].lv >= 10) { // Copy only if the linked skill has been mastered
+					sd->status.skill[dancer_song_skillidx[i]].id = skill->dbs->db[dancer_song_skillidx[i]].nameid;
+					sd->status.skill[dancer_song_skillidx[i]].lv = sd->status.skill[bard_song_skillidx[i]].lv; // Set the level to the same as the linking skill
+					sd->status.skill[dancer_song_skillidx[i]].flag = SKILL_FLAG_TEMPORARY; // Tag it as a non-savable, non-uppable, bonus skill
+				}
+			}
+		}
+		else {
```
skyleo, line 1643: put the else on the same line as the }

src/map/pc.c
```diff
@@ -1624,33 +1624,30 @@ static int pc_calc_skilltree(struct map_session_data *sd)
...
-		  || (skill->dbs->db[i].nameid >= DC_HUMMING && skill->dbs->db[i].nameid <= DC_SERVICEFORYOU))
-		) {
-			//Enable Bard/Dancer spirit linked skills.
-			int linked_nameid = skill->get_linked_song_dance_id(skill->dbs->db[i].nameid);
-			if (linked_nameid == 0) {
-				Assert_report("Linked bard/dance skill not found");
-				continue;
+	//Enable Bard/Dancer spirit linked skills.
+	if (sd->sc.count && sd->sc.data[SC_SOULLINK] && sd->sc.data[SC_SOULLINK]->val2 == SL_BARDDANCER) {
+		int bard_song_skillidx[4] = {skill->get_index(BA_WHISTLE), skill->get_index(BA_ASSASSINCROSS), skill->get_index(BA_POEMBRAGI), skill->get_index(BA_APPLEIDUN)};
```
skyleo, line 1631: Split line, or more all skill->get_index to the outside and do them once u need the index.
Add const qualifier to the array.

You could add sth like this if you rename the the arrays accordingly:
```c
		const int *from_nameid = NULL;
		const int *to_nameid = NULL;
		if (sd->status.sex == SEX_MALE) {
			from_nameid = bard_skills;
			to_nameid = dancer_skills;
		} else {
			from_nameid = dancer_skills;
			to_nameid = bard_skills;
		}
```

Also add this
```c
STATIC_ASSERT(ARRAYLENGTH(bard_skills) == ARRAYLENGTH(dancer_skills), "bard_skills and dancer_skills must be the same size");
```

Then you can loop over ARRAYLENGHT(bard_skills)
and inside the loop you use that if that you have accordingly but before you do sth like this:
```c
			int copy_from_index = skill->get_index(from_nameid[i]);
			int copy_to_index = skill->get_index(to_nameid[i]);
			if (copy_from_index == 0 || copy_to_index == 0) {
				Assert_report("Linked bard/dancer skill not found");
				continue;
			}
```
Which is also nice because you add more safety in case a skill->get_index() fails.

Credits to this solution go to Haru btw.

skyleo (changes requested):
Generally I suggest moving all the Soul Link bard-dancer logic to a separate function so that pc->calc_skilltree is easier to read and the code becomes easier to maintain as well.
with that I mean moving the `if (sd->sc.count && sd->sc.data[SC_SOULLINK] && sd->sc.data[SC_SOULLINK]->val2 == SL_BARDDANCER) {`
and all its content to its own function.

src/map/pc.c
skyleo: nice solution, good thinking.

src/map/skill.h
```diff
@@ -2292,6 +2292,8 @@ struct skill_interface {
 	int (*check_npc_chaospanic) (struct block_list *bl, va_list args);
 	int (*count_wos) (struct block_list *bl, va_list ap);
 	int (*get_linked_song_dance_id) (int skill_id);
+	void (*bard_dancer_song_soullink) (struct map_session_data *sd);
+
```
skyleo, line 2296: remove added empty line.

src/map/skill.c
```diff
@@ -20612,6 +20612,50 @@ static bool skill_parse_row_changematerialdb(char *split[], int columns, int cur
 	return true;
 }
 
+/**
+ * Enable Bard/Dancer to learn songs from their counterpart when spirit link
+ */
+static void skill_bard_dancer_song_soullink(struct map_session_data *sd)
```
skyleo, line 20618: name should start with an imperative verb. Rename your function.

skyleo: It should say briefly what it's doing basically.

src/map/skill.c
```diff
@@ -20612,6 +20612,50 @@ static bool skill_parse_row_changematerialdb(char *split[], int columns, int cur
...
+	
+	int bard_song_skillid[4] = {BA_WHISTLE, BA_ASSASSINCROSS, BA_POEMBRAGI, BA_APPLEIDUN};
+	int dancer_song_skillid[4] = {DC_HUMMING, DC_DONTFORGETME, DC_FORTUNEKISS, DC_SERVICEFORYOU};
+	
+	STATIC_ASSERT(ARRAYLENGTH(bard_song_skillid) == ARRAYLENGTH(dancer_song_skillid), "bard_song_skillid and dancer_song_skillid must be the same size");
+
+	int copy_from_index;
+	int copy_to_index;
+	for (int i = 0; i < 4; i++) {
+		copy_from_index = 0;
```
skyleo, line 20633: why are you setting it to 0 when you'll always overwrite it after that? This is unnecessary.

src/map/skill.c
```diff
@@ -20612,6 +20612,50 @@ static bool skill_parse_row_changematerialdb(char *split[], int columns, int cur
...
+		copy_to_index = 0;
```
skyleo, line 20634: why are you setting it to 0 when you'll always overwrite it after that? This is unnecessary.

src/map/skill.c
```diff
@@ -20612,6 +20612,50 @@ static bool skill_parse_row_changematerialdb(char *split[], int columns, int cur
...
+		return;
+	
+	int bard_song_skillid[4] = {BA_WHISTLE, BA_ASSASSINCROSS, BA_POEMBRAGI, BA_APPLEIDUN};
+	int dancer_song_skillid[4] = {DC_HUMMING, DC_DONTFORGETME, DC_FORTUNEKISS, DC_SERVICEFORYOU};
+	
+	STATIC_ASSERT(ARRAYLENGTH(bard_song_skillid) == ARRAYLENGTH(dancer_song_skillid), "bard_song_skillid and dancer_song_skillid must be the same size");
+
+	int copy_from_index;
+	int copy_to_index;
+	for (int i = 0; i < 4; i++) {
```
skyleo, line 20632: use ARRAYLENGTH(bard_song_skillid) instead of 4 here.

skyleo (changes requested):
Looks good, but some changes will need to be done.

src/map/skill.c
AnnieRuru: yeah, just double check the `skill_get_index` function
it does return 0 if skill ID not found, so set to 0 is unnecessary

src/map/skill.c
```diff
@@ -20612,6 +20612,47 @@ static bool skill_parse_row_changematerialdb(char *split[], int columns, int cur
 	return true;
 }
 
+/**
+ * Enable Bard/Dancer to learn songs from their counterpart when spirit link
+ */
```
skyleo, line 20617: add `@param sd description` accordingly.

src/map/skill.c
```diff
@@ -20612,6 +20612,47 @@ static bool skill_parse_row_changematerialdb(char *split[], int columns, int cur
...
+static void skill_add_bard_dancer_soullink_songs(struct map_session_data *sd)
+{
+	nullpo_retv(sd);
+	
+	if (sd->sc.count == 0 || sd->sc.data[SC_SOULLINK] == NULL || sd->sc.data[SC_SOULLINK]->val2 != SL_BARDDANCER)
+		return;
+	
+	int bard_song_skillid[4] = {BA_WHISTLE, BA_ASSASSINCROSS, BA_POEMBRAGI, BA_APPLEIDUN};
```
skyleo, line 20625: change to `const int[4]`

src/map/skill.c
```diff
@@ -20612,6 +20612,47 @@ static bool skill_parse_row_changematerialdb(char *split[], int columns, int cur
...
+	int dancer_song_skillid[4] = {DC_HUMMING, DC_DONTFORGETME, DC_FORTUNEKISS, DC_SERVICEFORYOU};
```
skyleo, line 20626: change to `const int[4]`

src/map/skill.c
```diff
@@ -20612,6 +20612,47 @@ static bool skill_parse_row_changematerialdb(char *split[], int columns, int cur
...
+	
+	STATIC_ASSERT(ARRAYLENGTH(bard_song_skillid) == ARRAYLENGTH(dancer_song_skillid), "bard_song_skillid and dancer_song_skillid must be the same size");
+
+	int copy_from_index;
```
skyleo, line 20630: change to `const int`

src/map/skill.c
```diff
@@ -20612,6 +20612,47 @@ static bool skill_parse_row_changematerialdb(char *split[], int columns, int cur
...
+	int copy_to_index;
```
skyleo, line 20631: change to `const int`

skyleo (commented):
Looking good, just these 4 changes that aren't really required but are nice.

src/map/skill.c
skyleo: strike that.

src/map/skill.c
skyleo: strike that

MishimaHaruna (approved):

Issue #2853, opened 2020-09-29
the reason why all pull request return [X] <-- built failed


AnnieRuru:
```
	CC	itemdb.c
	CC	log.c
	CC	mail.c
	CC	map.c
	CC	mapreg_sql.c
	CC	mercenary.c
	CC	mob.c
map.c: In function ‘map_cell2gat’:
map.c:3010:12: note: parameter passing for argument of type ‘struct mapcell’ changed in GCC 9.1
 3010 | static int map_cell2gat(struct mapcell cell)
      |            ^~~~~~~~~~~~
	CC	npc.c
	CC	npc_chat.c
	CC	party.c
	CC	path.c
	CC	pc.c
	CC	pc_groups.c
	CC	pet.c
	CC	quest.c
	CC	refine.c
	CC	rodex.c
	CC	script.c
In file included from HPMHooking.c:234:
HPMHooking/HPMHooking_map.Hooks.inc: In function ‘HP_map_cell2gat’:
HPMHooking/HPMHooking_map.Hooks.inc:49114:5: note: parameter passing for argument of type ‘struct mapcell’ changed in GCC 9.1
49114 | int HP_map_cell2gat(struct mapcell cell) {
      |     ^~~~~~~~~~~~~~~
	CC	searchstore.c
	CC	skill.c
	CC	status.c
	CC	storage.c
	CC	stylist.c
	CC	trade.c
	CC	unit.c
```
someone needs to look into this, I'm not familiar how travis works

AnnieRuru:
https://github.com/HerculesWS/Hercules/blob/1354b01c7a24e9ffd9ea570cbaec668721fc2a6c/src/map/map.c#L3010-L3019
wait, isn't the struct should have asterisk on it ?

```
static int map_cell2gat(struct mapcell *cell)
```

4144:
where you get this warning? ci not reporting it

AnnieRuru:
1. click pull request section, a lot of them built failed
![](https://raw.githubusercontent.com/AnnieRuru/customs/master/screenshots/travis-ci_gcc%201.png)

2. check my PR, I saw the Travis CI said built failed
![](https://raw.githubusercontent.com/AnnieRuru/customs/master/screenshots/travis-ci_gcc%202.png)

3. click on the details, one of them said built failed
![](https://raw.githubusercontent.com/AnnieRuru/customs/master/screenshots/travis-ci_gcc%203.png)

4. scroll down and I found this
[...]

PR #2854, opened 2020-09-30
TESTING ONLY !! DO NOT MERGE ... testing hercules travis-ci

AnnieRuru:
Issue https://github.com/HerculesWS/Hercules/issues/2853

 try remove the map_cell2get function and see if this PR return built success

if this actually return success, then this function needs a rewrite

### DIFF
--- a/src/map/map.c
+++ b/src/map/map.c
@@ -3007,17 +3007,6 @@ static struct mapcell map_gat2cell(int gat)
 	return cell;
 }
 
-static int map_cell2gat(struct mapcell cell)
-{
-	if( cell.walkable == 1 && cell.shootable == 1 && cell.water == 0 ) return 0;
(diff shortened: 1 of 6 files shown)


AnnieRuru:
yeah seriously travis works at turtle speed ....

4144:
what the sense to change this function?
you afraid huge memory copy on each call? it's not, because struct less than int size.
compiler note? as i explained before it don't make sense to build hercules with one compiler and plugins with another to get run time bug.

AnnieRuru:
> it don't make sense to build hercules with one compiler and plugins with another to get run time bug.

then there is a bug in travis =/

4144:
i wrote in other pr about.
this is not issue with this method. issue in timeout. because compiled in ARM arch, it much slower than amd64, and travis allow run compilation for fixed time only for 1h. if runs longer it show timeout.
most time this issue can be fixed by pressing ``retry`` on build with timeout.

Issue #2834, opened 2020-09-15
Possible bug at MD_LOOTER check in status.c file


EPuncker:
**Describe the bug**
monster_loot_type setting not working

**To Reproduce**
1. Go to your battle/monster.conf file and change **monster_loot_type** setting to 1
2. Create a few porings or any other monster that "eat" items from the ground
3. Throw any item in the ground
4. See error, poring will consume it regardless of this setting being set to 0 or 1

**Expected behavior**
Monster shouldn't loot the item regardless of his MD_LOOTER mode when setting is set to 1

**System specs:**
 - OS: Windows 10
 - Hercules Version latest
 - Mode: pre-renewal
 - Packet version: 20190322
 - Client type: main

skyleo:
Will fix, this is broken since 2004 btw.

PR #2402, opened 2019-03-10
Fix @bodystyle crash client

AnnieRuru:
[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style
### Pull Request Prelude
- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR will be closed if the above-mentioned criteria are not fulfilled.

### 2 Issues addressed
1. https://github.com/HerculesWS/Hercules/issues/2383#issuecomment-467607801
2. if the player on 3rd job, having alternate bodystyle, and `@jobchange` into novice for example,will crash the client.
This is extremely annoying when we testing stuffs

### Changes Proposed
1. Apply temporary fix so Job_Super_Novice_E no longer crash client 
2. Fix atcommand @jobchange when the player changing from 3rd job into non-3rd job, crashing client

### Affected Branches
* Master

PR #2855, opened 2020-10-01
Fix incorrect looting behavior with monster_loot_type battle config

skyleo:
### Pull Request Prelude

### Changes Proposed

Fix battle config variable monster_loot_type not working as intended

When it was set to 1, monsters would still loot, even though they
aren't supposed to.
When it was set to 0 and the monster couldn't loot anymore, monsters
would still try to loot, even though they should avoid doing so.

**Issues addressed:** #2834

### DIFF
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -625,7 +625,7 @@ static int mob_once_spawn(struct map_session_data *sd, int16 m, int16 x, int16 y
 
 			if (gc != NULL) {
 				struct guild *g = guild->search(gc->guild_id);
-				
+
 				md->guardian_data = (struct guardian_data*)aCalloc(1, sizeof(struct guardian_data));
 				md->guardian_data->castle = gc;
 				md->guardian_data->number = MAX_GUARDIANS;
@@ -1771,9 +1771,8 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 		return true;
 
 	// Scan area for targets
-	if (!tbl && mode&MD_LOOTER && md->lootitem && DIFF_TICK(tick, md->ud.canact_tick) > 0
-	 && (md->lootitem_count < LOOTITEM_SIZE || battle_config.monster_loot_type != 1)
-	) {
+	if (battle_config.monster_loot_type != 1 && tbl == NULL && (mode & MD_LOOTER) != 0x0 && md->lootitem != NULL
+	    && DIFF_TICK(tick, md->ud.canact_tick) > 0 && md->lootitem_count < LOOTITEM_SIZE) {
 		// Scan area for items to loot, avoid trying to loot if the mob is full and can't consume the items.
 		map->foreachinrange (mob->ai_sub_hard_lootsearch, &md->bl, view_range, BL_ITEM, md, &tbl);
 	}


EPuncker (approved):
tested and working
2020-10-01 11_41_43-PerflexRO v2

src/map/mob.c
```diff
@@ -1771,9 +1771,8 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 		return true;
 
 	// Scan area for targets
-	if (!tbl && mode&MD_LOOTER && md->lootitem && DIFF_TICK(tick, md->ud.canact_tick) > 0
-	 && (md->lootitem_count < LOOTITEM_SIZE || battle_config.monster_loot_type != 1)
-	) {
+	if (battle_config.monster_loot_type != 1 && tbl == NULL && (mode & MD_LOOTER) != 0x0 && md->lootitem != NULL
```
4144, line 1774: current option (if set to 0) allow loot items if mob already looted max allowed items. With value 1, looting can be if configured in mob db but only for max looting size.

Changed option, allow looting only if option is 0 and mob configured as looter. is this correct?
may be better change option to flag values? one bit for allow loot more than max value, other flag for disable looting at all, and may be other flags if you want something else?

4144 (changes requested):

src/map/mob.c
AnnieRuru: although I also support the name change for the battle_config,
but this should be an on/off behavior

currently 0 means yes (will loot) and 1 means no (will not loot)
which I support the name change, but set to the number of loots ?
I think currently max loot carry limit is 10 
-> #define LOOTITEM_SIZE 10 <- why this is set inside mapdefines.h ??
but poring will still continue to eats everything on the floor even if it already has loot more than 10

AnnieRuru (changes requested):
4144 is right.
Although currently this patch actually solve the bug,
but it doesn't make sense for that name inside battle_config

since this thing has been broken since the early days,
1. we can deprecate this `monster_loot_type` name in battle_config
and make a new one that will make sense in name
eg: monster_can_loot ... ?

2. or just silently switch 1 and 0 on the battle_config side ?

skyleo:
Can you guys open a separate issue for the suggestions you had.
This pull request merely focuses on fixing the buggy behavior.
Changing this battle configuration would be done in separate commits anyways, if one was to follow proper coding standards.

Please merge this finally ...

src/map/mob.c
skyleo: see https://github.com/HerculesWS/Hercules/pull/2855#issuecomment-739331348

AnnieRuru:
> Can you guys open a separate issue for the suggestions you had.

unfortunately, this PR exposed the code that doesn't comply with Hercules code styling
however if you add a TODO or FIXME, then yeah, will reminds others to add a follow up

https://github.com/HerculesWS/Hercules/pull/2402#issuecomment-471363064

skyleo:
> 
> 
> > Can you guys open a separate issue for the suggestions you had.
> 
> unfortunately, this PR exposed the code that doesn't comply with Hercules code styling
> however if you add a TODO or FIXME, then yeah, will reminds others to add a follow up
> 
> [#2402 (comment)](https://github.com/HerculesWS/Hercules/pull/2402#issuecomment-471363064)

90% of Hercules doesn't follow our Coding Style.
I'd rather not add a TODO or FIXME in the code directly, an issue should suffice.

dastgirp:
> > 
> > 
> > > Can you guys open a separate issue for the suggestions you had.
> > 
> > unfortunately, this PR exposed the code that doesn't comply with Hercules code styling
> > however if you add a TODO or FIXME, then yeah, will reminds others to add a follow up
> > 
> > [#2402 (comment)](https://github.com/HerculesWS/Hercules/pull/2402#issuecomment-471363064)
> 
> 90% of Hercules doesn't follow our Coding Style.
> I'd rather not add a TODO or FIXME in the code directly, an issue should suffice.

Though I haven't checked what line doesn't follow coding style in this PR.
But in general, if line changes in any PR, then that changed lines should follow the code style.

@Kenpachi2k13 might be better aware.

skyleo:
> 
> 
> > > > Can you guys open a separate issue for the suggestions you had.
> > > 
> > > 
> > > unfortunately, this PR exposed the code that doesn't comply with Hercules code styling
> > > however if you add a TODO or FIXME, then yeah, will reminds others to add a follow up
> > > [#2402 (comment)](https://github.com/HerculesWS/Hercules/pull/2402#issuecomment-471363064)
> > 
> > 
> > 90% of Hercules doesn't follow our Coding Style.
> > I'd rather not add a TODO or FIXME in the code directly, an issue should suffice.
> 
> Though I haven't checked what line doesn't follow coding style in this PR.
> But in general, if line changes in any PR, then that changed lines should follow the code style.
> 
> @Kenpachi2k13 might be better aware.

every line changed follows the Coding Style.

What they mean is that battle_config.monster_loot_type has  a misleading name because it only uses 0 and 1. They think that means it has a boolean meaning. I think the person that added this considered adding more options/types, which is pretty much why the order of it doesn't matter.
Their suggestion is of entire aesthetic nature and I don't see how it's related to coding style.

AnnieRuru:
what I'm trying to say is, if adding a fixme like example below, this can get accepted and can remind to do a follow up
```c
	// Scan area for targets. FIXME: battle_config.monster_loot_type has misleading naming behavior (PR#2383)
	if (battle_config.monster_loot_type != 1 && tbl == NULL && (mode & MD_LOOTER) != 0x0 && md->lootitem != NULL
	    && DIFF_TICK(tick, md->ud.canact_tick) > 0 && md->lootitem_count < LOOTITEM_SIZE) {
```

yes any line that has changed must follow coding style,
if just want a quick fix to be merge into github can use a TODO or FIXME

skyleo:
> 
> 
> what I'm trying to say is, if adding a fixme like example below, this can get accepted and can remind to do a follow up
> 
> ```c
> 	// Scan area for targets. FIXME: battle_config.monster_loot_type has misleading naming behavior (PR#2855)
> 	if (battle_config.monster_loot_type != 1 && tbl == NULL && (mode & MD_LOOTER) != 0x0 && md->lootitem != NULL
> 	    && DIFF_TICK(tick, md->ud.canact_tick) > 0 && md->lootitem_count < LOOTITEM_SIZE) {
> ```
> 
> yes any line that has changed must follow coding style,
> if just want a quick fix to be merge into github can use a TODO or FIXME



> 
> 
> what I'm trying to say is, if adding a fixme like example below, this can get accepted and can remind to do a follow up
> 
> ```c
> 	// Scan area for targets. FIXME: battle_config.monster_loot_type has misleading naming behavior (PR#2855)
> 	if (battle_config.monster_loot_type != 1 && tbl == NULL && (mode & MD_LOOTER) != 0x0 && md->lootitem != NULL
> 	    && DIFF_TICK(tick, md->ud.canact_tick) > 0 && md->lootitem_count < LOOTITEM_SIZE) {
> ```
> 
> yes any line that has changed must follow coding style,
> if just want a quick fix to be merge into github can use a TODO or FIXME

According to this : https://github.com/HerculesWS/Hercules/pull/2855#issuecomment-739811466
It's not a misleading name though.

MishimaHaruna (approved):

Issue #2824, opened 2020-08-30
Error on skill Abracadrabra/Hocus-Pocus (SA_ABRACADABRA)


haarenas:
**Describe the bug**
When you use the skill (that is not learned by the sage) and try to select a target or ground (depending on the randomized spell) the skill generated isn't casted (It doesnt generate any error on console log)

**To Reproduce**
1. Login with a Sage Class.
2. Use the skill Abracadabra.
3. Try execute a skill that you dont have learned and it will fail.

**Expected behavior**
It should be cast any of the randomized spells.

**System specs (please complete the following information):**
 - OS: Windows 10 Pro 64bit, version 1903.
 - Hercules Version: v2020.08.23
 - Mode: Renewal
 - Packet version: 20180620
 - Client type: RE

**Plugins used or source modifications**
No modifications.

**Aditional Note**
Possibly the origin of problem is the same as https://github.com/HerculesWS/Hercules/issues/2823

Issue #2823, opened 2020-08-29
Error on skill Improvised Song (WM_RANDOMIZESPELL)


haarenas:
**Describe the bug**
When you use the skill and try to select a target or ground (depending on the randomized spell) the spell isn't casted (It doesnt generate any error on console log)

**To Reproduce**
1. Login with a Ministrel/Wanderer.
2. Use the skill Improvised Song.
3. Try execute, but it will fail.

**Expected behavior**
It should be cast any of the randomized spells.

**System specs (please complete the following information):**
 - OS: Windows 10 Pro 64bit, version 1903.
 - Hercules Version: v2020.08.23
 - Mode: Renewal
 - Packet version: 20180620
 - Client type: RE

**Plugins used or source modifications**
No modifications.

PR #2859, opened 2020-10-02
Fixes Abracadabra and Improvised Skill

dastgirp:
### Pull Request Prelude
The Autocast by abra was cleared before the actual casting, which resulted in failure of skill if not learned.

### Changes Proposed

Adding ret = 0 check while casting skill, to clear the autocast data

**Issues addressed:** 
* [x] #2823
* [x] #2824

### DIFF
--- a/src/map/unit.c
+++ b/src/map/unit.c
@@ -1155,7 +1155,7 @@ static int unit_skilluse_id(struct block_list *src, int target_id, uint16 skill_
 	int ret = unit->skilluse_id2(src, target_id, skill_id, skill_lv, casttime, castcancel);
 	struct map_session_data *sd = BL_CAST(BL_PC, src);
 
-	if (sd != NULL)
+	if (sd != NULL && ret == 0)
 		pc->autocast_remove(sd, sd->auto_cast_current.type, sd->auto_cast_current.skill_id,
 				    sd->auto_cast_current.skill_lv);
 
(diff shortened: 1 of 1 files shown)


src/map/unit.c
```diff
@@ -1155,7 +1155,7 @@ static int unit_skilluse_id(struct block_list *src, int target_id, uint16 skill_
 	int ret = unit->skilluse_id2(src, target_id, skill_id, skill_lv, casttime, castcancel);
 	struct map_session_data *sd = BL_CAST(BL_PC, src);
 
-	if (sd != NULL)
+	if (sd != NULL && ret == 0)
```
4144, line 1158: this mean after this change auto cast never cleaned if skill was used without errors?
i think this wrong

Kenpachi2k13: `unit->skilluse_id2()` returns `0` on failure, so this is okay.

Maybe the
```c
else if (sd != NULL && ret != 0 && skill_id != SA_ABRACADABRA && skill_id != WM_RANDOMIZESPELL)
		skill->validate_autocast_data(sd, skill_id, skill_lv);
```
block I removed in #2699 [(LINK)](https://github.com/HerculesWS/Hercules/pull/2699/files?file-filters%5B%5D=.c&file-filters%5B%5D=.inc#diff-fc6e84e111da95e50d06e579cf274053) should be re-added, too.
I currently can't have a closer look at this, so this is just a suggestion...

dastgirp: @4144 the problem is, this function is triggered before actual casting of skill, so removing this check caused autocast struct to be emptied, and skill failed to cast.

Kenpachi2k13 (approved):

AnnieRuru (approved):

src/map/unit.c
skyleo: @dastgirp This definitely needs to be adressed what 4144 mentioned.

Kenpachi2k13: I just had a look at this and the condition should be changed to:
```c
if (sd != NULL && (ret == 0 || (skill_id != SA_ABRACADABRA && skill_id != WM_RANDOMIZESPELL)))
```

MishimaHaruna:
I'm closing this in favor of the more complete fix (extensively tested on a production server) from #2972

PR #2865, opened 2020-10-05
Changed the default value of max_body_style

EPuncker:
### Pull Request Prelude

### Changes Proposed
Changed the default value of max_body_style to 1 since there is only one body style AFAIK, otherwise it will cause a gravity error if someone set a value higher than 1 without having the proper sprites

### DIFF
--- a/conf/map/battle/client.conf
+++ b/conf/map/battle/client.conf
@@ -44,13 +44,13 @@ min_chat_delay: 0
 
 // Valid range of dyes and styles on the client.
 min_hair_style: 0
-max_hair_style: 29
+max_hair_style: 31
 min_hair_color: 0
 max_hair_color: 8
 min_cloth_color: 0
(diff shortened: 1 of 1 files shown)


EPuncker:
it actually has 4 body style, but the 2 to 4 are only mado gear

4144:
probably better add per job body style limits?
because some jobs may have sprites for 4 body styles and some other only 2 or less.

AnnieRuru:
I think the limit should be on the stylist script and atcommand `@bodystyle` instead

EPuncker:
> probably better add per job body style limits?
> because some jobs may have sprites for 4 body styles and some other only 2 or less.

yes I was also thinking about it, but then I remembered we also need a separated hair style/dyes to doram class... or should we handle this on stylist npc only? 🤷

PR #2867, opened 2020-10-06
Update PULL_REQUEST_TEMPLATE.md

EPuncker:
Just a small update to the PR template file because it appears that github have changed the text, so I changed it to match it to prevent any confusion.

2020-10-06 02_39_09-Comparing HerculesWS_stable EPuncker_issues2 · HerculesWS_Hercules


AnnieRuru:
why allow access to secrets ??
you want to know my birthday ?

I don't see the point of this PR, and this is Haru's only

PR #2868, opened 2020-10-06
Add define with current hercules version

4144:
### Pull Request Prelude

### Changes Proposed

Add new define HERCULES_VERSION with current hercules version.
It can be used in plugins for define different features based on version.


AnnieRuru:
I NEED A F**KING SCRIPT CONSTANT FOR IT !!!!!!!!

just add a constant for it in script.c and output it in constant.md .........

I remember I talked to Haru about this and suggest using a bot to automatically update the date whenever someone merge a PR
and Haru dislike the idea having a bot necro-post whenever someone merge PR

but now .... there is an unspoken rule that only Haru can merge commit,
and Haru only shows up once a month,
this idea is actually viable

4144:
this is not merge date. but release date. it changed once per month with new release only

src/map/script.c
```diff
@@ -27946,6 +27946,7 @@ static void script_hardcoded_constants(void)
 
 	script->constdb_comment("Server defines");
 	script->set_constant("PACKETVER",PACKETVER,false, false);
+	script->set_constant("HERCULES_VERSION",HERCULES_VERSION,false, false);
```
AnnieRuru, line 27949: https://github.com/HerculesWS/Hercules/blob/1354b01c7a24e9ffd9ea570cbaec668721fc2a6c/src/map/script.c#L27947-L27965
please do white space edit :P

see a few lines down there, got space after comma, here don't have

AnnieRuru (changes requested):

Asheraf:
A script constant for this is useless, the engine will throw an error if a command is not called properly anyway.. you cant do
```
if (HERCULES_VERSION >= 20201230) {
	command(arg1);
} else {
	command(arg1, arg2);
}
```

AnnieRuru:
> A script constant for this is useless

how can this be useless ? I have been getting restless with a lot of members complain my script is bug
when it was their fault for using outdated emulator
```
if (HERCULES_VERSION <= 20201230)
consolemes CONSOLEMES_ERROR, "This script require HERCULES_VERSION over 20201230 to run";
```
something like this can stop a lot of ~this shit~ the complains that I had

Asheraf:
The example you provided is just "yet another warning", it doesn't make the script compatible with multiple versions of the emulator, if someone uses a script that requires a new command with an older version of hercules an error stating that `foo` command does not exist is already shown.

MrKeiKun:
If this is release date, i think reading the release name from a file(like `VERSION` file name) would be nice.

4144:
> If this is release date, i think reading the release name from a **file**(like `VERSION` file name) would be nice.

release version. it's bit different with date.

src/common/mmo.h
```diff
@@ -64,7 +64,8 @@
 // 20120307 - 2012-03-07aRagexeRE+ - 0x970
 
 #ifndef PACKETVER
-	#define PACKETVER 20190530
+	// from tag vXXXXXXXX.Y -> XXXXXXXXY
+	#define PACKETVER 201905300
```
AnnieRuru, line 68: are you sure about this ?
isn't this change extremely drastic ?

for example the `<ITEMLINK>` function
https://github.com/HerculesWS/Hercules/blob/stable/npc/other/Global_Functions.txt#L484-L489
or all the `#if PACKETVER >= 20080102` `#endif` inside clif.c, all these should adds an extra digit ?

AnnieRuru:
@Asheraf , in some cases the error will not be thrown, better known as .... CONSTANT is the same as player variable !!

let's see ... take this script for example
https://github.com/HerculesWS/Hercules/commit/58e96dfb28b186f66409417a3233cb45302e5e1d
if I do a script with
```
mes "this is "+ getiteminfo(Apple, ITEMINFO_AEGISNAME);
```
on old emulator, if they don't have ITEMINFO_AEGISNAME constant defined, it will return 0
and 0 value on `*getiteminfo` is ITEMINFO_BUYPRICE, which return an integer
I want to return the name of an item, not the buying price of the item !

src/common/mmo.h
4144: ops, i changed wrong constant, sorry

4144: now correct constant used

src/config/core.h
```diff
@@ -21,6 +21,9 @@
 #ifndef CONFIG_CORE_H
 #define CONFIG_CORE_H
 
+// from tag vXXXXXXXX.Y -> XXXXXXXXY
+#define HERCULES_VERSION 202009200
```
AnnieRuru, line 25: I'm guessing the 0 at the end of 20200920 is the increment counter of commit of that day =/
why HERCULES_VERSION wants to be different than PACTEVER is anyone guess
but this might do the trick

AnnieRuru (approved):

src/config/core.h
Asheraf: the last number is for multiple releases on the same day eg: `v2020.03.08+2`, `v2020.03.08+1`, `v2020.03.08 `

 and for full list you can check https://github.com/HerculesWS/Hercules/tags

as for why HERCULES_VERSION is not the same as PACKETVER it's because this is the emulator version and not the client version.

PR #2874, opened 2020-10-17
Add more options to the quest database 

Asheraf:
### Pull Request Prelude

### Changes Proposed
These changes add support for hunting quests for any monsters, level range restricted, map restricted missions.

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -18325,20 +18325,23 @@ static void clif_quest_send_list(struct map_session_data *sd)
 		for (j = 0; j < qi->objectives_count; j++) {
 			Assert_retb(j < MAX_QUEST_OBJECTIVES);
 
-			int mob_id = 0;
-			if (qi->objectives[j].mob == 0 && qi->objectives[j].mapid >= 0)
-				mob_id = QUEST_MAPWIDE_ID;
-			else
-				mob_id = qi->objectives[j].mob;
+			int mob_id = qi->objectives[j].mob;
+			if (mob_id == 0) {
+				if (quest_mobtypeisenabled(qi->objectives[j].mobtype)) {
+					mob_id = QUEST_MOBTYPE_ID;
+				} else if (qi->objectives[j].mapid >= 0) {
+					mob_id = QUEST_MAPWIDE_ID;
+				}
+			}
 			real_len += sizeof(info->objectives[j]);
 
 #if PACKETVER_ZERO_NUM >= 20181010 || PACKETVER >= 20181017
 			info->objectives[j].huntIdent = sd->quest_log[i].quest_id;
 			info->objectives[j].huntIdent2 = j;
-			info->objectives[j].mobType = 0; // Info Needed
+			info->objectives[j].mobType = quest_mobtype2client(qi->objectives[j].mobtype);
 #elif PACKETVER >= 20150513
 			info->objectives[j].huntIdent = (sd->quest_log[i].quest_id * 1000) + j;
-			info->objectives[j].mobType = 0; // Info Needed
+			info->objectives[j].mobType = quest_mobtype2client(qi->objectives[j].mobtype);
 #endif
 			info->objectives[j].mob_id = mob_id;
 #if PACKETVER >= 20150513
@@ -18349,6 +18352,8 @@ static void clif_quest_send_list(struct map_session_data *sd)
 			info->objectives[j].maxCount = qi->objectives[j].count;
 			if (mob_id == QUEST_MAPWIDE_ID) {
 				safestrncpy(info->objectives[j].mobName, map->list[qi->objectives[j].mapid].name, NAME_LENGTH);
+			} else if (mob_id == QUEST_MOBTYPE_ID) {
+				safestrncpy(info->objectives[j].mobName, "", NAME_LENGTH);
 			} else {
 				const struct mob_db *monster = mob->db(qi->objectives[j].mob);
 				safestrncpy(info->objectives[j].mobName, monster->jname, NAME_LENGTH);
@@ -18425,19 +18430,22 @@ static void clif_quest_add(struct map_session_data *sd, struct quest *qd)
 	packet->count = qi->objectives_count;
 
 	for (i = 0; i < qi->objectives_count; i++) {
-		int mob_id = 0;
-		if (qi->objectives[i].mob == 0 && qi->objectives[i].mapid >= 0)
-			mob_id = QUEST_MAPWIDE_ID;
-		else
-			mob_id = qi->objectives[i].mob;
+		int mob_id = qi->objectives[i].mob;
+		if (mob_id == 0) {
+			if (quest_mobtypeisenabled(qi->objectives[i].mobtype)) {
+				mob_id = QUEST_MOBTYPE_ID;
+			} else if (qi->objectives[i].mapid >= 0) {
+				mob_id = QUEST_MAPWIDE_ID;
+			}
+		}
 
 #if PACKETVER_ZERO_NUM >= 20181010 || PACKETVER >= 20181017
 		packet->objectives[i].huntIdent = qd->quest_id;
 		packet->objectives[i].huntIdent2 = i;
-		packet->objectives[i].mobType = 0; // Info Needed
+		packet->objectives[i].mobType = quest_mobtype2client(qi->objectives[i].mobtype);
 #elif PACKETVER >= 20150513
 		packet->objectives[i].huntIdent = (qd->quest_id * 1000) + i;
-		packet->objectives[i].mobType = 0; // Info Needed
+		packet->objectives[i].mobType = quest_mobtype2client(qi->objectives[i].mobtype);
 #endif
 		packet->objectives[i].mob_id = mob_id;
 #if PACKETVER >= 20150513
@@ -18448,6 +18456,8 @@ static void clif_quest_add(struct map_session_data *sd, struct quest *qd)
 
 		if (mob_id == QUEST_MAPWIDE_ID) {
 			safestrncpy(packet->objectives[i].mobName, map->list[qi->objectives[i].mapid].name, NAME_LENGTH);
+		} else if (mob_id == QUEST_MOBTYPE_ID) {
+			safestrncpy(packet->objectives[i].mobName, "", NAME_LENGTH);
 		} else {
 			const struct mob_db *monster = mob->db(qi->objectives[i].mob);
 			safestrncpy(packet->objectives[i].mobName, monster->jname, NAME_LENGTH);
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -18323,11 +18323,15 @@ static void clif_quest_send_list(struct map_session_data *sd)
 		info->hunting_count = qi->objectives_count;
 
 		for (j = 0; j < qi->objectives_count; j++) {
-			struct mob_db *mob_data;
 			Assert_retb(j < MAX_QUEST_OBJECTIVES);
+
+			int mob_id = 0;
+			if (qi->objectives[j].mob == 0 && qi->objectives[j].mapid >= 0)
+				mob_id = QUEST_MAPWIDE_ID;
+			else
+				mob_id = qi->objectives[j].mob;
 			real_len += sizeof(info->objectives[j]);
 
-			mob_data = mob->db(qi->objectives[j].mob);
 #if PACKETVER_ZERO_NUM >= 20181010 || PACKETVER >= 20181017
 			info->objectives[j].huntIdent = sd->quest_log[i].quest_id;
 			info->objectives[j].huntIdent2 = j;
@@ -18336,14 +18340,19 @@ static void clif_quest_send_list(struct map_session_data *sd)
 			info->objectives[j].huntIdent = (sd->quest_log[i].quest_id * 1000) + j;
 			info->objectives[j].mobType = 0; // Info Needed
 #endif
-			info->objectives[j].mob_id = qi->objectives[j].mob;
+			info->objectives[j].mob_id = mob_id;
 #if PACKETVER >= 20150513
 			info->objectives[j].levelMin = qi->objectives[j].level.min;
 			info->objectives[j].levelMax = qi->objectives[j].level.max;
 #endif
 			info->objectives[j].huntCount = sd->quest_log[i].count[j];
 			info->objectives[j].maxCount = qi->objectives[j].count;
-			safestrncpy(info->objectives[j].mobName, mob_data->jname, sizeof(info->objectives[j].mobName));
+			if (mob_id == QUEST_MAPWIDE_ID) {
+				safestrncpy(info->objectives[j].mobName, map->list[qi->objectives[j].mapid].name, NAME_LENGTH);
+			} else {
+				const struct mob_db *monster = mob->db(qi->objectives[j].mob);
+				safestrncpy(info->objectives[j].mobName, monster->jname, NAME_LENGTH);
+			}
 		}
 #endif // PACKETVER >= 20141022
 	}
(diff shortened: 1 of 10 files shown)


Kenpachi2k13:
Instead of using 0 for any monster I'd suggest to use the `mob_group` enumeration in https://github.com/HerculesWS/Hercules/blob/1354b01c7a24e9ffd9ea570cbaec668721fc2a6c/src/map/mob.h#L328 to be more flexible.

Asheraf:
@Kenpachi2k13 there is no client support for those values so we'll end up having them all represented as `Monster (%d/%d)` , i think having same values as the client is better.

Kenpachi2k13:
Ah, okay. I didn't notice this. In this case you're right of course...

src/map/quest.c
```diff
@@ -519,6 +524,22 @@ static struct quest_db *quest_read_db_sub(struct config_setting_t *cs, int n, co
 			RECREATE(entry->objectives, struct quest_objective, ++entry->objectives_count);
 			entry->objectives[entry->objectives_count-1].mob = mob_id;
 			entry->objectives[entry->objectives_count-1].count = count;
+
+			const struct config_setting_t *lvt = libconfig->setting_get_member(tt, "Level");
+			if (lvt != NULL) {
+				if (config_setting_is_aggregate(lvt)) {
+					int min = libconfig->setting_get_int_elem(lvt, 0);
+					int max = libconfig->setting_get_int_elem(lvt, 1);
+					if (min > max) {
```
4144, line 534: probably also check ``max`` for ``< 0``

4144 (changes requested):

src/map/quest.h
```diff
@@ -29,8 +29,10 @@ struct block_list;
 struct config_setting_t;
 struct map_session_data;
 struct questinfo;
+struct mob_data;
 
 #define MAX_QUEST_DB (60355+1) // Highest quest ID + 1
+#define QUEST_MAPWIDE_ID 10363
```
Kenpachi2k13, line 35: Would be nice if you'd add a short description.

Kenpachi2k13 (changes requested):

4144 (approved):

Kenpachi2k13 (approved):

src/map/quest.c
```diff
@@ -314,10 +317,14 @@ static void quest_update_objective(struct map_session_data *sd, int mob_id)
...
 
 		for (j = 0; j < qi->objectives_count; j++) {
-			if (qi->objectives[j].mob == mob_id && sd->quest_log[i].count[j] < qi->objectives[j].count) {
-				sd->quest_log[i].count[j]++;
-				sd->save_quest = true;
-				clif->quest_update_objective(sd, &sd->quest_log[i]);
+			if ((qi->objectives[j].mob == 0 || qi->objectives[j].mob == md->class_) &&
+				sd->quest_log[i].count[j] < qi->objectives[j].count &&
+				(qi->objectives[j].level.min == 0 || qi->objectives[j].level.min <= md->level) &&
+				(qi->objectives[j].level.max == 0 || qi->objectives[j].level.max >= md->level) &&
```
AnnieRuru, line 323: can you clarify something ?
the level range is set on the player side or the monster level ?

why check the monster level ?

Asheraf: this is not meant to be used when a monster id is specified (i will add a warning for this later), the goal is to be able to have a quest which requires killing monsters of a certain level range, for example https://www.divine-pride.net/database/quest/16230

AnnieRuru (approved):

src/map/clif.c
```diff
@@ -18321,20 +18321,29 @@ static void clif_quest_send_list(struct map_session_data *sd)
...
+			}
 			real_len += sizeof(info->objectives[j]);
 
 #if PACKETVER_ZERO_NUM >= 20181010 || PACKETVER >= 20181017
 			info->objectives[j].huntIdent = sd->quest_log[i].quest_id;
 			info->objectives[j].huntIdent2 = j;
-			info->objectives[j].mobType = 0; // Info Needed
+			info->objectives[j].mobType = qi->objectives[j].mobtype.size |
+							quest->mobele2client(qi->objectives[j].mobtype.race) |
+							quest->mobrace2client(qi->objectives[j].mobtype.ele);
```
4144, line 18341: probably blocks like this better move into separate function or into macro?

src/map/quest.h
```diff
@@ -29,8 +29,40 @@ struct block_list;
...
 #define MAX_QUEST_DB (60355+1) // Highest quest ID + 1
+#define QUEST_MAPWIDE_ID 10363 // MobId handled by the client to display MapName
+#define QUEST_MOBTYPE_ID 31999 // MobId handled by the client to display Mob properties
+
+#define quest_mobtype2client(type) ((type).size | quest->mobele2client((type).ele) | quest->mobrace2client((type).race))
+#define quest_mobtypeisenabled(type) ((type).size != 0 || (type).ele_enabled || (type).race_enabled)
+
+enum quest_mobtype {
+	// Monster Races
+	TYPE_RC_DEMIHUMAN     = 0x80,
```
Kenpachi2k13, line 43: In my opinion the prefix `TYPE_` is too general. I'd prefer a more specific one like we're using in other enumerations. (`SERVER_TYPE_`, `RANKTYPE_`, `MAIL_TYPE_`, etc.)
Maybe something like `QUEST_MOB_TYPE_` or `QMT_`?

src/map/quest.c
```diff
@@ -509,13 +519,69 @@ static struct quest_db *quest_read_db_sub(struct config_setting_t *cs, int n, co
...
+					ShowWarning("quest_read_db_sub: Invalid format for Level in \"%s\", for quest (%d).\n", source, entry->id);
+				}
+			}
+			const char *map_name = NULL;
+			int16 mapid = -1;
+
+			if (libconfig->setting_lookup_string(tt, "MapName", &map_name) != CONFIG_FALSE) {
+				if ((mapid = map->mapname2mapid(map_name)) < 0) {
+					ShowWarning("quest_read_db_sub: Invalid MapName \"%s\" in \"%s\", for quest (%d), ignoring.\n", map_name, source, entry->id);
+				}
```
Kenpachi2k13, line 562: Unnecessary curly brackets.

Kenpachi2k13 (changes requested):
Your changes in quest.c contain some lines which exceed the 120 columns length limit significantly. Please have a look at it.

src/map/clif.c
```diff
@@ -18319,28 +18319,41 @@ static void clif_quest_send_list(struct map_session_data *sd)
...
-			struct mob_db *mob_data;
 			Assert_retb(j < MAX_QUEST_OBJECTIVES);
+
+			int mob_id = qi->objectives[j].mob;
+			if (qi->objectives[j].mob == 0) {
+				if (quest_mobtypeisenabled(qi->objectives[j].mobtype)) {
+					mob_id = QUEST_MOBTYPE_ID;
+				} else if (qi->objectives[j].mapid >= 0) {
+					mob_id = QUEST_MAPWIDE_ID;
+				}
```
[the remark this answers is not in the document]
Asheraf: It's a matter of preference here but I don't like free `else` cases.

src/map/clif.c
```diff
@@ -18319,28 +18319,41 @@ static void clif_quest_send_list(struct map_session_data *sd)
...
 			info->objectives[j].maxCount = qi->objectives[j].count;
-			safestrncpy(info->objectives[j].mobName, mob_data->jname, sizeof(info->objectives[j].mobName));
+			if (mob_id == QUEST_MAPWIDE_ID) {
+				memcpy(info->objectives[j].mobName, map->list[qi->objectives[j].mapid].name, NAME_LENGTH);
+			} else if (mob_id == QUEST_MOBTYPE_ID) {
+				safestrncpy(info->objectives[j].mobName, "", NAME_LENGTH);
+			} else {
+				const struct mob_db *monster = mob->db(qi->objectives[j].mob);
+				memcpy(info->objectives[j].mobName, monster->jname, NAME_LENGTH);
+			}
```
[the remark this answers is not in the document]
Asheraf: I just kept the existing code, but since now I added the empty string to string copy maybe it need to be changed.

src/map/clif.c
```diff
@@ -18413,26 +18426,38 @@ static void clif_quest_add(struct map_session_data *sd, struct quest *qd)
...
-		struct mob_db *monster;
-
-		monster = mob->db(qi->objectives[i].mob);
+		int mob_id = qi->objectives[i].mob;
+		if (qi->objectives[i].mob == 0) {
+			if (quest_mobtypeisenabled(qi->objectives[i].mobtype)) {
+				mob_id = QUEST_MOBTYPE_ID;
+			} else if (qi->objectives[i].mapid >= 0) {
+				mob_id = QUEST_MAPWIDE_ID;
+			}
```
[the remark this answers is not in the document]
Asheraf: same as above

src/map/quest.c
```diff
@@ -314,10 +317,17 @@ static void quest_update_objective(struct map_session_data *sd, int mob_id)
...
-				sd->save_quest = true;
-				clif->quest_update_objective(sd, &sd->quest_log[i]);
+			if ((qi->objectives[j].mob == 0 || qi->objectives[j].mob == md->class_) &&
+				sd->quest_log[i].count[j] < qi->objectives[j].count &&
+				(qi->objectives[j].level.min == 0 || qi->objectives[j].level.min <= md->level) &&
+				(qi->objectives[j].level.max == 0 || qi->objectives[j].level.max >= md->level) &&
+				(qi->objectives[j].mapid < 0 || qi->objectives[j].mapid == md->bl.m) &&
+				(qi->objectives[j].mobtype.size == 0 || qi->objectives[j].mobtype.size == md->status.size) &&
+				(qi->objectives[j].mobtype.race_enabled == false || qi->objectives[j].mobtype.race == md->status.race) &&
+				(qi->objectives[j].mobtype.ele_enabled == false || qi->objectives[j].mobtype.ele == md->status.def_ele)) {
```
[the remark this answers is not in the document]
Asheraf: This paragraph is wrong, the style guide is based on LK and they break after the operator.

src/map/quest.c
```diff
@@ -509,13 +519,69 @@ static struct quest_db *quest_read_db_sub(struct config_setting_t *cs, int n, co
 				break;
 			if (!config_setting_is_group(tt))
 				continue;
-			if (!libconfig->setting_lookup_int(tt, "MobId", &mob_id) || mob_id <= 0)
+			if (libconfig->setting_lookup_int(tt, "MobId", &mob_id) != CONFIG_FALSE && mob_id < 0) {
```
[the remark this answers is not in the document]
Asheraf: It is intended behavior, see the following example, no mob id is given so it fallthough here without a warning and the result is a quest that allows you to kill any monster.
```
Targets: (
	{
		Count: 20
	},
)
```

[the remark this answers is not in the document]
Kenpachi2k13: I see. But what if a monster ID is given and `libconfig->setting_lookup_int()` for whatever unknown reason returns `CONFIG_FALSE`?

And I just noticed the monster ID isn't checked for validity/existence if it's greater than `0`.

src/map/clif.c
```diff
@@ -18319,28 +18319,41 @@ static void clif_quest_send_list(struct map_session_data *sd)
...
 #endif
 			info->objectives[j].huntCount = sd->quest_log[i].count[j];
 			info->objectives[j].maxCount = qi->objectives[j].count;
-			safestrncpy(info->objectives[j].mobName, mob_data->jname, sizeof(info->objectives[j].mobName));
+			if (mob_id == QUEST_MAPWIDE_ID) {
+				memcpy(info->objectives[j].mobName, map->list[qi->objectives[j].mapid].name, NAME_LENGTH);
+			} else if (mob_id == QUEST_MOBTYPE_ID) {
+				safestrncpy(info->objectives[j].mobName, "", NAME_LENGTH);
+			} else {
+				const struct mob_db *monster = mob->db(qi->objectives[j].mob);
```
Kenpachi2k13, line 18354: Maybe add a check if `mob->db(qi->objectives[j].mob)` is `mob->dummy`?

src/map/quest.c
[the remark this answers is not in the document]
Asheraf: If libconfig fails then we got bigger problems, as for validating the id it's outside the scope of the changes i want to do here ("enable 0 for any monster"), also it's done in #2886 .

src/map/clif.c
Asheraf: `mob->dummy` is already initialized so it's safe to copy from it even if mob not existing, also outside the scope of my changes.

src/map/quest.c
```diff
@@ -927,6 +994,62 @@ static bool quest_questinfo_validate_mercenary_class(struct map_session_data *sd
...
+	case RC_PLANT:
+		return QMT_RC_PLANT;
+	case RC_DEMON:
+		return QMT_RC_DEMON;
+	case RC_ANGEL:
+		return QMT_RC_ANGEL;
+	case RC_UNDEAD:
+		return QMT_RC_UNDEAD;
+	case RC_FORMLESS:
+		return QMT_RC_FORMLESS;;
```
AnnieRuru, line 1045: extra ;;

AnnieRuru (changes requested):
I'm sorry, I know my C isn't that good

but please add a documentation file like `doc\quest_db.md` and explain what each field does
this isn't just a simple _kill 10 poring quest complete_ stuff anymore

from my understanding some fields can be left empty, some can't
like `Count` field cannot be 0 right ?

and can we use multiple `MobId:` or `MapName:` field ?
```c
MapName: prt_fild01
MapName: prt_fild02
MapName: prt_fild03
```

explain all this in a documentation

Asheraf:
> I'm sorry, I know my C isn't that good
> 
> but please add a documentation file like `doc\quest_db.md` and explain what each field does
> this isn't just a simple _kill 10 poring quest complete_ stuff anymore
> 
> from my understanding some fields can be left empty, some can't
> like `Count` field cannot be 0 right ?
> 
> and can we use multiple `MobId:` or `MapName:` field ?
> 
> ```c
> MapName: prt_fild01
> MapName: prt_fild02
> MapName: prt_fild03
> ```
> 
> explain all this in a documentation

I have added documentation about mandatory and optional fields to the database header, as for defining multiple keys that is not possible.. libconfig does not work that way.

src/map/quest.c
```diff
@@ -321,7 +321,10 @@ static void quest_update_objective(struct map_session_data *sd, const struct mob
 				sd->quest_log[i].count[j] < qi->objectives[j].count &&
 				(qi->objectives[j].level.min == 0 || qi->objectives[j].level.min <= md->level) &&
 				(qi->objectives[j].level.max == 0 || qi->objectives[j].level.max >= md->level) &&
-				(qi->objectives[j].mapid < 0 || qi->objectives[j].mapid == md->bl.m)) {
+				(qi->objectives[j].mapid < 0 || qi->objectives[j].mapid == md->bl.m) &&
+				(qi->objectives[j].mobtype.size == 0 || (qi->objectives[j].mobtype.size & md->status.size) != 0) &&
```
MishimaHaruna, line 325: Note that `md->status.size` isn't a bitmask! For example, `SZ_SMALL` (or `Size_Small` in scripts/DBs) is defined as 0, making this bitmask comparison always fail for small sized monsters.

If you want this to support bitmasks on the quest db side, then you need to do the comparison for example as `(qi->objectives[j].mobtype.size & (1 << md->status.size)) != 0` (and document to the user how the bits are mapped).

The same goes for elements and races (which are defined in a similar way to the sizes).

Otherwise, if a bit mask isn't necessary, the comparison can be done directly with i.e. `qi->objectives[j].mobtype.size == md->status.size `.

MishimaHaruna (changes requested):

src/map/quest.c
Asheraf: I seem to have mixed client and server values for this, for now I removed the bitmask support for these fields.

4144 (approved):

MishimaHaruna (approved):

PR #2885, opened 2020-10-30
Fix nullpo error in searchstore

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
Fixes 2 checks in search store function:
1. it prevents trying to search with an empty item list, this is not even allowed by the client, and makes no sense in the process, as it always start the search by an item. (that's why I put the check on clif)

2. When searching for an item without asking for cards, it was giving us a nullpo because card_list would be NULL (as it becomes a "0-length array"), but it is a totally valid search (e.g. Searching for an ETC item would never have cards to check for). So I disabled the null pointer check when the card count is zero, from my tests and checking the code, every code called from this would check card count before trying to access the array, so it looks safe.

**Issues addressed:** 
None, i think

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -19750,6 +19750,9 @@ static void clif_parse_SearchStoreInfo(int fd, struct map_session_data *sd)
 		return;
 	}
 
+	if (p->itemsCount < 1)
+		return; // Should never happen
+
 	type       = p->searchType;
 	max_price  = p->maxPrice;
 	min_price  = p->minPrice;
--- a/src/map/searchstore.c
+++ b/src/map/searchstore.c
@@ -135,8 +135,14 @@ static void searchstore_query(struct map_session_data *sd,
 		return;
 	}
 
+	if (item_count < 1) {
+		clif->search_store_info_failed(sd, SSI_FAILED_NOTHING_SEARCH_ITEM);
+		return;
+	}
+
 	nullpo_retv(itemlist);
-	nullpo_retv(cardlist);
+	if (card_count > 0)
+		nullpo_retv(cardlist);
 
 	// validate lists
 	for( i = 0; i < item_count; i++ ) {


src/map/clif.c
```diff
@@ -19778,6 +19778,9 @@ static void clif_parse_SearchStoreInfo(int fd, struct map_session_data *sd)
 
 	cardlist = RFIFOP(fd, sizeof(struct PACKET_CZ_SEARCH_STORE_INFO) + blocksize * item_count);
 
+	if (item_count < 1)
+		return; // Should never happen
```
Kenpachi2k13, line 19782: In my opinion this check should be placed some lines above, prior to assigning a value to `item_count`.

```c
	...

	if (packet_len < sizeof(struct PACKET_CZ_SEARCH_STORE_INFO))
	{// minimum packet length
		ShowError("clif_parse_SearchStoreInfo: Malformed packet (expected length=%u, length=%d, account_id=%d).\n", (uint32)sizeof(struct PACKET_CZ_SEARCH_STORE_INFO), packet_len, sd->bl.id);
		return;
	}

	if (p->itemsCount < 1) // Should never happen
		return;

	type       = p->searchType;
	max_price  = p->maxPrice;
	min_price  = p->minPrice;
	item_count = p->itemsCount;
	card_count = p->cardsCount;
	itemlist   = &p->items[0];

	...
```

src/map/searchstore.c
```diff
@@ -136,7 +136,8 @@ static void searchstore_query(struct map_session_data *sd,
 	}
 
 	nullpo_retv(itemlist);
```
Kenpachi2k13, line 138: I'd also check `item_count` prior to `nullpo_retv(itemlist);`.

```c
	if (item_count < 1) {
		clif->search_store_info_failed(sd, SSI_FAILED_NOTHING_SEARCH_ITEM);
		return;
	}

	nullpo_retv(itemlist);
```

Kenpachi2k13 (changes requested):

guilherme-gm:
Thanks for reviewing it! fixed and rebased

Kenpachi2k13 (approved):

4144 (approved):

MishimaHaruna (approved):

PR #2886, opened 2020-10-31
Allow constants on QuestDB

guilherme-gm:
### IMPORTANT
These changes were made using Asheraf's #2874 as base, so it must wait for it before being merged.

### Pull Request Prelude

### Changes Proposed

This change allows the use of constants for monster IDs and Item Ids. Also it adds some error messages when ignoring entries, so the user knows what happened.

After doing the change, some quests were giving errors on pre-re, because they have special RE drops, so I copied those items to pre-re db, as you shouldn't get the quest.

**Issues addressed:** 
None, I think.

### DIFF
--- a/db/pre-re/item_db.conf
+++ b/db/pre-re/item_db.conf
@@ -55875,6 +55875,34 @@ item_db: (
 	Weight: 1
 },
 {
+	Id: 6498
+	AegisName: "Jejellopy"
+	Name: "Jejellopy"
+	Buy: 200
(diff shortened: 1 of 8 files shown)


4144 (approved):

MishimaHaruna:
Marking it as a draft since it depends on another PR that is marked as draft - please mark them as ready once it is if it slips past my radar

MishimaHaruna (approved):

Issue #1621, opened 2017-03-11
Usage of bonus2 bSubRace on renewal db


iRagno:
[//]: # (**********************************)
[//]: # (** Fill in the following fields **)
[//]: # (**********************************)

[//]: # (Note: Lines beginning with syntax such as this one, are comments and will not be visible in your report!)

### Issue Prelude

[//]: # (Please complete these mandatory steps and check the following boxes by putting an `x` inside the brackets _before_ filing your issue)

-  [x] I have not modified the source prior to reproducing this issue.
-  [x] I am using the latest version of Hercules.
-  [x] I am aware that this report will be closed or deleted if it becomes obvious that I am stating the false.

### Description

DB renewal (item_db.conf and item_combo_db) are still using pre-renewal `bonus2 bSubRace` bonus script instead of `bonus2 bAddRaceTolerance` intended for renewal.

### Current Behavior

Having bonus2 bSubRace causes a weird behaviour when reducing damage, it doesn't reduce all the damage that should do.

### Expected Behavior

Have a proper reduction damage acording to race resistance.

**Branch(es):**
- [x] master
- [ ] other

**Hercules rev. hash/commit:** 

Git revision src: 9fba7e8c5966bb0aa5ce1c31ebe5cec96180e1d5

```
[Info]: Hercules 32-bit for Windows
[Info]: Exported revision (src): 'Unknown'
[Info]: Exported revision (scripts): 'Unknown'
```

### Operating System

- [Info]: OS version: 'Windows 7 Home Premium Service Pack 1 (build 7601) [x86_64]'

[//]: # (Mac OS X 10.12.3 16D32 [x86_64])
[//]: # (Thank you for adhering to this process! It ensures your issue is resolved quickly and that neither your nor our time is needlessly wasted.)
[//]: # (This template is for problem reports. For other types of report, edit it accordingly.)
[//]: # (For fixes please create a Pull Request.)

iRagno:
I'm almost sure this is a bug, but has been for so many time that I need to ask if there is a chance of that may be intended behaviour.

PR #2889, opened 2020-11-01
Use bAddRaceTolerance instead of bSubRace in RE database files

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

- Replace `bSubRace` with `bAddRaceTolerance` in _db/re/item_combo_db.conf_.
- Replace `bSubRace` with `bAddRaceTolerance` in _db/re/item_db.conf_.

**Issues addressed:** #1621

### DIFF
--- a/db/re/item_combo_db.conf
+++ b/db/re/item_combo_db.conf
@@ -545,8 +545,8 @@ combo_db: (
 	Script: <"
 		bonus bMatkRate, 10;
 		bonus bDex, 2;
-		bonus2 bSubRace, RC_Demon, 10;
-		bonus2 bSubRace, RC_Undead, 10;
+		bonus2(bAddRaceTolerance, RC_Demon, 10);
+		bonus2(bAddRaceTolerance, RC_Undead, 10);
 	">
 },
 {
@@ -554,8 +554,8 @@ combo_db: (
 	Script: <"
 		bonus bMatkRate, 10;
 		bonus bDex, 2;
-		bonus2 bSubRace, RC_Demon, 10;
-		bonus2 bSubRace, RC_Undead, 10;
+		bonus2(bAddRaceTolerance, RC_Demon, 10);
+		bonus2(bAddRaceTolerance, RC_Undead, 10);
 	">
 },
 {
@@ -970,7 +970,7 @@ combo_db: (
 {
 	Items: ["Freyja_Overcoat", "Freyja_Boots", "Freyja_Cape", "Freyja_Crown"]
 	Script: <"
-		bonus2 bSubRace, RC_DemiPlayer, 10;
+		bonus2(bAddRaceTolerance, RC_DemiPlayer, 10);
 		bonus bMaxHPrate, 20;
 		bonus2 bResEff, Eff_Freeze, 10000;
 		skill WZ_FIREPILLAR, 10;
@@ -1007,7 +1007,7 @@ combo_db: (
 {
 	Items: ["Assaulter_Plate", "Battle_Greave", "Commander_Manteau"]
 	Script: <"
-		bonus2 bSubRace, RC_NonDemiPlayer, -300;
+		bonus2(bAddRaceTolerance, RC_NonDemiPlayer, -200);
 		bonus bVit, 3;
 		bonus bMaxHPrate, 12;
 		bonus bHealPower2, 10;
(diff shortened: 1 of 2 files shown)


db/re/item_db.conf
```diff
@@ -151058,9 +151067,12 @@ item_db: (
 	Buy: 10
 	Loc: "EQP_SHADOW_SHIELD"
 	Script: <"
-		bonus2 bSubRace,RC_Insect,1;
-		if (getrefine()>=7) { bonus2 bSubRace,RC_Insect,1; }
-		if (getrefine()>=9) { bonus2 bSubRace,RC_Insect,1; }
+		if (getrefine() >= 9)
+			bonus2(bAddRaceTolerance, RC_Insect, 3);
+		if (getrefine() >= 7)
+			bonus2(bAddRaceTolerance, RC_Insect, 2);
```
AnnieRuru, line 151073: this line ... I think its wrong
if the equipment refine is 9, the value become 5 (3+2)

it should be ELSE IF ?

well logic wise, should use ELSE IF
but try check Aegis file, maybe it was really done like this officially

AnnieRuru (commented):
someone needs to grab an Aegis to confirm this

4144 (approved):

db/re/item_db.conf
Kenpachi2k13: You're right, Annie. It should be `else if`.
I'll fix this later today.

AnnieRuru:
```c
bonus2(bAddRaceTolerance, RC_Insect, 1 + (getrefine()>=7) + (getrefine()>=9));
```
hmm ??
just a suggestion

Kenpachi2k13: Fixed.

> dunno if this kind of code is accepted

Nope, using a boolean expression in a calculation isn't valid.
But using a short-if would be valid:
```c
bonus2(bAddRaceTolerance, RC_Insect, 1 + ((getrefine() >= 7) ? 1 : 0) + ((getrefine() >= 9) ? 1 : 0));
```

db/re/item_combo_db.conf
```diff
@@ -1073,7 +1073,7 @@ combo_db: (
 {
 	Items: ["Elite_Shooter_Suit", "Battle_Boots", "Sheriff_Manteau"]
 	Script: <"
-		bonus2 bSubRace, RC_NonDemiPlayer, -300;
+		bonus2(bAddRaceTolerance, RC_NonDemiPlayer, -200);
```
AnnieRuru, line 1076: is there a reason to change -300 into -200 ?

Kenpachi2k13: Yes, `` is wrong and `-200` is right. :grin:
```
item COMBIITEM_Elite_Shooter_Set
	event OnStartEquip:
		SubRaceTolerace RACE_All 200
		AddRaceTolerace RACE_Human 200
		AddRaceTolerace RACE_Player 200
		AddExtParam User VAR_MAXHPPERCENT 12
		AddExtParam User VAR_DEXAMOUNT 3
		SubRangeAttackDamage User 10
		SubSpellDelay 25
	return
	event OnFinishEquip:
		AddRaceTolerace RACE_All 200
		SubRaceTolerace RACE_Human 200
		SubRaceTolerace RACE_Player 200
		SubExtParam User VAR_MAXHPPERCENT 12
		SubExtParam User VAR_DEXAMOUNT 3
		AddRangeAttackDamage User 10
		AddSpellDelay 25
	return
```

AnnieRuru (approved):

db/re/item_db.conf
```diff
@@ -5307,7 +5307,7 @@ item_db: (
 	EquipLv: 40
 	Subtype: "W_1HSWORD"
 	Script: <"
-		bonus2 bSubRace,RC_NonBoss,-10;
+		bonus2(bAddRaceTolerance, RC_NonBoss, -10);
```
hemagx, line 5310: in Aegis this is a class instead of race tolerance (and shall be fixed to class or left as is)

db/re/item_db.conf
```diff
@@ -5926,7 +5926,7 @@ item_db: (
 	Script: <"
 		bonus bIgnoreDefRace,RC_DemiPlayer;
 		bonus2 bAddRace,RC_DemiPlayer,20;
-		bonus2 bSubRace,RC_DemiPlayer,-10;
+		bonus2(bAddRaceTolerance, RC_DemiPlayer, -10);
```
hemagx, line 5929: In aegis this is a normal sub race not race tolerance

hemagx (changes requested):
Reviewing the whole thing will be pain, but this said even in renewal normal subrace is still used, so each item have to manually checked, i only marked two.

Kenpachi2k13:
After some investigation it turned out that `bAddRaceTolerance` shouldn't exist at all and `bSubRace` should be used in RE, too.
Additionally we need a new bonus to represent the Aegis `ClassAddDamage` command, since it's not related to races and doesn't stack with race bonuses.
So this PR is obsolete -> `close();`

Issue #2898, opened 2020-11-13
*rodex_sendmail script command can send with non-existent char ID


AnnieRuru:
**Describe the bug**
1. you can use script command to send a mail to non-existent char ID
2. if send using the script command below will cause char-server.exe spam failed assertion

```
[Error]: --- failed assertion --------------------------------------------
[Error]: F:\Ragnarok\Hercules\src\char\mapif.c:1734: 'receiver_id + receiver_accountid > 0' in function `mapif_rodex_send'
[Error]: --- end failed assertion ----------------------------------------
```

**To Reproduce**
```c
prontera,155,185,5	script	kjsdfhks	1_F_MARIA,{
	rodex_sendmail 123, "someone", "title", "body\r leg", 0, Apple, 1; // send successfully and produce a line in `rodex_mail` table
	rodex_sendmail getcharid(CHAR_ID_CHAR, 0), "someone", "title", "body\r leg", 0, Apple, 1; // failed assertion
}
```

**Expected behavior**
1. not sure if its possible to use `char` table to check if the char ID is non-existent
but I think it can be work as intended as it's scripter's fault to send to wrong char ID

.... PS: however try send a mail normally, there is a validate name button to forbid sending to non-existent char though

2. failed assertion is another thing, this should be fixed

**System specs (please complete the following information):**
 - OS: Windows 10 pro 64-bit
 - Hercules Version: latest (always pull latest git)
 - Packet version: 20190530
 - Mode: Renewal
 - Client type: [2019-05-30aRagexeRE](http://nemo.herc.ws/downloads/2019-05-30aRagexeRE/)
 - Branch: Main

**Plugins used or source modifications**
................

**Additional context**

Kenpachi2k13:
The problem is `rodex_sendmail_sub()` (script.c) blindly takes the receiver's character/account ID from the argument passed by the script command without doing any evaluation.
```c
	if (strcmp(script->getfuncname(st), "rodex_sendmail_acc") == 0 || strcmp(script->getfuncname(st), "rodex_sendmail_acc2") == 0)
		msg->receiver_accountid = script_getnum(st, 2);
	else
		msg->receiver_id = script_getnum(st, 2);
```

To make this process safer `script_getnum(st, 2)` should be checked against `START_CHAR_NUM` respectively `START_ACCOUNT_NUM`/`END_ACCOUNT_NUM` prior to assigning the receiver's character/account ID.


[...]

Issue #2897, opened 2020-11-13
*rodex_sendmail script command attach with 0 item amount can't delete mail


AnnieRuru:
**Describe the bug**
when using script command `*rodex_sendmail`, send with item amount 0,
will cause the mail cannot retrieve the item or delete the mail

```sql
SELECT * FROM rodex_items;
delete from rodex_items where amount = 0 and id > -1;
```
I have to manually remove like this

**To Reproduce**
```c
prontera,155,185,5	script	kjsdfhks	1_F_MARIA,{
	rodex_sendmail getcharid(CHAR_ID_CHAR), "someone", "title", "body\r leg", 0, Apple, 0;
}
```

**Expected behavior**
this script command should put some check with item amount with 0

**System specs (please complete the following information):**
 - OS: Windows 10 pro 64-bit
 - Hercules Version: latest (always pull latest git)
 - Packet version: 20190530
 - Mode: Renewal
 - Client type: [2019-05-30aRagexeRE](http://nemo.herc.ws/downloads/2019-05-30aRagexeRE/)
 - Branch: Main

**Plugins used or source modifications**
.... nothing concern here

**Additional context**

Kenpachi2k13:
Well... Again missing data evaluation.
Neither `BUILDIN(rodex_sendmail)` nor `BUILDIN(rodex_sendmail2)` checks the passed item amount for validity.

PR #2901, opened 2020-11-14
Add some data evaluation to the RODEX send mail script commands

Kenpachi2k13:
### Pull Request Prelude

### Changes Proposed

- Add check for receiver ID is in valid range to `rodex_sendmail_sub()`.
- Add check for valid item amount to `BUILDIN(rodex_sendmail)`.
- Add check for valid item amount to `BUILDIN(rodex_sendmail2)`.

**Issues addressed:** #2897 #2898

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -26454,11 +26454,24 @@ static BUILDIN(navigateto)
 static bool rodex_sendmail_sub(struct script_state *st, struct rodex_message *msg)
 {
 	const char *sender_name, *title, *body;
+	const char *func_name = script->getfuncname(st);
+	int receiver_id = script_getnum(st, 2);
 
-	if (strcmp(script->getfuncname(st), "rodex_sendmail_acc") == 0 || strcmp(script->getfuncname(st), "rodex_sendmail_acc2") == 0)
-		msg->receiver_accountid = script_getnum(st, 2);
-	else
-		msg->receiver_id = script_getnum(st, 2);
+	if (strcmp(func_name, "rodex_sendmail_acc") == 0 || strcmp(func_name, "rodex_sendmail_acc2") == 0) {
+		if (receiver_id < START_ACCOUNT_NUM || receiver_id > END_ACCOUNT_NUM) {
+			ShowError("script:rodex_sendmail: Invalid receiver account ID %d passed!\n", receiver_id);
+			return false;
+		}
+
+		msg->receiver_accountid = receiver_id;
+	} else {
+		if (receiver_id < START_CHAR_NUM) {
+			ShowError("script:rodex_sendmail: Invalid receiver character ID %d passed!\n", receiver_id);
+			return false;
+		}
+
+		msg->receiver_id = receiver_id;
+	}
 
 	sender_name = script_getstr(st, 3);
 	if (strlen(sender_name) >= NAME_LENGTH) {
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -26523,6 +26523,18 @@ static BUILDIN(rodex_sendmail)
 			return false;
 		}
 
+		if (!data_isint(script_getdata(st, param + 1))) {
+			ShowError("script:rodex_sendmail: Passed amount for item %d is not a number!\n", i + 1);
+			return false;
+		}
+
+		int amount = script_getnum(st, param + 1);
+
+		if (amount < 1 || amount > min(MAX_AMOUNT, SHRT_MAX)) {
+			ShowError("script:rodex_sendmail: Invalid amount %d passed for item %d!\n", amount, i + 1);
+			return false;
+		}
+
 		++item_count;
 		if (data_isstring(script_getdata(st, param)) == false) {
 			int itemid = script_getnum(st, param);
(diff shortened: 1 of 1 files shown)


src/map/script.c
```diff
@@ -26523,6 +26523,18 @@ static BUILDIN(rodex_sendmail)
...
 		}
 
+		if (!data_isint(script_getdata(st, param + 1))) {
+			ShowError("script:rodex_sendmail: Passed amount for item %d is not a number!\n", i + 1);
+			return false;
+		}
+
+		int amount = script_getnum(st, param + 1);
+
+		if (amount < 1 || amount > SHRT_MAX) {
```
AnnieRuru, line 26533: here should be MAX_AMOUNT

https://github.com/HerculesWS/Hercules/blob/32045249e6f91f1d09db2b63ad9eefb04fd1c1e9/src/map/pc.c#L4648-L4672

AnnieRuru (changes requested):

src/map/script.c
Kenpachi2k13: Indeed. Fixed...

src/map/script.c
```diff
@@ -26454,11 +26454,24 @@ static BUILDIN(navigateto)
...
+	if (strcmp(func_name, "rodex_sendmail_acc") == 0 || strcmp(func_name, "rodex_sendmail_acc2") == 0) {
+		if (receiver_id < START_ACCOUNT_NUM || receiver_id > END_ACCOUNT_NUM) {
+			ShowError("script:rodex_sendmail: Invalid receiver account ID %d passed!\n", receiver_id);
+			return false;
+		}
+
+		msg->receiver_accountid = receiver_id;
+	} else {
+		if (receiver_id < START_CHAR_NUM) {
+			ShowError("script:rodex_sendmail: Invalid receiver character ID %d passed!\n", receiver_id);
```
AnnieRuru, line 26469: `ShowError("script:rodex_sendmail:...`
should be
`ShowError("buildin_%s:...", func_name);`

look back at your old commit on getiteminfo 
https://github.com/HerculesWS/Hercules/commit/58e96dfb28b186f66409417a3233cb45302e5e1d
```
	ShowError("buildin_getiteminfo: Invalid item info type %d.\n", type);
```
or haru's one 
https://github.com/HerculesWS/Hercules/commit/58e96dfb28b186f66409417a3233cb45302e5e1d
``` 
	default:
		ShowError("buildin_setiteminfo: invalid type %d.\n", n);
```

src/map/script.c
```diff
@@ -26510,6 +26523,18 @@ static BUILDIN(rodex_sendmail)
...
 		}
 
+		if (!data_isint(script_getdata(st, param + 1))) {
+			ShowError("script:rodex_sendmail: Passed amount for item %d is not a number!\n", i + 1);
+			return false;
+		}
+
+		int amount = script_getnum(st, param + 1);
+
+		if (amount < 1 || amount > min(MAX_AMOUNT, SHRT_MAX)) {
```
AnnieRuru, line 26533: are you sure about using `min(MAX_AMOUNT, SHRT_MAX)` here ?
if let's say gravity decided to pump up the MAX_AMOUNT to INT_MAX

--> NOTE gravity already pump up their MAX_ITEM_ID from short into int

then it will become SHRT_MAX instead of actual MAX_AMOUNT ...
MAX_AMOUNT is define in _src\common\mmo.h_ for this reason

src/map/script.c
```diff
@@ -26567,6 +26592,19 @@ static BUILDIN(rodex_sendmail2)
...
+
+		if (!data_isint(script_getdata(st, param + 1))) {
+			ShowError("script:rodex_sendmail: Passed amount for item %d is not a number!\n", i + 1);
+			return false;
+		}
+
+		int amount = script_getnum(st, param + 1);
+
+		if (amount < 1 || amount > min(MAX_AMOUNT, SHRT_MAX)) {
+			ShowError("script:rodex_sendmail: Invalid amount %d passed for item %d!\n", amount, i + 1);
```
AnnieRuru, line 26604: that previous one is under a subfunction
but the main script function `BUILDIN_DEF(rodex_sendmail)` runs from both
https://github.com/HerculesWS/Hercules/blob/32045249e6f91f1d09db2b63ad9eefb04fd1c1e9/src/map/script.c#L27884-L27887
and needs to print actual `buildin_rodex_sendmail` or `buildin_rodex_sendmail_acc` like said before

if can't do then nevermind, this is optional

AnnieRuru (changes requested):

src/map/script.c
Kenpachi2k13: For the sake of consistency I used the same "function name" as the other error messages in that function.

src/map/script.c
Kenpachi2k13: Same here. For the sake of consistency I used the same "function name" as the other error messages in that function.

src/map/script.c
Kenpachi2k13: `msg.items[i].item.amount` is of type `short` and thus this limitation if correct for now.

4144 (approved):

src/map/script.c
AnnieRuru: so in the future if gravity change this into INT ...
they have to change this line too ?

Kenpachi2k13: **IF** Gravity one day will change this (which I strongly doubt) we'd have to change the data type of field `amount` in `struct item` (mmo.h) and field `count` in `struct PACKET_CZ_ADD_ITEM_TO_MAIL`/`struct PACKET_ZC_ADD_ITEM_TO_MAIL` (packets_struct.h).

MishimaHaruna (approved):

PR #2920, opened 2020-12-18
Adjusted Fire Expansion Skill Effects.

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

Added missing effects of Fire Expansion Level 3 and 4.

**Issues addressed:** 


4144:
look like you forgot define SC_FIRE_EXPANSION_TEAR_GAS_SOB

MrKeiKun:
> 
> 
> look like you forgot define SC_FIRE_EXPANSION_TEAR_GAS_SOB

already sending the fix :D

src/map/status.h
```diff
@@ -733,6 +733,7 @@ typedef enum sc_type {
 	SC_STRANGELIGHTS,
 	SC_DECORATION_OF_MUSIC,
 
+	SC_FIRE_EXPANSION_TEAR_GAS_SOB,
```
MishimaHaruna, line 736: Inserting this here changes the ID of all the SCs that follow (and the comments don't match the generated IDs anymore). It would be better to put it at the end, so that any SCs saved to the database won't turn into the wrong effect on the next restart after applying this patch

MishimaHaruna (changes requested):

MishimaHaruna (approved):

MishimaHaruna:
I rebased the branch and fixed a conflict in skill.c

PR #2928, opened 2021-01-01
Haste Effect ignores some statuses

MrKeiKun:
### Pull Request Prelude

### Changes Proposed
This PR should fix guyak item(haste/boost movement speed) where it should ignore certain statuses.

**Issues addressed:** 


MrKeiKun:
> Can you explain to me what the previous behavior was?
> 
> Judging from your commit it seems like it should already work as intended. What was wrong about it?

The Haste Potion was supposed to block a few specific slow effects, but a bug in the old code accidentally erased all your speed effects whenever it and one of those slows were active together. The fix changes it so those slows simply can't be applied to you at all while the potion is active, instead of trying to erase them afterward.

PR #2930, opened 2021-01-05
Adds enum for client side action types

Asheraf:
### Pull Request Prelude

### DIFF
--- a/src/map/packets_struct.h
+++ b/src/map/packets_struct.h
@@ -3945,6 +3945,13 @@ struct PACKET_ZC_PERSONAL_INFOMATION {
 DEFINE_PACKET_HEADER(ZC_PERSONAL_INFOMATION, 0x097b);
 #endif  // PACKETVER_MAIN_NUM >= 20120503 || PACKETVER_RE_NUM >= 20120502 || defined(PACKETVER_ZERO)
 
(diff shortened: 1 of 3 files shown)


src/map/packets_struct.h
```diff
@@ -3945,6 +3945,12 @@ struct PACKET_ZC_PERSONAL_INFOMATION {
 DEFINE_PACKET_HEADER(ZC_PERSONAL_INFOMATION, 0x097b);
 #endif  // PACKETVER_MAIN_NUM >= 20120503 || PACKETVER_RE_NUM >= 20120502 || defined(PACKETVER_ZERO)
 
+struct PACKET_CZ_REQUEST_ACT {
```
4144, line 3948: this change will break 2008 clients and older. See offsets in ``packets.h`` for ``clif->pActionRequest``

technically hercules still support some of them.

but may be it's time to drop them?

Asheraf: Oh right I missed that since I only looked at newer versions, I dont think we should drop support yet for old stuff if it's still working.. perhaps ifdef the struct with padding fields for those versions.

4144: yes because we don't know offsets for other old clients, you can only copy padding from already added packet versions

MishimaHaruna (approved):

Issue #2940, opened 2021-01-30
Skill Missing on Plagirism/Reproduce


MrKeiKun:
**Describe the bug**
Skill overwriting when copying skills between Plagirism and Reproduce makes one skill disappear.

**To Reproduce**
1. Copy Fireball via Plagiarism
2. Use Reproduce (enabled now)
3. Copy Napalm Beat via Reproduce
4. Use Reproduce again (disabled now)
5. Use Napalm Beat Again via Plagiarism
6. Use Reproduce again (enabled now)
7. Copy Fireball via Reproduce
8. Use Reproduce again (disabled now)
9. Napalm Beat Nowhere to be found in skill window

**Expected behavior**
This is after character select.. skill went back..
![image](https://user-images.githubusercontent.com/4362134/106359014-91379200-634a-11eb-9b25-dc1065780c64.png)

**Screenshots of the problem**
![image](https://user-images.githubusercontent.com/4362134/106358920-153d4a00-634a-11eb-8a65-b12697fd276b.png)

**System specs (please complete the following information):**
 - OS: Windows 10 Pro
 - Hercules Version: 2021-01-11
 - Mode: Renewal

PR #2948, opened 2021-02-12
Refactor Skill Copy via Plagiarism and Reproduce

### LINKED ISSUE #989: [RE] Unable to Reproduce skill when Preserve is active
In official you can copy skill when Preserve and Reproduce are active,
so it's possible to get 2 OF 1st or 2nd class skills at the same time.
Source: experience from idRO and iRO

MrKeiKun:
### Pull Request Prelude

### Changes Proposed
PR should:

- [x] Remove hard coded list of skills that can be copied by Reproduce
- [x] Refactor method of what skills can be plagiarized and reproduced
- [x] Remove `copyskill_restrict` config
- [x] Fix #989 and #2940
- [x] Players can copy skills from another player or monster(Except: **Reproduce** cannot copy skills from monsters)
- [x] Recheck total list of skills that can be copied under 3rd jobs. 

**Issues addressed:** 

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -1231,11 +1231,14 @@ static int skill_calc_heal(struct block_list *src, struct block_list *target, ui
  **/
 static int can_copy(struct map_session_data *sd, uint16 skill_id)
 {
-	int idx = skill->get_index(skill_id);
+	int cidx = skill->get_index(skill_id);
 	nullpo_ret(sd);
 
 	/// Checks if preserve is active and if skill can be copied by Plagiarism
-	if (!idx)
+	if (!cidx)
+		return 0;
+
+	if (sd->status.skill[cidx].id && sd->status.skill[cidx].flag == SKILL_FLAG_PLAGIARIZED)
 		return 0;
 
 	// Checks if preserve is active and if skill can be copied by Plagiarism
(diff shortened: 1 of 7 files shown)


src/map/skill.c
```diff
@@ -3514,12 +3518,12 @@ static int skill_attack(int attack_type, struct block_list *src, struct block_li
 	map->freeblock_lock();
 
 	// Plagiarism and Reproduce Code Block [KeiKun]
-	if (damage > 0 && dmg.flag&BF_SKILL && tsd
+	if (bl->type == BL_PC && damage > 0 && dmg.flag&BF_SKILL && tsd
```
skyleo, line 3524: Talking only about Plagiarism:
Um so you checked that in pre-renewal and renewal one can't copy skills from monsters, even if they're normal playerskills?

This seems wrong to me, at least for pre-renewal.

skyleo (changes requested):

src/map/skill.c
MrKeiKun: after weeks, i was able to test this under renewal. `bl->type` here is the target which is the player. you can still copy skills from monsters.

MishimaHaruna (approved):

PR #2956, opened 2021-02-22
Added SC_NO_RECOVER_STATE

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 

Added handicap state no recover item bonus.


src/map/pc.c
```diff
@@ -4052,7 +4052,21 @@ static int pc_bonus3(struct map_session_data *sd, int type, int type2, int type3
 			}
 
 			break;
+		case SP_STATE_NO_RECOVER_RACE:
+		{
+			uint32 race_mask = map->race_id2mask(type2);
```
4144, line 4057: look like wrong validation for type2
if ``race > RC_FORMLESS``, map->race_id2mask will return ``1 << race``. for example for ``RC_NONBOSS`` it will be ``1 << 12`` and this equal ``4096``

but size of no_recover_state_race up to RC_MAX

4144 (changes requested):

db/re/item_db.conf
```diff
@@ -126639,10 +126639,10 @@ item_db: (
 	EquipLv: 95
 	Subtype: "W_MACE"
 	Script: <"
-		bonus bUnbreakableWeapon,0;
-		bonus3 bAutoSpell,NPC_CRITICALWOUND,5,10;
-		bonus bHit,getrefine();
-		bonus bCritical,getrefine();
+		bonus bUnbreakableWeapon, 0;
+		bonus3 bStateNoRecoverRace, RC_Player, 10000, 10000;
+		bonus bHit, getrefine() / 2;
```
MishimaHaruna, line 126644: Is this bHit change intentional or accidental? There's no mention of it in the commit message

src/map/pc.c
```diff
@@ -9014,6 +9030,11 @@ static int pc_itemheal(struct map_session_data *sd, int itemid, int hp, int sp)
 		if (sd->sc.data[SC_BITESCAR]) {
 			hp = 0;
 		}
+		
+		if (sd->sc.data[SC_NO_RECOVER_STATE]) {
+			hp -= hp;
+			sp -= sp;
```
MishimaHaruna, line 9036: Any reason to do `hp -= hp` instead of just `hp = 0` (and the same for sp)? I think that'd be more clear to the reader

src/map/skill.c
```diff
@@ -1236,6 +1236,8 @@ static int skill_calc_heal(struct block_list *src, struct block_list *target, ui
 			hp += hp / 10;
 		if (sc->data[SC_VITALITYACTIVATION])
 			hp = hp * 150 / 100;
+		if (sc->data[SC_NO_RECOVER_STATE])
+			hp -= hp;
```
MishimaHaruna, line 1240: This looks like another `hp = 0` in disguise

src/map/skill.c
```diff
@@ -8067,16 +8072,26 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 						hp += hp / 10;
 						sp += sp / 10;
 					}
+					if (tsc->data[SC_NO_RECOVER_STATE]) {
+						hp -= hp;
+						sp -= sp;
```
MishimaHaruna, line 8077: Same here

src/map/skill.c
```diff
@@ -8787,6 +8804,10 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 						hp += hp / 10;
 						sp += sp / 10;
 					}
+					if (tsc->data[SC_NO_RECOVER_STATE]) {
+						hp += hp;
+						sp += sp;
```
MishimaHaruna, line 8809: What's the purpose of this? Why does it double the values of `hp` and `sp`?

MishimaHaruna (changes requested):

db/re/item_db.conf
MrKeiKun: oh i will separate this under another commit within the PR. this was from the sc file.

MishimaHaruna (approved):

PR #2958, opened 2021-02-28
Report to console if a missing message is requested

pazkero:
### Pull Request Prelude

### Changes Proposed

If show_msg() requests a message code which is not present in messages.conf, a "??" will be provided silently.
This raises a warning so maintainers and fork owners can debug and fix their messages.conf file.

(This should not happen normally).

**Issues addressed:** None

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -126,6 +126,8 @@ static const char *atcommand_msg(int msg_number)
 	if(atcommand->msg_table[0][msg_number] != NULL && atcommand->msg_table[0][msg_number][0] != '\0')
 		return atcommand->msg_table[0][msg_number];
 
+	ShowWarning("atcommand_msg: Invalid message number was specified: %d", msg_number);
+	Assert_report(0);
 	return "??";
 }
 


src/map/atcommand.c
```diff
@@ -126,6 +126,7 @@ static const char *atcommand_msg(int msg_number)
 	if(atcommand->msg_table[0][msg_number] != NULL && atcommand->msg_table[0][msg_number][0] != '\0')
 		return atcommand->msg_table[0][msg_number];
 
+	ShowWarning("atcommand_msg: Invalid message number was specified: %d", msg_number);
```
4144, line 129: probably need also show stack, because can be too hard find source of wrong message id.

for this you can add ``Assert_report(0)``

4144 (changes requested):

src/map/atcommand.c
pazkero: Done, thanks :)

4144 (approved):

MishimaHaruna (approved):

PR #2968, opened 2021-03-20
Add 3th missing jobs and do some other small fixes

4144:
### Pull Request Prelude

### Changes Proposed

Add missing 3th jobs and baby jobs.
Add enum values for new 4th stats.
Add some checks for missing job data.
Convert SP_ enum into plain list.
Convert JOB_ enum into macro list and use in some functions.
Add warning if in switch used not all enum values.


Zarbony:
Nice :)
Does this will add the missing and fix the wrong Rebellion and Summoner skills too?

4144:
skills? no. it add jobs/configs etc.

src/common/mmo.h
```diff
@@ -1291,15 +1304,39 @@ enum e_class {
...
+	JOB_BABY_GUNSLINGER = 4228,
+
+	JOB_BABY_REBELLION = 4229,
+
+	JOB_BABY_STAR_GLADIATOR2 = 4238,
+
+	JOB_STAR_EMPEROR = 4239,
+	JOB_SOUL_REAPER = 4240,
+
+	JOB_BADY_STAR_EMPEROR = 4241,
```
Jedzkie: `BABY`

4144: this changed in other commits to other list

conf/messages.conf
```diff
@@ -658,6 +658,19 @@
...
+675: Baby Star Gladiator
+676: Baby Soul Linker
+677: Baby Gunslinger
+678: Baby Rebellion
+679: Star Emperor
+680: Baby Star Emperor
+681: Soul Reaper
+682: Baby Soul Reaper
 
 //670-849 FREE (please start using from the top if you need, leave the 670+ range for new jobs)
```
EPuncker, line 675:
```suggestion
//683-849 FREE (please start using from the top if you need, leave the 683+ range for new jobs)
```

db/constants.conf
```diff
@@ -130,6 +130,7 @@ constants_db: {
 	Job_Paladin2:           4022
 
 	Job_Baby:               4023
+	Job_Baby_Novice:        4023
```
MishimaHaruna, line 133: Is there any difference between Job_Baby and Job_Baby_Novice? Any reason to have both? The baby variant of novice has always been Job_Baby so far

db/pre-re/job_db.conf
```diff
@@ -846,6 +846,47 @@ Soul_Linker: {
 			 900, 900, 900, 900, 900, 900, 900, 900, 900, 900,	// 131 - 140
 			 900, 900, 900, 900, 900, 900, 900, 900, 900, 900]	// 141 - 150
 }
+Soul_Linker: {
```
MishimaHaruna, line 849: There's already a `Soul_Linker` key in this db (see line 808)

MishimaHaruna (changes requested):

db/constants.conf
4144: in code was used one constant in scripts or db another. because this i add both

4144: or better remove it?

MishimaHaruna: Ah I see, makes sense. What do you think about choosing one and marking the other one as deprecated, so that scripts that use it will get a warning?

4144: probably simpler remove new constant

4144: removed

MishimaHaruna (approved):

PR #2972, opened 2021-04-06
Fix Abracadabra / Improvise / itemskill validation

MishimaHaruna:
### Pull Request Prelude

### Changes Proposed

This introduces various fixes to the Abracadabra / Improvise skill validation:

- autocast code is moved to after homun/merc skill ranges so that when any of those are triggered, autocast data is not purged
- abracadabra/improvise is no longer purged immediately after cast, by calling autocast_clear_current in the proper manner
- the abracadabra/improvise skill requirement bypass that was removed in #2657 and related pull requests is re-implemented
- the condition that would get the character stuck and unable to cast other skills until teleporting or relogging when rolling AL_WARP from abracadabra (i.e. because AL_WARP lands in clif_parse_UseSkillMap in 2 steps) is fixed, by not immediately clearing the data after the first one

All credits for the fixes go to Heka of Origins

**Issues addressed:**
- Replaces #2859
- #2823
- #2824

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -13236,14 +13236,14 @@ static void clif_parse_UseSkillToPosSub(int fd, struct map_session_data *sd, uin
 
 	pc->autocast_set_current(sd, skill_id);
 
-	/**
+	/*
 	 * When using clif_item_skill() to initiate the execution of ground skills,
-	 * the client sometimes passes 0 for the skill level in packet 0x0af4.
+	 * the client sometimes passes 0 for the skill level in packet 0x0af4 (and in some clients,
+	 * this may be memory garbage instead of 0).
 	 * In that case sd->autocast.skill_lv is used for the auto-cast data validation,
 	 * since clif_item_skill() is only used for auto-cast skills.
-	 *
-	 **/
-	skill->validate_autocast_data(sd, skill_id, (skill_lv == 0) ? sd->auto_cast_current.skill_lv : skill_lv);
+	 */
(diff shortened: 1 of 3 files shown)


src/map/clif.c
```diff
@@ -13108,6 +13100,15 @@ static void clif_useSkillToIdReal(int fd, struct map_session_data *sd, int skill
 	// Whether skill fails or not is irrelevant, the char ain't idle. [Skotlex]
 	pc->update_idle_time(sd, BCIDLE_USESKILLTOID);
 
+	pc->autocast_set_current(sd, skill_id);
```
4144, line 13103: moving this line after return not open any exploits with ``return``?
because after change auto cast not clean and not set, but for example homun or mercenacy skills can be invoked.
also return on non existing skill not clean auto cast.

src/map/clif.c
```diff
@@ -13258,6 +13248,17 @@ static void clif_parse_UseSkillToPosSub(int fd, struct map_session_data *sd, uin
 		return;
 	}
 
+	pc->autocast_set_current(sd, skill_id);
```
4144, line 13251: here look like same movement as in previous comment

src/map/clif.c
MishimaHaruna: From my understanding, the intention here was to make sure that if your homunculus or mercenary AI triggers a skill while you're taking aim to cast the skill from hocus pocus, it won't remove your autocast data. If you think that's unsafe (and it probably is, if autocast data can contain mercenary or homunculus skills), I'll undo this part

MishimaHaruna: I undid this part

src/map/clif.c
MishimaHaruna: Undid it, same as above

Issue #2982, opened 2021-04-12
item_rate_adddrop doesn't effect chain items


olsonpm:
**Describe the bug**
Cards which grant drops via `bAddMonsterDropChainItem` aren't affected by rates

**To Reproduce**
1. In `conf/map/battle/drops.conf` set `item_rate_adddrop` to 5000 (50x)
2. In `db/pre-re/item_chain.conf` set `ITMCHAIN_GEM` to 

```txt
ITMCHAIN_GEM: {
  Dark_Red_Jewel: 200
}
```

this will reduce the variability in the test where the base rate for Dark_Red_Jewel should be 2%

3. Set @autoloot, equip Jewel_Sword and kill a few porings.

**Expected behavior**
Dark_Red_Jewel should be gained 100% of the time

**System specs (please complete the following information):**
- OS: windows 10 version 10.0.18363 Build 18363
- Hercules Version v2021.03.08
- Mode: pre-renewal
- Packet version: 20200902
- Client type: sak

**Plugins used or source modifications**
n/a

**Additional context**
rates in `db/pre-re/item_chain.conf` are unaffected by item_rate_adddrop.  Personally I feel the configuration should be modified to encompass all rates via `src/map/itemdb.c -> itemdb_read_chains`.  This would mean the comment in `conf/map/battle/drops.conf` should be updated

current docs
```txt
// The rate adjustment for card-granted item drops.
```

suggested docs
```txt
// The rate adjustment for skill and script-granted extra drop bonuses
```

I'm willing to open a PR if that solution sounds good to you

Jedzkie:
@olsonpm sure go ahead. Go make a PR for this issue.

olsonpm:
thanks will do

PR #2983, opened 2021-04-14
Apply rates to chain item drops

olsonpm:
### Pull Request Prelude

### Changes Proposed
apply `battle_config.item_rate_adddrop` to rates defined in `item_chain.conf`

**Issues addressed:** #2982 


src/map/itemdb.c
```diff
@@ -1571,7 +1571,12 @@ static void itemdb_read_chains(void)
 				prev->next = &itemdb->chains[count].items[c - 1];
 
 			itemdb->chains[count].items[c - 1].id = data ? data->nameid : 0;
-			itemdb->chains[count].items[c - 1].rate = data ? libconfig->setting_get_int(entry) : 0;
+
+			int rate = data ? libconfig->setting_get_int(entry) : 0;
+			if (battle_config.item_rate_adddrop != 100)
+				itemdb->chains[count].items[c - 1].rate = rate*battle_config.item_rate_adddrop/100;
```
4144, line 1577: missing spaces near ``*`` and near ``/``

4144 (changes requested):

src/map/itemdb.c
olsonpm: so I copied that [from here](https://github.com/HerculesWS/Hercules/blob/fbd9c8f9df8022619dcae158664bb5e81c7ae4bf/src/map/pc.c#L2163)

```c
rate = rate*battle_config.item_rate_adddrop/100;
```

is there a linter/formatter/other tool I can use to catch these ?

4144: many code in hercules with wrong style. if if line changed or added in commit it should use correct style
don't need search other style issues. only follow style in new/changed code

olsonpm: thanks that makes sense.  for future prs though, are there any tools which will catch style issues like this?  like linters or formatters?

4144: no, only manual checking

dastgirp:
I Strongly suggest to add new config, as this would break custom changes for many servers whenever they update it.

There isn't any warning or info displayed to them that this flag now affects other drops as well.

src/map/itemdb.c
olsonpm: okay no problem, i'll make sure to check next time

olsonpm:
@dastgirp i am willing to change the pr to affect only cards, not all chain items, but as far as a new config is concerned - this is a bug fix as the current config says it affects all card drops should be affected by the rate.  and currently cards which give drops via item cain groups are not impacted

dastgirp:
> @dastgirp i am willing to change the pr to affect only cards, not all chain items, but as far as a new config is concerned - this is a bug fix as the current config says it affects all card drops should be affected by the rate.  and currently cards which give drops via item cain groups are not impacted

it says card-granted item drops, and not card drops.
That means bonuses which add additional drops.

Commit in which it was implemented: https://github.com/HerculesWS/Hercules/commit/ee8093a79840edc7169291ca69ac3d256190fc83 (commit message/readme have much more clear description)

olsonpm:
card granted item drops don't include item chain?  sorry i'm not in front of my computer, i can check the commit message another time

olsonpm:
brazer card is a card which applies an item chain drop.  i feel this config should cover that card

olsonpm:
Okay I read the commit message and changelog entry.  I still don't understand why the config would omit rates defined in the item-chain when they are dropped via cards.  That feels like a bug to me.

4144:
after your changes server look like crashing

About config option i think new option was requested because if bug exists for too long it converted to feature. I mean all servers know this bug exists and using it as is.

But i not sure is really need another option or not. may be ask other devs?

olsonpm:
Yeah I'll take a look at the pr errors when I'm back at my comp.

And a new config option makes sense considering the bug has existed for a long time.  I can do that

olsonpm:
So the error is caused by my lack of c knowledge - I didn't realize structs were copy when assigned (i.e. pass by value).

Is there a recommended way to avoid the boilerplate of `itemdb->chains[count].items[c - 1]` everywhere ?  I would like to use a variable for this but the only thing I can come up with is storing a pointer to the struct e.g.

```c
struct item_chain_entry *item = &itemdb->chains[count].items[c - 1];

if( prev )
  prev->next = item;

(*item).id = data ? data->nameid : 0;

int rate = data ? libconfig->setting_get_int(entry) : 0;
if (battle_config.item_rate_adddrop != 100)
  (*item).rate = rate * battle_config.item_rate_adddrop / 100;
else
  (*item).rate = rate;
// ...
```

that feels weird, but I also am not a c programmer so I don't know what's preferred

olsonpm:
@4144 @dastgirp the new config option was implemented.  If there are build errors I may not have time to get to them til monday/tuesday

4144 (approved):

4144:
probably better squash commits into one or into two

olsonpm:
sure thing

MishimaHaruna (approved):

Issue #2976, opened 2021-04-10
"Free cast" not affected by speed modifiers such as curse


olsonpm:
#### I'm unsure whether this is a bug because I don't know how to test on an official server.  I'm willing to test on an official server if someone tells me how

----

**Describe the bug**
When cursed, a sage can cast a spell and get 75% original movement speed due to free cast

**To Reproduce**
1.
```
@jobchange professor
@allskill
```
there's probably a better way to get cursed, but I ensured I had enough stats to survive a teddy bear with
```
str: 99
agi: 99
int: 99
dex: 19 (low to get longer cast time)

// can't be too high since otherwise you won't get cursed)
luk: 19
vit: 39
```
2. @warp ein_dun02
3. aggro a teddy bear and `@heal` until you get cursed
4. when cursed, warp somewhere safe such as `@go 0`
5. cast a spell that takes a while such as Heaven's Drive and notice you walk around as though curse doesn't affect you

**Expected behavior**
I would expect free cast to modify movement speed, not set it

**System specs (please complete the following information):**
 - OS: windows 10 version 10.0.18363 Build 18363
 - Hercules Version v2021.03.08
 - Mode: pre-renewal
 - Packet version: 20200902
 - Client type: sak

**Plugins used or source modifications**
n/a

**Additional context**
the code responsible is here

https://github.com/HerculesWS/Hercules/blob/stable/src/map/status.c#L5064

```c
if (sd && sd->ud.skilltimer != INVALID_TIMER)
{
  // ...
  } else if (pc->checkskill(sd, SA_FREECAST) > 0) {
    speed_rate = 175 - 5 * pc->checkskill(sd,SA_FREECAST);
  }
}
if (speed_rate == -1)
{
```

because speed_rate is set due to FREECAST being active, all the other speed modifiers are skipped.

PR #2987, opened 2021-04-20
ensure free cast modifies speed rate instead of sets it

olsonpm:
### Pull Request Prelude

\* not sure if `skillTimerIsValid` is considered a frowned-upon mixed case name.  Please recommend a better name if you'd prefer one

### Changes Proposed
Ensure free cast modifies the walking speed of a unit rather than sets it.  This is to fix a (potential) bug where a cursed unit gets 75% of their original walk speed by using free cast.

**Issues addressed:** 
#2976 

### DIFF
--- a/src/map/status.c
+++ b/src/map/status.c
@@ -5051,19 +5051,20 @@ static unsigned short status_calc_speed(struct block_list *bl, struct status_cha
 	if( sc == NULL || ( sd && sd->state.permanent_speed ) )
 		return (unsigned short)cap_value(speed,MIN_WALK_SPEED,MAX_WALK_SPEED);
 
-	if (sd && sd->ud.skilltimer != INVALID_TIMER)
-	{
-		if (sd->ud.skill_id == LG_EXEEDBREAK) {
-			speed_rate = 160 - 10 * sd->ud.skill_lv;
-		} else if ((skill->get_inf2(sd->ud.skill_id) & INF2_FREE_CAST_REDUCED) != 0) {
-			speed_rate = 175 - 5 * sd->ud.skill_lv;
-		} else if (pc->checkskill(sd, SA_FREECAST) > 0) {
-			speed_rate = 175 - 5 * pc->checkskill(sd,SA_FREECAST);
-		}
-	}
+	bool skillTimerIsValid = sd != NULL && sd->ud.skilltimer != INVALID_TIMER;
+
+	if (skillTimerIsValid && sd->ud.skill_id == LG_EXEEDBREAK)
+		speed_rate = 160 - 10 * sd->ud.skill_lv;
+
 	if (speed_rate == -1)
 	{
 		speed_rate = 100;
+		if (skillTimerIsValid) {
+			if ((skill->get_inf2(sd->ud.skill_id) & INF2_FREE_CAST_REDUCED) != 0)
+				speed_rate = 175 - 5 * sd->ud.skill_lv;
+			else if (pc->checkskill(sd, SA_FREECAST) > 0)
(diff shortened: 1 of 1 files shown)


src/map/status.c
```diff
@@ -5051,19 +5051,20 @@ static unsigned short status_calc_speed(struct block_list *bl, struct status_cha
...
-	{
-		if (sd->ud.skill_id == LG_EXEEDBREAK) {
-			speed_rate = 160 - 10 * sd->ud.skill_lv;
-		} else if ((skill->get_inf2(sd->ud.skill_id) & INF2_FREE_CAST_REDUCED) != 0) {
-			speed_rate = 175 - 5 * sd->ud.skill_lv;
-		} else if (pc->checkskill(sd, SA_FREECAST) > 0) {
-			speed_rate = 175 - 5 * pc->checkskill(sd,SA_FREECAST);
-		}
-	}
+	bool skillTimerIsValid = sd && sd->ud.skilltimer != INVALID_TIMER;
```
4144, line 5054: in all lines you forgot to add check for ``sd`` non null (``sd != NULL``). Only if sure what ``sd`` is not null it can be used as pointer

4144 (changes requested):

src/map/status.c
olsonpm: are you saying there is a bug in the existing logic ?  Because I didn't change that.  `sd && sd->ud.skilltimer != INVALID_TIMER` would resolve to false if sd is null, correct ?

olsonpm: here's a fiddle showing sd as NULL should work fine

https://www.mycompiler.io/view/Hzv9nGJ

4144: ah i not saw this one check for null. Then it can works
but still better check it like ``sd != NULL && sd-> ...`` for follow style
and some missing spaces below.

olsonpm: I can add the null check but I don't see any mention of it in the style guide.

Also what spaces am I missing ?  I'm not seeing anything missing [per the whitespace section](https://github.com/HerculesWS/Hercules/wiki/Coding-Style#32-spaces)

4144: ``sd != NULL`` because ``sd`` is not bool and better check it with some non bool value. and for pointers 0 is NULL.

for missing spaces this ``pc->checkskill(sd,SA_FREECAST)`` -> ``pc->checkskill(sd, SA_FREECAST)``

olsonpm: Thanks for explaining the null style, I'll try to keep types consistent.  And thanks for pointing out the space

olsonpm:
okay should be good.  Btw if you have time I'd appreciate if you explained why these commits were implemented

[64c4d5e](https://github.com/HerculesWS/Hercules/commit/64c4d5e) - adds FreeCastReduced/Normal to skill_db.conf
[8953417](https://github.com/HerculesWS/Hercules/commit/8953417) - implements FreeCastReduced/Normal

I was confused because no skills have those attributes assigned, so I was hoping to understand whether they were from original ragnarok or intended for custom use.

4144:
i know not much how it should be in official kro, because this i cant say is this flag official or custom

4144 (approved):
Code is fine, better squash into one commit.
About logic is this is official way or not idk

olsonpm:
yeah if someone has an official account and can test free cast that would help.  It feels like a bug which is why I wrote this PR, but I understand wanting to keep things the way they are.

I'll squash shortly

MishimaHaruna (approved):

PR #3001, opened 2021-05-06
Add reading and checking client flags

4144:
### Pull Request Prelude

### Changes Proposed
Client may send special flags in client_version field.
In this pr hercules will compare current server flags and client flags, and if flags different then client will be disconnected.
This will prevent from strange bugs if server or client was created with different features.

For now only supported feature flag is cash show with preview support.

For enable flags in client need enable this patch http://nemo.herc.ws/patches/SendClientFlags/

### DIFF
--- a/conf/login/login-server.conf
+++ b/conf/login/login-server.conf
@@ -146,6 +146,9 @@ login_configuration: {
 		// What version we would allow to connect? (if check_client_version is enabled)
 		client_version_to_connect: 20
 
+		// Check if client flags are the same with compiled hercules version
+		check_client_flags: true
+
 		//==================================================================
 		// Client hash checking system
 		//==================================================================
--- a/conf/login/login-server.conf
+++ b/conf/login/login-server.conf
@@ -146,8 +146,11 @@ login_configuration: {
 		// What version we would allow to connect? (if check_client_version is enabled)
 		client_version_to_connect: 20
 
-		// Check if client flags are the same with compiled hercules version
+		// Check if client flags are same with compiled hercules version.
 		check_client_flags: true
+
+		// Report error on console if wrong client flags detected.
+		report_client_flags_error: true
 
 		//==================================================================
 		// Client hash checking system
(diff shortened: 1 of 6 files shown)


conf/login/login-server.conf
```diff
@@ -146,6 +146,9 @@ login_configuration: {
 		// What version we would allow to connect? (if check_client_version is enabled)
 		client_version_to_connect: 20
 
+		// Check is client flags same with compiled hercules version
```
EPuncker, line 149:
```suggestion
		// Check if client flags are the same with compiled hercules version
```

EPuncker (changes requested):

conf/login/login-server.conf
Jedzkie: @EPuncker maybe remove the word `the`?

Samuel23:
@4144 Is it possible with this like cash shop changes?

If server side cash shop was reloaded and it when client opens the cash shop as it's not the same values with the reloaded server side it will be disconnected?

4144:
> @4144 Is it possible with this like cash shop changes?
> 
> If server side cash shop was reloaded and it when client opens the cash shop as it's not the same values with the reloaded server side it will be disconnected?

you want update cash shop after player connected to game?
then need send changed cash shop tabs to all players.

Samuel23:
> > @4144 Is it possible with this like cash shop changes?
> > If server side cash shop was reloaded and it when client opens the cash shop as it's not the same values with the reloaded server side it will be disconnected?
> 
> you want update cash shop after player connected to game?
> then need send changed cash shop tabs to all players.

I created a custom @reloadcashdb to refresh cash shop items, the thing is it doesn't reload in game unless you relog in the game, so what happens is when I change anything in the server side and a player who doesn't relog yet, see's the old values which can lead to a mismatch for example price or item changes

MishimaHaruna (approved):

Issue #3002, opened 2021-05-08
@reloadscript clears map flags for current instances


olsonpm:
**Describe the bug**
@reloadscript clears map flags from current instances

**To Reproduce**
1. go to the endless tower ([here are instructions](https://irowiki.org/wiki/Endless_Tower))
2. enter `@item Wing_Of_Fly`
3. try to use the fly wing and - you will be **unable** to.  This is due to the mapflag `noteleport`
4. enter `@reloadscript`
5. use the fly wing - you will be **able** to because the mapflags are not being re instantiated

**Expected behavior**
I'd expect the fly wing to fail after `@reloadscript`

**Screenshots**
n/a

**System specs (please complete the following information):**
- OS: windows 10 version 10.0.18363 Build 18363
- Hercules Version v2021.03.08
- Mode: pre-renewal
- Packet version: 20200902
- Client type: sak

**Plugins used or source modifications**
n/a

**Additional context**
I'm looking into why this happens - will create a PR if I figure it out

PR #3003, opened 2021-05-10
reload instance map flags

olsonpm:
### Pull Request Prelude

### Changes Proposed

Reloads instance map flags from the source map

Note: I spent a decent amount of time making sure this wasn't being done elsewhere during the reload script process.  However the instance code/structures still don't make a ton of sense to me so I could be misunderstanding.

**Issues addressed:**
#3002

### DIFF
--- a/src/map/instance.c
+++ b/src/map/instance.c
@@ -801,7 +801,7 @@ static void instance_force_destroy(struct map_session_data *sd)
  */
 static void instance_reload_map_flags(int instance_id)
 {
-	Assert_ret(instance->valid(instance_id));
+	Assert_retv(instance->valid(instance_id));
 
 	struct instance_data *curInst = &instance->list[instance_id];
 
--- a/src/map/instance.c
+++ b/src/map/instance.c
@@ -803,11 +803,11 @@ static void instance_reload_map_flags(int instance_id)
 {
 	Assert_retv(instance->valid(instance_id));
 
-	struct instance_data *curInst = &instance->list[instance_id];
+	const struct instance_data *curInst = &instance->list[instance_id];
 
-	for (int i = 0; i < instance->list[instance_id].num_map; i++) {
+	for (int i = 0; i < curInst->num_map; i++) {
 		struct map_data *dstMap = &map->list[curInst->map[i]];
-		struct map_data *srcMap = &map->list[dstMap->instance_src_map];
+		const struct map_data *srcMap = &map->list[dstMap->instance_src_map];
 
 		memcpy(&dstMap->flag, &srcMap->flag, sizeof(struct map_flag));
 
--- a/src/map/instance.c
+++ b/src/map/instance.c
@@ -794,6 +794,26 @@ static void instance_force_destroy(struct map_session_data *sd)
 	}
 }
 
+/**
+ * reloads the map flags from the source map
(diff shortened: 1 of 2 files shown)


src/map/instance.c
```diff
@@ -806,11 +826,12 @@ static void do_reload_instance(void)
 				break;
 		}
 
-		if( k != instance->list[i].num_map ) /* any (or all) of them were disabled, we destroy */
+		if( k != instance->list[i].num_map ) /* any (or all) of them were disabled, we destroy */ {
```
4144, line 829: should be ``if (k != instance->list[i].num_map) /* any (or all) of them were disabled, we destroy */ {``

4144 (changes requested):

src/map/instance.c
olsonpm: ah - sorry forgot that I should be updating the styles when I copy/paste

src/map/instance.c
```diff
@@ -794,6 +794,26 @@ static void instance_force_destroy(struct map_session_data *sd)
...
 }
 
+/**
+ * reloads the map flags from the source map
+ *
+ * @param instance_id
+ */
+static void instance_reload_map_flags(int instance_id)
+{
+	struct instance_data *curInst = &instance->list[instance_id];
```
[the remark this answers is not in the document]
olsonpm: thanks will do.

olsonpm:
Alright the feedback should be addressed

src/map/instance.c
```diff
@@ -794,6 +794,27 @@ static void instance_force_destroy(struct map_session_data *sd)
...
+/**
+ * reloads the map flags from the source map
+ *
+ * @param instance_id
+ */
+static void instance_reload_map_flags(int instance_id)
+{
+	Assert_retv(instance->valid(instance_id));
+
+	struct instance_data *curInst = &instance->list[instance_id];
```
Asheraf, line 806: please mark this as `const`

src/map/instance.c
```diff
@@ -794,6 +794,27 @@ static void instance_force_destroy(struct map_session_data *sd)
...
+
+	for (int i = 0; i < instance->list[instance_id].num_map; i++) {
```
Asheraf, line 808: use `curInst->num_map`

src/map/instance.c
```diff
@@ -794,6 +794,27 @@ static void instance_force_destroy(struct map_session_data *sd)
...
+		struct map_data *dstMap = &map->list[curInst->map[i]];
+		struct map_data *srcMap = &map->list[dstMap->instance_src_map];
```
Asheraf, line 810: mark this as const as well

Asheraf (changes requested):

src/map/instance.c
olsonpm: thanks - I haven't used const in c before.  Interesting it can pick up dstMap shouldn't be const.  I'm used to javascript

src/map/instance.c
olsonpm: ah - thanks

olsonpm:
alright feedback round 2 should be set

MishimaHaruna (approved):

PR #3010, opened 2021-06-05
Update .mailmap

EPuncker:
### Pull Request Prelude

### Changes Proposed
Updated my old email with a valid one

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -2,7 +2,7 @@
 
 <Angelmelody@gmail.com> <cutepeipei@hotmail.com.tw>
 <Asheraf@users.noreply.github.com> <acheraf1998@gmail.com>
-Evil Puncker <EPuncker@users.noreply.github.com> <gm.perflex@gmail.com>
+Evil Puncker <EPuncker@users.noreply.github.com> <epuncker@gmail.com>
 <cannelle@users.noreply.github.com> <starfish2700@gmail.com>
 <dev@herc.ws> <Hercules@efficiently.awesome>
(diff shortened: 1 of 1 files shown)


.mailmap
```diff
@@ -2,7 +2,7 @@
 
 <Angelmelody@gmail.com> <cutepeipei@hotmail.com.tw>
 <Asheraf@users.noreply.github.com> <acheraf1998@gmail.com>
-Evil Puncker <EPuncker@users.noreply.github.com> <gm.perflex@gmail.com>
```
4144, line 5: You sure you never used this email in any commits before? ``gm.perflex@gmail.com``
if it was used, it will be visible as is in old commits.

probably you should add new line with new email?

also you can check how ``.mailmap`` affecting commits, by using for example ``git log`` emails what you see is emails changed based on ``.mailmap`` file

MishimaHaruna: I believe what he meant to do is to make `epuncker@gmail.com` visible instead of `EPuncker@users.noreply.github.com` and `gm.perflex@gmail.com` (@EPuncker please confirm) - if so, I'll amend the commit so that it actually does that since right now it tries to do the opposite (mailmap syntax is `<new> <old>`, not the other way around)

EPuncker: > I believe what he meant to do is to make `epuncker@gmail.com` visible instead of `EPuncker@users.noreply.github.com` and `gm.perflex@gmail.com` (@EPuncker please confirm) - if so, I'll amend the commit so that it actually does that since right now it tries to do the opposite (mailmap syntax is `<new> <old>`, not the other way around)

yes, that is it, I confirm

PR #3013, opened 2021-06-07
Convert random monster databases to libconfig

Asheraf:
### Pull Request Prelude


tools/mobgroupconverter.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python
```
4144, line 1: better change here to ``python3``
because code using python3 only code

4144 (changes requested):

4144 (approved):

src/map/mob.c
```diff
@@ -5367,70 +5363,109 @@ static void mob_mobavail_removal_notice(void)
...
+	int i = 0;
+	struct config_setting_t *entry = NULL;
+
+	while ((entry = libconfig->setting_get_elem(it, i++)) != NULL) {
+		int class_ = 0;
+		const char *name = libconfig->setting_get_string_elem(entry, 0);
+
+		if (!script->get_constant(name, &class_)) {
+			ShowWarning("%s: Invalid monster '%s' in \"%s\", skipping..\n", __func__, name, source);
+			return false;
```
MishimaHaruna, line 5448: This should probably `continue` instead of returning since you wrote "skipping" in the message above

src/map/mob.c
```diff
@@ -5367,70 +5363,109 @@ static void mob_mobavail_removal_notice(void)
...
 		}
-		if (i && !summon[i].qty) { //At least have the default here.
-			summon[i].class_[0] = mob->db_data[0]->summonper[i];
-			summon[i].qty = 1;
+
+		struct mob_db *m = mob->db(class_);
+		if (m == mob->dummy) {
+			ShowWarning("%s: Invalid monster '%s' in \"%s\", skipping..\n", __func__, name, source);
+			return false;
```
MishimaHaruna, line 5454: Same here, this should probably `continue`

src/map/mob.c
```diff
@@ -5965,6 +6000,11 @@ static void mob_reload(void)
 			item_drop_ratio_db[i] = NULL;
 		}
 	}
+	for (i = 0; i < MOBG_MAX_GROUP; i++) {
+		VECTOR_CLEAR(*mob->mob_groups[i]);
+		aFree(mob->mob_groups[i]);
```
MishimaHaruna, line 6005: This might be attempting to free a NULL pointer if a specific group was never initialized.

But does this have to be an array of pointers? Have you considered making it an array of structs (i.e. `struct mob_group mob_groups[MOBG_MAX_GROUP];`)? Any drawbacks?

MishimaHaruna (changes requested):

src/map/mob.c
Asheraf: Initially I wanted to remove the limit which is why it's a pointer, but with the current use cases you're right an array of structs is more appropriate.

MishimaHaruna (approved):

PR #3043, opened 2021-09-21
De-hardcode skill range modifier in skill_get_range2.

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

Moved hard coded skill based range modifier to `skill_db.conf`

**Issues addressed:** 

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -1072,6 +1072,7 @@ static int skill_get_casttype2(int index)
 static int skill_get_range2(struct block_list *bl, int skill_id, int skill_lv)
 {
 	int range;
+	int inf = 0;
 	struct map_session_data *sd = BL_CAST(BL_PC, bl);
 	if( bl->type == BL_MOB && battle_config.mob_ai&0x400 )
 		return 9; //Mobs have a range of 9 regardless of skill used.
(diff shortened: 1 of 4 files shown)


src/map/skill.c
```diff
@@ -1084,73 +1085,35 @@ static int skill_get_range2(struct block_list *bl, int skill_id, int skill_lv)
...
-		case HT_CLAYMORETRAP:
-		case RA_CLUSTERBOMB:
-		case RA_FIRINGTRAP:
-		case RA_ICEBOUNDTRAP:
-			if (sd != NULL)
-				range += (1 + pc->checkskill(sd, RA_RESEARCHTRAP))/2;
+	inf = skill->get_inf2(skill_id);
+
+	if (inf & INF2_RANGE_VULTURE) {
+		if (sd != NULL)
```
4144, line 1091: for avoid multiply checks for ``sd != NULL`` probably better move sd check before ``inf`` checks.
should be something like this
```
if (sd != NULL) {
// inf checks where sd is not NULL
} else {
// inf checks where sd is NULL
}
```

4144 (changes requested):

src/map/skill.c
MrKeiKun: great idea <3

src/map/skill.c
```diff
@@ -1084,73 +1085,29 @@ static int skill_get_range2(struct block_list *bl, int skill_id, int skill_lv)
...
-		case HT_LANDMINE:
-		case HT_FREEZINGTRAP:
-		case HT_BLASTMINE:
-		case HT_CLAYMORETRAP:
-		case RA_CLUSTERBOMB:
-		case RA_FIRINGTRAP:
-		case RA_ICEBOUNDTRAP:
-			if (sd != NULL)
-				range += (1 + pc->checkskill(sd, RA_RESEARCHTRAP))/2;
+	inf = skill->get_inf2(skill_id);
```
4144, line 1088: if ``inf`` used from this line only here better define it and not add at start of function

4144 (changes requested):

src/map/skill.c
MrKeiKun: oh yeah. i might one line the whole declare, assign and removing the initialized value. But after checking, `skill_get_inf2` will return `0` as long skill_idx aint `0`.

4144 (approved):

MishimaHaruna (approved):

PR #3049, opened 2021-09-30
Update some party related packets to use structures

Asheraf:
### Pull Request Prelude

### Changes Proposed

change some party packets to use structures.


src/map/clif.c
```diff
@@ -7469,76 +7456,73 @@ static void clif_party_option(struct party_data *p, struct map_session_data *sd,
...
-		clif->send(buf, len+9, &sd->bl, PARTY);
+		char buf[CHAT_SIZE_MAX + sizeof(struct PACKET_ZC_NOTIFY_CHAT_PARTY)];
+		struct PACKET_ZC_NOTIFY_CHAT_PARTY *packet = (struct PACKET_ZC_NOTIFY_CHAT_PARTY *)&buf;
+
+		packet->PacketType = HEADER_ZC_NOTIFY_CHAT_PARTY;
+		packet->PacketLength = len + sizeof(struct PACKET_ZC_NOTIFY_CHAT_PARTY);
+		packet->AID = account_id;
+		safestrncpy(packet->chatMsg, mes, len + 1);
+
+		const struct map_session_data *sd = p->data[i].sd;
```
4144, line 7509: look like ci failed because here ``const``

src/map/clif.c
```diff
@@ -7337,31 +7330,25 @@ static void clif_party_invite(struct map_session_data *sd, struct map_session_da
...
-	WFIFOW(fd,0) = 0x2c5;
-	safestrncpy(WFIFOP(fd,2),nick,NAME_LENGTH);
-	WFIFOL(fd,26) = result;
-	WFIFOSET(fd,packet_len(0x2c5));
-#endif
+	WFIFOHEAD(fd, sizeof(struct PACKET_ZC_PARTY_JOIN_REQ_ACK));
+	struct PACKET_ZC_PARTY_JOIN_REQ_ACK *p = WFIFOP(fd, 0);
+
+	p->PacketType = HEADER_ZC_PARTY_JOIN_REQ_ACK;
+	memcpy(p->characterName, nick, NAME_LENGTH);
```
MishimaHaruna, line 7349: This can copy past the end of the buffer, in case the function is called with `nick` pointing to a buffer shorter than NAME_LENGTH (see for example party.c:407: `clif->party_inviteack(sd, "", 7);`). Returning to safestrncpy would be safer.

There are other functions as well that do this risky operation (and are currently only called with fixed size buffers) that we might want to sanitize as well...

MishimaHaruna (changes requested):

PR #3070, opened 2021-10-23
Convert elemental damage table to libconfig

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
This PR converts `(pre-)re/attr_fix.txt` to libconfig, becoming `(pre-)-re/attr_fix.conf`. Although it is not a file that people usually touches, reading and editing it is quite hard and confusing. A converter script is also included.

The database file structure was chosen thinking about how most sites usually lists elemental damage, for example, you see a monster that has element X LvY, and then you have a table with how much damage each attacking element will give you. I think it is the most natural way for people to view it, but I am open to suggestions.

I chose to add the new tables in a separate commit than the conversion so the commits are smaller, but if you prefer, I can squash them, no problem.

To check if the converted table is correct, you can use this code after the table is read. It works for both master and this branch, and must generate the same output:
```C
for (i = 0; i < 4; i++) {
    for (j = ELE_NEUTRAL; j < ELE_MAX; j++) {
        for (k = ELE_NEUTRAL; k < ELE_MAX; k++) {
            printf("%d, ", battle->attr_fix_table[i][j][k]);
        }
        printf("\n");
    }
    printf("\n");
}
```


MishimaHaruna (approved):

MrKeiKun:
@guilherme-gm @MishimaHaruna 
```
pc.c: In function ‘pc_read_attr_fix_db_entry’:
pc.c:11994:20: warning: ‘%d’ directive writing between 1 and 10 bytes into a region of size 3 [-Wformat-overflow=]
   sprintf(name, "Lv%d", i);
                    ^~
pc.c:11994:17: note: directive argument in the range [1, 2147483646]
   sprintf(name, "Lv%d", i);
                 ^~~~~~
pc.c:11994:3: note: ‘sprintf’ output between 4 and 13 bytes into a destination of size 5
   sprintf(name, "Lv%d", i);
   ^~~~~~~~~~~~~~~~~~~~~~~~
        LD      map-server
```

PR #3083, opened 2021-11-06
Fix a regression in the mob drop rate calculation

MishimaHaruna:
### Pull Request Prelude

### Changes Proposed

The value was accidentally always overridden to be 900%. With the intent to cap to 90%, `max(rate, 90000)` erroneously used instead of `min(rate, 9000)`.

**Issues addressed:** N/A (reported in discord)


MrKeiKun:
@MishimaHaruna why change to min? that means even if `@iteminfo` calculates it to **100%** it will just drop to **90%**
since before my PR it was `drop_rate = max(drop_rate, cap_value((int)(0.5 + drop_rate * (sd->sc.data[SC_OVERLAPEXPUP]->val2) / 100.), 0, 9000));` which is `max`...

Issue #3097, opened 2021-12-13
Out of range homun exp using @homlevel


MrKeiKun:
**Describe the bug**
Trying to level homunculus to max using `@homlevel`

**To Reproduce**
1. `@makehomun 6004` or whatever id
2. `@homlevel 150` or `99` is fine depends on max level of homunculus
3. character select
4. check **char-server** and **map-server** console

**Expected behavior**
`out of range` warning should not show

**Logs**
* Map-Server
```
[SQL]: DB error - Out of range value for column 'exp' at row 1
[Debug]: at C:\Users\Kei\Desktop\HerculesWS\src\char\int_homun.c:102 - UPDATE `homunculus` SET `char_id`='150000', `class`='6004',`prev_class`='0',`name`='Vanilmirth',`level`='99',`exp`='14829735431805717965',`intimacy`='2100',`hunger`='32', `str`='1130', `agi`='1210', `vit`='1200', `int`='950', `dex`='1230', `luk`='1200', `hp`='9413',`max_hp`='9413',`sp`='322',`max_sp`='322',`skill_point`='33', `rename_flag`='0', `vaporize`='0', `autofeed`='0' WHERE `homun_id`='14'
```
* Char-Server
```
[Error]: homunculus data save failure for account 2000000
[Error]: homunculus data save failure for account 2000000
```

**System specs (please complete the following information):**
 - OS: Windows 10 21H1
 - Hercules Version: 202112010
 - Mode: Renewal and Pre-Renewal

PR #3098, opened 2021-12-14
Fix an issue with saving homunculus exp when it reaches max level

Asheraf:
### Pull Request Prelude

### Changes Proposed

This fixes an issue where `homun->get_exp` does into take into consideration that the exp value for max level is uninitialized, so it has been replaced with `homun->get_exp_next` which works the same way as `pc->nextbaseexp` and resolves the issue.

**Issues addressed:** 
#3097


4144:
look like need HPM update

4144 (approved):

MishimaHaruna (approved):

PR #3107, opened 2022-01-10
Few small optimizations to packets_struct.h

Lemongrass3110:
### Pull Request Prelude

### Changes Proposed
Moved some packet headers to correct locations and added their structs
Fixed some of the names to names that were leaked previously

**Issues addressed:** None


4144:
look like you put packets for wrong packet versions

Asheraf:
HPMDataCheck is an auto generated file using `tools/HPMHookGen/HPMDataCheckGen.pl`, do not manually change.

src/map/clif.c
```diff
@@ -23308,17 +23308,17 @@ static void clif_rodex_read_mail(struct map_session_data *sd, int8 opentype, str
 
 	int fd = sd->fd;
 	int body_len = (int)strlen(msg->body) + 1;
-	int size = sizeof(struct PACKET_ZC_ACK_READ_RODEX);
+	int size = sizeof(struct PACKET_ZC_ACK_READ_MAIL);
```
4144, line 23311: rename rodex into mail probably not good idea
because in herc rodex and mail packets based on different code. and because this should be not mixed.
technically in herc at same time can works old mail system and rodex.

4144 (changes requested):

src/map/clif.c
Lemongrass3110: grafik

No it cannot technically work at the same time in herc. :-)

Lemongrass3110: But feel free to adjust it to your likings. It was just a suggestion.
I have allowed edits by maintainers, so you can easily change it as well.

4144: i mean in herc mail and rodex is separate things, it may be not good, but it's like this for now.
this mean mail contains own packets for send/read etc and rodex contains own packets for do same. it more like duplication.
and while mail and rodex not merged, better name mail and rodex packets in different way.

4144 (changes requested):

4144 (commented):
look like fine, but better squash commits by features or all in one commit

4144 (approved):

src/map/packets_struct.h
```diff
@@ -401,6 +389,60 @@ DEFINE_PACKET_ID(ZC_PAR_4JOB_CHANGE, 0x0b25);
...
+DEFINE_PACKET_ID(ZC_COUPLESTATUS, 0x0141);
+
+#if PACKETVER_MAIN_NUM >= 20170906 || PACKETVER_RE_NUM >= 20170830 || defined(PACKETVER_ZERO)
+// TODO: Unconfirmed name
+struct PACKET_ZC_LONGLONGPAR_CHANGE {
+	int16 PacketType;
+	uint16 varID;
+	int64 amount;
+} __attribute__((packed));
+DEFINE_PACKET_ID(ZC_LONGLONGPAR_CHANGE, 0x0acb);
```
4144, line 443: here name should be ``ZC_LONGLONGPAR_CHANGE``

4144 (changes requested):

src/map/packets_struct.h
Lemongrass3110: You want me to remove the TODO?
Because the name is exactly like you wrote it

4144: ah yes sorry, i not saw what name is correct already

Asheraf: @Lemongrass3110 you can remove the TODO the name is confirmed.

src/map/packets_struct.h
```diff
@@ -236,10 +236,11 @@ enum packet_headers {
 #endif
 #if PACKETVER < 20100105
 	vendinglistType = 0x133,
+#elif PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200723
```
Asheraf, line 239: It's probably better to put this at the end instead of middle of the list so the reader can figure out which is the latest more easily.

src/map/packets_struct.h
```diff
@@ -3084,6 +3135,10 @@ struct PACKET_ZC_PC_PURCHASE_ITEMLIST_FROMMC_sub {
 #if PACKETVER >= 20160921
 	uint32 location;
 	uint16 viewSprite;
+#if PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200723
```
Asheraf, line 3138: move this out from being nested under the `#if PACKETVER >= 20160921`

src/map/packets_struct.h
```diff
@@ -3075,7 +3124,9 @@ struct PACKET_ZC_PC_PURCHASE_ITEMLIST_FROMMC_sub {
 #endif
 	uint8 identified;
 	uint8 damaged;
+#if !( PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200723 )
```
Asheraf, line 3127:
```suggestion
#if !(PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200723)
```

Asheraf (changes requested):

src/map/packets_struct.h
Lemongrass3110: Not really confirmed, but at least it seems to match the disassembly. :-)

src/map/packets_struct.h
Lemongrass3110: I thought it was not worth it, because I thought you will change it soon, to be defined where the struct is defined. With HEADER_*

src/map/packets_struct.h
Asheraf: It is confirmed :)

src/map/packets_struct.h
Asheraf: Yes but we also tend to copy the same ifdefs when doing the conversions, so for sake of clarity keep it sorted.

Lemongrass3110:
@Asheraf fixed :)

Asheraf (approved):

src/map/packets_struct.h
Lemongrass3110: Leaky leaky?
Guessing from function name is not confirming real names.

Lemongrass3110:
Rebased as requested from @4144 :)

MishimaHaruna (approved):

PR #3030, opened 2021-08-12
rA-Merge: Implemented Star Emperor and Soul Reaper

MrKeiKun:
### Pull Request Prelude

- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
Merged rAthena's Star Emperor and Soul Reaper

**Issues addressed:** 

[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style

PR #3128, opened 2022-04-04
Spiritball Update

MrKeiKun:
### Pull Request Prelude

### Changes Proposed
Follow-up to #3030. Merged both `clif_soulball` and `clif_spiritball` into one.

**Issues addressed:** 

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -3501,7 +3501,7 @@ ACMD(spiritball)
 	if( sd->spiritball > 0 )
 		pc->delspiritball(sd, sd->spiritball, 1);
 	sd->spiritball = number;
-	clif->spiritball(&sd->bl);
+	clif->spiritball(&sd->bl, BALL_TYPE_SPIRIT, AREA);
 	// no message, player can look the difference
 
 	return true;
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -1593,9 +1593,9 @@ static bool clif_spawn(struct block_list *bl)
 			struct map_session_data *sd = BL_UCAST(BL_PC, bl);
 			int i;
 			if (sd->spiritball > 0)
-				clif->spiritball(&sd->bl);
+				clif->spiritball(&sd->bl, BALL_TYPE_SPIRIT, AREA);
 			if (sd->soulball > 0)
-				clif->soulball(sd, NULL, AREA);
+				clif->spiritball(&sd->bl, BALL_TYPE_SOUL, AREA);
 			if (sd->state.size == SZ_BIG) // tiny/big players [Valaris]
 				clif->specialeffect(bl,423,AREA);
 			else if (sd->state.size == SZ_MEDIUM)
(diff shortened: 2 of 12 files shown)


src/map/map.h
```diff
@@ -439,6 +439,16 @@ enum elements {
...
 };
 
+/**
+ * Types of Ball Types
+ * Used by clif_spiritball [KeiKun]
+ */
+enum spirit_ball_types {
+	BALL_TYPE_NONE = 0,
+	BALL_TYPE_SPIRIT,
+	BALL_TYPE_SOUL
```
4144, line 449: you sure it using names like this?

4144 (changes requested):

src/map/map.h
MrKeiKun: I just used those to avoid using `int` as type. is there an official names for it?

4144: i cant find real enum, but is this can support ball type bigger than 2? is client draw it in some way?

MrKeiKun:
will squash later after another HPM Hook Update~

4144 (approved):

src/map/atcommand.c
```diff
@@ -3501,7 +3501,7 @@ ACMD(spiritball)
 	if( sd->spiritball > 0 )
 		pc->delspiritball(sd, sd->spiritball, 1);
 	sd->spiritball = number;
-	clif->spiritball(&sd->bl, BALL_TYPE_SPIRIT, AREA);
+	clif->spiritball(&sd->bl, 1, AREA);
```
MishimaHaruna, line 3504: I'm not really a fan of magic numbers, can you revert this to use an enum? It makes the code a lot easier to read

MishimaHaruna (changes requested):

src/map/atcommand.c
4144: this parameter set how many spirits should be spawned.
i don't think make any sense to use enum here.

MishimaHaruna: But then why does it say this

```
 * Ball Type:
 *      0 = None
 *      1 = Spirit
 *      2 = Soul
```

4144: ah here it not assigned directly to ``num`` then yes can be used enum

src/map/clif.c
```diff
@@ -1593,9 +1593,9 @@ static bool clif_spawn(struct block_list *bl)
 			struct map_session_data *sd = BL_UCAST(BL_PC, bl);
 			int i;
 			if (sd->spiritball > 0)
-				clif->spiritball(&sd->bl);
+				clif->spiritball(&sd->bl, 1, AREA);
 			if (sd->soulball > 0)
-				clif->soulball(sd, NULL, AREA);
+				clif->spiritball(NULL, 2, AREA);
```
Lemongrass3110, line 1598: You lost the &sd->bl here

PR #3136, opened 2022-04-28
Update mob.c

zaid-naim:
Right now if you try to put 100% drop rate on drops.conf, it will show correctly on @mi and @rates but monsters will not really drop items by 100% chance. Monsters will only drop items by 90% chance even if @mi and @rates say 100%. This change fixes that.

### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 

### DIFF
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -2745,7 +2745,7 @@ static int mob_dead(struct mob_data *md, struct block_list *src, int type)
 
 					// When PK Mode is enabled, increase item drop rate bonus of each items by 25% when there is a 20 level difference between the player and the monster.[KeiKun]
 					if (battle_config.pk_mode && (md->level - sd->status.base_level >= 20))
-						drop_rate_bonus += 25; // flat 25% bonus 
+						drop_rate_bonus += 25; // flat 25% bonus
 
 					drop_rate_bonus += sd->dropaddrace[md->status.race] + (is_boss(src) ? sd->dropaddrace[RC_BOSS] : sd->dropaddrace[RC_NONBOSS]); // bonus2 bDropAddRace[KeiKun]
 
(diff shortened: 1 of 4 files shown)


src/map/mob.c
```diff
@@ -2757,8 +2757,8 @@ static int mob_dead(struct mob_data *md, struct block_list *src, int type)
 
 					drop_rate = (int)(0.5 + drop_rate * drop_rate_bonus / 100.);
 
-					// Limit drop rate, default: 90%
-					drop_rate = min(drop_rate, 9000);
+					// Make sure drop rate does not go beyond 100% after the drop rate modifiers are applied
+					drop_rate = min(drop_rate, 10000);
```
MrKeiKun, line 2761: I am not sure why previous developers decided to use cap rate of 90%, but one thing is better to de-hardcode the value 90% into battle-conf.

MrKeiKun (changes requested):

src/map/mob.c
guilherme-gm: I think the 90% cap is the official behavior: https://irowiki.org/wiki/Drop_System
> There is a cap of 90% drop chance on items boosted by anything except database edits. That means if an item has a 50% chance to drop and it receives a 300% bonus, it will still be capped at 90% instead of 100%.

MrKeiKun:
I now understand the real problem here.

```
			droprate = monster->dropitem[i].p;

#ifdef RENEWAL_DROP
			if( battle_config.atcommand_mobinfo_type ) {
				droprate = droprate * pc->level_penalty_mod(monster->lv - sd->status.base_level, monster->status.race, monster->status.mode, 2) / 100;

				if (droprate <= 0 && !battle_config.drop_rate0item)
					droprate = 1;
			}
```
since mobinfo itself doesn't display the idea of the 90% cap, it provides rates when using `@mi` that are not similar to rates when monster died and tries to drop an item (which in mob.c).

dastgirp:
90% cap is official behavior for boosted items as far as I know.

The possible changes for this to work as you want could be:
1. Make 90% configurable (just like other drop rates)
2. Fix mi to show 90% wherever applicable.
(Or both)

MishimaHaruna:
I made this configurable and adjusted the behavior to be closer to what irowiki describes. Database edits and server drop rates will bypass this, as it used to be before commit 5401f961cf (except for an unintended behavior that made it possible for a positive bonus to actually lower the drop rate by making the cap kick in)

PR #3141, opened 2022-05-11
Fixes on instance information/destruction

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
This PR fixes some issues around instance information and destruction, that could cause visual glitches and assertion failures. Also, I have changed the response numbers into an enum to make it clear. Those are the issues addressed:

**Issue 1:** When players were moved out of an instance due to it being expired, they were receiving the information that the instance actually went Idle instead of destroyed. That was caused because when moving players out of the instance, it is still considered a valid instance, and this triggers the idle check.

To fix that, instances are now marked as "being destroyed", and thus invalid, before moving players out of it.

**Issue 2:** When a player had a Player Instance attached to them, even after being destroyed, they were still receiving the instance info when logging back in. That was caused because the check for player instance on login was not taking into account if the instance was freed or not (via `.state`).

To fix that, instance data is now fully cleaned up upon destruction (using `memset` instead of only zeroing a few fields) and the state check was added to the login process. `memset` step is not necessarily required, but it would avoid future issues if someone forget to check the state.

**Issue 3:** An assert failure when reloading scripts after playing an instance. That was caused due to instance reload trying to recreate a freed instance.

**Issue 4:** An assert failure when shutting the server down after playing an instance. That was caused due to hercules trying to free the already freed instance map.

**Issue 5:** In some cases, the wrong reason code was provided when an instance gets destroyed. It now checks for instance idle state to help choosing the proper reason.

### DIFF
--- a/src/map/instance.c
+++ b/src/map/instance.c
@@ -170,7 +170,7 @@ static int instance_create(int owner_id, const char *name, enum instance_owner_t
 		}
 	}
 
-	clif->instance(i, 1, 0); // Start instancing window
+	clif->instance(i, INSTANCE_WND_INFO_CREATE, 0); // Start instancing window
 	return i;
 }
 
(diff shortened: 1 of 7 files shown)


src/map/instance.h
```diff
@@ -31,10 +31,19 @@ struct map_session_data;
 
 #define INSTANCE_NAME_LENGTH (60+1)
 
+/**
+ * true if instance is in an active/playable state.
+ * In other words, if a player can interact with it.
+ * 
+ * @param inst instance_data to be checked
+ */
+#define instance_is_active(inst) (inst.state == INSTANCE_IDLE || inst.state == INSTANCE_BUSY)
```
4144, line 40: here better add ``(``) for ``inst``
```
#define instance_is_active(inst) ((inst).state == INSTANCE_IDLE || (inst).state == INSTANCE_BUSY)
```

4144 (changes requested):

src/map/instance.h
guilherme-gm: thanks, fixed and rebased

src/map/instance.h
```diff
@@ -31,10 +31,19 @@ struct map_session_data;
...
+ * @param inst instance_data to be checked
+ */
+#define instance_is_active(inst) ((inst).state == INSTANCE_IDLE || (inst).state == INSTANCE_BUSY)
+
 typedef enum instance_state {
 	INSTANCE_FREE,
 	INSTANCE_IDLE,
-	INSTANCE_BUSY
+	INSTANCE_BUSY,
+	INSTANCE_DESTRYOING,
```
MrKeiKun, line 46: `INSTANCE_DESTROYING`

src/map/instance.c
```diff
@@ -619,6 +619,9 @@ static void instance_destroy(int instance_id)
 			iptr[i] = -1;
 	}
 
+	// mark it as being destroyed so server doesn't try to give players more information about it
+	instance->list[instance_id].state = INSTANCE_DESTRYOING;
```
MrKeiKun, line 623: `INSTANCE_DESTROYING`

src/map/instance.h
guilherme-gm: oops. thanks! fixed/rebased in both places

MishimaHaruna (approved):

PR #3148, opened 2022-08-03
Replace safesnprintf

MishimaHaruna:
### Pull Request Prelude

### Changes Proposed

Since Hercules requires C99, `snprintf()` is always available and has a predictable truncation behavior identical to `safesnprtinf()`, making the latter redundant.

They only differ in return value in case of truncation (`snprintf()` returns the amount of characters that would have been written if the buffer wasn't limited, while `safesnprintf()` returns a negative value). Every place where it is used in Hercules currently ignores the return value.

This PR removes `safesnprintf()` and replaces it with `snprintf()`, since they're equivalent in every existing call location, but the latter allows the compiler (especially gcc) to produce better diagnostics about too small buffers.

Code has been adjusted to remove the warnings that were generated, either by increasing buffer sizes in order to make truncation impossible, or specifically handling it after the fact. In the case of gcc-7, where the -Wformat-truncation diagnostic is more naive (ignores the actual values assigned to variables and always assumes maximum range for their type), the warning is now disabled since it's less meaningful than the ones generated by the currently supported versions of the compiler.

**Issues addressed:** N/A


src/common/strlib.h
```diff
@@ -70,6 +70,19 @@
...
 
+/**
+ * Workaround for gcc issues 80354, 89312, that make it difficult to ignore
+ * -Wformat-truncation warnings in those rare cases where truncation is
+ * irrelevant to the program's correctness.
+ *
+ * @remark Do not use unless you're sure about the implications of ignoring
+ * this warning.
+ */
+#define snprintf_ignore_truncation(str, size, format, ...) do { \
```
4144, line 81: pragma with ignore -Wformat-truncation not works?
i mean in same was as in enum switch for example.

src/char/int_mail.c
```diff
@@ -337,7 +337,11 @@ static bool inter_mail_return_message(int char_id, int mail_id, int *new_mail)
 		safestrncpy(msg.dest_name, temp_, NAME_LENGTH);
 
 		// set reply message title
-		snprintf_ignore_truncation(temp_, MAIL_TITLE_LENGTH, "RE:%s", msg.title); // String truncation expected and ignored
+		PRAGMA_GCC7(GCC diagnostic push)
```
4144, line 340: i not sure, but may be pragma can works inside macro?

src/common/strlib.h
MishimaHaruna: Good idea, it worked. I removed the macro and replaced it with the pragmas in the few places where it was used (I didn't like the macro since it'd make it too tempting to ignore possible real issues)

src/char/int_mail.c
MishimaHaruna: they could, with the alternate _Pragma syntax - but I didn't like the macro at all, it made it too tempting to use it instead of resolving an actual issue with the code

4144: ok

4144 (approved):

Issue #2109, opened 2018-06-27
Mobs (visually) reappear after death under certain circumstances


drdaxxy:
[//]: # (**********************************)
[//]: # (** Fill in the following fields **)
[//]: # (**********************************)

[//]: # (Note: Lines beginning with syntax such as this one, are comments and will not be visible in your report!)

### Issue Prelude

[//]: # (Please complete these mandatory steps and check the following boxes by putting an `x` inside the brackets _before_ filing your issue)

-  [x] I have not modified the source prior to reproducing this issue.
-  [x] I am using the latest version of Hercules.
-  [x] I am aware that this report will be closed or deleted if it becomes obvious that I am stating the false.

### Description

Under some circumstances, e.g. with certain skills causing knockback like *Magnum Break*, players may see mobs that have died reappear with 0 HP in an idle animation.

### Current Behavior

Hercules may send `ZC_NOTIFY_MOVEENTRYn` packets concerning units that have already died. If the client has already removed the unit from its state before receiving that packet, it will recreate it, thus displaying a stuck mob with 0 HP to the player until they relog. (I believe this is not a client bug as this packet type may also be used to notify clients of units they have legitimately not seen before.)

A fairly consistent the problem manifests is when mobs whose movement speed has changed through *Increase AGI* die to some knockback skills. In these cases, the offending packet is sent at https://github.com/HerculesWS/Hercules/blob/c419726752ebd1b532f487e5683f63232a6c237b/src/map/clif.c#L1866 However, this is not necessarily the *only* place that can send it. I'm aware of MvP slaves sometimes displaying the glitch after their master dies on a live server (with source modifications), however I've not been able to reproduce that on unmodified Hercules, personally.

Lowering the unit removal notification delay in https://github.com/HerculesWS/Hercules/blob/c419726752ebd1b532f487e5683f63232a6c237b/src/map/mob.c#L2728 makes the issue occur more frequently but is not required.

Issue has been reproduced on `2018-06-21aRagexeRE` and `2018-04-04cRagexeRE` but most likely affects clients many years older as well.

### Expected Behavior

`ZC_NOTIFY_MOVEENTRYn` packets, as crafted in https://github.com/HerculesWS/Hercules/blob/c419726752ebd1b532f487e5683f63232a6c237b/src/map/clif.c#L1303 should never be sent after `ZC_NOTIFY_VANISH` (unless the mob is intended to be recreated on client).

### Steps To Reproduce The Issue

[...]

Issue #2047, opened 2018-05-25
Knockback effects on monsters.


AnisotropicDefixation:
There are a few issues regarding this and they're all very old.

1. Knocking back a monster that's affected by a hard status effect (namely SC_STUN) will cause it to "walk" back to you and then teleport back to its real location once the status ends. This is most easily reproduced by skills that both knockback and stun like CR_SHIELDCHARGE or RL_SLUGSHOT used on monsters with fast movement speed. It can also happen when a monster was previously stunned by something and then knocked back by another skill. 

2. Using knockback effects on monsters that are unable to move (SC_STOP, SC_ANKLESNARE, etc) causes them to move a bit and sometimes act in completely crazy ways like walking long distances backwards and similar.

3. Applying any status effect that disables movement, then knocking the monster back, then hitting it with a ranged attack. This will cause the monster to "dance" back and forth, moving a couple of cells when hit.

In all cases, those glitches are merely visual but impair your ability to deal with monsters as their real location isn't what it seems. Also, while it is possible to fix those issues on a local server by adding a short canmove tick upon a knockback or removing amotion/dmotion (seems to be related to the flinch animation?) from skills that knock back, it won't work on a real server due to the latency involved.

PR #3163, opened 2022-10-11
Stop units from moving when dead

csnv:
### Pull Request Prelude

### Changes Proposed

Stop units from moving and, consequently, notifying the client when the unit is already dead and the server is going to send a notification to vanish such unit. Fixes mobs walking around or standing when killed.

This issue has been already discussed here https://github.com/HerculesWS/Hercules/issues/2109 with some insights, however, I think this patch is fine as this "move request" should never be done after death. Nonetheless and, as precaution, this check could be made against BL_MOB only, which uses a timer and it's the main reason this visual bug appears.

**Issues addressed:** 
https://github.com/HerculesWS/Hercules/issues/2047
https://github.com/HerculesWS/Hercules/issues/2109

### DIFF
--- a/src/map/unit.c
+++ b/src/map/unit.c
@@ -175,6 +175,9 @@ static int unit_walk_toxy_sub(struct block_list *bl)
 	if (ud == NULL)
 		return 2;
 
+	if (status->isdead(bl))
+		return 1;
+
 	struct walkpath_data wpd = {0};
 
 	if (!path->search(&wpd, bl, bl->m, bl->x, bl->y, ud->to_x, ud->to_y, ud->state.walk_easy, CELL_CHKNOPASS))
--- a/src/map/unit.c
+++ b/src/map/unit.c
@@ -361,6 +361,9 @@ static int unit_walk_toxy_timer(int tid, int64 tick, int id, intptr_t data)
 	if (ud->walkpath.path_pos >= ud->walkpath.path_len)
 		return 1;
 
+	if (status->isdead(bl)) // Should not be able to move
+		return 1;
+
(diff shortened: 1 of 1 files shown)


skyleo:
I think this likely doesn't fix all cases. Imagine the timer has already been created and while the timer hasn't triggered yet, the mob is marked dead, then the walking command is still send when timer triggers.

csnv:
> I think this likely doesn't fix all cases. Imagine the timer has already been created and while the timer hasn't triggered yet, the mob is marked dead, then the walking command is still send when timer triggers.

That's exactly the case I am able to reproduce consistently and the dead condition check is included in the function that is called from the timer. Unless there's another walking command I'm unaware of, this should fix it.

4144:
> > I think this likely doesn't fix all cases. Imagine the timer has already been created and while the timer hasn't triggered yet, the mob is marked dead, then the walking command is still send when timer triggers.
> 
> That's exactly the case I am able to reproduce consistently and the dead condition check is included in the function that is called from the timer. Unless there's another walking command I'm unaware of, this should fix it.

Probably need add also dead check in start of ``unit_walk_toxy_timer``.
After ``unit_walk_toxy_sub`` complete with error status from function ``unit_walk_toxy_timer`` can be sent packets related to mob. I think this is wrong.

csnv:
Done, I've not been able to reproduce any more bugs before this last check, but it makes sense that we don't let things move around when they're already dead.
I'll leave the check in `unit_walk_toxy_sub` too as it's called in some other functions.

4144 (approved):

MishimaHaruna (approved):

PR #3165, opened 2022-10-14
Fixed signedness of image size

Lemongrass3110:
### Pull Request Prelude

- [X] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

**Issues addressed:** 
Image size is 59454, which is > INT16_MAX.
Therefore uint16 should be used.
The rest of the code should also be checked, as it uses int32 instead of int16 in general.

### DIFF
--- a/src/map/packets_struct.h
+++ b/src/map/packets_struct.h
@@ -5032,7 +5032,7 @@ DEFINE_PACKET_HEADER(ZC_MYGUILD_BASIC_INFO, 0x014c);
 struct PACKET_CZ_REQ_UPLOAD_MACRO_DETECTOR {
 	int16 PacketType;
 	char answer[16];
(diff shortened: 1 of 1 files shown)


4144 (approved):

MishimaHaruna (approved):
Thank you. The rest of the code seems reasonably safe. The only places that could use some improvement are the `clif_macro_detector_request_init()` and `clif_captcha_preview_request_init()` functions where the downcast to uint16 occurs.
Not strictly necessary since their input value is originally 16 bit so it should in theory never be larger than that, but to be on the safe side, an `Assert_retv(image_size >= 0 && image_size <= UINT16_MAX)` would be enough

dastgirp (approved):

PR #3166, opened 2022-10-17
quest log fixes and quest atcommand

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
This PR introduces a few fixes to quest log:
1. Quests with 3 objectives doesn't fail an assertion anymore.

2. Quests using monster size, race or element as condition were giving out some wrong labelling.

For example, a quest for "hunting fire monsters" would give something like "fire formless monsters", while actually just checking for fire element. That was due to the code not checking whether a race/element of 0 meant no race/element, or the race/element that actually has ID = 0 (as they are 0-based).

3. Fixes the bitmask value for size

Quests that asked for a monster size were not being sent correctly to the client. I tested on 2019-12-24. I will be honest... I went by trial and error until the label showed up... so I am not sure if this changes in newer clients (@Asheraf do you know if they change at some future client?)

4. Fixes mobid not being sent on ZC_HUNTING_QUEST_INFO for special cases (map name / race / size / element)

This prevented the client from displaying the targets when the player first got the quest, requiring the user to log out / back in to actually see their objectives.

---

Also, I took the chance to add `@quest` atcommand. This atcommand allows the user to manipulate their quest log, I find it useful for debugging/development purposes (saves the need to make npcs/db changes to add/remove quests from your character)

**Issues addressed:** 
None, I think

### DIFF
--- a/conf/messages.conf
+++ b/conf/messages.conf
@@ -1692,5 +1692,20 @@
 1529: 1 item has been graded.
 1530: %d items have been graded.
 
+// @quest
+1531: Usage: @quest <action> <quest id>. Action:
+1532: -- "add" or "set": adds this quest to the player (as a new quest)
+1533: -- "del" or "delete": removes this quest from player's log
+1534: -- "complete": marks this quest as complete
+1535: This quest does not exist.
+1536: You already have this quest.
+1537: Failed to add quest.
+1538: Quest added to your log.
+1539: You don't have this quest in your log.
+1540: Failed to erase quest.
+1541: Quest was removed from your log.
+1542: Failed to complete quest.
+1543: Quest was marked as complete in your log.
+
 // Custom translations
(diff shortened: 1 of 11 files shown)


4144 (approved):

Asheraf (approved):
Looks good, I also checked in a couple clients and it does seem to use 0x10, 0x20, 0x40 for sizes. not sure from where I came up with the other ones, might have been just a mistake on my side.

4144:
> Looks good, I also checked in a couple clients and it does seem to use 0x10, 0x20, 0x40 for sizes. not sure from where I came up with the other ones, might have been just a mistake on my side.

it was probably from char sizes constants.

guilherme-gm:
Thanks for checking!

MishimaHaruna (approved):

PR #3183, opened 2022-12-17
Added skeleton function for generating user token

csnv:
### Pull Request Prelude

### Changes Proposed
Includes function for generating an auth token.

Intended for plugin use.

**Issues addressed:** 


src/login/lclif.c
```diff
@@ -293,6 +293,7 @@ static bool lclif_send_server_list(struct login_session_data *sd)
 	packet->packet_id = HEADER_AC_ACCEPT_LOGIN;
 #else
 	packet->packet_id = HEADER_AC_ACCEPT_LOGIN2;
+	login->generate_token(sd, packet->auth_token);
```
dastgirp, line 296: Is this needed? You can use postHook and create a method in the plugin itself.

csnv: AFAIK, unless we override the entire function with a custom one, we cannot hook into the packet info to send the token, as the token is not currently being sent.
Otherwise the plugin would just send this packet again after herc has sent it as response to the user logging in, and I don't know if the client would update the token in that case.

src/login/login.c
```diff
@@ -1368,6 +1368,11 @@ static void login_auth_failed(struct login_session_data *sd, int result)
 	lclif->auth_failed(fd, ban_time, result);
 }
 
+// Generates auth token for web service
+static void login_generate_token(struct login_session_data* sd, unsigned char *auth_token) {
```
4144, line 1372: at function start ``{`` should be on new line

csnv: Fixed.

src/login/lclif.c
4144, line 296: here need add packet version check. because in old packet version field ``auth_token`` not exists and will be compilation error

4144 (changes requested):

src/login/lclif.c
csnv: This line is inside the `#else` of `#if PACKETVER < 20170315` so it should be compiling correctly. Also Github's checks api hasn't found any problem when compiling.

4144: ah yes, sorry

4144 (approved):

EPuncker (approved):

MishimaHaruna (approved):

Issue #3186, opened 2023-01-12
RODEX Bug (Items not reaching Player and Disappearing from mail)


Mojacko:
**Describe the bug**
Rodex mark the mail as you receive the Item even if you did not actually get the item. And making the Mail Item gone.

**To Reproduce**
1. Get a about 93~96 items in your inventory
2. Must have about 20 Mail with items.
3. Try "retrieve all"
4. You will only get few items but the Items from Mail is gone too.

**Expected behavior**
Should not remove items from the mail if the player did not successfully receive the Items.

**Screenshots**
This is the screenshot of Bug.
I have 20 Mail with each having 1 Egg Item. 
[Image 1](https://i.imgur.com/alimy8K.png)
You can see in Image 1 that I only get 7 Eggs but in mail all Item is gone even though the Box icon is still showing.

I tried to Relog my char and open mail
[Image 2](https://i.imgur.com/tyhD7YR.png)

You can see that all mail have mark as Item received but my inventory still has only 7 Eggs.

**System specs (please complete the following information):**
 - OS: [Ubuntu 18.04]
 - Hercules Version v2022.12.07
 - Mode: Pre-renewal
 - Packet version: 2019-06-05fRagexe
 - Client type: [main]

**Plugins used or source modifications**
No Plugin Loaded

Mojacko:
The same issue with [3009](https://github.com/HerculesWS/Hercules/issues/3009)

dastgirp:
This happens because 
1. There's no inventory space left
2. Weight exceeds

There needs to be additional check while doing retrieve all to ensure there's no loss.

Mojacko:
This only happens to "Retrieve all" button. but if trying to retrieve with single mail the weight checking is correct.

PR #3189, opened 2023-01-21
Queue up rodex item claim requests

### LINKED ISSUE #3009: Rodex retrieve-all inventory-slot check
**Describe the bug**
Rodex's retrieve all doesn't check the available number of slots of the character forces some items to be retrieved from the mail and the leftovers are permanently stucked in the mail itself.

**To Reproduce**
1. Force inventory to have 99/100 inventory slots
2. **Retrieve All** a particular rodex mail that has 2 or more distinct items attached to it

**Expected behavior**
Restrict user to **Retrieve all** when the available inventory slots cannot handle all the items attached to the mail.

 - OS: CentOS 8 x64
 - Hercules Version: master
 - Mode: Renewal

Asheraf:
### Pull Request Prelude

### Changes Proposed

attempt to fix #3009 & #3186, likely needs to be applied for zeny as well.

### DIFF
--- a/src/map/rodex.c
+++ b/src/map/rodex.c
@@ -30,6 +30,7 @@
 #include "common/nullpo.h"
 #include "common/sql.h"
 #include "common/memmgr.h"
+#include "common/showmsg.h"
 
 
 // NOTE : These values are hardcoded into the client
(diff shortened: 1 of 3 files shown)


src/map/rodex.c
```diff
@@ -570,6 +588,14 @@ static void rodex_get_items(struct map_session_data *sd, int8 opentype, int64 ma
 		return;
 	}
 
+	// Queue the mail ID to be claimed
+	VECTOR_ENSURE(sd->rodex.claim_list, 1, 1);
+	VECTOR_PUSH(sd->rodex.claim_list, mail_id);
```
MishimaHaruna, line 593: When this function is called recursively by rodex_getItemsAck (while items are queued), this causes the next queued item to be inserted twice. I think the push should only happen when this is called directly, and not when this is (re)called during queue processing.

Also, if the second call to this function fails (because i.e. inventory has become full after retrieving one item and moving on to the second item in queue), queue processing gets aborted since rodex_getItemsAck is never called again, but we still have leftover items in the queue that will not be removed since the VECTOR_ERASE only happens there

MishimaHaruna (changes requested):

MishimaHaruna (approved):

PR #3215, opened 2023-06-13
Some skill fixes

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
Small fixes to some skills. Applies to renewal & pre-renewal.

Commit messages copy/pasted below with some formatting

#### MO_KITRANSLATION
- the skill should fail and not consume requirements when:
   1. being cast on Gunslinger class
   2. being cast in a player that already have 5 spheres
   3. being cast in non-players (but friend -- e.g. Mercenary)

for those cases it should send USESKILL_FAIL cause

Currently, they were consuming SP and not showing message.

#### MO_ABSORBSPIRITS
- it should fail and not consume SP when
   1. the target player doesn't have spheres
   2. the target player is of Gunslinger class (due to never having
      spheres)
   3. the target player is friendly
- it should fail and consume SP when
   1. Used in BOSS monster
   2. Used in non-boss monster, but failing the 20% check
   3. Using in non-player units (like mercenaries)

Currently, they were consuming SP and not showing message.

#### WM_POEMOFNETHERWORLD
- it should not be placed over another instance of the skill
- when trying to do so, it should show a position error to client

Reference: kRO 2012.02.22 patch note ([Link](https://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=6162&curpage=65))

#### Message when you don't have enough Coins
Right now, when the player doesn't have enough coins to use a skill, it shows "Missing Spirit Spheres" error.

With this change, we now show "Needs x coins" / "Needs coins" (client messages)

**Issues addressed:** 
None, I think

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -6471,7 +6471,7 @@ static int skill_castend_id(int tid, int64 tick, int id, intptr_t data)
 
 		if( sd )
 		{
-			if( !skill->check_condition_castend(sd, ud->skill_id, ud->skill_lv) )
+			if (!skill->check_condition_castend(sd, ud->skill_id, ud->skill_lv, target))
 				break;
 			else
 				skill->consume_requirement(sd,ud->skill_id,ud->skill_lv,1);
@@ -11843,7 +11843,7 @@ static int skill_castend_pos(int tid, int64 tick, int id, intptr_t data)
 
 		if( sd )
 		{
-			if( ud->skill_id != AL_WARP && !skill->check_condition_castend(sd, ud->skill_id, ud->skill_lv) ) {
+			if (ud->skill_id != AL_WARP && !skill->check_condition_castend(sd, ud->skill_id, ud->skill_lv, NULL)) {
 				if( ud->skill_id == SA_LANDPROTECTOR )
 					clif->skill_poseffect(&sd->bl,ud->skill_id,ud->skill_lv,sd->bl.x,sd->bl.y,tick);
 				break;
@@ -12065,7 +12065,7 @@ static int skill_castend_map(struct map_session_data *sd, uint16 skill_id, const
 					return 0;
 				}
 
-				if(!skill->check_condition_castend(sd, sd->menuskill_id, lv)) { // This checks versus skill_id/skill_lv...
+				if (!skill->check_condition_castend(sd, sd->menuskill_id, lv, NULL)) { // This checks versus skill_id/skill_lv...
 					skill_failed(sd);
 					return 0;
 				}
@@ -16475,7 +16475,16 @@ static bool skill_items_required(struct map_session_data *sd, int skill_id, int 
 	return false;
 }
 
-static int skill_check_condition_castend(struct map_session_data *sd, uint16 skill_id, uint16 skill_lv)
+/**
+ * Checks conditions for a skill to be executed. This check happens after the cast time was completed.
+ * 
+ * @param sd The character who cast the skill.
+ * @param skill_id The skill's ID.
+ * @param skill_lv The skill's level.
+ * @param target The unit who was targeted by this skill (MAY BE NULL, if there is no target)
+ * @return 1 if conditions were satisfied. 0 otherwise (errors are also sent to client)
+ */
+static int skill_check_condition_castend(struct map_session_data *sd, uint16 skill_id, uint16 skill_lv, struct block_list *target)
 {
 	struct skill_condition require;
 	struct status_data *st;
@@ -16593,7 +16602,7 @@ static int skill_check_condition_castend(struct map_session_data *sd, uint16 ski
 			}
 			break;
 		default:
-			if (!skill->check_condition_castend_unknown(sd, &skill_id, &skill_lv))
+			if (!skill->check_condition_castend_unknown(sd, &skill_id, &skill_lv, target))
 				break;
 			return 0;
 	}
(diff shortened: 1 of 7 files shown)


src/map/skill.c
```diff
@@ -16321,8 +16332,14 @@ static int skill_check_condition_castbegin(struct map_session_data *sd, uint16 s
 			break;
 		default:
 			if (sd->spiritball < require.spiritball) {
-				clif->skill_fail(sd, skill_id, USESKILL_FAIL_SPIRITS, require.spiritball, 0);
-			return 0;
```
guilherme-gm, line 16325: note: this return was wrongly indented (see line 16323 opens a `{`)

src/map/skill.c
```diff
@@ -16321,8 +16332,14 @@ static int skill_check_condition_castbegin(struct map_session_data *sd, uint16 s
...
+				if ((sd->job & MAPID_BASEMASK) == MAPID_GUNSLINGER) {
+					// For some reason, the client only shows the singular message when it bType is 0 (instead of 1).
+					int btype = (require.spiritball == 1 ? 0 : require.spiritball);
```
guilherme-gm, line 16337: I was wondering whether we should make a new variable for coins (be it for requirements only or entirely). but since they seem to always force gunslinger to be away of spirit spheres, I guess we are safe..

src/map/skill.c
```diff
@@ -7708,24 +7708,31 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 		case MO_ABSORBSPIRITS:
 		{
 			int sp = 0;
+			bool success = false;
 			if (dstsd != NULL && dstsd->spiritball != 0
 			 && (sd == dstsd || map_flag_vs(src->m) || (sd && sd->duel_group && sd->duel_group == dstsd->duel_group))
 			 && (dstsd->job & MAPID_BASEMASK) != MAPID_GUNSLINGER
 			 ) {
```
guilherme-gm, line 7715: I think some of those checks could be cleaned up because they are being checked in `skill_check_condition_castend` now.

But in the other hand, `skill_check_condition_castend` has a few checks for autocast at the beginning that could allow the "invalid skill" to get here... so I am not sure if it is safe to clear the checks here.

src/map/skill.c
```diff
@@ -16582,6 +16589,23 @@ static int skill_check_condition_castend(struct map_session_data *sd, uint16 ski
...
+				clif->skill_fail(sd, skill_id, USESKILL_FAIL, 0, 0);
+				return 0;
+			}
+
+			if (sd != tgtsd && battle->check_target(&sd->bl, &tgtsd->bl, BCT_NOENEMY) == 1) {
+				clif->skill_fail(sd, skill_id, USESKILL_FAIL, 0, 0);
+				return 0;
+			}
+			break;
+		}
```
guilherme-gm, line 16617: I am not sure what is the proper way to use blocks around case per Herc's style guide. I couldn't find a mention for this. And this function actually uses 2 different modes. (see NC_MAGICDECOY below vs AM_SPHEREMINE above) I picked based on my tastes / avoiding an extra tab

src/map/skill.c
MishimaHaruna: I agree it's a bit confusing that they share the same variable. It might have been implemented this way because I believe even officially they're the same thing (until they realized that there were unexpected interactions between them, through bestowment or card-granted spheres).

If we split them, we need to explicitly take care of all those interactions, and convert between spheres and coins where necessary. Otherwise they can stay as they are, until the day we'll need to have both at once

guilherme-gm: yeah, I agree. guess it is better to leave this way for now. thanks!

src/map/skill.c
guilherme-gm: those: (reformatted for comments)
```C
if (
    /* Only for player targets */
    dstsd != NULL
    /* AND must have spheres (1) */
    && dstsd->spiritball != 0
    /* AND must be (self OR enemy) (2) */
    && (sd == dstsd || map_flag_vs(src->m) || (sd && sd->duel_group && sd->duel_group == dstsd->duel_group))
    /* AND NOT gunslinger (3) */
    && (dstsd->job & MAPID_BASEMASK) != MAPID_GUNSLINGER
) {
       /* Skill works */
}
```

from the updated `skill_check_condition_castend` (when not being from auto cast):
```
case MO_ABSORBSPIRITS: {
	struct map_session_data *tgtsd = BL_CAST(BL_PC, target);
	// Non-player targets would just continue
	if (tgtsd == NULL)
 		break;

	// Fail if 0 spheres (1)
	// Fail if gunslinger (3)
	if (tgtsd->spiritball == 0 
	    || (tgtsd->job & MAPID_BASEMASK) == MAPID_GUNSLINGER) 
		/* Fail */

	// Fail if using in a friend (not self) (2)
	if (sd != tgtsd
	    && battle->check_target(&sd->bl, &tgtsd->bl, BCT_NOENEMY) == 1)
		/* Fail*/

	/* Pass */
```

so in a normal skill usage, used against a `sd`, we would only get here when:
1. target is not gunslinger
2. target has 1 or more spheres
3. target is an enemy or self

so both checks are equivalent and simply checking if `dstsd != NULL` should be enough.

but we have these lines for auto cast: https://github.com/HerculesWS/Hercules/pull/3215/files#diff-759fea9dcd40522e86217dd352aadd687488dff49c67aa880610d4aeaaeb89c0R16515-R16521

src/map/skill.c
guilherme-gm: good point! thanks :D

src/map/skill.c
MishimaHaruna: I see what you mean. I agree that some of the check are redundant, but I wouldn't remove all of them since I believe that some here have a slightly different purpose, especially the gunslinger one.

In `skill_check_condition_castend()` they're deciding whether to make the skill fail entirely, while here they're preventing unintended side effects caused by the shared implementation between spirit spheres and gunslinger coins.

Let's say that tomorrow we introduce a new SC (or, alternatively, a plugin introduces a new SC) that makes the caster bypass requirements and cast skills unconditionately, bypassing code in `skill_check_condition_castend`. If the gunslinger check is removed from here, the skill would have a different, implementation-defined, behavior consuming coins even if it's supposed to consume spirit spheres. I'd feel like that behavior is a bug caused by the missing check here, and the correct behavior would be for the skill to have no effect since there are no spheres to remove.

The vs check though can be removed from here, since it isn't protecting implementation details that would break the skill's behavior, but simply implementing a game mechanic that imposed the skill should only be used on friendly units.

I believe that each piece of code should not rely on the way it is currently called, and should assume that plugins or future changes may call it differently, as long as they don't do anything illegal. And it should offer protection against illegal inputs that could lead to unexpected behavior or crashes (and document what's illegal) through assertions.

guilherme-gm: that's a very good point of view! thank you, Haru.
I have update the commit remove the vs check only

MishimaHaruna (approved):

PR #3216, opened 2023-06-18
Add skill_db2.conf and some SkillDB code fix/simplification

guilherme-gm:
### Pull Request Prelude

#### Add skill_db2.conf
This follows the same idea of the existing item_db2.conf / mob_db2.conf, allowing users to create skill_db entries (or override existing ones).

Just like the item/mob dbs, it also includes the `Inherit` field to allow changing only parts of a skill.

#### Code simplification
This is a suggestion to reduce the amount of if/else's in the flags parsing. Instead replacing them with an array of key/values.

I applied it to the functions related to skill_db.conf that had a long list and were only doing that. There is a few other cases with smaller lists which I skipped.

I have used this small diff to validate if the changes did not cause any side effect:
[compare_skills.diff.txt](https://github.com/HerculesWS/Hercules/files/11782956/compare_skills.diff.txt)

If you apply it to master, build and run, you will get a list of skill flags. Save it. Change to this branch and do the same thing. Both should present the same output.

#### Case enforcing and fix
During the simplification I also changed the comparison from `strcmpi` to `strcmp`. For our base code, it only triggered issues with `SameGuild` flag (which was being used as `Sameguild`. Custom changes will trigger warnings for people to work and use the correct case.

**Issues addressed:** 
None, I think


src/map/skill.c
```diff
@@ -21927,159 +21927,57 @@ static void skill_validate_skillinfo(struct config_setting_t *conf, struct s_ski
...
-					sk->inf2 &= ~INF2_RANGE_RESEARCHTRAP;
-			} else if (strcmpi(skill_info, "AllowPlagiarism") == 0) {
-				if (on)
-					sk->inf2 |= INF2_ALLOW_PLAGIARIZE;
-				else
-					sk->inf2 &= ~INF2_ALLOW_PLAGIARIZE;
-			} else if (strcmpi(skill_info, "None") != 0) {
-				ShowWarning("%s: Invalid sub-type %s specified for skill ID %d in %s! Skipping sub-type...\n",
+			int j = 0;
+			ARR_FIND(0, ARRAYLENGTH(info_list), j, strcmpi(skill_info, info_list[j].name) == 0);
```
guilherme-gm, line 21970: Since we have a validation for invalid types, this array ends up being linear-searched for each key in config, but from my tests it doesn't seem to get too slow (striked times were ignored due to being too far from the average):
|         | Current code  | New code      |
| ------- | ------------- | ------------- |
|         | 0.000418s     | ~~0.000589s~~ |
|         | 0.000436s     | 0.000431s     |
|         | 0.000483s     | 0.000441s     |
|         | 0.000477s     | 0.000445s     |
|         | 0.000443s     | 0.000438s     |
|         | 0.000450s     | 0.000445s     |
|         | ~~0.000567s~~ | 0.000440s     |
|         | ~~0.000731s~~ | 0.000466s     |
|         | 0.000434s     | 0.000470s     |
|         | 0.000415s     | 0.000477s     |
|         |               |               |
| Fastest | 0.000415s     | 0.000431s     |
| Slowest | 0.000477s     | 0.000477s     |

I ran it in WSL2, calculating the time around the function call -- I can try in a real Linux if needed.

MishimaHaruna: Linear search on the array shouldn't be very different from the chain of if-else conditions (that has the same length as the array and mostly does the same operations), so indeed I wouldn't expect much of a difference, if at all.

To me it basically looks like the opposite of a loop-unroll optimization, but it doesn't make sense to unroll a loop with this amount of operations, at the risk of typos or swapped values in the long if-else chain, so I like what you did here.

I agree it should be strcmp since the DB keys are meant to be case sensitive

guilherme-gm: Thank you, Haru!

In this case, I will take the chance to change this too

MishimaHaruna (approved):

PR #3219, opened 2023-07-17
Guillotine Cross Poisoning Weapon skill fixes

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
Solves some issues with Guillotine Cross' Poisoning Weapon skill:
1. It had some issues with the effect rates and duration calculation
2. It was applying a very wrong duration for Venom Bleed poison
3. It was not showing a result message upon poison application

References:
- https://github.com/rathena/rathena/pull/6263
- https://www.divine-pride.net/database/item/12724/venom-bleed

**Issues addressed:** 
None, I think. Bug was found by Lexi on discord

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -3878,11 +3878,8 @@ static int skill_attack(int attack_type, struct block_list *src, struct block_li
 			case GC_VENOMPRESSURE:
 			{
 				struct status_change *ssc = status->get_sc(src);
-				if( ssc && ssc->data[SC_POISONINGWEAPON] && rnd()%100 < 70 + 5*skill_lv ) {
-					short rate = 100;
-					if ( ssc->data[SC_POISONINGWEAPON]->val1 == 9 )// Oblivion Curse gives a 2nd success chance after the 1st one passes which is reducible. [Rytech]
-						rate = 100 - tstatus->int_ * 4 / 5;
-					sc_start(src, bl, ssc->data[SC_POISONINGWEAPON]->val2, rate, ssc->data[SC_POISONINGWEAPON]->val1, skill->get_time2(GC_POISONINGWEAPON, 1) - (tstatus->vit + tstatus->luk) / 2 * 1000, skill_id);
+				if (ssc != NULL && ssc->data[SC_POISONINGWEAPON] != NULL && rnd() % 100 < 70 + 5 * skill_lv) {
+					sc_start(src, bl, ssc->data[SC_POISONINGWEAPON]->val2, 100, ssc->data[SC_POISONINGWEAPON]->val1, skill->get_time2(GC_POISONINGWEAPON, 1), skill_id);
 					status_change_end(src, SC_POISONINGWEAPON, INVALID_TIMER);
 					clif->skill_nodamage(src,bl,skill_id,skill_lv,1);
 				}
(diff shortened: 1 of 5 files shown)


src/map/skill.c
```diff
@@ -14423,9 +14425,10 @@ static int skill_unit_onplace_timer(struct skill_unit *src, struct block_list *b
 		 * 3rd stuff
 		 **/
 		case UNT_POISONSMOKE:
-			if (battle->check_target(ss,bl,BCT_ENEMY) > 0 && !(tsc != NULL && tsc->data[sg->val2] != NULL)
+			if (battle->check_target(ss, bl, BCT_ENEMY) > 0 && !(tsc != NULL && tsc->data[sg->val2] != NULL)
```
guilherme-gm, line 14428: I noticed that the effect of those poisons are also implemented in the item scripts, I was thinking about making a `SkillScript` field in item db so we could move the time calculation/etc to the item itself instead of hardcoding the poison logic.

But this condition makes things somewhat complicated:
```
(tsc != NULL && tsc->data[sg->val2] != NULL)
```

I would need to keep it hardcoded, or pass a lot of info to the item script... so I went with the easy approach of just using SkillData2.

If you want to see the initial implementation, I will keep this draft in my own repo for now: https://github.com/guilherme-gm/Hercules/pull/5

src/map/skill.c
```diff
@@ -20323,26 +20323,32 @@ static int skill_poisoningweapon(struct map_session_data *sd, int nameid)
...
+	int msg = 0;
+	switch (nameid) {
+	case ITEMID_POISON_PARALYSIS:     msg = MSG_PARALYZE;      type = SC_PARALYSE;      break;
+	case ITEMID_POISON_FEVER:         msg = MSG_PHYREXIA;      type = SC_PYREXIA;       break;
+	case ITEMID_POISON_CONTAMINATION: msg = MSG_DEATHHURT;     type = SC_DEATHHURT;     break;
+	case ITEMID_POISON_LEECH:         msg = MSG_RICHEND;       type = SC_LEECHESEND;    break;
+	case ITEMID_POISON_FATIGUE:       msg = MSG_VENOMBLEED;    type = SC_VENOMBLEED;    break;
+	case ITEMID_POISON_NUMB:          msg = MSG_TOXIN;         type = SC_TOXIN;         break;
+	case ITEMID_POISON_LAUGHING:      msg = MSG_MAGICMUSHROOM; type = SC_MAGICMUSHROOM; break;
+	case ITEMID_POISON_OBLIVION:      msg = MSG_OBLIANCURSE;   type = SC_OBLIVIONCURSE; break;
```
guilherme-gm, line 20336: I was conflicted between following the standard and putting each assignment in a line or keeping it like that...

It does look easier to read when adding everything in a single line. But let me know your thoughts and I can change.

If there is an even better way than the switch, I am also open to suggestions :)

MishimaHaruna: I agree that this tabular form makes it easier on the eye (and it's short enough so that it's feasible to keep it aligned). The day we switch to clang-format to auto-format the code, we can put ignore directives around blocks similar to this so that their custom format is preserved.

MishimaHaruna (approved):

PR #3220, opened 2023-07-21
Double HIT bonus from BS_WEAPONRESEARCH in pre-renewal

wiryawanadipa:
This condition is for renewal

For the pre-renewal bonus hit, the rate bonus is in status.c

If both are active (battle.c and status.c) it will give +20 HIT first and then add +20% from (I'm not sure which one) the total hit or total hit rate.

For the pre-renewal bonus hit, just use the bonus from status.c

### Pull Request Prelude

### Changes Proposed

- add RENEWAL if condition for BS_WEAPONRESEARCH HIT bonus

**Issues addressed:** 
- fix double bonus from BS_WEAPONRESEARCH


skyleo:
Please reword the commit message to something meaningful such as "Fix hit bonus from BS_WEAPONRESEARCH being applied twice in pre-re"

PR #3221, opened 2023-07-21
Fix hit bonus from BS_WEAPONRESEARCH being applied twice in pre-renewal

wiryawanadipa:
### Pull Request Prelude

This condition is for renewal

For the pre-renewal bonus hit, the rate bonus is already calculated in status.c

If both are active (battle.c and status.c), in pre-renewal it will give +20 HIT (this is BS_WEAPONRESEARCH pre-renewal behavior) first and then add +20% accuracy after the hit rate is determined (this is BS_WEAPONRESEARCH renewal behavior).

For the pre-renewal bonus hit, just use the bonus from status.c

Note: Sorry for the previous commit, I made a mess. That was my first commit to other people's project.

### Changes Proposed

- Add renewal if condition for BS_WEAPONRESEARCH HIT bonus in battle.c since it's for renewal

Based on this wiki
https://irowiki.org/classic/Weaponry_Research
https://irowiki.org/wiki/Weaponry_Research

The renewal uses a percentage (+20% accuracy after the hit rate determine) while the pre-renewal use fix HIT number (+20 HIT)

**Issues addressed:** 

- Fix hit bonus from BS_WEAPONRESEARCH being applied twice in pre-renewal

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -5108,9 +5108,12 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 
 		if( sd ) {
 			// Weaponry Research hidden bonus
+			// This condition is for renewal
+			// For pre-renewal hit rate bonus is in status.c
+#ifdef RENEWAL
 			if ((temp = pc->checkskill(sd,BS_WEAPONRESEARCH)) > 0)
 				hitrate += hitrate * ( 2 * temp ) / 100;
-
+#endif
 			if ((sd->weapontype == W_1HSWORD || sd->weapontype == W_DAGGER) && (temp = pc->checkskill(sd, GN_TRAINING_SWORD)) > 0)
 				hitrate += 3 * temp;
 		}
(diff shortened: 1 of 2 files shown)


skyleo:
> Note: Sorry for the previous commit, I made a mess. That was my first commit to other people's project.

Don't worry about it, you're working on your own fork so no mess can bother us. :)

Although you could change the `removing comment` commit to `Remove comment` to fit our commit naming style.
And the line you touched could have applied our coding style to it:
```
if((skill_lv=pc->checkskill(sd,BS_WEAPONRESEARCH))>0)
```
-> 
```
if ((skill_lv = pc->checkskill(sd, BS_WEAPONRESEARCH)) > 0)
```

But don't worry about that. You might want to read up on Git's rebase feature and I suggest you to use the CLI for Git.
(It can be used to reword commits and edit them)

skyleo (approved):

MishimaHaruna (approved):

skyleo:
After researching this further in Aegis, I've come to the conclusion that this is indeed official behavior.

It adds a flat + 2 * skill_lv HIT which is seen in client too. And a hidden + 2 * skill_lv % on top of the final hitrate (Which is used internally).

Thus this PR should not be merged.

skyleo:
If you think this is a logical bug in the official server software itself.

Feel free to rewrite your fix in such a way that it can be toggled in the battle configurations, in case one wants the *Hercules* behavior instead of the *Aegis* one.

PR #3236, opened 2023-10-06
Add logic to generate item links and getitemlink script command

guilherme-gm:
### Pull Request Prelude

### Changes Proposed

Implements the logic to generate "item links" on the server side (like when a player shift-clicks an item in-game). This uses a client-specific base62 string representing the item, refine, cards, etc.

With this PR it will be possible to create item links using script and source (which will be needed for some features)

Special thanks to Asheraf for sharing references with me for the code used in clif; Also credits to rA since I looked into some of their implementation to do this.

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -64,6 +64,7 @@
 #include "map/vending.h"
 #include "map/achievement.h"
 #include "common/HPM.h"
+#include "common/base62.h"
 #include "common/cbasetypes.h"
 #include "common/conf.h"
 #include "common/ers.h"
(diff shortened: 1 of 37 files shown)


src/map/clif.c
```diff
@@ -25955,6 +25956,111 @@ static void clif_parse_adventuterAgencyJoinResult(int fd, struct map_session_dat
...
+ * 
+ * @param buf uninitialized buffer where the string will be written to. Caller MUST free it.
+ * @param it item data to be formatted
+ */
+static void clif_format_itemlink(StringBuf *buf, const struct item *it)
+{
+	nullpo_retv(buf);
+	nullpo_retv(it);
+
+	StrBuf->Init(buf);
```
guilherme-gm, line 25971: I wonder, maybe it would be better to ask the caller to give an initialized buffer? this way the caller may choose to give an already filled buffer and this function would only add extra things to it.

And it would be clear for the caller that it needs to Init + Destroy and not only Destroy

MishimaHaruna: I think so. If the caller is passing it, then it'd make sense if it'd be already initialized (that'd also allow a caller to pass a StringBuf that already contains a string it wants the item link appended to).

Otherwise, if you want to initialize it in this function, then I think the signature should be `StringBuf *clif_format_itemlink(const struct item *it)`, so that it's allocated and initialized here (and it's care of the caller to destroy it afterwards), but that'd lose the convenience of appending to an already existing string.

I'm fine with either of these two options

MishimaHaruna (changes requested):

src/map/clif.c
guilherme-gm: thanks! I changed it so the caller needs to send the initialized buffer

MishimaHaruna (approved):

PR #3230, opened 2023-08-24
2-2 jobs skill rebalance - Sage (2018 patch/Renewal)

guilherme-gm:
### Pull Request Prelude
- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
The main purpose of this PR is to introduce the rebalance of Sage job skills. This change affects Renewal-only.

On official servers this came along with rebalances of 1st, 2nd jobs and transclass too. I am working in additional PRs for the remaining 2-2, including bard/dancer, being in separate PRs in order to keep those PRs in a reasonable size.

The implementation in this PR is based on kRO and kRO zero patch notes, iRO Wiki, rAthena and divine pride info, along with some in-game testing. I can't say everything is 100% accurate because there were discrepancies between different sources, and I could not test everything in kRO, but should be quite close.

Since clients before 2018-10 could only support up to 7 skills in Auto Spell, I added a define that allows people in Renewal to load pre-re autospell db and use the old skill list (in a mix of new/old behavior as documented in general.h) in case they can't migrate now.

I won't list all the rebalance changes in the PR description, but it may be checked in each commit text. Also, the commits are in the same order as they appear in the references below.

**Affected skills**
- SA_VOLCANO
- SA_DELUGE 
- SA_VIOLENTGALE 
- SA_FLAMELAUNCHER
- SA_FROSTWEAPON
- SA_LIGHTNINGLOADER
- SA_SEISMICWEAPON
- SA_AUTOSPELL 

References:
- [kRO 2018.10.31 patch note](https://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=7033&curpage=21)
- [kRO 2018.11.28 patch note / translated](https://board.herc.ws/topic/17043-kro-patch-2018-11-28/)
- [Divine Pride](https://www.divine-pride.net/forum/index.php?/topic/3453-kro-mass-skills-balance-1st-2nd-and-transcendent-classes-skills/)

---

Since AutoSpell is tricky to test, I made some sort of "unit tests" for Auto Spell in a separate branch to try to cover as many cases. You can check the testing code in this PR in my fork: https://github.com/guilherme-gm/Hercules/pull/9/files#diff-57e0249af7272bf23858146fc95e4674d48e2ffdf89e0824333cb54e232dd56d 

[...]

PR #3237, opened 2023-10-08
Refactor Auto Spell and move its config to libconfig

guilherme-gm:
### Pull Request Prelude

### Changes Proposed

This PR was mostly part of #3230 where I have refactored Auto Spell to make easier implement the renewal rebalance. Since these rebalance PRs are taking some time for me to finish and with Keikun's suggestion of moving the config to libconfig, I decided to make this as a separate PR so we can move this part earlier.

This PR moves Auto Spell skill list to LibConfig and reworks the listing/selection logic to remove the if hard to read if chains to something more generic. This way we can implement the renewal rebalance changes with a few lines instead of a huge ugly `#if`/`#else` blocks.

To avoid as much as possible from breaking something with this rewrite, I made some sort of "unit tests" for Auto Spell to try to cover as many cases as I could. You can check the testing code in this PR in my fork: https://github.com/guilherme-gm/Hercules/pull/9/files#diff-57e0249af7272bf23858146fc95e4674d48e2ffdf89e0824333cb54e232dd56d  (Only PRE-RE version of it applies to this PR, and it should work for both pre-re and re)

It is kinda hackish, but this testing PR only extracts the functions to smaller ones with the same names as in this PR so I can test it using a plugin to run as many cases as possible. The same plugin can be used in this PR and it should pass for PRE-RE and RE (if you tweak to use pre-re version of the plugin).

About the decisions on the rewrite:
- The decision of which skills to show was moved from clif to skill so skill logic stays on skill.c (as a bonus, this clif could now be used for other things, I guess)
- Cast end logic was moved out of the huge skill castend function so we can have smaller and clearer functions (and also to make it testable and pluggable)
- `skill->autospell` renamed to make it clear at which moment this function takes part
- Everything now uses the same list of skills, so changing the list doesn't require changing 3 different places

If you are wondering how this will cover the rebalance PR, my idea is to add a constant that may be used in `SkillLevel` to indicate to use half of AutoSpell level. From my tests, it should be a simple change and quite clean.

**Issues addressed:** 
None


src/map/skill.c
```diff
@@ -24724,14 +24724,224 @@ static bool skill_read_skilldb(const char *filename)
...
+	sk->skill_id = skill_id;
+}
+
+/**
+ * Reads a AutoSpell skill's usable levels (SkillLevel) when reading the autospell DB.
+ *
+ * @param conf The libconfig settings block which contains the skill's data.
+ * @param sk The autospell_skill struct where the level should be set.
+ */
+static void skill_read_autospell_skill_level(struct config_setting_t *conf, struct s_autospell_db *sk)
```
guilherme-gm, line 24813: **thought/possible improvement:** we have several functions very similar to this one in `skill.c`. I am wondering whether we could make some generalization to clear those replications...

I can see the advantage of having them separate since plugins can hook into one of the fields without worrying about the other ones.

But since we are at startup, where performance is not extremely critical, maybe we could use a parameter so plugins could simply use an if to tell whether it is the one they want. and combine this with a function received by parameter for validation/actually setting the value.

Something like: (pseudo code)
```
enum config_identifier {
   autospell_skillevel = 0,
   skill_range = 1,
   // ...
}

load_autospell_skill_level(int *arr, int level, int val, struct info) {
   // validate ... ("info" can be used to provide context for the message)
   arr[level] = val;
}

read_grouped_value(enum config_identifier, conf, field_name, load_fn) {
   // check if it is a single value or a group, read and call load_fn for each level
}
```

MishimaHaruna: I like this idea

guilherme-gm: Thanks :) I will work on this in a separate PR so we don't mix unrelated things into the autospell one

src/map/skill.c
```diff
@@ -24724,14 +24760,231 @@ static bool skill_read_skilldb(const char *filename)
...
+static int skill_autospell_db_entry_compare(const void *entry1, const void *entry2)
+{
+	nullpo_ret(entry1);
+	nullpo_ret(entry2);
+
+	struct s_autospell_db *entry1_ = (struct s_autospell_db *) entry1;
+	struct s_autospell_db *entry2_ = (struct s_autospell_db *) entry2;
+
+	// autospell_level == 0 is always at the end of the list, because they are unused entries
+	if (entry1_->autospell_level == 0)
```
MishimaHaruna, line 24881: This may lead to undefined behavior when called with a value of `0` in both entry1 and entry2 (should return `0`, but it will return `1` here).

MishimaHaruna (changes requested):

src/map/skill.c
guilherme-gm: Fixed. thank you!!

MishimaHaruna (approved):

PR #3238, opened 2023-10-19
Fixes to itemlink and fixes/improvements to CI

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
This applies some changes to our actions that should make most of them now pass.

1. AppVeyor doesn't have "mysql56" anymore. I changed to "mysql" (following their docs)

2. Fixed clang-13 build
 The package was never found. I am not skilled in linux packages nor actions, but per my understanding, `debian:unstable` no longer has clang-13 (checked here: https://pkgs.org/search/?q=clang ). I changed build pipeline (which includes clang-13) to use `debian:bookworm` instead. I don't know if we really need to use unstable here...

3. Added concurrency settings to actions so they now stop the current CI of a PR if a new commit is pushed to this PR.
  I have followed these docs: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency and from my tests everything seems fine

With those changes I expect to have most checks passing, MySQL will continue failing because of `MYSQL_OPT_RECONNECT` deprecation. We use it and this generates a Warning, failing the CI. I am unsure how to better approach this one... I am open to suggestions (and I can work in a separate PR for that).

https://dev.mysql.com/doc/c-api/8.0/en/c-api-auto-reconnect.html

Finally, fixing these has shown a few errors in the CI. Some were caused by me 😅  in #3236 , so I am fixing here. Another one was in tests, I fixed it too (hopefully the right way).

**Edit:** now that all checks finished running.
- "mysql" group is expected to fail (as noted above)
- the other failures are due to a random freeze that happens in some runs that I have no idea why. Back on #3217 I reduced its timeout, but could not find the root cause, since it happens in random tests and at random stages.

**Issues addressed:** 
None, I think

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -25993,12 +25993,12 @@ static void clif_format_itemlink(StringBuf *buf, const struct item *it)
 #if PACKETVER >= 20161116
 	static const char view_sep = '&';
 #endif
-#if PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200723
+#if PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200723 || PACKETVER_ZERO_NUM >= 20221024
 	static const char grade_sep = '\'';
 #endif
 
 	// Separators that were updated along the time
-#if PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200723
+#if PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200723 || PACKETVER_ZERO_NUM >= 20221024
 	static const char *tag_name = "ITEML";
 	static const char card_sep = ')';
 	static const char optid_sep  = '+';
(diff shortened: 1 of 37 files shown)


src/map/clif.c
```diff
@@ -25993,12 +25993,12 @@ static void clif_format_itemlink(StringBuf *buf, const struct item *it)
 #if PACKETVER >= 20161116
 	static const char view_sep = '&';
 #endif
-#if PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200723
+#if PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200723 || PACKETVER_ZERO_NUM >= 20221024
```
guilherme-gm, line 25996: @4144 / @Asheraf do you know if I've got the zero packetver right? I am assuming this date because `grade` fields were added to packets at this date. But I could not test

4144: for packets with grade field should be correct date. for item link idk need test in client

MishimaHaruna (approved):

src/map/clif.c
guilherme-gm: ok. thanks! I will leave like that for now as an "educated guess" and try to test later. Should work for most cases.

PR #3242, opened 2023-10-29
Added config option to keep direction for free cells

### LINKED ISSUE #1241: Unable to get item
Picking up an item that is across you while there is something (player/NPC/mob) blocking the cell between you and the item will result to making you back and forth trying to get the item.
This bug only happens when the item you're trying to pick up was on your left side of the screen. (Just like in the video.)
Hercules:
https://youtu.be/QmmX846hu30

Aegis Video:
https://youtu.be/vbMF2s8xUEo

csnv:
### Pull Request Prelude

### Changes Proposed

Added configuration option `keep_dir_free_cell`. Although it is set to false, I would like it to be enabled by default in Hercules.
This option allows discovery of free cells based on the unit's current facing direction instead of always using EAST, which have the following advantages:
- Improves distribution of mobs when dispersing after being grouped. This setting allows monsters to spread in a circular fashion instead of forming a long horizontal line.
- Allows characters to move more naturally if they came from the east, instead of walking back.
- In some instances, fixes issue #1241 allowing the user to obtain the item.

This commit also fixes #1241 in all scenarios not allowing the character to enter a walk loop. Fix based on rAthena's.

**Issues addressed:** 
#1241 

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -11831,9 +11831,14 @@ static void clif_parse_WalkToXY(int fd, struct map_session_data *sd)
 	if(sd->sc.data[SC_RUN] || sd->sc.data[SC_WUGDASH])
 		return;
 
-	pc->delinvincibletimer(sd);
+	RFIFOPOS(fd, packet_db[RFIFOW(fd, 0)].pos[0], &x, &y, NULL);
 
-	RFIFOPOS(fd, packet_db[RFIFOW(fd,0)].pos[0], &x, &y, NULL);
+	// Do not allow one cell move commands if the target cell is not free
+	if (battle_config.official_cell_stack_limit > 0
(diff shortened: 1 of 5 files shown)


skyleo:
Love the idea!

src/map/clif.c
```diff
@@ -11831,9 +11831,14 @@ static void clif_parse_WalkToXY(int fd, struct map_session_data *sd)
 	if(sd->sc.data[SC_RUN] || sd->sc.data[SC_WUGDASH])
 		return;
 
-	pc->delinvincibletimer(sd);
+	RFIFOPOS(fd, packet_db[RFIFOW(fd, 0)].pos[0], &x, &y, NULL);
```
skyleo, line 11834: Can you separate this change to its own commit as it doesn't do what your commit message states?

skyleo (changes requested):

skyleo (approved):

4144 (approved):

MishimaHaruna (approved):

Issue #640, opened 2015-08-12
Dispell skill is not working on @duel state


kyeme:
SA_DISPELL skill is not working on @duel state ^^

MishimaHaruna:
Could you elaborate a little more on this?

- Does it work between an user in duel and one outside duel? (intended not to work)
- Does it work between two users in the same duel? (intended to work)

kyeme:
Does it work between two users in the same duel? (intended to work) <-- This is not working in latest. Sage class cannot dispell even if there is gemstone availabe.

AnisotropicDefixation:
That's due to the skill being only usable on party members outside of PvP flagged maps.

[code]
if( sd && dstsd && !map_flag_vs(sd->bl.m)
					&& (sd->status.party_id == 0 || sd->status.party_id != dstsd->status.party_id) ) {
					// Outside PvP it should only affect party members and no skill fail message.
					break;
{
[/code]

Can just add a duel_group exception or something.

PR #3246, opened 2023-11-01
Fixed SA_DISPEL not working in duels

csnv:
### Pull Request Prelude

### Changes Proposed

Fixed Sage's dispel to work in duels.

**Issues addressed:** 
#640 

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -8619,8 +8619,9 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 			int splash;
 			if (flag&1 || (splash = skill->get_splash(skill_id, skill_lv)) < 1) {
 				int i;
-				if( sd && dstsd && !map_flag_vs(sd->bl.m)
-					&& (sd->status.party_id == 0 || sd->status.party_id != dstsd->status.party_id) ) {
+				if (sd != NULL && dstsd != NULL && !map_flag_vs(sd->bl.m)
+					&& (sd->status.party_id == 0 || sd->status.party_id != dstsd->status.party_id)
+					&& (sd->duel_group != 0 && sd->duel_group != dstsd->duel_group)) {
 					// Outside PvP it should only affect party members and no skill fail message.
 					break;
 				}


src/map/skill.c
```diff
@@ -8619,8 +8619,9 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 			int splash;
 			if (flag&1 || (splash = skill->get_splash(skill_id, skill_lv)) < 1) {
 				int i;
-				if( sd && dstsd && !map_flag_vs(sd->bl.m)
-					&& (sd->status.party_id == 0 || sd->status.party_id != dstsd->status.party_id) ) {
+				if (sd && dstsd && !map_flag_vs(sd->bl.m)
+					&& (sd->status.party_id == 0 || sd->status.party_id != dstsd->status.party_id)
+					&& (sd->duel_group != dstsd->duel_group)) {
```
skyleo, line 8624: I think you will need to check for sd->duel_group being != 0 (and don't mistreat it as a boolean as all the other lines of code do :( )
Otherwise for players that both aren't in a duel, dispell will always "work".

skyleo (changes requested):

src/map/skill.c
```diff
@@ -8619,8 +8619,9 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 			int splash;
 			if (flag&1 || (splash = skill->get_splash(skill_id, skill_lv)) < 1) {
 				int i;
-				if( sd && dstsd && !map_flag_vs(sd->bl.m)
-					&& (sd->status.party_id == 0 || sd->status.party_id != dstsd->status.party_id) ) {
+				if (sd && dstsd && !map_flag_vs(sd->bl.m)
```
MrKeiKun, line 8622: since this was modified, go extra mile and expand these first two conditions.

skyleo: True. Stuff like `sd` should be `sd != NULL`, same for dstsd.

csnv: These checks are actually not needed as all of them are previously checked already. The only detail left is guild mates.
Currently, the flag GuildOnly in skill_db allows casting it on guild mates in and outside of gvg, I'm thinking on adding a new flag `GuildPvpOnly` it applies to guild mates only in pvp scenarios.

csnv: I've finally discarded the idea of adding yet another flag for 1 or 2 skills that will make use of it.
Fixes related to these comments, done.

skyleo (approved):

MishimaHaruna (approved):

Issue #2727, opened 2020-05-18
2nd classes's skills update


haurenburu:
***Skills balance contents***
2nd classes
- [x] Knight (still need to confirm some formulas)
- [x] Blacksmith
- [x] Assassin
- [x] Wizard (i`m not sure about vermilion hits)
- [x] Priest
- [x] Hunter
- [x] Crusader
- [x] Alchemist
- [x] Rogue
- [x] Sage
- [x] Monk (missing double damage on GF i dont know how to do it :[ )
- [x] Bard & Dancer

https://www.divine-pride.net/forum/index.php?/topic/3453-kro-mass-skills-balance-1st-2nd-and-transcendent-classes-skills/

neverfoldaq:
hey, do you have any progress here?

MishimaHaruna:
The 2nd class skills rebalance is now available for testing on the rebalance branch (see v2024.06-rebalance pre-release)

PR #3253, opened 2023-11-15
2-2 jobs skill rebalance - Monk (2018 patch/Renewal)

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
The main purpose of this PR is to introduce the rebalance of Monk job skills. This change affects Renewal-only.

On official servers this came along with rebalances of 1st, 2nd jobs and transclass too. I am working in additional PRs for the remaining 2-2, including bard/dancer, being in separate PRs in order to keep those PRs in a reasonable size.

The implementation in this PR is based on kRO and kRO zero patch notes, iRO Wiki, rAthena and divine pride info, along with some in-game testing. I can't say everything is 100% accurate because there were discrepancies between different sources, and I could not test everything in kRO, but should be quite close.

I won't list all the rebalance changes in the PR description, but it may be checked in each commit text. Also, the commits are in the same order as they appear in the references below.

Additionally, `MO_INVESTIGATE` has a separate fix for how its damage works in Renewal. This is made in a separate commit because iRO Wiki reports this formula since 2016, so it is probably a bug that we didn't know. 

**Affected skills**
- MO_TRIPLEATTACK (Raging Trifecta Blow) 
- MO_CHAINCOMBO (Raging Quadruple Blow) 
- MO_COMBOFINISH (Raging Thrust) 
- MO_EXTREMITYFIST (Asura Strike) 
- MO_BLADESTOP (Root) 
- MO_INVESTIGATE (Occult Impact) 
- MO_FINGEROFFENSIVE (Throw Spirit Sphere) 
- MO_ABSORBSPIRITS (Absorb Spirit Sphere) 

**Other effects**
- Changed size penalty for Knuckles (Renewal only)

References:
- [kRO 2018.10.31 patch note](https://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=7033&curpage=21)
- [kRO 2018.11.28 patch note / translated](https://board.herc.ws/topic/17043-kro-patch-2018-11-28/)
- [Divine Pride](https://www.divine-pride.net/forum/index.php?/topic/3453-kro-mass-skills-balance-1st-2nd-and-transcendent-classes-skills/)

**Issues addressed:** 
Part of #2727 

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -2276,7 +2276,11 @@ static int battle_calc_skillratio(int attack_type, struct block_list *src, struc
 #endif
 					break;
 				case MO_COMBOFINISH:
+#ifndef RENEWAL
 					skillratio += 140 + 60 * skill_lv;
+#else
+					skillratio += 350 + 150 * skill_lv + status_get_str(src) * 5;
+#endif
 					break;
 				case BA_MUSICALSTRIKE:
 				case DC_THROWARROW:
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -6717,13 +6717,17 @@ static enum damage_lv battle_weapon_attack(struct block_list *src, struct block_
 	}
 
 	if(sd && (skillv = pc->checkskill(sd,MO_TRIPLEATTACK)) > 0) {
-		int triple_rate= 30 - skillv; //Base Rate
-		if (sc && sc->data[SC_SKILLRATE_UP] && sc->data[SC_SKILLRATE_UP]->val1 == MO_TRIPLEATTACK) {
-			triple_rate += triple_rate * (sc->data[SC_SKILLRATE_UP]->val2) /100;
+#ifndef RENEWAL
+		int triple_rate = 30 - skillv; // Base Rate
+#else
+		int triple_rate = 30; // Base Rate
+#endif
+		if (sc != NULL && sc->data[SC_SKILLRATE_UP] != NULL && sc->data[SC_SKILLRATE_UP]->val1 == MO_TRIPLEATTACK) {
+			triple_rate += triple_rate * (sc->data[SC_SKILLRATE_UP]->val2) / 100;
 			status_change_end(src, SC_SKILLRATE_UP, INVALID_TIMER);
 		}
 		if (rnd() % 100 < triple_rate) {
-			if (skill->attack(BF_WEAPON, src, src, target, MO_TRIPLEATTACK, skillv, tick, 0))
+			if (skill->attack(BF_WEAPON, src, src, target, MO_TRIPLEATTACK, skillv, tick, 0) != 0)
 				return ATK_DEF;
 			return ATK_MISS;
 		}
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -826,6 +826,13 @@ static int64 battle_calc_masteryfix(struct block_list *src, struct block_list *t
 	nullpo_ret(src);
 	nullpo_ret(target);
 
+#ifdef RENEWAL
+	// In renewal, Occult Impact doesn't get extra damage by any mastery, not even weapon ones
+	if (skill_id == MO_INVESTIGATE)
+		return damage;
+#endif
+
+
 	sc = status->get_sc(src);
 	sd = BL_CAST(BL_PC, src);
 	tstatus = status->get_status_data(target);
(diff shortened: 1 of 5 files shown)


src/map/battle.c
```diff
@@ -6627,6 +6627,39 @@ static bool battle_check_arrows(struct map_session_data *sd)
...
+	struct status_change *tsc = status->get_sc(target);
+	if (tsc == NULL || tsc->data[SC_BLADESTOP_WAIT] == NULL)
+		return false; // Target is not in BladeStop wait mode
+
+	if (is_boss(src))
+		return false; // Boss monsters are not affected
+
+	// CHECKME: Is that right?
+	if (src->type == BL_PC)
+		return true; // Player gets into BladeStop regardless of distance
```
guilherme-gm, line 6679: I tested this for RE and this behavior is wrong.

In Aegis, if a ranged monster (e.g. Orc Archer) or a ranged player (e.g. a Hunter) hits me from many cells away, Blade Stop won't start.

I will hold the fix for after rebalance to avoid adding non-rebalance things into the mix of this PR chain.

@skyleo may you check how this works for pre-re?

skyleo: copy from our chat:
```
So yeah, if attacker player and attacking bladestopped player -> No range check. if attacking mob -> no bladestop check

If attacker mob or anything else non-player and attacking bladestopped anything -> range check of 2 

```

skyleo: In Aegis 11.2 the code is written in a way that it assumes that mobs cannot use Bladestop, so when someone is in Bladestop there is code that assumes that someone is a player, without checking if it's player type.

guilherme-gm: Thank you very much! moved the refactor + fix to #3254 so it may go to master earlier. And I will rebase this PR later on

violent01:
MO_FINGEROFFENSIVE is broken or I'm just mistaken? Is this deliberately implemented `ratio /= skill->get_num(MO_FINGEROFFENSIVE, skill_lv);` or what? If that line of code is implemented Finger Offensive only does 1 hit instead of 5 and the damage is so low. Can you also revisit MO_INVESTIGATE as I'm not convinced that its damage would be just 2k+ only against a high-def and vit enemy like an ancient sea horse (2182 MobID)

guilherme-gm:
Thanks for checking.

I double checked in kRO Zero sakray, here are my findings:

## `MO_FINGEROFFENSIVE`
I think there is indeed a small mistake in it. I shouldn't divide the ratio itself, but the damage. But, the damage is close to right. I will try to find what is causing this extra damage to slipping in, but the skill will not be buffed.

Unfortunately they had changed the skill ratio in Zero:
- from: 800 / 1000 / 1200 / 1400 / 1600 
- to: 1600 / 2000 / 2400 / 2800 / 3200

But we can still compare the overall behavior.

Using a Monk Lv 56/11 with:
- Iron Fist Lv1 (Zero skill tree is different, Lv1 is enough)
- STR 4
- DEX 2
- LUK 1

This gives base StatusATK = 18, and 3 Mastery ATK from Iron Fist, and each sphere gives +3 ATK too

I've got the following results against `DUMMY_10`:
- with 1 Sphere, Throw Sphere Lv1 (SkillRatio = 1600), Damage = 667 (5x 133)
- with 1 Sphere, Throw Sphere Lv2 (SkillRatio = 2000), Damage = 835 (5x 167)
- with 1 Sphere, Throw Sphere Lv3 (SkillRatio = 2400), Damage = 1000 (5x 200)
- with 2 Sphere, Throw Sphere Lv1 (SkillRatio = 1600), Damage = 715 (5x 143)

These values are correct following iRO Wiki formula: https://irowiki.org/wiki/ATK

```
StatusATK = 18
Mastery = 6 (with 1 sphere)
ATK = 18 * 2 + 6 = 42

SkillDamage(Lv1) = 1600 * 42 / 100 = 672
SkillDamage(Lv2) = 2000 * 42 / 100 = 840
SkillDamage(Lv3) = 2400 * 42 / 100 = 1008

// (Enemy Lv10 = 5 def)
FinalDamage(Lv1) = 672 - 5 = 667
FinalDamage(Lv2) = 840 - 5 = 835
FinalDamage(Lv3) = 1008 - 5 = 1003 (!!)
```

Now, this damage is given in a bundle of 5 hits. So we have to `FinalDamage/5` and apply it 5 times. Thus: (Always rounding down)
```
HitDamage(Lv1) = 667 / 5 = 133
HitDamage(Lv2) = 835 / 5 = 167
HitDamage(Lv3) = 1003 / 5 = 200
```

Note that Lv3 damage will go down to 1000 (from 1003) because of the division. This will be delivered as 5 hits of 200 (it is a single hit, but the damage is applied 5 times).

This was for Zero. For main, we just need to change the Ratio:
```
StatusATK = 18
Mastery = 6 (with 1 sphere)
ATK = 18 * 2 + 6 = 42

SkillDamage(Lv1) = 800 * 42 / 100 = 336

// (Enemy Lv10 = 5 def)
FinalDamage(Lv1) = 336 - 5 = 331

HitDamage(Lv1) = 331 / 5 = 66
```

On the current code, it is giving a damage of 5x 81 instead of 5x 66.

So yeah, something is still wrong, but the damage should actually be lower, not higher


## `MO_INVESTIGATE`
I have no way to validate it against a high defense target, but I could do some extra tests against some weak mobs (Pupa / Condor / Poring), and looks like it should bypass Target's SoftDEF and get boosted by Target's HardDEF.

So `FinalDamage = (ATK + (Target HardDEF/2)) * SkillRatio` (bypassing the step of reducing SoftDEF)

I will edit this message and detail my tests later, but that's right, the damage should be higher.

guilherme-gm:
Updated the PR with fixes for `MO_FINGEROFFENSIVE` and `MO_INVESTIGATE` as described in the previous comment.

guilherme-gm:
updated the PR so it no longer builds over previous one, but is an independent PR built over rebalance. I think this makes things a bit better than chaining all of them, and the 3 remaining PRs for 2nd classes should not conflict between each other; they may also be merged in any order

MishimaHaruna (approved):

PR #3254, opened 2023-11-15
Fixes to Blade Stop (Root) trigger

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
Extract the conditions to start SC_BLADESTOP to its own function, since it has several conditions, and with the rebalance (#3253), more RE/Pre-RE differences.

And also fixes some mistakes in the logic, making it closer to official behavior. This affects both Pre-RE and RE.

Changes:
1. The target's equipped weapon no longer affects the range of Blade Stop. This applies for both Pre-RE and RE. In current code, a bare-handed monk would only blade-stop targets 1-cell away
2. Player attackers are only blade stopped from far away (> 2 cells) in Pre-RE -- My tests in renewal showed that a Hunter will only get blade-stopped if they are up to 2 cells away from the Monk.

Discussion here: https://github.com/HerculesWS/Hercules/pull/3253#discussion_r1394281487 . Copying skyleo findings:
```
So yeah, if attacker player and attacking bladestopped player -> no extra range check(still needs to be in attack range of course). if attacking mob -> no bladestop check

If attacker mob or anything else non-player and attacking bladestopped anything -> range check of 2 

In Aegis 11.2 the code is written in a way that it assumes that mobs cannot use Bladestop, so when someone is in Bladestop there is code that assumes that someone is a player, without checking if it's player type.
```

Huge thanks to skyleo for helping with Pre-RE validation.

**This change is NOT related to the rebalance of classes made in RE.**

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -1479,7 +1479,7 @@ static int64 battle_calc_defense(int attack_type, struct block_list *src, struct
 				def1 = 0;
 			if (def2 < 1)
 				def2 = 1;
-			
+
 			//Vitality reduction from rodatazone: http://rodatazone.simgaming.net/mechanics/substats.php#def
 			if (tsd) {
 				//Sd vit-eq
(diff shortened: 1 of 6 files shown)


skyleo (approved):
LGTM.

nitpick: Commits should ideally start with a capitalized imperative verb.

PR #3256, opened 2023-11-25
Fix sc statuses not loaded on time when under heavy load

csnv:
### Pull Request Prelude

### Changes Proposed
Added a new flag to indicate sc data has been loaded and postpone clif->parse_LoadEndAck until all requirements are met.

When under heavy load, pc->reg_received runs before pc->scdata_received, thus invoking character calculation without taking any sc statuses like SC_PUSH_CART into consideration.

Along with incorrect character stats or substats, this issue makes the player not able to see visual effects or elements related to statuses, as well as the cart, until the player changes maps or uses a fly wing.

**Issues addressed:** 

### DIFF
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -12260,14 +12260,19 @@ static void pc_scdata_received(struct map_session_data *sd)
 		pc->expire_check(sd);
 	}
 
-	if( sd->state.standalone ) {
-		clif->pLoadEndAck(0,sd);
-		pc->autotrade_populate(sd);
-		pc->autotrade_start(sd);
-	}
 
 	if (sd->sc.data[SC_SOULENERGY] != NULL)
 		sd->soulball = sd->sc.data[SC_SOULENERGY]->val1;
+
+	sd->state.scloaded = 1;
+	if (sd->state.standalone) {
+		clif->pLoadEndAck(0,sd);
+		pc->autotrade_populate(sd);
+		pc->autotrade_start(sd);
+	} else {
+		clif->pLoadEndAck(sd->fd, sd);
+	}
+	
 }
 static int pc_expiration_timer(int tid, int64 tick, int id, intptr_t data)
 {
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -11161,6 +11161,10 @@ static void clif_parse_LoadEndAck(int fd, struct map_session_data *sd)
 		return;
 	}
 
+	if (sd->state.scloaded == 0) { // SC data was not received yet. pc->scdata_received will reinvoke
+		return;
+	}
+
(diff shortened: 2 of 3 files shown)


src/map/pc.c
```diff
@@ -12260,14 +12260,19 @@ static void pc_scdata_received(struct map_session_data *sd)
 		pc->expire_check(sd);
 	}
 
-	if( sd->state.standalone ) {
+
+	if (sd->sc.data[SC_SOULENERGY] != NULL)
```
skyleo, line 12264: nit: This seems unrelated to the commit it's in. Split into individual commit.

csnv: There's no additional logic here, I reordered it so clif->pLoadEndAck is called last.

skyleo: Hmmm. okay. I was looking at it more in the terms of you moved SC_SOULENERGY check.

EPuncker (approved):

src/map/pc.c
```diff
@@ -12260,14 +12260,19 @@ static void pc_scdata_received(struct map_session_data *sd)
...
+		sd->soulball = sd->sc.data[SC_SOULENERGY]->val1;
+
+	sd->state.scloaded = 1;
+	if (sd->state.standalone) {
 		clif->pLoadEndAck(0,sd);
 		pc->autotrade_populate(sd);
 		pc->autotrade_start(sd);
+	} else {
+		clif->pLoadEndAck(sd->fd, sd);
```
MishimaHaruna, line 12273: I'm not sure if I'm overlooking it, but I don't think I can see any mechanism that prevent loadendack execution if the client isn't ready yet (i.e. this should probably be called from pc_scdata_received if the client aleady triggered LoadEndAck and its execution was deferred bythe `sd->state.scloaded == 0` condition?)

MishimaHaruna (changes requested):

src/map/pc.c
csnv: That's exactly the case in this change. pLoadEndAck will be called but under normal circumstances not run entirely until the client has loaded, aka `sd->state.loaded == 1`

The aim of this change is only for those rare cases where the server is under heavy stress and the client loads faster than the server can receive SC data.

MishimaHaruna: Ok, it looks fine. I was worried about the execution of `connect_new = 1` before the `pc_regs_received` call, but in the code paths I examined it doesn't seem to cause any side effects. And the same mechanisms that prevent double execution through `pc_regs_received` seem to be adequate to prevent double execution through `pc_scdata_received`

MishimaHaruna (approved):

PR #3259, opened 2023-12-04
Fix MO_TRIPLEATTACK animation, one cell movement, PA_SACRIFICE delay

csnv:
### Pull Request Prelude

### Changes Proposed
- Fix MO_TRIPLEATTACK animation not working when there's no possible chainable combo skill.
- Fix one cell movements when the target cell is occupied by an invisible object (like a gm using the hide atcommand).
- Remove incorrect afterCastActDelay in PA_SACRIFICE for pre-re.
- Removed fixed target on *some* combo skills. Monk skills act on the current targeted enemy even if last skill was used on a different enemy. Proof https://www.youtube.com/watch?v=XuQ775Eu9-o thanks to akiyammio.
All combo skills are set to current enemy unless we can prove otherwise.
- Fixed attackabletime being overwritten when an autotriggered skill increased the timing (e.g. MO_TRIPLEATTACK)
- Fixed combo delays applying to all SC_COMBOATTACK based skills and not sending the proper time to client.

**Issues addressed:** 

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -11835,7 +11835,8 @@ static void clif_parse_WalkToXY(int fd, struct map_session_data *sd)
 
 	// Do not allow one cell move commands if the target cell is not free
 	if (battle_config.official_cell_stack_limit > 0
-		&& check_distance_blxy(&sd->bl, x, y, 1) && map->count_oncell(sd->bl.m, x, y, BL_CHAR | BL_NPC, 1))
+		&& check_distance_blxy(&sd->bl, x, y, 1)
+		&& map->count_oncell(sd->bl.m, x, y, BL_CHAR | BL_NPC, 0x1 | 0x2) >= battle_config.official_cell_stack_limit)
 		return;
 
 	pc->delinvincibletimer(sd);
--- a/db/pre-re/skill_db.conf
+++ b/db/pre-re/skill_db.conf
@@ -11208,7 +11208,6 @@ skill_db: (
 		IgnoreFlee: true
 	}
 	InterruptCast: true
-	AfterCastActDelay: 2000
 	CoolDown: 0
 	Requirements: {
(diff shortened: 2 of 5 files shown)


src/map/clif.c
```diff
@@ -11835,7 +11835,7 @@ static void clif_parse_WalkToXY(int fd, struct map_session_data *sd)
 
 	// Do not allow one cell move commands if the target cell is not free
 	if (battle_config.official_cell_stack_limit > 0
-		&& check_distance_blxy(&sd->bl, x, y, 1) && map->count_oncell(sd->bl.m, x, y, BL_CHAR | BL_NPC, 1))
+		&& check_distance_blxy(&sd->bl, x, y, 1) && map->count_oncell(sd->bl.m, x, y, BL_CHAR | BL_NPC, 2))
```
MishimaHaruna, line 11838: Is the removal of the `0x1` bit (meant to skip units that are in transit) intentional here?

MishimaHaruna (changes requested):

src/map/clif.c
csnv: It was meant to skip those units under `sc->option&OPTION_INVISIBLE`, like characters using `@hide`, and other invisible units that players should not be aware of.

MishimaHaruna: But since it uses `official_stack_limit`, shouldn't it follow its mechanics? We always ignore moving units in those checks (on a side note, why does this enforce a count of 0 regardless of the value of official_stack_limit? i.e. `official_stack_limit = 5` would consider a cell with 4 unit stacked as still available to walk to)

csnv: Makes sense. I've updated the PR with that taken into consideration. Thanks!

MishimaHaruna (approved):

PR #3260, opened 2023-12-04
Added combo skill cache mechanic toggle

csnv:
### Pull Request Prelude

### Changes Proposed
Currently, combo skills can only be triggered in a small window of 300ms (by default) where SC_COMBOATTACK is active and the previous combo skill delay has terminated. This is the official behavior and makes triggering combo skills after hard delays quite difficult and unintuitive, specially if the user has latency problems.

This change adds a toggle that, if enabled, allows combo skill users to trigger combo skills automatically after the previous combo skill delay has finished if it was requested during its delay.

This optional behavior allows chaining combo skills more intuitive and easier, without bypassing delays or changing mechanics.

**Issues addressed:** 
None.


EPuncker (approved):

src/map/clif.c
```diff
@@ -13562,6 +13562,18 @@ static void clif_useSkillToIdReal(int fd, struct map_session_data *sd, int skill
...
 	} else if (DIFF_TICK(tick, sd->ud.canact_tick) < 0) {
+		// Cannot perform skill yet, delay execution until canact_tick allows it
+		// Only for combo skills with implicit delays, otherwise we create a speedhack that bypasses client's animation delay		
+		if (battle_config.combo_cache_skill
+			&& !skip_combo_check
+			&& sd->sc.data[SC_COMBOATTACK] != NULL
+			&& skill->is_combo(skill_id)
+			&& skill->get_delaynodex(skill_id, skill_lv) != 0
+		) {
+			timer->add(sd->ud.canact_tick, clif->combo_delay_timer, sd->bl.id, (intptr_t)MakeDWord((uint16)skill_id, (uint16)skill_lv));
```
MishimaHaruna, line 13573: Any mechanisms to prevent enqueuing this multiple times by spamming the skill hotkey?

MishimaHaruna (changes requested):

src/map/clif.c
csnv: Enqueing by spamming the hotkey is part of the design I followed, in case the user did not press the correct hotkey at the time.
The important aspect is that even if multiple skills are queued, all of them will have their conditions re-checked. If it fails a second time, it will not re-run again.

MishimaHaruna (approved):

Issue #2749, opened 2020-05-29
rodex_sendmail_acc


tlacson7:
**Describe the bug**
Mail notification is received but no msg on mail.

**To Reproduce**
prontera,150,150,4	script	Rodex Mail	1_M_01,{
	// Sends a messsage to attached player "Account Box" from "Rodex Test", with title "Rodex Test3",
	// with message "Hello World, How are You?", 1000 Zenies and 5 Red Potions attached
	rodex_sendmail_acc(
		getcharid(CHAR_ID_ACCOUNT), "Rodex Test", "Rodex Test3", "Hello World, How are You?",
		1000,
		Red_Potion, 5
	);
	end;
}

**System specs (please complete the following information):**
 - OS: Centos 7x64
 - Hercules Version latest
 - Mode: Prere
 - Packet version: 20151029
 - Client type: main

guilherme-gm:
Your client version is too old to support account mails. Hercules did make it and sent the mail alert, but your client is not able to request for account mails, so you don't see them.

I made a PR (#3263) adding warnings when using it in unsupported clients so people can know why it is not working. not exactly a "solution" but I think this is the best that can be done in this case :)

There was a discussion in the past whether we should redirect account mails to the main box, but was decided not to. Maybe a plugin could be made to support this.

PR #3263, opened 2023-12-18
Add date checks and warns to RoDEX mail box and enable it by default

guilherme-gm:
### Pull Request Prelude

### Changes Proposed

Add safety checks to prevent users from enabling RoDEX account box in unsupported clients and improve documentation around it.

Also enables the account box by default. It was disabled back them because herc default packetver didn't support. But since we moved to default to a 2019 main client, account mail has been enabled on the client side, so there is no reason to keep it disabled by default.

With these changes:
1. Enabling the account mail in an unsupported client will give a warning and disable it
2. Users using supported RE versions need to also enable `ENABLE_RODEX_ACCOUNT_MAIL_RE_PATCH` because they must patch their exe ( https://gitlab.com/4144/Nemo/-/merge_requests/83 )
3. Trying to send account mails while the feature is disabled will show a warning, but still send the mail -- This is intentional, I don't want to cause huge issues for someone who is already using it in a broken way, and I see no harm in delivering the mail
4. Account box will be enabled by default in a fresh Hercules

**Issues addressed:** 
"Solves" #2749 

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -8014,6 +8014,18 @@ static void battle_adjust_conf(void)
 	}
 #endif
 
+#if !(PACKETVER_MAIN_NUM >= 20160323 || (PACKETVER_RE_NUM >= 20160316 && defined(ENABLE_RODEX_ACCOUNT_MAIL_RE_PATCH)) || defined(PACKETVER_ZERO))
+	if (battle_config.feature_rodex_use_accountmail == 1) {
+#ifndef BUILDBOT /* Don't show these warnings to CI */
+		ShowWarning("conf/map/battle/feature.conf:features/rodex_use_accountmail RoDEX account mail is enabled but it requires 2016-03-16 RagexeRE / 2016-03-23 Ragexe or newer, disabling...\n");
+#ifdef PACKETVER_RE
+		ShowWarning("conf/map/battle/feature.conf:features/rodex_use_accountmail For RE clients, you should also enable ENABLE_RODEX_ACCOUNT_MAIL_RE_PATCH\n");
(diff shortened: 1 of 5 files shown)


src/map/battle.c
```diff
@@ -8014,6 +8014,16 @@ static void battle_adjust_conf(void)
 	}
 #endif
 
+#if !(PACKETVER_MAIN_NUM >= 20160323 || (PACKETVER_RE_NUM >= 20160316 && defined(ENABLE_RODEX_ACCOUNT_MAIL_RE_PATCH)) || defined(PACKETVER_ZERO))
+	if (battle_config.feature_rodex_use_accountmail == 1) {
+		ShowWarning("conf/map/battle/feature.conf:features/rodex_use_accountmail RoDEX account mail is enabled but it requires 2016-03-16 RagexeRE / 2016-03-23 Ragexe or newer, disabling...\n");
```
4144, line 8019: need disable this warning for ci, because it failing builds

4144 (changes requested):

src/map/battle.c
guilherme-gm: oh, oops. right.

I am wondering, should I disable it with `#ifndef BUILDBOT` or add another grep to travis.sh like you did for mysql?

4144: yes BUILDBOT should be fine

guilherme-gm: hmm these ifs got a bit... messy, but I think it is fine since it is in a specific place. if you have any better suggestion let me know

MishimaHaruna (approved):

PR #3264, opened 2023-12-18
Add more details when char-server fails to connect

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
Add a few more info when char-server fails to connect. This helps users understand what is going on, and may also help with support on forums/discord.

I chose to make this PR after noticing the Account ID check that is never mentioned anywhere and a new user stumbled into an issue in it today.

Those cases are now more detailed:
- login-server now records details for failure due to account ID range and wrong sex
- char-server now informs about the Account ID range. It doesn't detail the maximum value since people may be using 2 different builds, and `MAX_SERVERS` is part of login server, not common/char.

**Issues addressed:** 
None

### DIFF
--- a/src/char/char.c
+++ b/src/char/char.c
@@ -6096,7 +6096,7 @@ static void char_ensure_online_char_data(struct online_char_data *character)
 {
 	nullpo_retv(character);
 	if (character->data == NULL) {
-		character->data = aCalloc(sizeof(struct online_char_data2), 1);
+		character->data = aCalloc(1, sizeof(struct online_char_data2));
 	}
 }
 
--- a/src/char/char.c
+++ b/src/char/char.c
@@ -2414,7 +2414,7 @@ static int char_parse_fromlogin_connection_state(int fd)
 	case 1: // Invalid username/password
 		ShowError("Can not connect to login-server.\n");
 		ShowError("The server communication passwords (default s1/p1) are probably invalid.\n");
-		ShowError("Also, please make sure your login db has the correct communication username/passwords and the gender of the account is S.\n");
+		ShowError("Also, please make sure your login db has the correct communication username/passwords, the gender of the account is S and its id is between 0 and (MAX_SERVERS - 1).\n");
 		ShowError("The communication passwords are set in /conf/map/map-server.conf and /conf/char/char-server.conf\n");
 		sockt->eof(fd);
 		return 1;
--- a/src/login/login.c
+++ b/src/login/login.c
@@ -1504,11 +1504,14 @@ static void login_parse_request_connection(int fd, struct login_session_data* sd
 	if (!sockt->allowed_ip_check(ipl)) {
 		ShowNotice("Connection of the char-server '%s' REFUSED (IP not allowed).\n", server_name);
 		login->char_server_connection_status(fd, sd, 2);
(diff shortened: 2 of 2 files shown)


src/char/char.c
```diff
@@ -2414,7 +2414,7 @@ static int char_parse_fromlogin_connection_state(int fd)
 	case 1: // Invalid username/password
 		ShowError("Can not connect to login-server.\n");
 		ShowError("The server communication passwords (default s1/p1) are probably invalid.\n");
-		ShowError("Also, please make sure your login db has the correct communication username/passwords and the gender of the account is S.\n");
+		ShowError("Also, please make sure your login db has the correct communication username/passwords, the gender of the account is S and its id is between 0 and (MAX_SERVERS - 1).\n");
```
guilherme-gm, line 2417: "0" is not a typo, the check on `login.c` really checks for `>= 0` (see below), and testing with an account id = 0, it does seem to work. I always thought 0 index was somehow reserved, but looks like it isn't.

Whether we want to keep this or not... I am in favor of changing the minimum ID to 1.

4144 (approved):

guilherme-gm:
@4144

gcc 14 seems to have gone crazy with `aCalloc`... 

```
char.c: In function 'char_ensure_online_char_data':
char.c:6099:33: error: allocation of insufficient size '1' for type 'struct online_char_data2' with size '24' [-Werror=alloc-size]
 6099 |                 character->data = aCalloc(sizeof(struct online_char_data2), 1);
      |                                 ^
	CC	HPMlogin.c
```

maybe it got lost expanding this?
```

// char.c
character->data = aCalloc(sizeof(struct online_char_data2), 1);

// define
#define aCalloc(m, n) (calloc_proxy((m), (n),ALC_MARK))

// calloc_proxy
void *calloc_proxy(size_t num, size_t size, const char *file, int line, const char *func) GCC11ATTR ((malloc, malloc (free_proxy, 1))) __attribute__ ((alloc_size (1, 2))) GCCATTR ((returns_nonnull));
```

this last line is too magical for me xD

4144:
try swap parameters in aCalloc call

EPuncker (approved):

guilherme-gm:
Looks like it works :)

I am a bit curious how it managed to do it... it looks like it managed to understand that the 2nd parameter is meant for the size, even though `alloc_size` attribute doesn't give that info (at least the docs just says it is the "product of the 2 parameters"). interesting...

src/char/char.c
MishimaHaruna: 0 seems indeed valid, even though I don't believe I've ever used it before. It makes sense it was implemented this way since we use it as array index for `login->dbs->server[]` (and also `login->dbs->api_server[]`), the 'invalid' value we use is `-1`.

I don't have a compelling reason to keep it this way or to change it. If you find it confusing, feel free to change how it works (in a separate PR)!

MishimaHaruna (approved):

src/char/char.c
guilherme-gm: Ah, I think I made a confusion between `fd = 0` and `account_id = 0`. `fd = 0` is reserved, but this only affects sockets and doesn't relate to account id. so it is fine

Issue #165, opened 2013-09-28
Centralize Job ID/Name/Data


shennetsind:
Doesn't really belong to performance but I didn't feel well on another milestone, anyway:
Its all over the place, really. By aggregating it smartly we could duck hundreds of lines of code and make the whole process of adding or modifying job data waaaay more flexible (and thus faster).
Glance job names:

```
const char* job_name(int class_)
{
    switch (class_) {
    case JOB_NOVICE:
    case JOB_SWORDMAN:
    case JOB_MAGE:
    case JOB_ARCHER:
    case JOB_ACOLYTE:
    case JOB_MERCHANT:
    case JOB_THIEF:
        return msg_txt(550 - JOB_NOVICE+class_);

    case JOB_KNIGHT:
    case JOB_PRIEST:
    case JOB_WIZARD:
    case JOB_BLACKSMITH:
    case JOB_HUNTER:
    case JOB_ASSASSIN:
        return msg_txt(557 - JOB_KNIGHT+class_);

    case JOB_KNIGHT2:
        return msg_txt(557);

    case JOB_CRUSADER:
    case JOB_MONK:
    case JOB_SAGE:
    case JOB_ROGUE:
    case JOB_ALCHEMIST:
    case JOB_BARD:
    case JOB_DANCER:
        return msg_txt(563 - JOB_CRUSADER+class_);

    case JOB_CRUSADER2:
        return msg_txt(563);

    case JOB_WEDDING:
    case JOB_SUPER_NOVICE:
    case JOB_GUNSLINGER:
    case JOB_NINJA:
    case JOB_XMAS:
        return msg_txt(570 - JOB_WEDDING+class_);

    case JOB_SUMMER:
        return msg_txt(621);

    case JOB_NOVICE_HIGH:
    case JOB_SWORDMAN_HIGH:
    case JOB_MAGE_HIGH:
    case JOB_ARCHER_HIGH:
    case JOB_ACOLYTE_HIGH:
    case JOB_MERCHANT_HIGH:
    case JOB_THIEF_HIGH:
        return msg_txt(575 - JOB_NOVICE_HIGH+class_);

    case JOB_LORD_KNIGHT:
    case JOB_HIGH_PRIEST:
    case JOB_HIGH_WIZARD:
    case JOB_WHITESMITH:
    case JOB_SNIPER:
    case JOB_ASSASSIN_CROSS:
        return msg_txt(582 - JOB_LORD_KNIGHT+class_);

[...]

PR #3265, opened 2023-12-22
Create constants for messages.conf and simplify switch job name switches

guilherme-gm:
### Pull Request Prelude

### 1. Enum for messages.conf
Replacing the hardcoded numbers in source for `conf/messages.conf` entries with a enum.

This is meant to ease the readability a bit with constants that we can understand instead of numbers + comments.

I did not perform any check whether there are unused message entries (Although I do think there are).

I am open for suggestions about naming. Naming 1000+ entries in one go is very hard, so some stuff may have slipped with not so good names.

Huge thanks to @TioAkima who helped me naming all those enum entries.

#### For people upgrading their copies
1. The messages.conf maximum value is now defined in `common/msgtable.h`, you can either add more items in there OR override it using a define:
```
#define MSGTBL_MAX 1600
```

2. If you want to update your customizations to also use constants, a "hacky" conversion script is available here (Linux-only): https://github.com/guilherme-gm/Hercules-releases/tree/master/PR%20Extras/PR%203265%20-%20messages%20conf%20enum -- Check the folder readme for details

--

### 2. Using XMacro for JobID => Job Name conversions
I took the chance to also update the 2 functions in source that converts job ID to name. They had a hardcoded switch/case + message id, and I replaced both with XMacro using the `class*.h` files (as it is done in a few other places). This should help prevent desynchronization in the future.

Still about desynchronization, there were a few values missing in both functions, and with this PR these values are now correctly mapped.

I made a small plugin to dump all the values into the terminal, and it may be found here:
https://github.com/guilherme-gm/Hercules-releases/tree/master/PR%20Extras/PR%203265%20-%20messages%20conf%20enum/

It dumps the list of jobID => jobName so we can compare with comparison tools. A dump of master vs this PR is also available here: https://github.com/guilherme-gm/Hercules-releases/tree/master/PR%20Extras/PR%203265%20-%20messages%20conf%20enum/jobname_dump

#### For people upgrading their copies
If you have custom jobs, you may need to update the enum macros in `src/common/class.h`, `src/common/class_special.h` and `src/common/class_hidden.h` to include the message ID of your jobs.

**Issues addressed:** 
Helps #165 (DON'T SOLVE IT)


src/char/inter.c
```diff
@@ -148,233 +149,18 @@ static void inter_do_final_msg(void)
 	for (i = 0; i < MAX_JOB_NAMES; i++)
 		aFree(msg_table[i]);
 }
-/* from pc.c due to @accinfo. any ideas to replace this crap are more than welcome. */
```
guilherme-gm, line 151: I removed this comment since I _think_ the main issue here was the manual duplication. If you guys think we should keep the comment for further improvement, just let me know (or if there is a suggestion to improve it already , I can do it)

4144:
In almost all lines like this
``JOB_ENUM_VALUE(NOVICE, 0, MSGTBL_JOB_NOVICE)``
probably better use for them simpler lines like this (MSGTBL_JOB_ can be added automatically)
``JOB_ENUM_VALUE(NOVICE, 0)``
and for exceptions use something like this
``JOB_ENUM_VALUE2(THIRDJOB_BEGIN, 4053, MSGTBL_UNKNOWN_JOB)``
?

guilherme-gm:
I was trying to make it more explicit so it doesn't hide too much info and gets confusing.

Those were my motivations:
1. Being forced to always write all fields would make me think when adding a value: "Ok to declare a job I need to give it a name + id + message enum". While having the 2 variations I would try to only use name/id, and then get an error and would have to understand that I am missing an enum in msgtbl_.
2. Also, when looking for where a enum value is being used, like, "who is using this MSGTBL_JOB_NOVICE? -- nothing, because it is hidden on the expansion", then I delete it and a include fails because of 2 levels of things being expanded ends up making it.

In my opinion X-macros are not too clear by default, but they do save a lot of code, so I am afraid of changing it like that and it becoming confusing.

But I can change if it is preferred to save those repetitions in the declaration.

4144:
I think one macro is ok, but probably better remove prefix ``MSGTBL_`` because for existing fields ``JOB_`` and ``MAPID_`` added automatically

guilherme-gm:
Ok. I still think it is adding unneeded abstraction, but it is much better than completely hiding them as the initial proposal. I am ok with this :) I will update the PR to remove MSGTBL_ prefix

src/common/msgtable.h
```diff
@@ -0,0 +1,2944 @@
...
+	/** [%d] seconds left until you can use */
+	MSGTBL_SECONDS_UNTIL_USE = 26,
+	/** Storage has been not loaded yet. */
+	MSGTBL_STORAGE_NOT_LOADED = 27,
+	/** No player found. */
+	MSGTBL_NO_PLAYER_FOUND = 28,
+	/** 1 player found. */
+	MSGTBL_1_PLAYER_FOUND = 29,
+	/** %d players found. */
+	MSGTBL_N_pLAYERS_FOUND = 30,
```
MishimaHaruna, line 68: Minor nitpick: a lowercase `p` slipped in

MishimaHaruna (changes requested):

src/common/msgtable.h
guilherme-gm: Ooops, I will fix now. nice catch :D

MishimaHaruna (approved):

PR #3271, opened 2024-01-14
Fix sysinfogen.sh get incorrect synctax SYSINFO_CPUCORES

vietlubu:
### Pull Request Prelude

### Issue:
When trying to run `make`. `sysinfo.inc` will be generated.
I'm using Mac M1 the output of `SYSINFO_CPUCORES` is
```
// CPU Cores (Platform-dependent)
#define i ( 10 (8 performance and 2 efficiency) )
```
That will be make error while compile process:
```
sysinfo.c:1020:25: error: expected ')'
        sysinfo->p->cpucores = SYSINFO_CPUCORES;
                               ^
../common/sysinfo.inc:33:34: note: expanded from macro 'SYSINFO_CPUCORES'
#define SYSINFO_CPUCORES ( 10 (8 performance and 2 efficiency) )
                                 ^
sysinfo.c:1020:25: note: to match this '('
../common/sysinfo.inc:33:31: note: expanded from macro 'SYSINFO_CPUCORES'
#define SYSINFO_CPUCORES ( 10 (8 performance and 2 efficiency) )
                              ^
1 error generated.
make[1]: *** [obj_all/sysinfo.o] Error 1
make: *** [common_sql] Error 2
```

### Changes Proposed
Add quotation marks to call `HERC_CORES` in `sysinfogen.sh`

**Issues addressed:**  no

### DIFF
--- a/sysinfogen.sh
+++ b/sysinfogen.sh
@@ -238,7 +238,7 @@ cat >> "$OUTFILE" << EOF
 #define SYSINFO_CPU "$( cleanstring "${HERC_CPU}" )"
 
 // CPU Cores (Platform-dependent)
-#define SYSINFO_CPUCORES ( $( cleanstring "${HERC_CORES}" ) )
+#define SYSINFO_CPUCORES ( "$( cleanstring "${HERC_CORES}" )" )
 
 EOF
 [ $? -eq 0 ] || do_fail
@@ -290,4 +290,3 @@ cat >> "$OUTFILE" << EOF
 
 EOF
 [ $? -eq 0 ] || do_fail
-
--- a/sysinfogen.sh
+++ b/sysinfogen.sh
@@ -190,7 +190,7 @@ case $HERC_PLATFORM in
 			HWDATA="$( system_profiler SPHardwareDataType )"
 			HWDATA_CPU="$( echo "$HWDATA" | grep "Processor Name:" | cut -d: -f2- )"
 			HWDATA_CPUSPEED="$( cleanstring "$( echo "$HWDATA" | grep "Processor Speed:" | cut -d: -f2- )" )"
-			HERC_CORES="$( echo "$HWDATA" | grep "Total Number of Cores:" | cut -d: -f2- )"
+			HERC_CORES="$( echo "$HWDATA" | grep "Total Number of Cores:" | cut -d: -f2- | sed -E 's/ *([0-9]+).*/\1/')"
 			HERC_CPU="${HWDATA_CPU} (${HWDATA_CPUSPEED})"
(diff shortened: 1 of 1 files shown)


4144 (approved):

EPuncker (approved):

sysinfogen.sh
```diff
@@ -238,7 +238,7 @@ cat >> "$OUTFILE" << EOF
 #define SYSINFO_CPU "$( cleanstring "${HERC_CPU}" )"
 
 // CPU Cores (Platform-dependent)
-#define SYSINFO_CPUCORES ( $( cleanstring "${HERC_CORES}" ) )
+#define SYSINFO_CPUCORES ( "$( cleanstring "${HERC_CORES}" )" )
```
MishimaHaruna, line 241: This should evaluate to an int rather than a string (it gets assigned to an int variable in [src/common/sysinfo.c](https://github.com/HerculesWS/Hercules/blob/b5a59b17f29a1e36ce815925218e8b9c96ed8f2b/src/common/sysinfo.c#L53))

I think keeping it as an integer has some value, since it can be used in plugins or custom code to take decisions based on the number of cores available.

What do you think about something like this instead? (in your case it'd output `( 10 )`)

```diff --git i/sysinfogen.sh w/sysinfogen.sh
index f96a1c492..fa5bcc3cf 100755
--- i/sysinfogen.sh
+++ w/sysinfogen.sh
@@ -190,7 +190,7 @@ case $HERC_PLATFORM in
 			HWDATA="$( system_profiler SPHardwareDataType )"
 			HWDATA_CPU="$( echo "$HWDATA" | grep "Processor Name:" | cut -d: -f2- )"
 			HWDATA_CPUSPEED="$( cleanstring "$( echo "$HWDATA" | grep "Processor Speed:" | cut -d: -f2- )" )"
-			HERC_CORES="$( echo "$HWDATA" | grep "Total Number of Cores:" | cut -d: -f2- )"
+			HERC_CORES="$( echo "$HWDATA" | grep "Total Number of Cores:" | cut -d: -f2- | sed -E 's/ *([0-9]+).*/\1/')"
 			HERC_CPU="${HWDATA_CPU} (${HWDATA_CPUSPEED})"
 		fi
 		;;

```

I'm not sure if it's worth also exposing the details of how many performance and efficiency cores are available (and I'd need to research how to do so on other OSes as well), if we want to add those as well to the sysinfo information, but I'd still keep those as separate integer values, not as a string

MishimaHaruna (changes requested):

MishimaHaruna (approved):

PR #3272, opened 2024-01-14
Allow call openlapineupgradeui() and openlapineddukddakboxui() without itemid

vietlubu:
### Pull Request Prelude

### Changes Proposed
Allow call `openlapineupgradeui()` and `openlapineddukddakboxui()` without itemid
Example:
```diff
{
	Id: 23926
	AegisName: "Shadow_9_Refine_Hammer"
	Name: "Shadow_9_Refine_Hammer"
	Type: "IT_DELAYCONSUME"
-	Script: <" openlapineupgradeui(23926); ">
+	Script: <" openlapineupgradeui(); ">
},
```

**Issues addressed:** 

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -27941,7 +27941,7 @@ static BUILDIN(openlapineddukddakboxui)
 	struct map_session_data *sd = script_rid2sd(st);
 	if (sd == NULL)
 		return false;
-	const int item_id = script_getnum(st, 2);
+	const int item_id = (script_getnum(st, 2) == 0) ? sd->itemid : script_getnum(st, 2);
 	struct item_data *it = itemdb->exists(item_id);
 	if (it == NULL) {
 		ShowError("buildin_openlapineddukddakboxui: Item %d is not valid\n", item_id);
@@ -27962,7 +27962,8 @@ static BUILDIN(openlapineupgradeui)
 	if (sd == NULL)
 		return false;
 
-	const int item_id = script_getnum(st, 2);
+	const int item_id = (script_getnum(st, 2) == 0) ? sd->itemid : script_getnum(st, 2);
+
 	struct item_data *it = itemdb->exists(item_id);
 	if (it == NULL || it->lapineupgrade == NULL) {
 		ShowError("buildin_openlapineupgradeui: Item %d is not valid\n", item_id);
@@ -29183,8 +29184,8 @@ static void script_parse_builtin(void)
 
 		BUILDIN_DEF(identify, "i"),
 		BUILDIN_DEF(identifyidx, "i"),
-		BUILDIN_DEF(openlapineddukddakboxui, "i"),
-		BUILDIN_DEF(openlapineupgradeui, "i"),
+		BUILDIN_DEF(openlapineddukddakboxui, "?"),
+		BUILDIN_DEF(openlapineupgradeui, "?"),
 
 		BUILDIN_DEF(callfunctionofnpc, "vs*"),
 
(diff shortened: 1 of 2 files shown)


guilherme-gm:
please, update `doc/script_commands.txt` too

EPuncker (approved):

src/map/script.c
```diff
@@ -27941,7 +27941,7 @@ static BUILDIN(openlapineddukddakboxui)
 	struct map_session_data *sd = script_rid2sd(st);
 	if (sd == NULL)
 		return false;
-	const int item_id = script_getnum(st, 2);
+	const int item_id = (script_getnum(st, 2) == 0) ? sd->itemid : script_getnum(st, 2);
```
4144, line 27944: missing parameters need check with ``script_hasdata``
and why it do two times check? may be better read item id once?
also what if ``sd->itemid`` is ``-1``?

4144 (changes requested):

src/map/script.c
vietlubu: What do you think if checking item like this?
```c
	struct item_data *it;
	if ( script_hasdata( st, 2 ) ) {
		if ( script_isstring( st, 2 ) ) {
			const char* item_name = script_getstr( st, 2 );
			it = itemdb->search_name(item_name);
		}else{
			it = itemdb->exists(script_getnum( st, 2 ));
		}
	} else {
		// Call function without param
		it = itemdb->exists(sd->itemid);
	}
```

4144: almost good
style with ``if`` look like from rathena style. need write like this ``if (script_hasdata(st, 2)) {``

``sd->itemid`` almost always equal to ``-1`` except some cases. this item never exists. may be better if equal ``-1`` return error?

vietlubu: @4144 your mean is check `sd->itemid` before check exists, like this?
```c
// Call function without param
if (sd->itemid == -1) {
	return false;
}
it = itemdb->exists(sd->itemid);
```
Because item id will be great then 0, can we check `sd->itemid <= 0` before check item exists?

4144: yes

4144 (approved):

MishimaHaruna (approved):

src/map/script.c
```diff
@@ -27941,16 +27941,37 @@ static BUILDIN(openlapineddukddakboxui)
...
+			it = itemdb->exists(script_getnum(st, 2));
+		}
+	} else {
+		if (sd->itemid > 0) {
+			it = itemdb->exists(sd->itemid);
+		}
+	}
+
+	if (it == NULL || it->lapineddukddak == NULL) {
+		ShowError("buildin_openlapineddukddakboxui: Item Id %d is not valid\n", it->nameid);
```
MishimaHaruna, line 27967: Using `it->nameid` here in case may cause a crash, if `it == NULL` (i.e. if you pass an invalid item id)

src/map/script.c
```diff
@@ -27962,17 +27983,36 @@ static BUILDIN(openlapineupgradeui)
...
+		}
+	} else {
+		if (sd->itemid > 0) {
+			it = itemdb->exists(sd->itemid);
+		}
+	}
+
 	if (it == NULL || it->lapineupgrade == NULL) {
-		ShowError("buildin_openlapineupgradeui: Item %d is not valid\n", item_id);
+		ShowError("buildin_openlapineupgradeui: Item Id %d is not valid\n", it->nameid);
```
MishimaHaruna, line 28008: Same as above about `it->nameid`

MishimaHaruna (changes requested):

MishimaHaruna (approved):

PR #3274, opened 2024-01-19
Fixed attack/chase regression when walking towards enemy

csnv:
### Pull Request Prelude

### Changes Proposed
Reverts changes made to unit.c in https://github.com/HerculesWS/Hercules/commit/964b47351ef2156423f6e0a68bfd3361283936c1

Credits for discovering the bug and its solution to @skyleo , as well as the following explanation:
> I think it affects for example moments where you use a skill out of range on a mob and you think you reached in range
and then it walks out of range
probably more like auto-attack though
But yeah chasing and atk behavior failing under certain situations.

**Issues addressed:** 

### DIFF
--- a/src/map/unit.c
+++ b/src/map/unit.c
@@ -526,7 +526,7 @@ static int unit_walk_toxy_timer(int tid, int64 tick, int id, intptr_t data)
 				unit->attack(bl, tbl->id, ud->state.attack_continue);
 			}
 		} else { // Update chase-path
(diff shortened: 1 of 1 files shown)


skyleo (approved):
lgtm


**Note: There may have been behavior introduced after that commit that tried to fix bugs this introduced, which are now obsolete and might actually do harm now. Applying this fix is still the right thing to do, but one should keep in mind about the kludges that may exist due to this and need to be removed.**

PR #3275, opened 2024-01-21
Add script function to work with item_group

lucasgolino:
### Pull Request Prelude

### Changes Proposed

This PR aims to create a script function to get all items without duplicated from a specific `group_id`. The idea behind this is be able to get all items and be capable to create more complex NPC with it. Ideal for this is: use more groups feature to handle with a list of group in a easiest way without needed to create arrays of a item listing all itens you need, and instead of it, use groups list as resource in NPC.

In same PR i create an utilities tool set to clean duplicated from group in order to get only unique list from group.

Note: I add script command on `docs/script_commands.txt`  

Example:
```c++
getitemgroupitems(.@item_list[0], "Old_Card_Album");
for ( .@i = 0; .@i < getarraysize(.@item_list); .@i++ ) {
    mes "^1E56BD\t"+getitemname(.@item_list[.@i])+" ^000000";
}
```

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -11142,14 +11142,17 @@ Check out some examples in `doc/sample/goldpc.txt`.
 
 *getitemgroupitems(<item_id_array>, <item_group_id>)
 
-Return boolean if success otherwise false.
-Will allocate Without duplicated item at <item_id_array> with all items in <item_group_id>.
+Fills the array <item_id_array> with the IDs of the items from the <item_group_id> item group, without duplicates.
 
-<item_group_id> can be defined on `db/(pre-)re/item_group_db.conf`
+Returns the amount of items inserted into the array.
+
+Note: The array will not be automatically cleared. It's recommended to use the returned value to loop over the array instead of relying on getarraysize.
+
+The possible values for <item_group_id> can be found in `db/(pre-)re/item_group.conf`
 
 Example:
 
-	getitemgroupitems(.@item_list[0], "Old_Card_Album");
-	for (.@i = 0; .@i < getarraysize(.@item_list); .@i++) {
-		mes "^1E56BD\t"+getitemname(.@item_list[.@i])+" ^000000";
+	.@count = getitemgroupitems(.@item_list[0], Old_Card_Album);
+	for (.@i = 0; .@i < .@count; ++.@i) {
+		mesf("^1E56BD\t%s^000000", getitemname(.@item_list[.@i]));
 	}
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -11149,7 +11149,7 @@ Will allocate Without duplicated item at <item_id_array> with all itens in <item
 
 Example:
 
-	getitemlistbygroup(.@item_list[0], "Old_Card_Album");
+	getitemgroupitems(.@item_list[0], "Old_Card_Album");
 	for ( .@i = 0; .@i < getarraysize(.@item_list); .@i++ ) {
 		mes "^1E56BD\t"+getitemname(.@item_list[.@i])+" ^000000";
 	}
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -8037,7 +8037,7 @@ false means OFF, and true means ON. See setmapflag() for examples of mapflags.
 *getbattleflag("<battle flag>")
 
 Sets or gets the value of the given battle flag.
-Battle flags are the flags found in the conf/map/battle/*.conf files and is
+Battle flags are the flags found in the conf/map/battle/\*.conf files and is
 also used in Lupus' variable rates script.
 
 Examples:
@@ -11136,3 +11136,20 @@ When <played_time> is not provided, it defaults to -1, and the current played ti
           If keeping the GoldPC mode between sessions is desired, you may use OnPCLoginEvent.
 
 Check out some examples in `doc/sample/goldpc.txt`.
+
+
+---------------------------------------
+
+*getitemgroupitems(<item_id_array>, <item_group_id>)
+
+Return boolean if success otherwise false.
+Will allocate Without duplicated item at <item_id_array> with all itens in <item_group_id>.
+
+<item_group_id> can be defined on `db/(pre-)re/item_group_db.conf`
+
+Example:
+
+	getitemlistbygroup(.@item_list[0], "Old_Card_Album");
+	for ( .@i = 0; .@i < getarraysize(.@item_list); .@i++ ) {
+		mes "^1E56BD\t"+getitemname(.@item_list[.@i])+" ^000000";
+	}
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -8037,7 +8037,7 @@ false means OFF, and true means ON. See setmapflag() for examples of mapflags.
 *getbattleflag("<battle flag>")
 
 Sets or gets the value of the given battle flag.
-Battle flags are the flags found in the conf/map/battle/\*.conf files and is
+Battle flags are the flags found in the conf/map/battle/*.conf files and is
 also used in Lupus' variable rates script.
 
 Examples:
@@ -11143,13 +11143,13 @@ Check out some examples in `doc/sample/goldpc.txt`.
 *getitemgroupitems(<item_id_array>, <item_group_id>)
 
 Return boolean if success otherwise false.
-Will allocate Without duplicated item at <item_id_array> with all itens in <item_group_id>.
+Will allocate Without duplicated item at <item_id_array> with all items in <item_group_id>.
 
 <item_group_id> can be defined on `db/(pre-)re/item_group_db.conf`
 
 Example:
 
 	getitemgroupitems(.@item_list[0], "Old_Card_Album");
-	for ( .@i = 0; .@i < getarraysize(.@item_list); .@i++ ) {
+	for (.@i = 0; .@i < getarraysize(.@item_list); .@i++) {
 		mes "^1E56BD\t"+getitemname(.@item_list[.@i])+" ^000000";
 	}
--- a/src/map/itemdb.c
+++ b/src/map/itemdb.c
@@ -347,6 +347,26 @@ static bool itemdb_in_group(struct item_group *group, int nameid)
 	return false;
 }
 
+/**
+ * Search for a group of item
+ * @param group return information about group
+ * @param nameid item group id to search
+ * @return bool true if found, false otherwise
+ */
+static bool itemdb_search_group(struct item_group *group, int nameid)
+{
+	for(int i = 0; i < itemdb->group_count; i++ ) {
+		struct item_group g = itemdb->groups[i];
+		
+		if( g.id == nameid ) {
+			*group = g;
+			return true;
+		}
+	}
+
+	return false;
+}
+
 /// Searches for the item_data.
 /// Returns the item_data or NULL if it does not exist.
 static struct item_data *itemdb_exists(int nameid)
@@ -3357,6 +3377,7 @@ void itemdb_defaults(void)
 	itemdb->option_exists = itemdb_option_exists;
 	itemdb->reform_exists = itemdb_reform_exists;
 	itemdb->in_group = itemdb_in_group;
+	itemdb->search_group = itemdb_search_group;
 	itemdb->group_item = itemdb_searchrandomid;
 	itemdb->chain_item = itemdb_chain_item;
 	itemdb->package_item = itemdb_package_item;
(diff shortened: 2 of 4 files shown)


doc/script_commands.txt
```diff
@@ -8037,7 +8037,7 @@ false means OFF, and true means ON. See setmapflag() for examples of mapflags.
 *getbattleflag("<battle flag>")
 
 Sets or gets the value of the given battle flag.
-Battle flags are the flags found in the conf/map/battle/*.conf files and is
+Battle flags are the flags found in the conf/map/battle/\*.conf files and is
```
EPuncker, line 8040:
```suggestion
Battle flags are the flags found in the conf/map/battle/*.conf files and is
```

doc/script_commands.txt
```diff
@@ -11134,3 +11134,20 @@ When <played_time> is not provided, it defaults to -1, and the current played ti
...
 
 Check out some examples in `doc/sample/goldpc.txt`.
+
+
+---------------------------------------
+
+*getitemgroupitems(<item_id_array>, <item_group_id>)
+
+Return boolean if success otherwise false.
+Will allocate Without duplicated item at <item_id_array> with all itens in <item_group_id>.
```
EPuncker, line 11144:
```suggestion
Will allocate Without duplicated item at <item_id_array> with all items in <item_group_id>.
```

doc/script_commands.txt
```diff
@@ -11134,3 +11134,20 @@ When <played_time> is not provided, it defaults to -1, and the current played ti
...
+
+<item_group_id> can be defined on `db/(pre-)re/item_group_db.conf`
+
+Example:
+
+	getitemgroupitems(.@item_list[0], "Old_Card_Album");
+	for ( .@i = 0; .@i < getarraysize(.@item_list); .@i++ ) {
```
EPuncker, line 11151:
```suggestion
	for (.@i = 0; .@i < getarraysize(.@item_list); .@i++) {
```

src/map/itemdb.c
```diff
@@ -347,6 +347,26 @@ static bool itemdb_in_group(struct item_group *group, int nameid)
...
 
+/**
+ * Search for a group of item
+ * @param group return information about group
+ * @param nameid item group id to search
+ * @return bool true if found, false otherwise
+ */
+static bool itemdb_search_group(struct item_group *group, int nameid)
+{
+	for(int i = 0; i < itemdb->group_count; i++ ) {
```
EPuncker, line 358:
```suggestion
	for(int i = 0; i < itemdb->group_count; i++) {
```

src/map/itemdb.c
```diff
@@ -347,6 +347,26 @@ static bool itemdb_in_group(struct item_group *group, int nameid)
...
+		struct item_group g = itemdb->groups[i];
+		
+		if( g.id == nameid ) {
```
EPuncker, line 361:
```suggestion
		if (g.id == nameid) {
```

src/map/script.c
```diff
@@ -9133,6 +9133,92 @@ static BUILDIN(grouprandomitem)
...
+	}
+
+	struct item_group group;
+	bool finded_group = itemdb->search_group(&group, item->nameid);
+	if (!finded_group) {
+		ShowWarning("buildin_getitemgroupitems: any group find\n");
+		return false;
+	}
+
+	if(!utils_itemgroup_remove_duplicate(&group)) {
```
EPuncker, line 9205:
```suggestion
	if (!utils_itemgroup_remove_duplicate(&group)) {
```

doc/script_commands.txt
lucasgolino: I change this because `/*` make entire code as comment.

src/map/itemdb.c
```diff
@@ -3357,6 +3377,7 @@ void itemdb_defaults(void)
 	itemdb->option_exists = itemdb_option_exists;
 	itemdb->reform_exists = itemdb_reform_exists;
 	itemdb->in_group = itemdb_in_group;
+	itemdb->search_group = itemdb_search_group;
```
csnv, line 3380: Please generate the required hook for this new function with the utilites from `/tools/HPMHookGen/`

src/map/script.c
```diff
@@ -9133,6 +9133,92 @@ static BUILDIN(grouprandomitem)
 	return true;
 }
 
+/*==========================================
+ * utils for item group
+ *------------------------------------------*/
+static bool utils_itemgroup_remove_duplicate(struct item_group *group)
```
csnv, line 9139: This should probably be added to interface and change the name to reflect the interface `script`

src/map/itemdb.c
```diff
@@ -347,6 +347,26 @@ static bool itemdb_in_group(struct item_group *group, int nameid)
...
 }
 
+/**
+ * Search for a group of item
+ * @param group return information about group
+ * @param nameid item group id to search
+ * @return bool true if found, false otherwise
+ */
+static bool itemdb_search_group(struct item_group *group, int nameid)
+{
```
csnv, line 357: `itemdb->search` returns the item directly instead of a bool, I think it's better to follow the same logic and return the group that was found (`NULL` if not found).

doc/script_commands.txt
csnv: I would say that's the fault of the code editor, as it's a txt file it should not auto-detect C comments.
This change may cause some legibility problems, as it's not clear what's the purpose of that backslash in a path.

doc/script_commands.txt
```diff
@@ -11134,3 +11134,20 @@ When <played_time> is not provided, it defaults to -1, and the current played ti
...
+Return boolean if success otherwise false.
+Will allocate Without duplicated item at <item_id_array> with all items in <item_group_id>.
+
+<item_group_id> can be defined on `db/(pre-)re/item_group_db.conf`
+
+Example:
+
+	getitemgroupitems(.@item_list[0], "Old_Card_Album");
+	for (.@i = 0; .@i < getarraysize(.@item_list); .@i++) {
+		mes "^1E56BD\t"+getitemname(.@item_list[.@i])+" ^000000";
```
MishimaHaruna, line 11152: Please encourage the use of a more modern syntax:

```suggestion
		mesf("^1E56BD\t%s^000000", getitemname(.@item_list[.@i]));
```

src/map/script.c
```diff
@@ -9133,6 +9133,92 @@ static BUILDIN(grouprandomitem)
...
+
+	if (!data_isreference(data)) {
+		ShowError("buildin_getitemgroupitems: not a variable\n");
+		script->reportdata(data);
+		script_pushnil(st);
+		st->state = END;
+		return false;// not a variable
+	}
+
+	struct item_data *item = itemdb->name2id(script_getstr(st, 3));
```
MishimaHaruna, line 9191: I'm not a fan of using strings where you could use integers. Can you make this function accept an integer instead? (i.e. the constant `Old_Card_Album` instead of the string `"Old_Card_Album"`).

If you think it can be of use, you can make it accept both, but item IDs are primarily numbers (for which we have constants) and we should never force scripters to use strings instead.

src/map/script.c
```diff
@@ -9133,6 +9133,92 @@ static BUILDIN(grouprandomitem)
...
+	if (!item) {
+		ShowError("buildin_getitemgroupitems: invalid item name [%s]\n", script_getstr(st, 2));
+		script_pushnil(st);
+		return false;
+	}
+
+	struct item_group group;
+	bool finded_group = itemdb->search_group(&group, item->nameid);
+	if (!finded_group) {
```
MishimaHaruna, line 9200: Nitpick:

```suggestion
	bool found_group = itemdb->search_group(&group, item->nameid);
	if (!found_group) {
```

src/map/script.c
```diff
@@ -9133,6 +9133,92 @@ static BUILDIN(grouprandomitem)
...
+		ShowWarning("buildin_getitemgroupitems: any group find\n");
```
MishimaHaruna, line 9201:
```suggestion
		ShowWarning("buildin_getitemgroupitems: item group not found\n");
```

src/map/script.c
```diff
@@ -9133,6 +9133,92 @@ static BUILDIN(grouprandomitem)
...
+		return false;
+	}
+
+	int32 idata = reference_getindex(data);
+	int32 dataid = reference_getid(data);
+	const char *data_name = reference_getname(data);
+	
+	for (int i = 0; i < group.qty; i++) {
+		int id = group.nameid[i];
+		script->set_reg(st, NULL, reference_uid(dataid, idata + i), data_name, (const void *)h64BPTRSIZE(id), (data)->ref);
```
MishimaHaruna, line 9216: Please note that by doing this, if the array the user passed was not empty and had more items than the current item group, they won't be cleared (and the user has no way to know which values are real and which are garbage left over).

I see two options here: either make this commandd return the amount of items you put in the array, so that the user can refer to that instead of getarray size, or warn the user in the documentation that it is their responsibility to ensure that the array is clean before passing it to this command.

So for example, if you choose the former, the command should be used this way:

```
.@count = getitemgroupitems(.@item_list[0], "Old_Card_Album");
	mes getitemname(.@item_list[.@i]);
```

or if you choose the latter:

```
deletearray(.@item_list[0]);
getitemgroupitems(.@item_list[0], "Old_Card_Album");
	mes getitemname(.@item_list[.@i]);
```

MishimaHaruna (changes requested):

MishimaHaruna (approved):

src/map/itemdb.c
```diff
@@ -348,23 +348,17 @@ static bool itemdb_in_group(struct item_group *group, int nameid)
 }
 
 /**
- * Search for a group of item
- * @param group return information about group
+ * Search for an item group
  * @param nameid item group id to search
- * @return bool true if found, false otherwise
+ * @return A pointere to the group, or NULL if not found.
```
skyleo, line 353: pointer*

maybe split via `@retval`  ?

src/map/itemdb.c
```diff
@@ -348,23 +348,17 @@ static bool itemdb_in_group(struct item_group *group, int nameid)
 }
 
 /**
- * Search for a group of item
- * @param group return information about group
+ * Search for an item group
  * @param nameid item group id to search
```
skyleo, line 352: Empty line between function description and params please.

src/map/script.c
```diff
@@ -9228,12 +9167,33 @@ static BUILDIN(getitemgroupitems)
...
+
 	int count = 0;
-	for (int i = 0; i < group.qty; i++) {
-		int id = group.nameid[i];
+	for (int i = 0; i < group->qty; i++) {
+		int id = group->nameid[i];
+		int j = 0;
+		ARR_FIND(0, i, j, group->nameid[i] == group->nameid[j]);
+		if (i != j) {
+			// Already encountered - skip duplicates
```
skyleo, line 9192: so it's possible to have duplicates in item_group stuff but this script command will ignore that fact? Is there no way to combine duplicate entries into one for this? Or do duplicates really have no effect?

skyleo (changes requested):

src/map/script.c
```diff
@@ -9212,6 +9212,22 @@ static BUILDIN(getitemgroupitems)
 	int32 idata = reference_getindex(data);
 	int32 dataid = reference_getid(data);
 	const char *data_name = reference_getname(data);
+
+	if (not_server_variable(*data_name)) {
```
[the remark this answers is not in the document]
MishimaHaruna: The name is misleading, but it's meant to differentiate between variables that require a RID attached from those that don't (npc, scope, global permanent, global temporary, instance)

src/map/script.c
```diff
@@ -9212,6 +9212,22 @@ static BUILDIN(getitemgroupitems)
...
+			return false;
+		}
+	}
+
+	if (is_string_variable(data_name)) {
+		// string array
+		ShowError("buildin_getitemgroupitems: not an integer array reference\n");
+		script->reportdata(data);
+		st->state = END;
+		return false;
```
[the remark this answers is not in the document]
MishimaHaruna: I think it's worth differentiating, this one is an error that correct code will never hit (so it's ok to abort execution and show a trace), while the other may be triggered only based on certain conditions (i.e. player logging out leaving behind a background script running)

src/map/script.c
MishimaHaruna: Duplicates affect the probability that a given item is selected. But we're returning a bare list of IDs, without caring about the probability. We could return both, but then we'd need the user to either pass two arrays, or to return an array of interleaved IDs and counts. I kept the original behavior from the PR author of only returning the IDs for now

src/map/script.c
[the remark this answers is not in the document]
skyleo: Hmm, seems reasonable. okay.

src/map/script.c
skyleo: Ah, okay. As long as the list doesn't contain information about the chances, it's fine.

PR #3279, opened 2024-02-09
2-2 jobs skill rebalance - Bard/Dancer (2018 Patch/Renewal)

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
The main purpose of this PR is to introduce the rebalance of Bard/Dancer job skills. This change affects Renewal-only.

On official servers this came along with rebalances of 1st, 2nd jobs and transclass too.

I am not sure whether this could be moved like that, because some skills from transclasses will not play well with the rebalanced ones.

The implementation in this PR is based on kRO and kRO zero patch notes, iRO Wiki, rAthena and divine pride info, along with some in-game testing. I can't say everything is 100% accurate because there were discrepancies between different sources, and I could not test everything in kRO, but should be quite close.

I won't list all the rebalance changes in the PR description, but it may be checked in each commit text. Also, the commits are in the same order as they appear in the references below.

**Notes**
1. Musical Strike and Slinging Arrow has different info about the new Cast Time in each source. (See commit for info). I am using Divine Pride data.
2. FrostJoker Cooldown: Sakray notes and Divine Pride says 5s. Main notes and iRO Wiki says 4s. Following 4s.
3. Bragi duration 60s -> 180s is from 2019-12-04 and was intentionally left out: https://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=7191&curpage=1
4. Some SP Cost changes is not mentioned in iRO wiki / DP / kRO notes, but it is set in kRO Lua and divine pride.
5. I am not 100% sure on how confusion on Rockisveil works
6. I am not 100% sure if drum battle ATK bonus really is equip atk like MC_LOUD. but It does come as SP_ATK2.
7. From my tests, attack bonus granted by DC_DANCINGLESSON shows in the UI right side of ATK. But in herc it does not. Since this part was not changed in the renewal rebalance, I am unsure whether Mastery was supposed to always show in the UI and we have a long term bug. I won't touch this for now and open an issue for further investigation.

**Affected skills**
- BA_APPLEIDUN (Song of Lutie)
- BA_ASSASSINCROSS (Impressive Riff)
- BA_DISSONANCE (Unchained Serenade)
- BA_FROSTJOKE (Unbarring Octave)
- BA_MUSICALLESSON (Music Lessons)
- BA_MUSICALSTRIKE (Melody Strike)
- BA_POEMBRAGI (Magic Strings)
- BA_WHISTLE (Perfect Tablature)
- BD_ADAPTATION (Amp)
- BD_DRUMBATTLEFIELD (Battle Theme)
- BD_ENCORE (Encore)
- BD_ETERNALCHAOS (Down Tempo)
- BD_INTOABYSS (Power Chord)
- BD_LULLABY (Lullaby)
- BD_RICHMANKIM (Mental Sensing)
- BD_RINGNIBELUNGEN (Harmonic Lick)
- BD_ROKISWEIL (Classical Pluck)
- BD_SIEGFRIED (Acoustic Rhythm)
- DC_DANCINGLESSON (Dance Lessons)
- DC_DONTFORGETME (Slow Grace)
- DC_FORTUNEKISS (Lady Luck)
- DC_HUMMING (Focus Ballet)
- DC_SERVICEFORYOU (Gypsy's Kiss)
- DC_THROWARROW (Slinging Arrow)
- DC_UGLYDANCE (Hip Shaker)

**References:**
- [kRO 2018.10.31 patch note](https://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=7033&curpage=21)
- [kRO 2018.11.28 patch note / translated](https://board.herc.ws/topic/17043-kro-patch-2018-11-28/)
- [Divine Pride](https://www.divine-pride.net/forum/index.php?/topic/3453-kro-mass-skills-balance-1st-2nd-and-transcendent-classes-skills/)

**Issues addressed:** 
Part of #2727 


src/map/status.c
```diff
@@ -7813,6 +7813,15 @@ static int status_change_start_sub(struct block_list *src, struct block_list *bl
 				val3 = total_tick/1000; //Tick duration
 				tick_time = 1000; // [GodLesZ] tick time
 				break;
+
+#ifdef RENEWAL
+			case SC_POEMBRAGI:
+				// val1 = skill lv
+				val2 = val1 * 2; // Cast time reduction
```
guilherme-gm, line 7955: I have been wondering whether we could/should/want to move these vals to `SkillData2` and `SkillData3` instead of defining them here. (For every skill, not only bragi)

src/map/skill.c
```diff
@@ -11709,6 +11863,103 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
...
+ * - flag & 1 :
+ *     when not set: it is the initial skill cast,
+ *     when set: it is in the reiteration over targets to get buffed.
+ * - enum e_battle_check_target flags: defines who will be affected by the reiteration. examples:
+ *     flag | BCT_PARTY -- affects party members (e.g. apply SC to all party members)
+ *     flag | BCT_ENEMY -- affects enemies (e.g. apply SC to all enemies)
+ *
+ * - @TODO: Document other possible flag values
+ */
+static void skill_castend_nodamage_id_sc_song(struct block_list *src, struct block_list *bl, uint16 skill_id, uint16 skill_lv, int64 tick, int flag)
```
guilherme-gm, line 11883: I tried to start making the idea of "separate function for each skill" a thing here, since most dance/songs are very similar. But I ended up with:
1. this one having a if for lullaby
2. `skill_castend_nodamage_id_ugly_dance` having a duplicated code for most of it
3. the caller function needing to give the target

I am open for suggestions on how I could improve that, I had a few random thoughts but didn't tried too much:
1. Moving chance to SkillData2 (but except for 1 case, it is always 100%...)
2. Making more functions, with their exclusive logic and those would call this one. Something like:
     - song_friend_sc --> sc_song w/ Party + chance = 100%
     - song_enemy_sc --> sc_song w/ enemy + chance = 100%
     - song_lullaby --> sc_song w/ enemy + chance = 15%
     - song_ugly_dance --> sc_song w/ enemy + special logic for flag 1

But well, I am not sure whether those would work nicely.

src/map/skill.c
```diff
@@ -13214,95 +13468,98 @@ static struct skill_unit_group *skill_unitsetting(struct block_list *src, uint16
 		case DC_UGLYDANCE:
 			val1 = 10; //FIXME: This value is not used anywhere, what is it for? [Skotlex]
 			break;
+
+#ifndef RENEWAL
```
guilherme-gm, line 13472: I am not sure whether I should add some toggle to rollback to previous music/dance behavior for RE. The new effects requires 2018-11+ clients or characters will show no visual effect.

But making a toggle for that much of changes is... complicated. I don't know where to set the boundaries, and we have several conf files to change, not only skill_db.

src/map/pc.c
```diff
@@ -6829,6 +6831,16 @@ static void pc_calcexp(struct map_session_data *sd, uint64 *base_exp, uint64 *jo
 		}
 #endif
 
+#ifdef RENEWAL
+		// after rebalance, boss monsters does give more EXP, but "MVP" exp is not increased
+		if (sd->sc.data[SC_RICHMANKIM] != NULL && src->type == BL_MOB && (flags & EXP_FLAG_MVP) == 0) {
+			buff_ratio += sd->sc.data[SC_RICHMANKIM]->val1;
+			// jexp is boosted by both buff_ration AND buff_job_ratio, so we should not use buff_job_ratio here
```
guilherme-gm, line 6838: @skyleo can you check if this part about buff_ratio / buff_job_ratio is correct on the general exp calc? I find it very weird that job exp boost is buff_ratio + buff_job_ratio: (near the end of the function)

```
	bexp += apply_percentrate64(bexp, buff_ratio, 100);
	jexp += apply_percentrate64(jexp, buff_ratio + buff_job_ratio, 100);
```

maybe @hemagx remembers something... but this change is from 8 years ago 😅

src/map/skill.c
```diff
@@ -15277,12 +15278,18 @@ static int skill_check_condition_char_sub(struct block_list *bl, va_list ap)
 				return 1;
 			default: //Warning: Assuming Ensemble Dance/Songs for code speed. [Skotlex]
 				{
-					uint16 skill_lv;
 					if(pc_issit(tsd) || !unit->can_move(&tsd->bl))
 						return 0;
+
+					uint16 skill_lv = pc->checkskill(tsd, skill_id);
+#ifdef RENEWAL // In Renewal, partner also gets the requirements consumed, so we must check it
```
guilherme-gm, line 15310: @skyleo another question, is it correct that in Pre-RE, when we cast an Ensemble skill, the partner doesn't pay the initial SP requirement?
Example: Ring Nibelungen costs 50 SP to be cast + 1 SP/3 second.

If player A casts it, and player B is forced to join it, then:
1. Player A pays 50 SP
2. After 3 seconds: Player A + Player B pays 1 SP
3. ...

Is that right for Pre-RE? or should Player B also pay 50 SP to start the ensemble skill? In Renewal after rebalance, I could observe that this is the behavior, but the existing code didn't consume the requirement from Player B.

src/map/skill.c
MishimaHaruna: While I was originally in favor of keeping the option to revert these rebalance changes to support older clients, I realize it's too much of a burden (in both implementation and future maintenance), so I'll take it back.

Considering that I expect a lot of people running renewal servers to have adopted 2018 or newer clients by now (if anything, for the UI improvements they bring), and that it's completely out of the question to have a third "pre-rebalance" build mode and db folder to maintain, I'm in favor of requiring a 2018-11+ client for renewal.

People that wish to run a renewal server on an older client (or want to avoid the rebalance changes), can revert these changes on their fork (or patch their clients if someone comes up with client backport patches). And people that want to include some of these rebalance changes in their pre-renewal servers, may also do so in their fork.

Let's keep it rolling - no reverts for renewal, while pre-renewal stays as it is without the rebalance changes

violent01:
One thing I noticed here is that Ensemble skills can only be used the old way(Bard & Dancer must be close to each other). The new way is that Ensemble skills can be used around 9x9 cell distance.

violent01:
Missing input in skill_db BD_INTOABYSS. You forgot to put `SplashRange: -1`.

db/re/skill_db.conf
```diff
@@ -9405,26 +9406,20 @@ skill_db: (
...
 	DamageType: {
 		NoDamage: true
+		SplashArea: true
 	}
-	SkillData1: 60000
-	SkillData2: 15000
+	SplashRange: 4
+	SkillData1: 15_000 // SC_SLEEP duration (in miliseconds)
 	FixedCastTime: 0
+	CoolDown: 20_000
```
violent01, line 9417: Missing input `CastTime: 1_000`

db/re/skill_db.conf
```diff
@@ -9639,27 +9639,20 @@ skill_db: (
 	AttackType: "Misc"
 	DamageType: {
 		NoDamage: true
+		SplashArea: true
 	}
-	SkillData1: 60000
-	SkillData2: 60000
```
violent01, line 9770: Missing input `SplashRange: -1`

src/map/pc.c
skyleo: A bit odd since this means the bonuses from SC_CASH_PLUSEXP and SC_OVERLAPEXPUP are applied twice for job exp.

Also since when does SC_RICHMANKIM affect players? The experience calculation for SC_RICHMANKIM is in mob.c and basically only mobs that are inside Richmankim, will get SC_RICHMANKIM on them and thus give more experience.

From briefly reading iro wiki it seems that this hasn't changed. I suggest you introduce your change in mob.c at the Experience Calculation around line 2580 in function mob_dead.

src/map/skill.c
skyleo: Yeah this seems to be the behavior in Pre-RE from what I've seen in Ghidra.

src/map/pc.c
guilherme-gm: Yeah, I found odd that SC_CASH_PLUSEXP and SC_OVERLAPEXPUP applies twice, I did not test it in-game to check, but looks like a bug to me... not exactly a re-balance issue though.

---

Regarding RICHMANKIM:

After Renewal re-balance, SC_RICHMANKIM is a player buff just like Increase AGI/etc, there is no longer ground effects (monsters does not get SC_RICHMANKIM now).

From my in-game tests, the EXP is applied per-player.

Consider the following examples for a party of 2 players with even sharing, killing a monster where penalty is applied equally to both members (to make explanation simpler)

**Example 1:** Player A and Player B **NOT** buffed
    - Result: Player A and Player B gets the same amount of EXP

**Example 2:** Player A **Buffed** / Player B **NOT** buffed
    - Result: Player A always gets more EXP than Player B (regardless of who performed the kill)

**Example 3:** Player A and Player B **Buffed**
    - Result: Player A and Player B gets the same amount of EXP (increased by the effect)

And MVP exp does not get boosted.

I did try to keep it in mob.c, since we are talking about mob exp, but doing it there is... complicated. EXP at `mob_dead` did not consider the party EXP split yet, and since my tests have shown that a non-buffed player will get less EXP, I would need to change the party sharing code.

guilherme-gm:
@violent01 thanks for checking, fixed the missing values

MishimaHaruna (approved):

PR #3291, opened 2024-04-27
Fix Soul Balls in newer clients

guilherme-gm:
### Pull Request Prelude

### Changes Proposed

Fixes soul balls not being properly displayed in newer clients. (2020+).

2020+ clients split the logic for spirit spheres and soul balls, players may now have both at the same time, and a new packet was introduced for soul balls only (being separated from the spirit spheres one).

Implementation based on rAthena logic/packet. also huge thanks to 4144 for packet dates.

I decided to split `clif_spiritball` into 2 functions and leave the number of balls to the caller so we can simplify those functions. There were 2 reasons for that:
1. We now have to deal with different packets for each case depending on client
2. Pretty much every call was converting sd/hd to bl, to convert back to sd/hd inside the function... leaving to the caller simplifies things and gives flexibility if someone is looking for some sort of customization

**Issues addressed:** 
None, I think (someone mentioned it in discord)


src/map/skill.c
```diff
@@ -5063,7 +5063,13 @@ static int skill_castend_damage_id(struct block_list *src, struct block_list *bl
 					}
 					clif->slide(src, src->x, src->y);
 					clif->fixpos(src);
-					clif->spiritball(src, BALL_TYPE_SPIRIT, AREA);
+
+					int spiritballs = 0;
+					if (sd != NULL)
+						spiritballs = sd->spiritball;
+					else if (src->type == BL_HOM)
+						spiritballs = BL_UCAST(BL_HOM, src)->homunculus.spiritball;
```
guilherme-gm, line 5071: maybe this should go to `status`? we do that twice

src/map/clif.c
```diff
@@ -8099,65 +8099,47 @@ static void clif_devotion(struct block_list *src, struct map_session_data *tsd)
...
-			p.num = sd->spiritball;
-			break;
-		case BALL_TYPE_SOUL:
-			p.AID = sd->bl.id;
-			p.num = sd->soulball;
-			break;
-		case BALL_TYPE_NONE:
-			break;
-		}
-		clif->send(&p, sizeof(struct PACKET_ZC_SPIRITS), ((bl == NULL && spirit == BALL_TYPE_SOUL) ? &sd->bl : bl), (spirit == BALL_TYPE_SPIRIT ? AREA : target));
```
guilherme-gm, line 8138: the logic here in previous version is somewhat confusing to me, but:
1. `nullpo_retv(bl);` at the start, so `(bl == NULL && ..)` is always false, thus `bl` is always the target (Well, it would be anyway)
2. the `target` for SPIRIT is always `AREA`, this is given by the caller already.


so simply changing to `bl, target` is equivalent.

MrKeiKun: 1. I agree
2. `target `is not always `AREA`

src/map/clif.c
```diff
@@ -9761,7 +9751,7 @@ static void clif_refresh(struct map_session_data *sd)
 	if (sd->charm_type != CHARM_TYPE_NONE && sd->charm_count > 0)
 		clif->charm_single(sd->fd, sd);
 	if (sd->soulball > 0)
-		clif->spiritball(&sd->bl, BALL_TYPE_SOUL, SELF);
+		clif->soulball(&sd->bl, sd->soulball, SELF);
```
MrKeiKun, line 9754: Like this one.

guilherme-gm: oh, what I tried to say is that for `BALL_TYPE_SPIRIT`, `clif->spiritball` is already called with `AREA` every time, no need to "hardcode" it in the ternary (thus, simply using `target` is enough). Just trying to justify that removing the ternary was safe, because `clif->spiritball` was already called with AREA every time it was called for "spirit ball" instead of "soul ball".

This one was not `AREA`, but it was also not `BALL_TYPE_SPIRIT`

src/map/clif.h
```diff
@@ -1209,7 +1209,8 @@ struct clif_interface {
 	void (*useskill) (struct block_list* bl, int src_id, int dst_id, int dst_x, int dst_y, uint16 skill_id, uint16 skill_lv, int casttime);
 	void (*produce_effect) (struct map_session_data* sd,int flag,int nameid);
 	void (*devotion) (struct block_list *src, struct map_session_data *tsd);
-	void (*spiritball) (struct block_list *bl, enum spirit_ball_types spirit, enum send_target target);
+	void (*soulball) (struct block_list *bl, int soulballs, enum send_target target);
+	void (*spiritball) (struct block_list *bl, int spiritballs, enum send_target target);
```
MishimaHaruna, line 1213: Can you change this signature (i.e. rename) so that it becomes incompatible with the previous version?

The goal is to avoid old custom code that was calling `clif->spiritball()` to compile without errors or warnings, and silently cause unexpected behavior since the meaning of the second argument has changed in an incompatible way

MishimaHaruna (changes requested):

src/map/clif.h
guilherme-gm: I thought the change from enum to int would be enough, but is not... updated by changing both to plural (`clif->spiritballs` / `clif->soulball`).

Thanks!

MishimaHaruna (approved):

PR #3292, opened 2024-04-28
Fix issues with Auto Shadow Spell

### LINKED ISSUE #3286: Auto Shadow Spell Nullpo Error
**Describe the bug**
Nullpo error when using Auto Shadow Spell skill. Hoping someone can reproduce. 

**To Reproduce**
1. Use Auto Shadow Spell.
2. Click cancel or ok without selecting a skill.

**Expected behavior**
Skill should not trigger nullpo error.

**Screenshots**
image

 - OS: [CentOS 7]
 - Hercules Version [v2024.02]
 - Mode: [pre-re]
 - Packet version: [20160113]
 - Client type: [main]

**Plugins used or source modifications**
N/A

**Additional context**
N/A

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
Fixes a few issues with Auto Shadow Spell.

1. When casting Auto Shadow Spell, we now properly check if there are skills copied with Intimidade/Reproduce before trying to get the skill_id
   This prevents assertion failures due to the skill id being "0" (no skill)
   In current code, this would still work, but cause a log about bad skill handling

2. Fixed a check that prevented you from switching an active shadow spell to another when their Skill ID was smaller (due to sc val1)

3. Fix assertion error when clicking "cance" or clicking "ok" in Auto Spell list without selecting a skill
   As far as I could analyze, the client is not cleaning up the field that stores the current selected skill, so when you click "ok" it just sends a garbage id that may (or may not) be a valid skill. We now validate before running the skill code.
   For cancel, it sends id "0", which is also fixed now since 0 is not a valid skill id.

As a bonus, I've converted both packets related to that flow to use structs.

**Issues addressed:** Closes #3286 

### DIFF
--- a/src/map/packets_struct.h
+++ b/src/map/packets_struct.h
@@ -3319,6 +3319,14 @@ struct PACKET_ZC_MAKINGARROW_LIST {
 } __attribute__((packed));
 DEFINE_PACKET_HEADER(ZC_MAKINGARROW_LIST, 0x01ad);
 
+struct PACKET_ZC_SKILL_SELECT_REQUEST {
+	int16 packetType;
+	int16 packetLength;
+	int32 flag; //< 0 = old code compatibility; 1 = Auto Shadow Spell; same value is received in CZ_SKILL_SELECT_RESPONSE
+	int16 skillIds[];
+} __attribute__((packed));
+DEFINE_PACKET_HEADER(ZC_SKILL_SELECT_REQUEST, 0x0442);
+
 #if PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200723 || PACKETVER_ZERO_NUM >= 20221024
 #define REPAIRITEM_INFO REPAIRITEM_INFO2
 struct PACKET_ZC_REPAIRITEMLIST {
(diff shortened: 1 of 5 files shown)


src/map/packets_struct.h
```diff
@@ -3319,6 +3319,20 @@ struct PACKET_ZC_MAKINGARROW_LIST {
 } __attribute__((packed));
 DEFINE_PACKET_HEADER(ZC_MAKINGARROW_LIST, 0x01ad);
 
+struct PACKET_ZC_SKILL_SELECT_REQUEST {
+	int16 packetType;
+	int16 packetLength;
+	int32 skillCount;
```
4144, line 3325: this is not skill count this is special flag.
Flag same with flag from PACKET_CZ_SKILL_SELECT_RESPONSE if zero client may send additional packets
```
0 - LOWERVER_COMPATIBILITY
1 - SC_AUTOSHADOWSPELL
```

4144 (changes requested):

src/map/packets_struct.h
guilherme-gm: hmm nice catch, I named it skillCount because we were writing `c` to it:
```C
WFIFOL(fd,4) = c;
```

and `c` here is the number of skills sent.

but you are right, checking a packet capture from official its value is 1. I will also fix the value sent

MishimaHaruna (approved):

PR #3300, opened 2024-06-08
Add script commands to generate some client <TAGS>

guilherme-gm:
### Pull Request Prelude

### Changes Proposed
This PR adds some script commands which generates strings for client tags:
- `mesnavigation`: General purpose command for `<NAVI>`;
- `mesmobspawn`: Special command for `<NAVI>` that opens a window "searching" for a monster -- This is a special behavior in navigation
- `mesurl`: For `<URL>` tag
- `mestipbox`: For `<TIPBOX>` tag

Using those commands takes into account client support and tries to fall back to some sensible results when something is not supported.

I decided to make them as script commands instead of Global Functions because:
1. We keep client-specific logic into `clif` -- I feel like this is too client specific to be handled by scripts
2. We can eventually use these from source directly if there is a need


doc/script_commands.txt
```diff
@@ -2612,6 +2612,79 @@ deleted.
 	// This would delete all elements of the array
 	deletearray(.@array);
 
+---------------------------------------
+//=====================================
+1.1 - Message formatting commands (tags)
```
guilherme-gm, line 2617: a bit off topic from this commit, but:
There is another tag, `<QUEST>` which exists in clients but I don't remember seeing it being used. I was unsure whether it was worth to create a `mesquest("label", questId)`

doc/script_commands.txt
```diff
@@ -2612,6 +2612,79 @@ deleted.
...
+	Note: does nothing when used with modes NAV_MODE_MAP or NAV_MODE_MOB
+
+- show_window: (bool, optional, defaults to false)
+	Whether to open the navigation UI (true) or not (false) when clicking the label.
+
+- mode: (int, optional, defaults to NAV_MODE_ALL (0))
+	The kind of navigation being performed. This mainly affects the messages shown by the client when you reach the map.
+	- NAV_MODE_ALL (0): "You have reached your destination"
+	- NAV_MODE_MAP (1): "You have reached the target map"
+	- NAV_MODE_NPC (2): "You have reached the map the NPC is at" / "You have reached the target NPC"
```
guilherme-gm, line 2656: I thought about making a `mesmapnavi("mapname"{, "label"})` and `mesnpcnavi("NPC Name"{, "label"})` which could check whether map or npc exists. But I don't remember seeing those being used in official scripts... so I am not sure if it is worth it.

doc/script_commands.txt
MishimaHaruna: I think the other tags might be interesting to have since they can open possibilities for custom scripts, but definitely at a lower priority if they're not used anywhere in official scripts (in a future, separate, PR)

MishimaHaruna (approved):

PR #3304, opened 2024-07-01
Fix server sync reading and wrong disconnection from API Server

guilherme-gm:
### Pull Request Prelude

### Changes Proposed

This fixes how packet 0x272D (CharServer -> LoginServer online accounts sync) is read by Login Server. Thus, preventing accounts from being wrongly marked as offline.

When login-server marks an account as offline, it will eventually tell API Server to disconnect them, thus also invalidating users' API Tokens while they are in game.

Huge thanks to AcidMarco in Herc's discord that provided a quite good explanation of what he has observed and this helped me a lot in finding the root cause (or at least I think it did...)

**Issues addressed:** 
None, I think.

### DIFF
--- a/src/login/login.c
+++ b/src/login/login.c
@@ -679,7 +679,7 @@ static void login_fromchar_parse_online_accounts(int fd, int id)
 	login->online_db->foreach(login->online_db, login->online_db_setoffline, id); //Set all chars from this char-server offline first
 	users = RFIFOW(fd,4);
 	for (i = 0; i < users; i++) {
-		int aid = RFIFOL(fd,6+i*4);
+		int aid = RFIFOL(fd, 8 + i * 4);
 		struct online_login_data *p = idb_ensure(login->online_db, aid, login->create_online_user);
 		p->char_server = id;
 		if (p->waiting_disconnect != INVALID_TIMER)
--- a/src/login/login.c
+++ b/src/login/login.c
@@ -32,6 +32,7 @@
 #include "common/HPM.h"
 #include "common/apipackets.h"
 #include "common/cbasetypes.h"
+#include "common/charloginpackets.h"
 #include "common/conf.h"
 #include "common/core.h"
 #include "common/db.h"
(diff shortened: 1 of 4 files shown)


src/login/login.c
```diff
@@ -679,7 +679,7 @@ static void login_fromchar_parse_online_accounts(int fd, int id)
 	login->online_db->foreach(login->online_db, login->online_db_setoffline, id); //Set all chars from this char-server offline first
 	users = RFIFOW(fd,4);
```
dastgirp, line 680: Probably change this to RFIFOL as well?
char.c has `WFIFOL(chr->login_fd,4) = i;`

dastgirp:
Probably unrelated to this PR, but there's one more issue wherein API server is called for autotraders on restart, and we get bunch of Account not online errors

src/login/login.c
guilherme-gm: nice! I ended up converting it to struct format so we don't have to worry about those details

guilherme-gm:
> Probably unrelated to this PR, but there's one more issue wherein API server is called for autotraders on restart, and we get bunch of Account not online errors

Yeah, I think this is a separate issue. I will take a look into it and open a new PR if I find something (maybe we should create an issue about it?)

MishimaHaruna (approved):

vBrenth:
Hello.. as im using it, I started getting this disconnection and after quite some times it reconnected.
image

guilherme-gm:
@vBrenth can you try using the first commit only? this one: https://github.com/HerculesWS/Hercules/pull/3304/commits/29d858c70589fdb8a3279d930103a43b8cc7d6c7

I am wondering if I may have made a mistake when converting to struct.

vBrenth:
@guilherme-gm  I've been running the first commit only and its still having some saving problem but less and no disconnection between char and login so far.

guilherme-gm:
> @guilherme-gm I've been running the first commit only and its still having some saving problem but less and no disconnection between char and login so far.

Thanks for the update, I made a fix on #3312 so the 2nd commit properly works.

Are the saving problems giving any logs? is there any reproduction steps for it? I think we need to figure out whether this is a server issue or a client one, since I noticed that there are some cases where the client doesn't send the changes to be saved

vBrenth:
I cant find a way to reproduce it. Its just happen after some time with players are online.

PR #3305, opened 2024-07-27
`getpartymember` upgrade

jasonch35:
### Pull Request Prelude

- [x ] I have followed [proper Hercules code styling][code].
- [x ] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x ] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
Hi, I'd like to share my version of the `getpartymember` script command, similar to `getunits`.

This updated version no longer uses global temporary variables, which are prone to being overwritten, especially when multiple players call the function simultaneously. Instead, it requires the scripter to declare an array as an argument in the 3rd parameter of the command.

Additionally, this change introduces the following constants: `PT_MEMBER_NAME`, `PT_MEMBER_CHARID`, and `PT_MEMBER_ACCID`. The function now also returns the size of the party.

Sample usage:
```
.@count = getpartymember(getcharid(CHAR_ID_PARTY), PT_MEMBER_NAME, .@name$);
for (.@i = 0; .@i < .@count; .@i++) {
	consolemes(CONSOLEMES_DEBUG, ""+.@name$[.@i]+"");
}
```

The existing script will be deprecated, but I will gladly update all current NPC scripts to use this new version if approved. I can apply similar updates to `getguildmember` and `getmobdrops` as well.

I think this approach aligns more closely with Herc's style.

**Issues addressed:** 
Wrong data when multiple players call the command.

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -9938,6 +9938,13 @@ static BUILDIN(getpartymember)
 		st->state = END;
 		return false;
 	}
+
+	if (type < PT_MEMBER_NAME || type > PT_MEMBER_ACCID) {
+		ShowError("buildin_getpartymember: Invalid type argument\n");
+		script->reportdata(data);
+		st->state = END;
+		return false;
+	}
 	
 	if (!is_int_variable(varname) && (type == PT_MEMBER_CHARID || type == PT_MEMBER_ACCID)) {
 		ShowError("buildin_getpartymember: Target argument is not an int variable\n");
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -9920,41 +9920,68 @@ static BUILDIN(getpartyname)
 
 /*==========================================
  * Get the information of the members of a party by type
- * @party_id, @type
- * return by @type :
- * - : nom des membres
- * 1 : char_id des membres
- * 2 : account_id des membres
+ * getpartymember(<party_id>, <type>, <array>);
  *------------------------------------------*/
 static BUILDIN(getpartymember)
 {
-	struct party_data *p;
-	int j=0,type=0;
+	struct map_session_data *sd = NULL;
+	struct party_data *p = party->search(script_getnum(st, 2));
+	enum partymember_type type = script_getnum(st, 3);
+	struct script_data *data = script_getdata(st, 4);
+	const char *varname = reference_getname(data);
+	int id = reference_getid(data);
+	int num = 0;
 
-	p=party->search(script_getnum(st,2));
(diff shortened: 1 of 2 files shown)


src/map/script.c
```diff
@@ -9920,41 +9920,68 @@ static BUILDIN(getpartyname)
...
+	if (not_server_variable(*varname)) {
+		sd = script->rid2sd(st);
+
+		if (sd == NULL) {
+			script_pushint(st, 0);
+			return true; // player variable but no player attached
+		}
+	}
+
+	if (p) {
```
MishimaHaruna, line 9965:
```suggestion
	if (p != NULL) {
```

src/map/script.c
```diff
@@ -9920,41 +9920,68 @@ static BUILDIN(getpartyname)
...
+		for (int i = 0; i < MAX_PARTY; i++) {
+			if (p->party.member[i].account_id) {
```
MishimaHaruna, line 9967:
```suggestion
			if (p->party.member[i].account_id != 0) {
```

src/map/script.c
```diff
@@ -9920,41 +9920,68 @@ static BUILDIN(getpartyname)
...
- * 2 : account_id des membres
+ * getpartymember(<party_id>, <type>, <array>);
  *------------------------------------------*/
 static BUILDIN(getpartymember)
 {
-	struct party_data *p;
-	int j=0,type=0;
+	struct map_session_data *sd = NULL;
+	struct party_data *p = party->search(script_getnum(st, 2));
+	enum partymember_type type = script_getnum(st, 3);
```
MishimaHaruna, line 9929: This is not checked against unknown types (i.e. the user passing a value different from the supported three).

I think it would be wise to abort without even looping through the party if an unsupported option was passed.

MishimaHaruna (changes requested):
Hello, and thank you for sharing! Please do update the scripts when you have time to do so (in a separate commit).

I think it looks good. Even if still in draft, I went alread and added some review comments. Please have a look at them.

src/map/script.c
jasonch35: please review:
a79cc0d8373fe44203f14b501c08c0d83422c621

MishimaHaruna (commented):
Looks good! Will you update the scripts (luckily there aren't a lot of calls to this command, I only see 16) and the script_commands.txt documentation?

MishimaHaruna (approved):

PR #3319, opened 2024-09-10
`getmobdrops` refactor

jasonch35:
### Pull Request Prelude

### Changes Proposed
Avoid use of global variables that tend to get overwritten when command is called by 2 or more players simultaneously.
The command now requires an array as the second or third argument.
The third argument is optional and, if provided, will copy the drop rates into the array.
Updated documentation to reflect these changes.

Usage:
```
.@count = getmobdrops(.@mob_id, .@item, .@rate);
if (.@count) {
	mes(getmonsterinfo(.@mob_id, MOB_NAME) + " - " + .@count + " drops found:");
		mes(.@item[.@i] + " (" + getitemname(.@item[.@i]) + ") " + .@rate[.@i]/100 + ((.@rate[.@i]%100 < 10) ? ".0":".") + .@rate[.@i]%100 + "%");
	}
} else {
	mes("No drops found.");
}
close();
```

**Issues addressed:** 
Wrong data when multiple players call the command.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -4333,54 +4333,25 @@ Example:
 
 ---------------------------------------
 
-*getmobdrops(<mob id>)
+*getmobdrops(<mob_id>, <item_id_array>{, <drop_rate_array>});
 
-This command will find all drops of the specified mob and return the item
-IDs and drop percentages into arrays of temporary global variables.
-getmobdrops() returns true if successful and false if the mob ID doesn't
-exist.
-
-Upon executing this,
-
-$@MobDrop_item[] is a global temporary number array which contains the
-                 item IDs of the monster's drops.
-
-$@MobDrop_rate[] is a global temporary number array which contains the
-                 drop percentages of each item. (1 = .01%)
-
-$@MobDrop_count  is the number of item drops found.
-
-Be sure to use $@MobDrop_count to go through the arrays, and not
-'getarraysize', because the temporary global arrays are not cleared
-between runs of 'getmobdrops'. If a mob with 7 item drops is looked up,
-the arrays would have 7 elements. But if another mob is looked up and it
-only has 5 item drops, the server will not clear the arrays for you,
-overwriting the values instead. So in addition to returning the 5 item
-drops, the 6th and 7th elements from the last call remain, and you will
-get 5+2 item drops, of which the last 2 don't belong to the new mob.
-$@MobDrop_count will always contain the correct number (5), unlike
-getarraysize() which would return 7 in this case.
+This command will find drops information of the specified <mob_id>, copy the
+item ids on to <item_id_array> and drop rate on to <drop_rate_array> if provided.
+Returns the number of items found. Will return 0 if <mob_id> is invalid.
 
 Example:
-
 	// get a Mob ID from the user
 	input(.@mob_id);
 
-	if (getmobdrops(.@mob_id)) { // getmobdrops() returns true on success
-		// immediately copy global temporary variables into scope
-		// variables, since we don't know when getmobdrops() will get
-		// called again for another mob, overwriting your global temporary
-		// variables.
-		.@count = $@MobDrop_count;
-		copyarray(.@item[0], $@MobDrop_item[0], .@count);
-		copyarray(.@rate[0], $@MobDrop_rate[0], .@count);
-
+	.@count = getmobdrops(.@mob_id, .@item, .@rate);
+	
+	if (.@count == 0) {
+		mes("No drops found.");
+	} else {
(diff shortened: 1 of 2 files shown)


skyleo:
May want to consider to deprecate the command instead, to not screw over existing scripts too much. And instead use a new name `getmonsterdrops`, so that when the deprecated command is used, a warning is shown on map-server saying to use `getmonsterdrops` instead.

Alternatively one can go the hard-route but you really have to ensure and be aware of what is shown to people using older scripts then, it has to not silently fail 100%. And it definitely would be good if it fails at parsing already and not at runtime, otherwise people may start the server and think all is fine after the update containing this change.

Ideally this PR should already replace all existing scripts with the new command.

jasonch35:
@skyleo Good point, I hadn’t considered that. However, both [getpartymember](https://github.com/HerculesWS/Hercules/pull/3315) and [getguildmember](https://github.com/HerculesWS/Hercules/pull/3318) underwent similar upgrades and were merged without deprecation. 😄 Should I still create a new command and deprecate getmobdrops?

skyleo:
> @skyleo Good point, I hadn’t considered that. However, both [getpartymember](https://github.com/HerculesWS/Hercules/pull/3315) and [getguildmember](https://github.com/HerculesWS/Hercules/pull/3318) underwent similar upgrades and were merged without deprecation. 😄 Should I still create a new command and deprecate getmobdrops?

Although I don't fully remember the behavior, there scripting engine may fail older scripts on parsing already due to the mandatory parameter amount being different between the older and your version of the script command.

So I guess you may be fine. May be worth testing though. ;)

skyleo (approved):

MishimaHaruna:
> Although I don't fully remember the behavior, the script engine may fail older scripts on parsing already due to the mandatory parameter amount being different between the older and your version of the script command.

That's right, since the old command had only one argument and the new one has two mandatory ones and an optional one, it'll already error out when trying to load an old script, forcing the user to notice and update the script. There's no need to deprecate and renamed the command in this case

doc/script_commands.txt
```diff
@@ -4333,54 +4333,25 @@ Example:
...
-		// variables, since we don't know when getmobdrops() will get
-		// called again for another mob, overwriting your global temporary
-		// variables.
-		.@count = $@MobDrop_count;
-		copyarray(.@item[0], $@MobDrop_item[0], .@count);
-		copyarray(.@rate[0], $@MobDrop_rate[0], .@count);
-
+	.@count = getmobdrops(.@mob_id, .@item, .@rate);
+	
+	if (.@count) {
```
MishimaHaruna, line 4348: Small nitpick in order to encourage the readers of this documentation to use better code style

```suggestion
	if (.@count == 0) {
```

src/map/script.c
```diff
@@ -12284,36 +12284,69 @@ static BUILDIN(monster)
...
 {
-	int class_ = script_getnum(st,2);
-	int i, j = 0;
-	struct mob_db *monster;
+	struct map_session_data *sd = NULL;
+	int mob_id = script_getnum(st, 2);
+	struct mob_db *monster = NULL;
+	struct script_data *data1 = script_getdata(st, 3);
+	struct script_data *data2 = NULL;
+	const char *varname1 = reference_getname(data1);
```
MishimaHaruna, line 12292: This is a dangerous operation on data that might not be a reference (might cause a crash depending on the value passed). Please only call `reference_getname()` after ensuring that `data_isreference()`.

src/map/script.c
```diff
@@ -12284,36 +12284,69 @@ static BUILDIN(monster)
...
+	const char *varname2;
+	int varid1 = reference_getid(data1);
+	int varid2;
```
MishimaHaruna, line 12295: Might be safer to initialize these even if they are not used in some cases

```suggestion
	const char *varname2 = NULL;
	int varid1 = reference_getid(data1);
	int varid2 = 0;
```

src/map/script.c
```diff
@@ -12284,36 +12284,69 @@ static BUILDIN(monster)
...
+	if (!data_isreference(data1) || reference_toconstant(data1) || !is_int_variable(varname1)) {
+		ShowError("buildin_getmobdrops: Target argument must be an integer variable\n");
+		script->reportdata(data1);
+		st->state = END;
+		return false;
+	}
+
+	if (script_hasdata(st, 4)) {
+		data2 = script_getdata(st, 4);
+		varname2 = reference_getname(data2);
```
MishimaHaruna, line 12307: Same remark as above about calling `reference_getname()` before checking whether `data_isreference()`

MishimaHaruna (changes requested):

MishimaHaruna (approved):
Thank you, looks good to me now, I'll merge it

PR #3327, opened 2024-10-21
mob_db2 inheritance doc

ahoyaharr:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 
- Adds `Inherit` field to documentation in `mob_db2` 
- Fixes a comment about Magnum Break's auto attack modifier

### DIFF
--- a/db/mob_db2.conf
+++ b/db/mob_db2.conf
@@ -54,6 +54,10 @@ mob_db: (
 		Int: intelligence             (int, defaults to 0)
 		Dex: dexterity                (int, defaults to 0)
 		Luk: luck                     (int, defaults to 0)
+                                              (if Inherit: true,
+                                               values default to
+                                               the inherited value
+                                               from mob_db)
 	}
 	ViewRange: view range                 (int, defaults to 1)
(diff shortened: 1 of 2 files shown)


db/mob_db2.conf
```diff
@@ -94,6 +94,15 @@ mob_db: (
 		// ...
 	}
 	DamageTakenRate: damage taken rate    (int, defaults to 100)
+	Inherit: true/false                   (bool, defaults to false.
+                                               When true, inherits the
+                                               values from mob_db using
+                                               Id as the primary key.
+                                               Individual items in a nested 
+                                               field (e.g., Drops) can be
```
MishimaHaruna, line 102: Is this bit about overriding drops documented elsewhere with this wording? Because I don't think it's exactly correct, I believe it's currently a bit complicated to properly override drops without breaking anything

If I'm not missing anything in the existing code, it's currently possible to override each drop slot sequentially, but there's no way to cleanly override a slot while leaving the previous ones untouched

ahoyaharr: Thanks, fixed now.

MishimaHaruna (approved):

PR #3328, opened 2024-10-29
Improved atCommand(mobinfo)

Venseer:
### Pull Request Prelude

- [ x ] I have followed [proper Hercules code styling][code].
- [ x ] I have read and understood the [contribution guidelines][cont] before making this PR.
- [ x ] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

Change mobinfo atcommand. Drop list now users item links.

20241029192904683 - CommonRigidAmurratsnake

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -7156,14 +7156,14 @@ ACMD(mobinfo)
 	unsigned char msize[3][7] = {"Small", "Medium", "Large"};
 	unsigned char mrace[12][11] = {"Formless", "Undead", "Beast", "Plant", "Insect", "Fish", "Demon", "Demi-Human", "Angel", "Dragon", "Boss", "Non-Boss"};
 	unsigned char melement[10][8] = {"Neutral", "Water", "Earth", "Fire", "Wind", "Poison", "Holy", "Dark", "Ghost", "Undead"};
-	char atcmd_output2[CHAT_SIZE_MAX];
+	StringBuf buf;
 	struct item_data *item_data;
 	struct mob_db *monster, *mob_array[MAX_SEARCH];
 	int count;
 	int i, k;
 
 	memset(atcmd_output, '\0', sizeof(atcmd_output));
-	memset(atcmd_output2, '\0', sizeof(atcmd_output2));
+	StrBuf->Init(&buf);
 
 	if (!*message) {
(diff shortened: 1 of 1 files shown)


src/map/atcommand.c
```diff
@@ -7253,12 +7253,16 @@ ACMD(mobinfo)
...
-
-			strcat(atcmd_output, atcmd_output2);
+			struct item link_item = { 0 };
+			link_item.nameid = monster->dropitem[i].nameid;
+			StringBuf buf;
+			StrBuf->Init(&buf);
+			StrBuf->AppendStr(&buf, " - ");
+			clif->format_itemlink(&buf, &link_item);
+			snprintf(atcmd_output2, sizeof(atcmd_output2), " %02.02f%%", (float)droprate / 100);
+			StrBuf->AppendStr(&buf, atcmd_output2);
```
MishimaHaruna, line 7263: This copy can be skipped by formatting the string directly into the stringbuf

```suggestion
			StrBuf->Printf(buf, " %02.02f%%", (float)droprate / 100);
```

src/map/atcommand.c
```diff
@@ -7253,12 +7253,16 @@ ACMD(mobinfo)
...
+			strcat(atcmd_output, StrBuf->Value(&buf));
+			StrBuf->Destroy(&buf);
```
MishimaHaruna, line 7265: This can be moved outside the for loop (both `Init()` and `Destroy()`), and here it could just `Clear()` it, to avoid unnecessary memory reallocations. See `ACMD(itemlist)` or `ACMD(who)` for examples.

If you want to further improve it, you can even remove the strcat() and use directly the StringBuf's `Value()` in `clif->message()` (again, examples of this in those same two functions I mentioned above)

MishimaHaruna (changes requested):
Thank you! I have a few suggestions for improvements, please have a look

src/map/atcommand.c
Venseer: This is changed now, I think I made it better memory wise.
Removed atcmd_output2 entirely.
Initialized StrBuffer at item start and destroyed at item end.

Reworked the clear and write logic for the String buffer.

Also had forgot MVP drops so that is new too.

src/map/atcommand.c
Venseer: This is now changed.

Venseer:
Made some adjustments regarding strbuffer initialization.
New: Added MVP drops item link too.

20241031170441664 - EasyInfantileGonolek

MishimaHaruna (approved):

PR #3329, opened 2024-10-29
Improved atCommand(iteminfo)

Venseer:
### Pull Request Prelude

- [ x ] I have followed [proper Hercules code styling][code].
- [ x ] I have read and understood the [contribution guidelines][cont] before making this PR.
- [ x ] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

Uses itemlink on iteminfo atcommand.
No longer list slots on iteminfo, itemlink contains slot information.

20241029192847108 - ElectricPeruSwan

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -7717,14 +7717,22 @@ ACMD(iteminfo)
 		clif->message(fd, atcmd_output);
 		count = MAX_SEARCH;
 	}
+	StringBuf buf;
+	StrBuf->Init(&buf);
 	for (i = 0; i < count; i++) {
 		struct item_data *item_data = item_array[i];
 		if (item_data != NULL) {
-			snprintf(atcmd_output, sizeof(atcmd_output), msg_fd(fd, MSGTBL_ITEMINFO_DETAILS), // Item: '%s'/'%s'[%d] (%d) Type: %s | Extra Effect: %s
-				item_data->name, item_data->jname, item_data->slot, item_data->nameid,
+
+			struct item link_item = { 0 };
+			link_item.nameid = item_data->nameid;
+			clif->format_itemlink(&buf, &link_item);
+
+			snprintf(atcmd_output, sizeof(atcmd_output), msg_fd(fd, MSGTBL_ITEMINFO_DETAILS), // Item: '%s'/'%s' (%d) Type: %s | Extra Effect: %s
+				item_data->name, StrBuf->Value(&buf), item_data->nameid,
 				itemdb->typename(item_data->type),
 				(item_data->script == NULL) ? msg_fd(fd, MSGTBL_ITEMINFO_NONE) : msg_fd(fd, MSGTBL_ITEMINFO_WITH_SCRIPT) // None / With script
 			);
+			StrBuf->Clear(&buf);
 			clif->message(fd, atcmd_output);
 
 			snprintf(atcmd_output, sizeof(atcmd_output), msg_fd(fd, MSGTBL_ITEMINFO_NPC_DETAILS), item_data->value_buy, item_data->value_sell, item_data->weight / 10.); // NPC Buy:%dz, Sell:%dz | Weight: %.1f
@@ -7741,6 +7749,7 @@ ACMD(iteminfo)
 			clif->message(fd, atcmd_output);
 		}
 	}
+	StrBuf->Destroy(&buf);
 	return true;
 }
 
--- a/conf/messages.conf
+++ b/conf/messages.conf
@@ -1323,7 +1323,7 @@
 
 // @iteminfo
 1276: Please enter an item name/ID (usage: @ii/@iteminfo <item name/ID>).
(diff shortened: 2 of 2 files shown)


dastgirp:
This needs to be PACKETVER specific, as it won't work for old Clients

Venseer:
Do you happen to have the packetver? I can work that out.

Venseer:
> This needs to be PACKETVER specific, as it won't work for old Clients

Just checked clif_format_itemlink. It verifies packetver and writes only whats compatible.
Check clif.c line 26044

src/map/atcommand.c
```diff
@@ -7720,11 +7720,19 @@ ACMD(iteminfo)
...
+			StrBuf->Init(&buf);
+			clif->format_itemlink(&buf, &link_item);
+
 			snprintf(atcmd_output, sizeof(atcmd_output), msg_fd(fd, MSGTBL_ITEMINFO_DETAILS), // Item: '%s'/'%s'[%d] (%d) Type: %s | Extra Effect: %s
-				item_data->name, item_data->jname, item_data->slot, item_data->nameid,
+				item_data->name, StrBuf->Value(&buf), item_data->nameid,
 				itemdb->typename(item_data->type),
 				(item_data->script == NULL) ? msg_fd(fd, MSGTBL_ITEMINFO_NONE) : msg_fd(fd, MSGTBL_ITEMINFO_WITH_SCRIPT) // None / With script
 			);
+			StrBuf->Destroy(&buf);
```
MishimaHaruna, line 7735: The StringBuf could be `Init()`ialized and `Destroy()`ed outside the for loop, and only `Clear()`ed here, in order to avoid unnecessary memory reallocations

src/map/atcommand.c
```diff
@@ -7720,11 +7720,19 @@ ACMD(iteminfo)
...
+
+			struct item link_item = { 0 };
+			link_item.nameid = item_data->nameid;
+			StringBuf buf;
+			StrBuf->Init(&buf);
+			clif->format_itemlink(&buf, &link_item);
+
 			snprintf(atcmd_output, sizeof(atcmd_output), msg_fd(fd, MSGTBL_ITEMINFO_DETAILS), // Item: '%s'/'%s'[%d] (%d) Type: %s | Extra Effect: %s
-				item_data->name, item_data->jname, item_data->slot, item_data->nameid,
+				item_data->name, StrBuf->Value(&buf), item_data->nameid,
```
MishimaHaruna, line 7731: This is reducing the amount of arguments passed to snprintf, but the actual format string hasn't been updated, resulting in errors, crashes or string glitches at runtime.

You can find the string in `conf/messages.conf`, entry id 1277. Please also update the format string in the comment here and the one in the comment in `src/common/msgtable.h` (search `MSGTBL_ITEMINFO_DETAILS`), so that they are all identical

MishimaHaruna (changes requested):
Thank you! Please check the following things before I can merge this

src/map/atcommand.c
Venseer: This PR came from a diff from my personal changes and I completely missed this, my bad.
This is now fixed.

src/map/atcommand.c
Venseer: Thanks, 
Performed those changes and tested it.

MishimaHaruna (approved):

kyeme:
Not working in 2016-01-06aRagexe client?

itemlink

guilherme-gm:
@kyeme may you try making an npc with this?
```
mesf("Your item is: %s", getitemlink(Apple));
```

does it render properly or also shows the tag like that?

I think it might be the case that not every client supports those tags in the same places 🫠  I remember testing and checking client for `mes`, but not for other places.

dastgirp:
> @kyeme may you try making an npc with this?
> 
> ```
> mesf("Your item is: %s", getitemlink(Apple));
> ```
> 
> does it render properly or also shows the tag like that?
> 
> I think it might be the case that not every client supports those tags in the same places 🫠 I remember testing and checking client for `mes`, but not for other places.

2020 client doesn't support getitemlink in "mes".
So 2016 might not support it as well

guilherme-gm:
> > @kyeme may you try making an npc with this?
> > ```
> > mesf("Your item is: %s", getitemlink(Apple));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > does it render properly or also shows the tag like that?
> > I think it might be the case that not every client supports those tags in the same places 🫠 I remember testing and checking client for `mes`, but not for other places.
> 
> 2020 client doesn't support getitemlink in "mes". So 2016 might not support it as well

hmm well noted, my 2019-12-24 client "partially" supports it, it doesn't make the text clickable, but it does render the item name. 
Maybe `dispbottom` would be a better test to check whether the tag is broken for the client or it just doesn't support it in `message`?

PR #3349, opened 2025-04-24
Bug fix: wrong behavior in hateffect when disabling effect

TioAkima:
### Pull Request Prelude

### Changes Proposed

Fixes a logical flaw in the `hateffect` script function, where calling it with `enabled = 0` (to remove an effect) would still add the `effectId` to the `hatEffectId` vector if it wasn't already present.

The correct behavior when `enabled = 0` is to remove the effect if present — and **do nothing** if it's not found.

This fix adds a `return true;` right after the `for` loop to prevent unintended addition of the `effectId` when the intent is to remove it.

**Issues addressed:** No related issues.

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -27870,6 +27870,9 @@ static BUILDIN(hateffect)
 			}
 		}
 	}
+	
+	if (enabled == 0)
+		return true;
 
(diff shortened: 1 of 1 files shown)


guilherme-gm (approved):
Thank you for the contribution!

Issue #701, opened 2015-09-06
Gentle Touch-Convert (Sura Skill) reduces HP by too much


Idunnr:
The Gentle Touch-Convert (Sura Skill) should reduce your max HP by 5% at level 5. I have tested it and found that it reduces max HP by 20% at level 5 instead. The reduction rate is off for all levels, starting at 4% at level 1 and incrementing by 4% for each level.

http://irowiki.org/wiki/Gentle_Touch-Convert - this corresponds with the current in-game skill description we have.

Cheers <3

guilherme-gm:
For future reference: I couldn't find why it was like that before, most likely this used to work like that back in 2010.

What I could see:

1. Since 2012-02, kRO description mentions 1% - 5% cost (I don't have data from before this)
2. iRO wiki started showing the 4% - 20% reduction in 2017, and removed in 2022, but no idea why
3. kRO Patch from 2010-03-31 mentions reduction in MaxHP cost -- https://ragnarok.fandom.com/wiki/RO_Patch_(2010_Mar._31)
> The effect of increasing damage and increasing speed options, MHP reduction options has been changed.


My guess is that when it was implemented in Herc, we got mixed information between 2010 and 2012 changes and thus it stayed with cost of 4% per level.

It looks like the skill got reworked in 2019, so no reason for us to stress too much about this history, as at some point we might end up changing it again.


Fixed in #3358

PR #3358, opened 2025-08-24
Fixed SC_GENTLETOUCH_CHANGE reduces way too much HP.

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** #701 

### DIFF
--- a/src/map/status.c
+++ b/src/map/status.c
@@ -5795,8 +5795,8 @@ static unsigned int status_calc_maxhp(struct block_list *bl, struct status_chang
 		maxhp += maxhp * 5 * sc->data[SC_INSPIRATION]->val1 / 100 + 600 * sc->data[SC_INSPIRATION]->val1;
 	if (sc->data[SC_RAISINGDRAGON])
 		maxhp += maxhp * (2 + sc->data[SC_RAISINGDRAGON]->val1) / 100;
-	if (sc->data[SC_GENTLETOUCH_CHANGE]) // Max HP decrease: [Skill Level x 4] %
-		maxhp -= maxhp * (4 * sc->data[SC_GENTLETOUCH_CHANGE]->val1) / 100;
+	if (sc->data[SC_GENTLETOUCH_CHANGE]) // Max HP decrease: [Skill Level x 1] %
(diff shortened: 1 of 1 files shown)


src/map/status.c
```diff
@@ -5795,8 +5795,8 @@ static unsigned int status_calc_maxhp(struct block_list *bl, struct status_chang
 		maxhp += maxhp * 5 * sc->data[SC_INSPIRATION]->val1 / 100 + 600 * sc->data[SC_INSPIRATION]->val1;
 	if (sc->data[SC_RAISINGDRAGON])
 		maxhp += maxhp * (2 + sc->data[SC_RAISINGDRAGON]->val1) / 100;
-	if (sc->data[SC_GENTLETOUCH_CHANGE]) // Max HP decrease: [Skill Level x 4] %
-		maxhp -= maxhp * (4 * sc->data[SC_GENTLETOUCH_CHANGE]->val1) / 100;
+		if (sc->data[SC_GENTLETOUCH_CHANGE]) // Max HP decrease: [Skill Level x 1] %
```
guilherme-gm: small indent mistake here

guilherme-gm (changes requested):

src/map/status.c
MrKeiKun: done

guilherme-gm (approved):
Thanks for the PR!

--

For future reference: I couldn't find why it was like that before, most likely this used to work like that back in 2010.

What I could see:

1. Since 2012-02, kRO description mentions 1% - 5% cost (I don't have data from before this)
2. iRO wiki started showing the 4% - 20% reduction in 2017, and removed in 2022, but no idea why
3. kRO Patch from 2010-03-31 mentions reduction in MaxHP cost -- https://ragnarok.fandom.com/wiki/RO_Patch_(2010_Mar._31)
> The effect of increasing damage and increasing speed options, MHP reduction options has been changed.


My guess is that when it was implemented in Herc, we got mixed information between 2010 and 2012 changes and thus it stayed with cost of 4% per level.

It looks like the skill got reworked in 2019, so no reason for us to stress too much about this history, as at some point we might end up changing it again.

Issue #2811, opened 2020-08-11
ServerCrashes upon reloadpcdb


MrKeiKun:
**Describe the bug**
Jobchanging -> @reloadpcdb and vice versa crashes server

**To Reproduce**
1. Create character
2. Select newly created character
3. Enter command `@job 4062`
4. Enter command `@reloadpcdb`
5. See Error

**Expected behavior**
Able to jobchange and `@reloadpcdb` (vice versa) without server crashing

**System specs (please complete the following information):**
 - OS: Centos 8 x64
 - Hercules Version v2020.07.26
 - Mode: Renewal
 - Packet version: 20180418
 - Client type: RE

**asan out**
```
=================================================================
==3364884==ERROR: AddressSanitizer: heap-use-after-free on address 0x61200002cf50 at pc 0x000000c9cb04 bp 0x7ffc66e53650 sp 0x7ffc66e53640
READ of size 4 at 0x61200002cf50 thread T0
    #0 0xc9cb03 in pc_maxbaselv /home/kei/Hercules/src/map/pc.c:7319
    #1 0xca6335 in pc_jobchange /home/kei/Hercules/src/map/pc.c:9200
    #2 0xca6335 in pc_jobchange /home/kei/Hercules/src/map/pc.c:9107
    #3 0x4f6f9f in atcommand_jobchange /home/kei/Hercules/src/map/atcommand.c:1028
    #4 0x5114ca in atcommand_exec /home/kei/Hercules/src/map/atcommand.c:10697
    #5 0xcb7a31 in pc_process_chat_message /home/kei/Hercules/src/map/pc.c:12564
    #6 0x7d7319 in clif_process_chat_message /home/kei/Hercules/src/map/clif.c:10475
    #7 0x77edf3 in clif_parse_GlobalMessage /home/kei/Hercules/src/map/clif.c:11593
    #8 0x75538c in clif_parse /home/kei/Hercules/src/map/clif.c:24314
    #9 0x1372784 in do_sockets /home/kei/Hercules/src/common/socket.c:1073
    #10 0x41abd1 in main /home/kei/Hercules/src/common/core.c:539
    #11 0x7f7b118f56a2 in __libc_start_main (/lib64/libc.so.6+0x236a2)
    #12 0x41c09d in _start (/home/kei/Hercules/map-server+0x41c09d)

0x61200002cf50 is located 144 bytes inside of 280-byte region [0x61200002cec0,0x61200002cfd8)
freed by thread T0 here:
[...]

PR #3377, opened 2025-08-30
Fixed server crash via @reloadpcdb

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

Fixes a heap-use-after-free in `pc_readdb()` (`@reloadpcdb`): `pc->clear_exp_groups()` freed the class exp group data before `pc->dbs->class_exp_table` entries pointing into it were cleared, so the table could be dereferenced after free. `pc_readdb()` now clears `pc->dbs->class_exp_table` before rebuilding the exp groups.

Repopulating `class_exp_table` requires calling `status->read_job_db()`. That call is made from `@reloadpcdb` specifically, not from `pc_readdb()` itself — `pc_readdb()` also runs during initial server startup, before `status->init()` has populated `status->dbs->unit_params`, so calling it there crashed the map-server on every boot.

**Issues addressed:** #2811 

### DIFF
--- a/src/map/atcommand.c
+++ b/src/map/atcommand.c
@@ -4038,6 +4038,11 @@ ACMD(reloadstatusdb)
 ACMD(reloadpcdb)
 {
 	pc->readdb();
+	// Repopulate class_exp_table, which pc->readdb() just cleared; safe here since the
+	// server has already finished booting and status->read_unit_params_db() has run.
+	// Must NOT be called from pc->readdb() itself, since that also runs during initial
+	// server startup, before status->init() has populated status->dbs->unit_params[].
+	status->read_job_db();
 	clif->message(fd, msg_fd(fd, MSGTBL_RELOAD_PLAYER_RELOADED));
 	return true;
 }
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -11806,8 +11806,12 @@ static int pc_readdb(void)
 	/**
 	 * Read and load into memory, the exp_group_db.conf file.
 	 */
+	pc->clear_class_exp_table();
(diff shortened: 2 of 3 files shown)


hemagx (changes requested):
From your asan report (I would prefer a crash dump), it says it was a use after free, which means it was  free'd else where, why it's not being nulled there and why it's a seperate function? I need more explanation here please.

MrKeiKun:
> From your asan report (I would prefer a crash dump), it says it was a use after free, which means it was free'd else where, why it's not being nulled there and why it's a seperate function? I need more explanation here please.

Why not nulled in pc_clear_exp_groups where the free happens?

pc_clear_exp_groups only owns and manages the VECTOR storage — it frees the exp arrays inside each class_exp_group entry and clears the vectors. It has no knowledge of class_exp_table, and by design it shouldn't: class_exp_table is a separate non-owning pointer array that acts as a class-indexed lookup into those vectors, and it is populated not by any function in pc.c but by status->read_job_db() in status.c. The two are in different abstraction layers — one owns the data, the other indexes it. Coupling pc_clear_exp_groups to nulling a lookup table it doesn't populate would mean it needs to know about a side effect that belongs to a different module.

Why a separate function?

Because the clear of class_exp_table needs to explicitly mirror its repopulation by status->read_job_db(). In pc_readdb the sequence is now: clear the lookup → free the storage → reload storage → repopulate the lookup. Making it a named function pointer in the pc interface keeps that intent readable and also follows the existing HPM hook pattern of the codebase, so plugins can override it if needed. An alternative would have been to inline the null loop directly in pc_readdb, but a named function makes the symmetry with read_job_db more obvious.

Issue #2973, opened 2021-04-07
status_calc_regen_rate_pc (NULLPO)


tlacson7:
**Describe the bug**
this triggers
`[Error]: --- nullpo info --------------------------------------------
[Error]: status.c:3229: 'regen->sitting' in function `status_calc_regen_pc'
[Error]: 0x63cdc6 nullpo_backtrace_print
[Error]:        /home/test/me/src/common/nullpo.c:77
[Error]: 0x63cdc6 assert_report
[Error]:        /home/test/me/src/common/nullpo.c:147
[Error]: 0x621844 status_calc_regen
[Error]:        /home/test/me/src/map/status.c:3420
[Error]: 0x52b839 pc_setpos
[Error]:        /home/test/me/src/map/pc.c:6054
[Error]: 0x51fa8d pc_authok
[Error]:        /home/me/restart/src/map/pc.c:1319
[Error]: 0x7f9f23232e99 HP_pc_authok
[Error]:        HPMHooking/HPMHooking_map.Hooks.inc:61677
[Error]: 0x454204 chrif_authok
[Error]:        /home/test/me/src/map/chrif.c:828
[Error]: 0x459850 chrif_parse
[Error]:        /home/test/me/src/map/chrif.c:1669
[Error]: 0x64f7cd do_sockets
[Error]:        /home/test/me/src/common/socket.c:1474
[Error]: 0x408af1 main
[Error]:        /home/croadmin/restart/src/common/core.c:539
[Error]: 0x7f9f257b3554 ???
[Error]:        ???:0
[Error]: 0x408cb2 ???
[Error]:        ???:0
[Error]: 0xffffffffffffffff ???
[Error]:        ???:0
[Error]: --- end nullpo info ----------------------------------------
`

**To Reproduce**
1. Go to guild map.
2. Relogin.

**System specs (please complete the following information):**
 - OS: [Centos 7x64]
 - Hercules Version [latest]
 - Mode: [pre-renewal]
 - Packet version: [20151029]
 - Client type: [main]

skyleo:
What's a guild map? Can you give an example?

tlacson7:
woe castles like payg_cas02

skyleo:
Can you describe more what class you used? and what skills you had?

skyleo:
Are you sure you have the latest version of Hercules?
Can you give us the code of `status_calc_regen` that you have?

tlacson7:
[...]

Issue #2847, opened 2020-09-24
annoying nullpo error on latest git


AnnieRuru:
**Describe the bug**
I hope I'm not the only one getting this error, every time I login my test server I keet seeing this annoying nullpo error

**To Reproduce**
- purposely reduce the hp of your character (via @heal -10000)
- logout
- login
- gets a nullpo warning

**Expected behavior**
I don't want to see any nullpo

**Screenshots**
```
[Error]: --- nullpo info --------------------------------------------
[Error]: F:\Ragnarok\Hercules\src\map\status.c:3373: 'regen->sitting' in function `status_calc_regen_pc'
[Error]: --- end nullpo info ----------------------------------------
[Error]: --- nullpo info --------------------------------------------
[Error]: F:\Ragnarok\Hercules\src\map\status.c:3598: 'regen->skill' in function `status_calc_regen_rate_pc'
[Error]: --- end nullpo info ----------------------------------------
```

**System specs (please complete the following information):**
 - OS: Windows 10 Pro 64-bit
 - Hercules Version: latest (always pull latest git)
 - Packet version: 20180620
 - Mode: Renewal
 - Client type: [2018-06-20eRagexeRE](https://rathena.org/board/topic/117168-packetver-20180620-client-release-2018-06-20eragexere/)
 - Branch: Main

**Plugins used or source modifications**
NONE - purposely unload every plugin and I still see this error !

**Additional context**

4144:
sadly on windows impossible to track where is issue really happened.

AnnieRuru:
I can reproduce this error 100% of the time,
just need a `gvg_castle` mapflag

EPuncker:
I think it is related to this commit https://github.com/HerculesWS/Hercules/commit/365ab4473c2ccf84d45cafcdb7cdc6a9a0701078#diff-08420bab8271f3965304fee42cf4b865R6056-R6061 by @hemagx 

I'm going to test it soon to confirm too

AnnieRuru:
LMAO !!!
yes after removing those lines the error is gone

hemagx:
@EPuncker nope, it should stay where it is we need to know the old and new map
@AnnieRuru well, the nullpo is an issue not in my code tho i need to figure out how or why this is null on player loading cause iirc it should not, also `regen.state.gc` is removed and its fine to trigger calculations from outside status.

PR #3378, opened 2025-08-30
Adjust some nullpo checks

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

- #2973
Adjust some nullpo check on status_calc_regen_pc and status_calc_regen_rate_pc

**Issues addressed:**
#2973
#2847 (probably?)  

### DIFF
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -1375,6 +1375,9 @@ static bool pc_authok(struct map_session_data *sd, int login_id2, time_t expirat
 
 	sd->lang_id = map->default_lang_id;
 
+	sd->regen.skill = &sd->skill_regen;
+	sd->regen.sitting = &sd->sitting_regen;
+
 	//warp player
 	if ((i=pc->setpos(sd,sd->status.last_point.map, sd->status.last_point.x, sd->status.last_point.y, CLR_OUTSIGHT)) != 0) {
 		ShowError ("Last_point_map %s - id %d not found (error code %d)\n", mapindex_id2name(sd->status.last_point.map), sd->status.last_point.map, i);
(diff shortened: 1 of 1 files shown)


src/map/status.c
```diff
@@ -2608,8 +2608,11 @@ static void status_calc_regen_pc(struct map_session_data *sd, struct status_data
 	nullpo_retv(sd);
 	nullpo_retv(st);
 	nullpo_retv(regen);
-	nullpo_retv(regen->sitting);
-	nullpo_retv(regen->skill);
+
+	if (regen->skill == NULL)
```
hemagx: The original design here is that `regen->skill`, `regen->sitting`  should never be null, but now you're defaulting to make it `sd` is there a rationale behind it? the commit messages doesn't give a rationale at all or explain how it fixes this issue, this seem more like a band-aid workaround than an actual fix for the issue mentioned.

src/map/status.c
```diff
@@ -2683,38 +2686,42 @@ static void status_calc_regen_pc(struct map_session_data *sd, struct status_data
 	}
 
 	// Skill HP/SP restore bonuses
-	struct regen_data_sub *skill_regen = regen->skill;
-	skill_regen->hp = 0;
-	skill_regen->sp = 0;
+	if (regen->skill != NULL) {
```
hemagx: It cannot be null still with your check this is redundant.

hemagx (changes requested):

src/map/status.c
MrKeiKun: you are kinda right here. its a bandaid solution. maybe i should solve the real problem.

src/map/status.c
MrKeiKun: same here.

kyeme:
Tested, and there are no more console errors. Thank you!

Issue #3285, opened 2024-03-10
Shield sprite view is not showing 


kyeme:
The shields sprite doesn’t show on the opponents view. However on my point of view I can see the shield sprite.

![screenMain040](https://github.com/HerculesWS/Hercules/assets/4656578/0f9d483b-48df-4116-80a3-5da9420b2944)
![screenMain041](https://github.com/HerculesWS/Hercules/assets/4656578/250e0f6d-1230-49d4-806b-e99a6e26b91d)

 - Mode: pre-renewal
 - Packet version: 2016-01-06
 - Client type: 2016-01-06aRagexe

csnv:
I cannot reproduce the issue with version 20190530. Is this something specific to 20160106?

kyeme:
This bug is happening even in other client date. What character are you using while testing? Can you test on champion job?

csnv:
This is with a High Priest in 20200520
![image](https://github.com/HerculesWS/Hercules/assets/890582/f0ca9898-fb08-4610-bc9c-2ffe889b92c6)

With a Champion it doesn't display the shield in either client, also the knuckle doesn't display properly.
![image](https://github.com/HerculesWS/Hercules/assets/890582/42458fce-723e-49b5-b8d3-658629a0e24e)

Issue #2967, opened 2021-03-17
Shields are not visible on other players.


Jedzkie:
**Describe the bug**
Shields are not visible on the other players.

**To Reproduce**
Use clients from 2014 up to mid 2018.

**Expected behavior**
Shields must be visible on other players.

**Screenshots**
Hercules:
![image](https://user-images.githubusercontent.com/4826387/111461221-e6433200-8757-11eb-8999-322a33f8c473.png)

Official:
![image](https://user-images.githubusercontent.com/4826387/111461310-0115a680-8758-11eb-99b1-4addae7f2e09.png)

**System specs (please complete the following information):**
 - OS: Windows 10 64bit (Version 10.0.19042.867)
 - Hercules Version: v2021.03.08
 - Mode: Renewal
 - Packet version: 2016-03-16aRagexeRE
 - Client type: RE

**Plugins used or source modifications**
N/A

**Additional context**
N/A

sgsilva:
Weapons to, dont show to other players.

MrKeiKun:
is this still an issue? cause right now im using latest clean herc. But 2019-05-30 ragexe, shield shows properly.
![image](https://user-images.githubusercontent.com/4362134/132973088-fbe48f47-2656-49a9-ad5a-344c037b2c59.png)

kyeme:
@MrKeiKun Not all jobs display shields correctly, some jobs have a bug. This can be easily confirmed when using the Champion job.

MrKeiKun:
> [@MrKeiKun](https://github.com/MrKeiKun) Not all jobs display shields correctly, some jobs have a bug. This can be easily confirmed when using the Champion job.

https://github.com/HerculesWS/Hercules/pull/3380

PR #3380, opened 2025-09-01
Fix Weapon and Shield Sprite

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:**
#3285
#2967 
 

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -1141,9 +1141,7 @@ static void clif_set_unit_idle(struct block_list *bl, struct map_session_data *t
 	p.head = vd->hair_style;
 	p.weapon = vd->weapon;
 	p.accessory = vd->head_bottom;
-#if PACKETVER < 7 || PACKETVER_MAIN_NUM >= 20181121 || PACKETVER_RE_NUM >= 20180704 || PACKETVER_ZERO_NUM >= 20181114
 	p.shield = vd->shield;
-#endif
 	p.accessory2 = vd->head_top;
 	p.accessory3 = vd->head_mid;
 	if (bl->type == BL_NPC && vd->class_ == FLAG_CLASS) {
@@ -1300,9 +1298,7 @@ static void clif_spawn_unit(struct block_list *bl, enum send_target target)
 	p.head = vd->hair_style;
 	p.weapon = vd->weapon;
 	p.accessory = vd->head_bottom;
-#if PACKETVER < 7 || PACKETVER_MAIN_NUM >= 20181121 || PACKETVER_RE_NUM >= 20180704 || PACKETVER_ZERO_NUM >= 20181114
 	p.shield = vd->shield;
-#endif
 	p.accessory2 = vd->head_top;
 	p.accessory3 = vd->head_mid;
 	if (bl->type == BL_NPC && vd->class_ == FLAG_CLASS) {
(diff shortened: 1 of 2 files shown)


skyleo:
What client versions did you test this for?

skyleo:
Okay, I believe this may be bad for older clients. So preferably you should test this for at least 3 different jobs. 
And this should be guarded by a PACKETVER ifdef. So you'll have to at least check one client per year for the range you'll be covering. I believe for older clients this is guaranteed to break stuff o-o

MrKeiKun:
> What client versions did you test this for?
> 
> I think this may be affecting the `LOOK_SHOES` case in clif_changelook as well if the same holds.

i have no way of testing right now. i only trace codes.

> Okay, I believe this may be bad for older clients. So preferably you should test this for at least 3 different jobs. And this should be guarded by a PACKETVER ifdef. So you'll have to at least check one client per year for the range you'll be covering. I believe for older clients this is guaranteed to break stuff o-o

@kyeme hopefully can test it.

MrKeiKun:
not sure if my latest commit fixes it @kyeme . hopefully you can try again :)

kyeme:
@MrKeiKun  Still not working

MrKeiKun:
I'll dig more, seems to be a packet thing

MrKeiKun:
@kyeme i adjusted the packet seems like older clients uses smaller size for weapon and shield i cannot pinpoint the dates i just used the modern condition dates as example. hopefully this works 😄

kyeme:
@MrKeiKun  
I applied the fix and it resolved the shield issue — now other players can see the shield without needing to re-equip.

However, the shield for Monk/Champion is still not visible. I tested this also on iRO official and confirmed that no shield is visible on Monk there either.

Could this be related to the client exe version being used?

MrKeiKun:
> @MrKeiKun I applied the fix and it resolved the shield issue — now other players can see the shield without needing to re-equip.
> 
> However, the shield for Monk/Champion is still not visible. I tested this also on iRO official and confirmed that no shield is visible on Monk there either.
> 
> Could this be related to the client exe version being used?

I need help about this @4144 @MishimaHaruna or @skyleo :D

MrKeiKun:
@kyeme try again.

kyeme:
I just want to add the other jobs where the shield sprite is not visible, aside from the Monk class:
• Sura
• Gunslinger
• Ninja
• Taekwon
• Soul Linker
• Star Gladiator
• Rebellion
• Kagerou
• Oboro

kyeme:
> @kyeme try again.

Thanks! Error fixed.

MrKeiKun:
> I just want to add the other jobs where the shield sprite is not visible, aside from the Monk class: • Sura • Gunslinger • Ninja • Taekwon • Soul Linker • Star Gladiator • Rebellion • Kagerou • Oboro

do this jobs really shows shield?

kyeme:
I'm using the 2016-01-06 client. While testing this, I checked whether shields are visible on the following jobs:

• Monk/Champion
• Sura
• Gunslinger
• Ninja
• Taekwon
• Soul Linker
• Star Gladiator
• Rebellion
• Kagerou
• Oboro

I think this is a client-side issue because I also have a Champion on iRO, and I cannot see the equipped shield there either.

I also tested with a 2022 client, and the shield is visible on Champion when NEMO's "Custom Shield" patch is enabled.

However, this pull request fixed the shield display for the other affected jobs.

Issue #166, opened 2013-09-28
pc_autosave interval improvement


shennetsind:
pc_autosave is a timer that is run continuously but unlike interval timers it resets itself on every run (which in my opinion, is a waste).

```
    interval = map->autosave_interval/(map->usercount()+1);
    if(interval < map->minsave_interval)
        interval = map->minsave_interval;
    timer->add(timer->gettick()+interval,pc->autosave,0,0);
```

wouldn't it be more efficient to have it be a interval timer, and then when it is run do: measure interval, check if different from current, and if so just update the timer's interval? that way the timer is reused instead of needlessly removing itself and adding it over

PR #3389, opened 2025-09-01
Improved pc_autosave interval

MrKeiKun:
### Pull Request Prelude

### Changes Proposed
1. Added set_interval function to timer interface:
2. Improved pc_autosave logic:

**Issues addressed:** #166 

### DIFF
--- a/src/common/timer.c
+++ b/src/common/timer.c
@@ -469,6 +469,40 @@ static int64 timer_settick(int tid, int64 tick)
 }
 
 /**
+ * Modifies a timer's interval.
+ *
+ * @param tid     The timer ID.
+ * @param interval New interval value.
+ * @return 0 on success, -1 on failure.
+ */
+static int timer_set_interval(int tid, int interval)
+{
+	if (tid < 1 || tid >= timer_data_num) {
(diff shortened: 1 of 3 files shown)


skyleo:
I don't really understand the point of this, can you try explaining the advantage?


Also doesn't this destroy the heap property that timers are in? (at least I think they are in a heap)

MrKeiKun:
> I don't really understand the point of this, can you try explaining the advantage?

The original code calls timer->add on every execution, unconditionally allocating a new timer slot even when the computed interval is identical (player count hasn't changed). With the persistent interval timer, the common case (no player count change) becomes a no-op. timer->set_interval is only called when the interval actually differs.

> Also doesn't this destroy the heap property that timers are in? (at least I think they are in a heap)

No, it doesn't disturb the heap. The heap (timer_heap) is a min-heap ordered by tick (next fire time), see the DIFFTICK_MINTOPCMP comparator in timer.c. timer_set_interval only modifies timer_data[tid].interval, which is never a heap key. The interval field is read only inside do_timer to compute the next tick after a fire event. Changing it mid-life has no effect on the current heap ordering.

but overall, you're correct to question it. there's no meaningful advantage, and the complexity cost outweighs the gain. We should close this.

Issue #3127, opened 2022-04-01
Combine mob_race2_db to mob_db


MrKeiKun:
```
// Monster Racial Groups Database
//
// Structure of Database:
// Race2ID,MobID1,MobID2,MobID3,...,MobID9

// Goblins
1,1122,1123,1124,1125,1126,1258,1299
// Kobolds
2,1133,1134,1135,1282,1296
// Orcs
3,1023,1152,1153,1189,1213,1273
// Golems
4,1040,1278,1366,1497,2024
// Guardians
5,1285,1286,1287
// Ninja Classes (Pirates_Pride)
6,1315,1364,1401,1560
```

It's time to combine this to mob_db marked as `RaceGroup`

PR #3394, opened 2025-09-06
Merged data from mob_race2_db to mob_db

MrKeiKun:
### Pull Request Prelude

### Changes Proposed
Merged data from mob_race_db.txt on both renewal and pre-renewal inside their respective mob_db.conf

**Issues addressed:** #3127 


MrKeiKun:
I noticed that `ASSULTER` is `RC2_NINJA` and `RC2_TURTLE` does it mean a monster can have two RaceGroup?

Asheraf (changes requested):
Please add removal notice error similar to what we have for mob_avail.txt

src/map/mob.c
```diff
@@ -5411,6 +5417,20 @@ static void mob_mobavail_removal_notice(void)
 	}
 }
 
+static void mob_race2_db_removal_notice(void)
+{
+	const char *folders[] = { "pre-re", "re" };
```
Asheraf: just use DBPATH no need to check for both

Asheraf (changes requested):

hemagx:
@MrKeiKun While official race2 is quite small and easy for hand migration, custom ones might not be, so we will also require a conversion script in place to move forward with this PR.

MrKeiKun:
I will PR later the script 😘

**Update:**
Done with the script.

hemagx (approved):

Issue #62, opened 2013-07-21
Ability to import/override libconfig files


shennetsind:
e.g. conf/atcommand.conf

PR #3409, opened 2026-04-20
Add libconfig import support to remaining conf files

MrKeiKun:
### Pull Request Prelude

### Changes Proposed
Adds the `import:` override mechanism to conf files that were missing it, consistent with the existing pattern used in `map-server.conf`, `char-server.conf`, etc.

- **`conf/atcommand.conf`** — added import lookup in `atcommand_config_read()`
- **`conf/groups.conf`** — refactored `read_config()` in `pc_groups.c` to use a static config pool so all configs stay alive until inheritance resolution and `atcommand->load_groups()` complete
- **`conf/channels.conf`** — split `read_channels_config()` into a `_from_file()` helper with recursive import support
- **`conf/battlegrounds.conf`** — same wrapper/helper split pattern
- **`conf/clans.conf`** — import recurses with `reload=false` to avoid clearing the DB on the secondary file
- **`conf/network.conf`** and **`conf/api/api_network.conf`** — added import lookup in `socket_net_config_read()` (already accepted a filename)
- **`conf/plugins.conf`** — replaced the old `fopen` replace-hack with proper merge support; HPMHooking-first ordering preserved via an `is_primary` flag

Each modified conf file gets an `import:` line pointing to `conf/import/<name>.conf`, and a corresponding skeleton is added to `conf/import-tmpl/`.

**Issues addressed:** https://github.com/HerculesWS/Hercules/issues/62 

### DIFF
--- a/src/common/socket.c
+++ b/src/common/socket.c
@@ -2124,6 +2124,10 @@ static void socket_net_config_read(const char *filename)
 		ShowNotice("Please edit your '%s' allowed list to fit your network configuration.\n", filename);
 	}
 #endif  // BUILDBOT
+
+	const char *import = NULL;
+	if (libconfig->lookup_string(&network_config, "import", &import) == CONFIG_TRUE)
+		sockt->net_config_read(import);
 	libconfig->destroy(&network_config);
 	return;
 }
--- a/src/common/socket.c
+++ b/src/common/socket.c
@@ -2087,10 +2087,9 @@ static int socket_net_config_read_sub(struct config_setting_t *t, struct s_subne
  *
  * @param filename The filename to read from.
  */
-static void socket_net_config_read(const char *filename)
+static void socket_net_config_read_file(const char *filename)
 {
 	struct config_t network_config;
-	int i;
 	nullpo_retv(filename);
 
 	if (!libconfig->load_file(&network_config, filename)) {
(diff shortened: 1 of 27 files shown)


src/common/socket.c
```diff
@@ -2124,6 +2124,12 @@ static void socket_net_config_read(const char *filename)
 		ShowNotice("Please edit your '%s' allowed list to fit your network configuration.\n", filename);
 	}
 #endif  // BUILDBOT
+
+	{
+		const char *import = NULL;
```
hemagx: there's no need to open a new scope, we're using C11 already.

src/map/battleground.c
```diff
@@ -351,11 +351,17 @@ static enum bg_queue_types bg_str2teamtype(const char *str)
 	return type;
 }
 
+static void bg_config_read_from_file(const char *config_filename);
+
 static void bg_config_read(void)
+{
+	bg_config_read_from_file("conf/battlegrounds.conf");
```
hemagx: I don't like that creating a function just to do this call? instead edit the original function to take a file name and edit the call with default file name.

src/map/battleground.c
```diff
@@ -516,6 +522,12 @@ static void bg_config_read(void)
 			bg->arenas = arena_count;
 		}
 	}
+
+	{
+		const char *import = NULL;
```
hemagx: there's no need to open a new scope, we're using C11 already.

src/map/channel.c
```diff
@@ -577,11 +577,17 @@ static void channel_quit_guild(struct map_session_data *sd)
 	}
 }
 
+static void read_channels_config_from_file(const char *config_filename);
+
 static void read_channels_config(void)
+{
+	read_channels_config_from_file("conf/channels.conf");
```
hemagx: Same, no need for new function.

src/map/channel.c
```diff
@@ -812,6 +818,12 @@ static void read_channels_config(void)
 
 		ShowStatus("Done reading '"CL_WHITE"%u"CL_RESET"' channels in '"CL_WHITE"%s"CL_RESET"'.\n", db_size(channel->db), config_filename);
 	}
+
+	{
```
hemagx: there's no need to open a new scope, we're using C11 already.

src/map/clan.c
```diff
@@ -974,6 +980,12 @@ static bool clan_config_read(bool reload)
 
 	clan->config_read_additional_settings(settings, config_filename);
 	clan->read_db(settings, config_filename, reload);
+
+	{
```
hemagx: there's no need to open a new scope, we're using C11 already.

src/map/pc_groups.c
```diff
@@ -58,28 +58,41 @@ static inline GroupSettings *name2group(const char *group_name)
 	return strdb_get(pcg->name_db, group_name);
 }
 
+/*
+ * Pool of config objects kept alive until all group processing is done.
+ * group_settings->commands/permissions/root point into these configs.
+ */
+#define PC_GROUPS_MAX_CONFIGS 2
```
hemagx: why is there a limit? a vector can be used here.

hemagx (changes requested):

src/map/pc_groups.c
MrKeiKun: true. let me update that

PR #3410, opened 2026-04-21
Fix CI builds, update it to match current support policy and fix several issues discovered by ci builds

hemagx:
### Pull Request Prelude

### Changes Proposed

* Removal of Appveyor builds as we will no longer use MSVS builds for tests
* Removal of Travis CI build files as they're no longer used
* Cleaning up of CI builds to only cover supported compiler versions and mariadb versions
* Removed macOS and windows builds from github actions until we support MSVS/XCode again
* Temporary moved builds to older distor until we fix pcre3 dependency
* Fixed map flag `magic_damage_rate` not being correctly read and its value overwriting several other mapflags when it's set
* Fixed a mix between sc_type/si_type enum in SC_ARMORPROPERTY
* Corrected an assertion in `status_get_sc_icon` that allowed for buffer overflow
* Disabled `-Wenum-enum-conversion` until we fix mixed enums flags in several places in battle/skill
* Disabled `-fno-finite-math-only` because of CJson library to be enabled later once we fix dependency builds
* Implemented a simple block-header in no memory manager mode to preserve `aReallocz` behavior cause the current method to determine the block size causes buffer overflow
* Update all actions used in github actions to their latest version
* Fixed a buffer overflow in `test_chuncked`
* Changed sanatizer library loading to happen exactly before the test target is loaded and off-loaded after to eliminate false positives
**Issues addressed:** 


src/common/memmgr.c
```diff
@@ -113,76 +113,116 @@ struct malloc_interface *iMalloc;
...
 	return ret;
 }
 
 static char *aStrdup_(const char *p, const char *file, int line, const char *func) __attribute__((nonnull (1))) GCCATTR ((returns_nonnull));
 static char *aStrdup_(const char *p, const char *file, int line, const char *func)
 {
-	char *ret = STRDUP(p, file, line, func);
+	// since we now have headers it can't be just a wraper
+	char *ret = iMalloc->malloc(strlen(p) + 1, file, line, func);
+	strcpy(ret, p);
```
MishimaHaruna: You already compute strlen of p, so this could be a memcpy, like in aStrndup_ below (and moved after the null check)

src/common/memmgr.c
```diff
@@ -113,76 +113,116 @@ struct malloc_interface *iMalloc;
...
+
+		if (ret != NULL) {
+			ret->size = size + sizeof(*ret);
+			size_t new_size = ret->size;
+
+			if (new_size > old_size) {
+				memset((uint8 *)ret + old_size, 0, new_size - old_size);
+			}
+
+			++ret;
```
MishimaHaruna: This and the other `++ret` in the else could be removed in favor of returning `ret + 1` (for clarity and consistency with the other functions that do the same)

src/map/mob.c
```diff
@@ -4453,11 +4453,11 @@ static bool mob_read_optdrops_db(void)
 	}
 
 	struct config_setting_t *its = libconfig->lookup(&option_groups, "option_drop_group_db");
-	struct config_setting_t *groups = NULL;
+	struct config_setting_t *groups = libconfig->setting_get_elem(its, 0);
+	int count = libconfig->setting_length(groups);
```
MishimaHaruna: I believe these two lines result in a crash if the lookup fails and `its` is NULL, or if `groups` is NULL (`config_setting_get_elem()` and `config_setting_length()` will happily try to dereference the pointer you pass to them).

Compare with the old version, where the calls were behind a NULL check of `its`.

src/map/mob.c
```diff
@@ -6202,6 +6205,7 @@ void mob_defaults(void)
 
 	memset(mob->db_data, 0, sizeof(mob->db_data));
 	mob->dummy = NULL;
+	mob->opt_drop_groups = NULL;
```
MishimaHaruna, line 6209: Please also set `mob->opt_drop_groups_count = 0;` since those should always match, and there are places where the access to the pointer/array is only guarded by its _count

tools/ci/travis.sh
```diff
@@ -50,7 +50,12 @@ function aborterror {
 function run_server {
 	echo "Running: $1 --run-once $2"
 	rm -rf core* || true
+	BACKUP_LD_PRELOAD="$LD_PRELOAD"
+	if [ $SANATIZER_BUILD -eq 1 ]; then
```
MishimaHaruna: `$SANITIZER_BUILD`?

tools/ci/travis.sh
```diff
@@ -102,9 +107,10 @@ DBNAME=ragnarok
 DBUSER=ragnarok
 DBPASS=ragnarok
 DBHOST=localhost
+SANATIZER_BUILD=0
 
 case "$MODE" in
-	createdb|importdb|test)
+	createdb|importdb|test|extratest)
```
MishimaHaruna: extratest should be handled separately since it doesn't take the DB arguments (but I don't see it called with this `1` flag either?)

To simplify it and avoid the positional arguments, I'd suggest using `--sanitizer-build` as first argument, and then testing with something similar to

```
if [ "$1" = "--sanitizer-build" ]; then
    SANITIZER_BUILD=1
    shift
fi
```

That way the script will support being called as

```
./travis.sh --sanitizer-build extrabuild
```

as well as

```
./travis.sh --sanitizer-build ragnarok ragnarok ragnarok blah
```

tools/ci/travis.sh
```diff
@@ -244,9 +253,10 @@ EOF
 			export ASAN_OPTIONS=leak_check_at_exit=1:detect_stack_use_after_return=true:strict_init_order=true:detect_odr_violation=0
 		fi
 		# run_test spinlock # Not running the spinlock test for the time being (too time consuming)
-		run_test libconfig
-		run_test chunked
+		# run_test libconfig
+		# run_test chunked
 		echo "run all servers without HPM"
+		sleep 1000000000000000
```
MishimaHaruna: This is totally not a test, right? Now I know what year you're from, time traveler

MishimaHaruna (changes requested):

tools/ci/travis.sh
guilherme-gm, line 1: I wonder if we should rename this (maybe later), since we are removing travis, but this is used by gitlab + actions

hemagx: let's keep it for legacy reasons for when our grandchildren ask us about it ~

PR #3413, opened 2026-05-23
Migrating build system to cmake

hemagx:
### Pull Request Prelude

### Changes Proposed

This Pull request changes our compilation generation tool to cmake so unify all platforms under a single umberalla, we also added conan as our dependency manager to get rid of as much of 3rdparty as possible and also the DLLs deployed for windows.

**Issues addressed:** 


CMakeLists.txt
```diff
@@ -0,0 +1,212 @@
...
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  set(RUNNING_ON_LINUX ON)
+else()
+  set(RUNNING_ON_LINUX OFF)
+endif()
+
+include(CMakeDependentOption)
+# Configuration options
+cmake_dependent_option(ENABLE_LIBBACKTRACE "Compiles with libbacktrace. (off by default - experimental)" ON ${RUNNING_ON_LINUX} OFF)
```
MishimaHaruna: I think it's more typical to express this as

```diff
-cmake_dependent_option(ENABLE_LIBBACKTRACE "Compiles with libbacktrace. (off by default - experimental)" ON ${RUNNING_ON_LINUX} OFF)
+cmake_dependent_option(ENABLE_LIBBACKTRACE "Compiles with libbacktrace. (off by default - experimental)" ON RUNNING_ON_LINUX OFF)
```

(let the condition parser expand the variable, so that if it were to contain weird stuff, it wouldn't break)

CMakeLists.txt
```diff
@@ -0,0 +1,212 @@
...
+set(WITH_HTTP_PARSER "http-parser" CACHE STRING "select http parser. Allowed values: http-parser, llhttp")
+option(ENABLE_GDB "Compiles with gdb specific optimzations for better debugging. (off by default)" OFF)
+
+if(NOT "${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+  if(${ENABLE_GDB})
+    message(FATAL_ERROR "ENABLE_GDB can be only used in debug builds")
```
MishimaHaruna: I think it might make sense to allow this in `RelWithDebInfo` as well

cmake_modules/LanguageFeatureHelper.cmake
```diff
@@ -0,0 +1,57 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+include(CheckIncludeFile)
+
+# Helpers to test existance of header and add definition for its existance
+# _header header name without .h
+# _def def to be defined if the header is found keep empty to use default scheme
+# _required boolean to fail if check doesn't pass
+function(test_header_def _header _def _required)
```
MishimaHaruna: Instead of a boolean `_required`, this could take a `REQUIRED` flag, to make the calls easier to understand

```
test_header_def("foo.h" "")
test_header_def("bar.h" "" REQUIRED)
```

is clearer than

```
test_header_def("foo.h" "" OFF)
test_header_def("bar.h" "" ON)
```

cmake_modules/LanguageFeatureHelper.cmake
```diff
@@ -0,0 +1,57 @@
...
+    message(FATAL_ERROR "Header ${_header}.h is required")
+  endif()
+endfunction()
+
+# Helpers to test existance of a function and add definition for its existance
+# _function function name
+# _def def to be defined if the function is found keep empty to use default scheme
+# _required boolean to fail if check doesn't pass
+include(CheckFunctionExists)
+function(test_function_def _function _def _required)
```
MishimaHaruna: Same as above about the `REQUIRED` flag

cmake_modules/CompilerFlagHelpers.cmake
```diff
@@ -0,0 +1,64 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Helpers to help test and set flags and warnings
+include(CheckCCompilerFlag)
+
+# Tests for warning compiler flag
+# _name warning flag name without the leading W
+# _required boolean to fail if check doesn't pass
+function(testadd_warning_compiler_flag _name _required)
```
MishimaHaruna: Same as below about the `_required` flag

cmake_modules/CompilerFlagHelpers.cmake
```diff
@@ -0,0 +1,64 @@
...
+    add_compile_options("-W${_name}")
+  elseif(${_required})
+    message(FATAL_ERROR "Compiler warning flag -W${_name} is required")
+  endif()
+endfunction()
+
+# Tests for compiler flag
+# _name flag name
+# _required boolean to fail if check doesn't pass
+function(testadd_compiler_flag _name _required)
```
MishimaHaruna: Same as below about the `_required` flag

cmake_modules/CompilerFlagHelpers.cmake
```diff
@@ -0,0 +1,64 @@
...
+    message(FATAL_ERROR "Compiler flag -${_name} is required")
+  endif()
+endfunction()
+
+include(CheckLinkerFlag)
+
+# Tests for linker flag
+# _name flag name
+# _required boolean to fail if check doesn't pass
+function(testadd_linker_flag _name _required)
```
MishimaHaruna: Same as below about the `_required` flag

CMakeLists.txt
```diff
@@ -0,0 +1,212 @@
...
+	#include <time.h>
+	#include <unistd.h>
+
+	int main(int argc, char** argv)
+	{
+		struct timespec tval;
+		return clock_gettime(CLOCK_MONOTONIC, &tval);
+	}
+]] HAVE_MONOTONIC_CLOCK)
+if(${HAVE_MONOTONIC_CLOCK})
```
MishimaHaruna:
```diff
-if(${HAVE_MONOTONIC_CLOCK})
+if(HAVE_MONOTONIC_CLOCK)
```

Let the condition parser expand your variables

CMakeLists.txt
```diff
@@ -0,0 +1,212 @@
...
+else()
+  message(FATAL_ERROR "unknown http_parser selected '${WITH_HTTP_PARSER}'. Allowed values: http-parser, llhttp. ... stopping")
+endif()
+
+# Check for compile settings
+
+# Debug setting
+if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+  add_compile_definitions(DEBUG)
+  if(${ENABLE_GDB})
```
MishimaHaruna, line 429:
```diff
-  if(${ENABLE_GDB})
+  if(ENABLE_GDB)
```

CMakeLists.txt
```diff
@@ -0,0 +1,212 @@
...
+
+# Check for functions
+test_function_def(uselocale "" OFF) # Libconfig
+test_function_def(newlocale "" OFF) # Libconfig
+test_function_def(freelocale "" OFF) # Libconfig
+test_function_def(setrlimit "" OFF)
+
+# Check for mandatory compile flags (currently it's just a list of all default flag using just ./configure)
+include(CompilerFlagHelpers)
+testadd_compiler_flag(pipe ON)
```
MishimaHaruna: All these compiler flags and warning flags are gcc and/or clang-specific, they should probably go into an `if(NOT "${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")` or similar branch

CMakeLists.txt
```diff
@@ -0,0 +1,212 @@
...
+
+# Check for compile settings
+
+# Debug setting
+if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+  add_compile_definitions(DEBUG)
+  if(${ENABLE_GDB})
+    testadd_compiler_flag(ggdb ON)
+  else()
+    testadd_compiler_flag(g ON)
```
MishimaHaruna: This should be added automatically (and with the correct flag for the detected compiler), no need to hack it here

CMakeLists.txt
```diff
@@ -0,0 +1,212 @@
...
+  endif()
+endif()
+
+# Run sysinfogen.sh whenever src/common/sysinfo.inc is required in a build
+add_custom_command(
```
MishimaHaruna: This will break on windows native builds - it needs to be in an if (not sure if hiding it behind a MSVC check is enough, might need some brainstorming on how to approach it properly)

3rdparty/http-parser/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+  FILES
+  ${HERC_HTTP_PARSER_HDRS}
+)
+
+add_library(hercules-http_parser STATIC
+  ${HERC_HTTP_PARSER_SRCS}
+  ${HERC_HTTP_PARSER_HDRS}
+)
+
+target_compile_options(hercules-http_parser PRIVATE -Wno-shadow -Wno-switch-enum)
```
MishimaHaruna: This seems gcc/clang-specific

3rdparty/libgif/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+  FILES
+  ${HERC_LIBGIF_HDRS}
+)
+
+add_library(hercules-libgif STATIC
+  ${HERC_LIBGIF_SRCS}
+  ${HERC_LIBGIF_HDRS}
+)
+
+target_compile_options(hercules-libgif PRIVATE -Wno-switch-enum -Wno-undef)
```
MishimaHaruna: This seems gcc/clang-specific

3rdparty/llhttp/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+  FILES
+  ${HERC_LLHTTP_HDRS}
+)
+
+add_library(hercules-llhttp STATIC
+  ${HERC_LLHTTP_SRCS}
+  ${HERC_LLHTTP_HDRS}
+)
+
+target_compile_options(hercules-llhttp PRIVATE -Wno-error -Wno-format -Wno-format-y2k -Wno-format-security -Wno-redundant-decls -Wno-undef -Wno-switch-enum)
```
MishimaHaruna: This seems gcc/clang-specific

src/common/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+  api.h
+  charloginpackets.h
+  charmappackets.h
+  mapcharpackets.h
+  chunked/rfifo.h
+  chunked/wfifo.h
+  config/defc.h
+  config/emblems.h
+  config/undefc.h
+  ../plugins/HPMHooking.h
```
MishimaHaruna: This will likely require redesign but let's leave it as is for now (can spawn an issue about it perhaps)

src/common/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+)
+
+list(APPEND COMMON_HDRS ${CONFIG_HDRS})
```
MishimaHaruna: Not a fan of having these defined in a separate variable that comes from a separate cmakelists, they should go into this file instead.

In general not a fan of listing source files in variables unless strictly necessary, I'd put them in add_library() directly

src/common/nullpo.c
```diff
@@ -30,8 +30,8 @@
 #include <stdlib.h>
 #include <string.h>
 #if defined(HAVE_LIBBACKTRACE)
-#include "libbacktrace/backtrace.h"
-#include "libbacktrace/backtrace-supported.h"
+#include "backtrace.h"
+#include "backtrace-supported.h"
```
MishimaHaruna: These belong to an external library so they should be `<backtrace.h>` and `<backtrace-supported.h>`

src/config/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+set(CONFIG_HDRS
+  "${CMAKE_SOURCE_DIR}/src/config/const.h"
+  "${CMAKE_SOURCE_DIR}/src/config/core.h"
+  "${CMAKE_SOURCE_DIR}/src/config/renewal.h"
+  "${CMAKE_SOURCE_DIR}/src/config/secure.h"
+  "${CMAKE_SOURCE_DIR}/src/config/classes/general.h"
+)
```
MishimaHaruna, line 29: As mentioned above, this file should probably be removed entirely (and these changed to relative paths, in the cmakelists that actually uses them)

cmake_modules/CompilerFlagHelpers.cmake
```diff
@@ -62,3 +62,51 @@ function(testadd_linker_flag _name _required)
     message(FATAL_ERROR "Linker flag -${_name} is required")
   endif()
 endfunction()
+
+# Tests for a sanitizer flag
+# _name flag name
+# _required boolean to fail if check doesn't pass
+function(testadd_sanitizer_flag _name _required)
```
MishimaHaruna: This could also use a `REQUIRED` flag

src/common/CMakeLists.txt
```diff
@@ -123,3 +123,7 @@ target_link_libraries(common PRIVATE hercules-libconfig)
 target_link_libraries(common PRIVATE libbacktrace::libbacktrace)
 target_link_libraries(common PRIVATE mariadb-connector-c::mariadb-connector-c)
 target_link_libraries(common PRIVATE ZLIB::ZLIB)
+
+if("${COMMON_MEMORY_MANAGER_LINK}")
```
MishimaHaruna:
```diff
-if("${COMMON_MEMORY_MANAGER_LINK}")
+if(COMMON_MEMORY_MANAGER_LINK)
```

CMakeLists.txt
```diff
@@ -16,7 +16,10 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-cmake_minimum_required(VERSION 3.24)
+cmake_minimum_required(VERSION 3.25)
+
+# CMake Policies
+cmake_policy(SET CMP0140 NEW)
```
MishimaHaruna: Thy is this `NEW` behavior called explicitly? I'm seeing that it's new in 3.25, so isn't it enough to say `cmake_minimum_required(VERSION 3.25)`?

CMakeLists.txt
```diff
@@ -36,6 +39,13 @@ if(NOT CMAKE_BUILD_TYPE)
   message(FATAL_ERROR "Build type not specified. Specify it with -DCMAKE_BUILD_TYPE=type (type can be one of Release, RelWithDebInfo, Debug)")
 endif()
 
+if(CMAKE_SIZEOF_VOID_P EQUAL 4)
+  message(FATAL_ERROR "32-bit builds are not supported. Please use a 64-bit generator/toolchain.")
+endif()
```
MishimaHaruna: I don't see a need to block 32 bit builds (which right now are possible, at least on msvc).
It probably makes very little sense to build 32 bit, I can understand that, but I wouldn't go out of my way to outright block them

CMakeLists.txt
```diff
@@ -116,12 +170,70 @@ testadd_warning_compiler_flag(zero-length-bounds OFF)
...
+  int foo(void) {
+		return 0;
+	}
+
+	int main(int argc, int *argv[]) {
+		int foo = 0;
+		return foo + 1;
+	}
+]] IS_WSHADOW_GOOD)
+if(${IS_WSHADOW_GOOD})
```
MishimaHaruna:
```diff
-if(${IS_WSHADOW_GOOD})
+if(IS_WSHADOW_GOOD)
```

CMakeLists.txt
```diff
@@ -116,12 +170,70 @@ testadd_warning_compiler_flag(zero-length-bounds OFF)
...
+  testadd_warning_compiler_flag(shadow OFF)
+endif()
+set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})
+
+# Certain versions of gcc implement -Wformat-truncation in a naive way,
+# disregarding the actual values that an integer can take in every code path
+# and assuming a worst case of INT_MAX. That makes it not very useful as it
+# produces meaningless warnings.
+has_compiler_flag(Wno-format-truncation HAS_NO_FORMAT_TRANSACTION)
+if(NOT ${HAS_NO_FORMAT_TRANSACTION})
```
MishimaHaruna:
```diff
-if(NOT ${HAS_NO_FORMAT_TRANSACTION})
+if(NOT HAS_NO_FORMAT_TRANSACTION)
```

CMakeLists.txt
```diff
@@ -116,12 +170,70 @@ testadd_warning_compiler_flag(zero-length-bounds OFF)
...
+			int foo(int argc, char *argv[]) {
+				char buf[3];
+				for (int i = 0; i < 4; i++) {
+					snprintf(buf, sizeof(buf), "a%d", i);
+				}
+				return buf[1];
+			}
+  ]] IS_FORMAT_TRANSACTION_GOOD)
+
+  if(NOT ${IS_FORMAT_TRANSACTION_GOOD})
```
MishimaHaruna:
```diff
-  if(NOT ${IS_FORMAT_TRANSACTION_GOOD})
+  if(NOT IS_FORMAT_TRANSACTION_GOOD)
```

CMakeLists.txt
```diff
@@ -116,12 +170,70 @@ testadd_warning_compiler_flag(zero-length-bounds OFF)
...
+    testadd_warning_compiler_flag(no-format-truncation OFF)
+  endif()
+  set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})
+endif()
+
+# Plugin flags
+testwrite_compiler_flag(fPIC OFF SOFLAGS)
```
MishimaHaruna: Seems compiler specific

CMakeLists.txt
```diff
@@ -162,6 +274,26 @@ else()
   message(STATUS "CLOCK_MONOTONIC is either unsupported or not working as expected")
 endif()
 
+# Check if epoll is available and we're on linux
+if(${ENABLE_EPOLL})
```
MishimaHaruna:
```diff
-if(${ENABLE_EPOLL})
+if(ENABLE_EPOLL)
```

CMakeLists.txt
```diff
@@ -162,6 +274,26 @@ else()
...
+		#error This is not Linux
+		#endif
+		#include <sys/epoll.h>
+    int main(int argc, char** argv)
+	  {
+	  	return 0;
+	  }
+  ]] HAVE_LINUX_EPOLL)
+
+  if(NOT ${HAVE_LINUX_EPOLL})
```
MishimaHaruna:
```diff
-  if(NOT ${HAVE_LINUX_EPOLL})
+  if(NOT HAVE_LINUX_EPOLL)
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
 
 # Check for compile settings
 
+# Memory manager setting
+set(ENABLE_MEMORY_MANAGER_ALLOWED
+  no
```
MishimaHaruna: `no` is a special value for cmake which becomes a boolean false in a boolean context, so it might be safer to avoid it

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  builtin
+  memwatch
+  dmalloc
+  gcollect
+  bcheck
+)
+
+if(NOT ${ENABLE_MEMORY_MANAGER} IN_LIST ENABLE_MEMORY_MANAGER_ALLOWED)
```
MishimaHaruna:
```diff
-if(NOT ${ENABLE_MEMORY_MANAGER} IN_LIST ENABLE_MEMORY_MANAGER_ALLOWED)
+if(NOT "${ENABLE_MEMORY_MANAGER}" IN_LIST ENABLE_MEMORY_MANAGER_ALLOWED)
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  message(FATAL_ERROR "unknown memory manager '${ENABLE_MEMORY_MANAGER}'... stopping")
+elseif(${ENABLE_MEMORY_MANAGER} STREQUAL "no")
```
MishimaHaruna:
```diff
-elseif(${ENABLE_MEMORY_MANAGER} STREQUAL "no")
+elseif("${ENABLE_MEMORY_MANAGER}" STREQUAL "no")
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  add_compile_definitions(NO_MEMMGR)
+elseif(${ENABLE_MEMORY_MANAGER} STREQUAL "memwatch")
```
MishimaHaruna:
```diff
-elseif(${ENABLE_MEMORY_MANAGER} STREQUAL "memwatch")
+elseif("${ENABLE_MEMORY_MANAGER}" STREQUAL "memwatch")
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  test_header_def(memwatch "" ON)
+  add_compile_definitions(MEMWATCH)
+elseif(${ENABLE_MEMORY_MANAGER} STREQUAL "dmalloc")
```
MishimaHaruna:
```diff
-elseif(${ENABLE_MEMORY_MANAGER} STREQUAL "dmalloc")
+elseif("${ENABLE_MEMORY_MANAGER}" STREQUAL "dmalloc")
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  test_header_def(dmalloc "" ON)
+  add_compile_definitions(DMALLOC DMALLOC_FUNC_CHECK)
+  set(COMMON_MEMORY_MANAGER_LINK "dmalloc")
+elseif(${ENABLE_MEMORY_MANAGER} STREQUAL "gcollect")
```
MishimaHaruna:
```diff
-elseif(${ENABLE_MEMORY_MANAGER} STREQUAL "gcollect")
+elseif("${ENABLE_MEMORY_MANAGER}" STREQUAL "gcollect")
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  test_header_def(gc "" ON)
+  add_compile_definitions(GCOLLECT)
+  set(COMMON_MEMORY_MANAGER_LINK "gc")
+elseif(${ENABLE_MEMORY_MANAGER} STREQUAL "bcheck")
```
MishimaHaruna:
```diff
-elseif(${ENABLE_MEMORY_MANAGER} STREQUAL "bcheck")
+elseif("${ENABLE_MEMORY_MANAGER}" STREQUAL "bcheck")
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  add_compile_definitions(BCHECK)
+endif()
+
+if(${ENABLE_PACKETVER})
```
MishimaHaruna:
```diff
-if(${ENABLE_PACKETVER})
+if(ENABLE_PACKETVER)
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  add_compile_definitions("PACKETVER=${ENABLE_PACKETVER}")
+endif()
+
+if(${ENABLE_PACKETVER_RE})
```
MishimaHaruna:
```diff
-if(${ENABLE_PACKETVER_RE})
+if(ENABLE_PACKETVER_RE)
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  add_compile_definitions(ENABLE_PACKETVER_RE)
+endif()
+
+if(${ENABLE_PACKETVER_ZERO})
```
MishimaHaruna:
```diff
-if(${ENABLE_PACKETVER_ZERO})
+if(ENABLE_PACKETVER_ZERO)
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  add_compile_definitions(ENABLE_PACKETVER_ZERO)
+endif()
+
+if(${ENABLE_PACKETVER_SAK})
```
MishimaHaruna:
```diff
-if(${ENABLE_PACKETVER_SAK})
+if(ENABLE_PACKETVER_SAK)
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  add_compile_definitions(ENABLE_PACKETVER_SAK)
+endif()
+
+if(${ENABLE_PACKETVER_AD})
```
MishimaHaruna:
```diff
-if(${ENABLE_PACKETVER_AD})
+if(ENABLE_PACKETVER_AD)
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  add_compile_definitions(ENABLE_PACKETVER_AD)
+endif()
+
+if(${CASHSHOP_PREVIEW_PATCH})
+  add_compile_definitions(ENABLE_CASHSHOP_PREVIEW_PATCH)
+endif()
+
+if(${OLD_CASHSHOP_PREVIEW_PATCH})
```
MishimaHaruna:
```diff
-if(${OLD_CASHSHOP_PREVIEW_PATCH})
+if(OLD_CASHSHOP_PREVIEW_PATCH)
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+  add_compile_definitions(ENABLE_OLD_CASHSHOP_PREVIEW_PATCH)
+endif()
+
+if(${ENABLE_EPOLL})
```
MishimaHaruna:
```diff
-if(${ENABLE_EPOLL})
+if(ENABLE_EPOLL)
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+if(${ENABLE_PACKETVER})
+  add_compile_definitions("PACKETVER=${ENABLE_PACKETVER}")
+endif()
+
+if(${ENABLE_PACKETVER_RE})
+  add_compile_definitions(ENABLE_PACKETVER_RE)
+endif()
+
+if(${ENABLE_PACKETVER_ZERO})
+  add_compile_definitions(ENABLE_PACKETVER_ZERO)
```
MishimaHaruna: This long sequence of if foo add_compile_definitions foo can be replaced with generator expression for a more compact and cleaner look. Something like:


```
add_compile_definitions(
  $<ENABLE_PACKETVER:PACKETVER=${ENABLE_PACKETVER}>
  $<ENABLE_PACKETVER_RE:ENABLE_PACKETVER_RE>
  $<ENABLE_PACKETVER_ZERO:ENABLE_PACKETVER_ZERO>
  $<ENABLE_PACKETVER_SAK:ENABLE_PACKETVER_SAK>
  ...
)
```

docs here: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:condition

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+if(${OLD_CASHSHOP_PREVIEW_PATCH})
+  add_compile_definitions(ENABLE_OLD_CASHSHOP_PREVIEW_PATCH)
+endif()
+
+if(${ENABLE_EPOLL})
+  add_compile_definitions(SOCKET_EPOLL)
+endif()
+
+if("${OBFUSCATIONKEY1}" AND "${OBFUSCATIONKEY2}" AND "${OBFUSCATIONKEY3}")
+  if(ON)
```
MishimaHaruna: Please doublecheck this, it's likely that some autoformatting tool killed this expression (i.e. cmake-fmt is awfully broken and shouldn't be used)

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+    add_compile_definitions("OBFUSCATIONKEY1=${obfuscationkey1}" "OBFUSCATIONKEY2=${obfuscationkey2}" "OBFUSCATIONKEY3=${obfuscationkey3}")
+  else()
+    message(FATAL_ERROR "obfuscation keys provided doesn't match the correct hexdecimal format")
+  endif()
+endif()
+
 # Debug setting
+if(NOT "${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+  if(${ENABLE_GDB})
```
MishimaHaruna, line 429:
```diff
-  if(${ENABLE_GDB})
+  if(ENABLE_GDB)
```

CMakeLists.txt
```diff
@@ -186,7 +318,82 @@ find_package(ZLIB REQUIRED)
...
+    message(FATAL_ERROR "ENABLE_GDB can be only used in debug builds")
```
MishimaHaruna, line 430: I already commented on this in a previous commit, but I believe github wasn't able to notice that the line was moved - It could make sense to allow this in RelWithDebInfo builds as well, not just Debug

CMakeLists.txt
```diff
@@ -196,6 +403,105 @@ if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
   endif()
 endif()
 
+if(${ENABLE_BUILDBOT})
```
MishimaHaruna:
```diff
-if(${ENABLE_BUILDBOT})
+if(ENABLE_BUILDBOT)
```

CMakeLists.txt
```diff
@@ -196,6 +403,105 @@ if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
...
+  add_compile_definitions(BUILDBOT)
+endif()
+
+if(${ENABLE_RDTSC})
```
MishimaHaruna:
```diff
-if(${ENABLE_RDTSC})
+if(ENABLE_RDTSC)
```

CMakeLists.txt
```diff
@@ -196,6 +403,105 @@ if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
...
+  add_compile_definitions(ENABLE_RDTSC)
+endif()
+
+if("${ENABLE_PROFILER}" STREQUAL "no")
```
MishimaHaruna: Probably a good idea to change this to a different string that has no special boolean meaning, or people will expect to also be able to pass OFF or FALSE

CMakeLists.txt
```diff
@@ -196,6 +403,105 @@ if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
...
+elseif("${ENABLE_PROFILER}" STREQUAL "gprof")
+  add_compile_options(pg)
+  add_link_options(pg)
+else()
+  message(FATAL_ERROR "Unknown profiler option ${ENABLE_PROFILER}")
+endif()
+
+if(${DISABLE_RENEWAL})
```
MishimaHaruna: I'd remove this DISABLE_RENEWAL hack, it's redundant and equivalent to -DENABLE_RENEWAL=FALSE

CMakeLists.txt
```diff
@@ -196,6 +403,105 @@ if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
...
+  set(ENABLE_RENEWAL OFF)
+endif()
+
+if(NOT ${ENABLE_RENEWAL})
```
MishimaHaruna:
```diff
-if(NOT ${ENABLE_RENEWAL})
+if(NOT ENABLE_RENEWAL)
```

CMakeLists.txt
```diff
@@ -196,6 +403,105 @@ if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
...
+  add_compile_definitions(DISABLE_RENEWAL)
+endif()
+
+if(${ENABLE_STATIC_LINKING})
```
MishimaHaruna: CMake already has a flag for this, see https://cmake.org/cmake/help/v4.3/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS

MishimaHaruna (changes requested):
Initial review - will continue later (I'm going commit by commit, so some comments might be outdated, feel free to close them)

CMakeLists.txt
hemagx: I was too lazy to implement the check in configure.ac for 32bit builds, if you can give it a look and see if it makes sense or not, if not i can just remove this check.

CMakeLists.txt
hemagx: This was a placheholder, corrected later.

CMakeLists.txt
hemagx: Removed later

src/common/CMakeLists.txt
```diff
@@ -127,3 +128,20 @@ target_link_libraries(common PRIVATE ZLIB::ZLIB)
...
   target_link_libraries(common PRIVATE ${COMMON_MEMORY_MANAGER_LINK})
 endif()
+
+# Run sysinfogen.sh whenever src/common/sysinfo.inc is required in a build
+
+# In original configure.ac CPFLAGS and DEFs can be mixed but LDFLAGS isn't included
+get_directory_property(SYSGEN_CFLAGS COMPILE_OPTIONS)
+get_directory_property(SYSGEN_DEFS COMPILE_DEFINITIONS)
+
+add_custom_command(
```
MishimaHaruna: OS-specific, will need to be put behind a condition

CMakeLists.txt
MishimaHaruna: The configure.ac check is not very useful, it's meant to enforce 32 bit builds when you don't want 64 bit ones, on 64 bit OSes/compilers. It can be safely removed, without replacing it with anything. People that want 32 bit builds, will provide the compiler flags themselves, or use a 32 bit cross-compiler

MishimaHaruna (changes requested):

src/char/CMakeLists.txt
```diff
@@ -0,0 +1,106 @@
...
+  int_quest.h
+  int_rodex.h
+  int_storage.h
+  int_userconfig.h
+  pincode.h
+  packets_hc_struct.h
+  capiif.h
+)
+
+list(APPEND CHAR_HDRS ${CONFIG_HDRS})
```
MishimaHaruna: These should probably be included indirectly by depending on `common` or a dedicated library

src/login/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+  account.h
+  apipackets.h
+  HPMlogin.h
+  ipban.h
+  lclif.h
+  loginlog.h
+  lapiif.h
+)
+
+list(APPEND LOGIN_HDRS ${CONFIG_HDRS})
```
MishimaHaruna: Same as above, these should just be a dependency on common or dedicated header-only library

src/map/CMakeLists.txt
```diff
@@ -0,0 +1,173 @@
...
+  skill.h
+  status.h
+  storage.h
+  stylist.h
+  trade.h
+  unit.h
+  vending.h
+)
+
+list(APPEND MAP_HDRS ${CONFIG_HDRS})
```
MishimaHaruna: Same as above

MishimaHaruna (changes requested):

CMakeLists.txt
```diff
@@ -29,12 +29,32 @@ set(CMAKE_C_STANDARD 11)
...
+    sset(CMAKE_INSTALL_RPATH "@rpath/${CMAKE_INSTALL_LIBDIR}/")
+  endif()
+  set(CMAKE_INSTALL_RPATH_USE_LINK_PATH OFF)
+endif()
+
 if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
   message(FATAL_ERROR "In-source builds are not allowed. Please create a separate build directory.")
 endif()
 
+if(NOT CMAKE_INSTALL_PREFIX)
```
MishimaHaruna: This is always set by cmake if not passed so the check will be always false. See https://cmake.org/cmake/help/v4.3/variable/CMAKE_INSTALL_PREFIX.html and https://cmake.org/cmake/help/v4.3/variable/CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT.html#variable:CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT

MishimaHaruna (changes requested):

cmake_modules/GeneralHelpers.cmake
```diff
@@ -0,0 +1,27 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Appends a list of headers to global property ALL_HDRS with their absloute path
+# _headers list of headers
+function(add_header_all_list _headers)
```
MishimaHaruna: What do we need this for? Can we avoid messing with lists of headers like this?

cmake_modules/GeneralHelpers.cmake
```diff
@@ -25,3 +25,14 @@ function(add_header_all_list _headers)
   endforeach()
   set_property(GLOBAL PROPERTY ALL_HDRS ${OLD_ALL_HDRS})
 endfunction()
+
+# Appends a list of headers to global property ALL_CONFIG_HDRS with their absloute path
+# _headers list of headers
+function(add_header_all_config_list _headers)
```
MishimaHaruna: It would be nice not to have to mess with lists of files like this. What do we need this for, and can we get rid of it?

src/plugins/HPMHooking/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+
+# List of header files of the plugin
+set(HERC_PLUGIN_H
+  HPMHooking.h
+)
+
+# List of include files
+set(HPMHOOKING_INCLUDES "")
+
+file(GLOB include_files RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} *.inc)
```
MishimaHaruna: Might be better to remove the glob since we know the names we expect

src/plugins/sample.c
```diff
@@ -18,27 +18,10 @@
...
-
-// check is TESTVAR defined from custom compile flags
-#ifndef TESTVAR
-#error TESTVAR not defined
-#endif
-
-// check is TESTVAR2 defined to value 2 from custom compile flags
-#if !(TESTVAR2 == 2)
-#error TESTVAR2 not defined
-#endif
```
MishimaHaruna, line 36: I'd keep these checks to demonstrate/test that the flags are added

src/plugins/sample/sample.c
```diff
@@ -18,27 +18,10 @@
...
-// check is TESTVAR2 defined to value 2 from custom compile flags
-#if !(TESTVAR2 == 2)
-#error TESTVAR2 not defined
-#endif
-
-#endif  // WIN32
-
 /// Sample Hercules Plugin
 
+#include "plugins/sample/sample.h"
```
MishimaHaruna: If this requires a full path, then we're doing something wrong in the cmake. We should ensure that each plugin has its own folder in its include path

src/plugins/sample/sample.h
```diff
@@ -0,0 +1,33 @@
...
+
+// check is TESTVAR defined from custom compile flags
+#ifndef TESTVAR
+#error TESTVAR not defined
+#endif
+
+// check is TESTVAR2 defined to value 2 from custom compile flags
+#if !(TESTVAR2 == 2)
+#error TESTVAR2 not defined
+#endif
```
MishimaHaruna: I don't think this needs to be in a header, it can be in the plugin's main .c file in my opinion (it's not defining anything that it needs to share)

If you want to show an example of a plugin that has its own .h file, then maybe add something defined or declared here that is used in the .c file

MishimaHaruna (changes requested):

tools/HPMHookGen/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+set(API_SERVER_PATH "${CMAKE_SOURCE_DIR}/src/api")
+
+if(DOXYGEN_FOUND AND PERL_FOUND)
+  add_custom_target(hooks
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    COMMAND
+      echo
+      "Preprocessing files"
+    COMMAND
+      ./preprocess.py
```
MishimaHaruna, line 29: This kind of invocation is a bit specific to unix systems, so the whole thing should probably go behind a check to prevent it from choking on windows.

Alternatively (and probably better), this could call python explicitly, like it does for perl below. That way it should probably run on any OS.

Also, since it checks for perl's presence it should also check for python

MishimaHaruna (changes requested):

3rdparty/libconfig/CMakeLists.txt
```diff
@@ -48,3 +48,5 @@ add_library(hercules-libconfig STATIC
   ${HERC_LIBCONF_SRCS}
   ${HERC_LIBCONF_HDRS}
 )
+
+target_compile_options(hercules-libconfig PRIVATE -Wno-null-dereference)
```
MishimaHaruna: gcc/clang-specific option, will require a conditional

MishimaHaruna (changes requested):

src/common/core.c
```diff
@@ -50,9 +50,20 @@
 
 #ifndef _WIN32
 #	include <unistd.h>
-#else
+#endif
+#if defined(WIN32)
```
MishimaHaruna: testing for `WIN32` right after testing for `_WIN32` will look odd to those in the future who will git blame this and end up finding this commit.

I would instead group all the unix checks within the else branch:

```
#if defined(WIN32)
#include "common/winapi.h"
#else
#include <unistd.h>
#if defined(__sun)
 //...
#elif // ...
#endif
#endif

```

3rdparty/libconfig/libconfig.h
```diff
@@ -27,17 +27,7 @@
...
-#if defined(_WIN32) || defined(__WIN32__) || defined(WIN32)
-#if defined(LIBCONFIG_STATIC)
 #define LIBCONFIG_API
-#elif defined(LIBCONFIG_EXPORTS)
-#define LIBCONFIG_API __declspec(dllexport)
-#else /* ! LIBCONFIG_EXPORTS */
-#define LIBCONFIG_API __declspec(dllimport)
-#endif /* LIBCONFIG_STATIC */
-#else /* ! WIN32 */
-#define LIBCONFIG_API
```
MishimaHaruna, line 39: Is there a need to kill the possibility to build this as a dll?

src/common/sysinfo.c
```diff
@@ -39,6 +39,7 @@
 #include <stdlib.h> // atoi
 #ifdef WIN32
 #	include <windows.h>
+#	include <winternl.h>
```
MishimaHaruna, line 42: What's this header providing here? It might be worth adding a comment to it since it's a quite unusual (and not very recommended) header to include

CMakeLists.txt
```diff
@@ -117,6 +117,10 @@ set(WITH_HTTP_PARSER "http-parser" CACHE STRING "select http parser. Allowed val
 option(ENABLE_TESTING "Enables compiling and running of tests" OFF)
 option(ENABLE_CLASSIC_AUTOSPELL "Enables classic auto spell list" OFF)
 
+if(ENABLE_TESTING)
+  enable_testing()
```
MishimaHaruna: This should probably not be part of this commit since it's not mentioned in the commit message.

It's also called twice (the second time in the src/test CMakeLists)

src/common/atomic.h
```diff
@@ -36,10 +36,7 @@
 #if defined(_MSC_VER)
 #include "common/winapi.h"
 
-// This checks if C/C++ Compiler Version is 18.00
-#if _MSC_VER < 1800
-
-#if !defined(_M_X64)
+#if _MSC_VER < 1800 && !defined(_M_X64)
```
MishimaHaruna: The endif matching this (line 95 at the moment) says `#endif //endif 32bit windows` - but that's no longer completely true after this change. It should be changed to avoid misleading readers

src/common/atomic.h
```diff
@@ -97,7 +94,12 @@ forceinline volatile int64 InterlockedExchange64(volatile int64 *target, int64 v
 
 #endif //endif 32bit windows
 
-#endif //endif _msc_ver check
+#if _M_X64 // Redefine the 32bit interlock function
```
MishimaHaruna: Is it intended for this to be an incomplete set?

The if above matches old MSVC x86 builds, this one matches x86_64 builds on both old and new MSVC. This leaves new MSVC x86 builds. What happens there?

MishimaHaruna (changes requested):

tools/ci/travis.sh
```diff
@@ -254,57 +243,55 @@ EOF
 		else
 			export ASAN_OPTIONS=leak_check_at_exit=1:detect_stack_use_after_return=true:strict_init_order=true:detect_odr_violation=0
 		fi
-		# run_test spinlock # Not running the spinlock test for the time being (too time consuming)
-		# run_test libconfig
-		# run_test chunked
+		run_test spinlock
```
MishimaHaruna: Was this uncommented accidentally?

tools/ci/travis.sh
```diff
@@ -254,57 +243,55 @@ EOF
...
-		run_server ./map-server "$PLUGINS --load-plugin db2sql --itemdb2sql"
-		run_server ./map-server "$PLUGINS --load-plugin db2sql --mobdb2sql"
+		run_server ./bin/map-server "$PLUGINS --load-plugin db2sql --db2sql"
+		run_server ./bin/map-server "$PLUGINS --load-plugin db2sql --itemdb2sql"
+		run_server ./bin/map-server "$PLUGINS --load-plugin db2sql --mobdb2sql"
 # look like works on windows only
 #		echo "run all servers with dbghelpplug"
 #		run_server ./login-server "$PLUGINS --load-plugin dbghelpplug"
 #		run_server ./char-server "$PLUGINS --load-plugin dbghelpplug"
 #		run_server ./map-server "$PLUGINS --load-plugin dbghelpplug"
```
MishimaHaruna: These lines should be removed along with the dbghelpplug plugin

.gitignore
```diff
@@ -121,7 +121,27 @@ Thumbs.db
...
-/src/plugins/Makefile
+/src/plugins/*
+!/src/plugins/HPMHooking/
+!/src/plugins/HPMHooking/*
+!/src/plugins/constdb2doc/
+!/src/plugins/constdb2doc/*
+!/src/plugins/db2sql/
+!/src/plugins/db2sql/*
+!/src/plugins/dbghelpplug/
+!/src/plugins/dbghelpplug/*
```
MishimaHaruna: These lines should be removed along with the dbghelpplug plugin

CMakeLists.txt
```diff
@@ -0,0 +1,570 @@
+# This file is part of Hercules.
```
MishimaHaruna, line 1: This script requires that the user passes some arguments (in particular the conan ones), that might be hard to remember. They should be documented in the project readme, and perhaps it would be nice to provide a convenience script (.sh and .bat/.ps1) with reasonable defaults

CMakeLists.txt
```diff
@@ -0,0 +1,570 @@
...
+  add_compile_options(/wd4018) # Disable signed/unsigned arthmatic mismatch
+  add_compile_options(/wd5287) # Disable enum conversion warning
+  add_compile_options(/wd4312) # Disable conversion to greater size warning (see 0xdeadbeef pointers)
+  # Definations
+  add_compile_definitions(WIN32 _WIN32 __WIN32 _CRT_SECURE_NO_DEPRECATE _CRT_NONSTDC_NO_DEPRECATE FD_SETSIZE=4096 WIN32_LEAN_AND_MEAN)
+  # Define arch
+  if(CMAKE_SIZEOF_VOID_P EQUAL 4)
+    add_compile_definitions(_X86_)
+  else()
+    add_compile_definitions(_ARM64_)
```
MishimaHaruna: This doesn't look right. It's defining `_ARM64_` on x86_64 builds?! (also, it doesn't seem to be used anywhere in the code)

CMakeLists.txt
```diff
@@ -0,0 +1,570 @@
...
+  add_compile_options(/wd4090) # Disable const qualifier mismatch because msvs doesn't like const passed to our aFree
+  add_compile_options(/wd4244 /wd4267) # Disable int conversion possible lose of data
+  add_compile_options(/wd4018) # Disable signed/unsigned arthmatic mismatch
+  add_compile_options(/wd5287) # Disable enum conversion warning
+  add_compile_options(/wd4312) # Disable conversion to greater size warning (see 0xdeadbeef pointers)
+  # Definations
+  add_compile_definitions(WIN32 _WIN32 __WIN32 _CRT_SECURE_NO_DEPRECATE _CRT_NONSTDC_NO_DEPRECATE FD_SETSIZE=4096 WIN32_LEAN_AND_MEAN)
+  # Define arch
+  if(CMAKE_SIZEOF_VOID_P EQUAL 4)
+    add_compile_definitions(_X86_)
```
MishimaHaruna: I don't see this macro used anywhere in the code. Is it necessary?

CMakeLists.txt
```diff
@@ -0,0 +1,570 @@
...
+else()
+  message(FATAL_ERROR "Unknown profiler option ${PROFILER}")
+endif()
+
+if(NOT ENABLE_RENEWAL)
+  add_compile_definitions(DISABLE_RENEWAL)
+endif()
+
+if(ENABLE_WERROR)
+  testadd_warning_compiler_flag(error REQUIRED)
```
MishimaHaruna: This flag is gcc/clang-specific. The MSVC equivalent is /WX.
Perhaps this option could be renamed to `WARNINGS_ARE_ERRORS` and made it work across compilers instead?

CMakeLists.txt
```diff
@@ -0,0 +1,570 @@
...
+endif()
+
+if(ENABLE_LTO)
+  testadd_compiler_flag(flto REQUIRED)
```
MishimaHaruna, line 462: gcc/clang-specific flags - this will fail with unexplanatory error messages on MSVC, so it could perhaps go into an if else with a proper message(FATAL_ERROR

CMakeLists.txt
```diff
@@ -0,0 +1,570 @@
...
+  testadd_compiler_flag(ffat-lto-objects REQUIRED)
+  testadd_linker_flag(flto REQUIRED)
+  testadd_linker_flag(ffat-lto-objects REQUIRED)
+endif()
+
+if(ENABLE_SANITIZE)
+  add_compile_definitions(SANITIZE)
```
MishimaHaruna, line 472: gcc/clang-specific flags - this will silently do nothing on MSVC, so it could perhaps go into an if else with a proper message(FATAL_ERROR) so that the user doesn't have any false expectations

MishimaHaruna (changes requested):

src/common/CMakeLists.txt
hemagx: quick question: if we include those headers just in `common.a` wouldn't that be enough for it to trigger a rebuild for any of the servers/plugins if those headers changed? if so i would just list them there and get rid of all of this variables mess.

hemagx: nvm this was answered before

src/plugins/sample/sample.c
hemagx: oh it doesn't require that that's a little screw up.

src/common/CMakeLists.txt
hemagx: that's already handled in a later commit where windows builds support were added.

src/common/sysinfo.c
hemagx: This is for the `NTSTATUS` macro used later in code, idk why it didn't require it before, but since move to cmake it no longer saw it until this header is used :x

MishimaHaruna (approved):
Seems mergeable to me now, great job!

PR #3414, opened 2026-06-07
Fix #166: Replace pc_autosave O(N) iterator with a save queue

### LINKED ISSUE #166: pc_autosave interval improvement

MrKeiKun:
The previous implementation iterated all connected players on every autosave tick (O(N)) to find the next one to save, and kept the timer running even with no players online.

This replaces that with a FIFO queue of bl.ids (O(1) lookup via map->id2sd). Players are enqueued on login and dequeued on logout. The timer is started by the first login and stops when the queue empties, eliminating idle wakeups on an empty server.

The interval is now computed as autosave_interval / queue_length, preserving the original intent of distributing saves evenly across the configured window.

### Pull Request Prelude

### Changes Proposed

**Issues addressed:** #166 

### DIFF
--- a/src/map/map.c
+++ b/src/map/map.c
@@ -2039,6 +2039,8 @@ static int map_quit(struct map_session_data *sd)
 		return 0;
 	}
 
+	pc->autosave_remove(sd->bl.id);
+
 	if( sd->expiration_tid != INVALID_TIMER )
 		timer->delete(sd->expiration_tid,pc->expiration_timer);
 
--- a/src/map/map.c
+++ b/src/map/map.c
@@ -2039,6 +2039,8 @@ static int map_quit(struct map_session_data *sd)
 		return 0;
 	}
 
+	pc->autosave_remove(sd->bl.id);
+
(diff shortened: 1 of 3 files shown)


skyleo:
Please don't merge this yet @hemagx .

It's a very nice idea, but there is a very similar nice idea already tested, adressing more edge cases and tested by time and a decent amount of players on Arcadia Online, which I will share and KeiKun and me will compare our solutions. Or rather we already are comparing as this PR came up due to a conversation about this optimization.

Issue #880, opened 2015-11-19
Spell Breaker


xBeret:
In aegis the skill SA_SPELLBREAKER applies damage when used on target.

Link post: http://herc.ws/board/tracker/issue-8457-spell-breaker/?gopid=24195#entry24195

xBeret:
@MishimaHaruna @hemagx 

Someone looking at this.

PR #3428, opened 2026-08-25
Make SA_SPELLBREAKER's HP siphon deal visible, non-lethal damage

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

In official Aegis, `SA_SPELLBREAKER` deals visible damage to the target as part of its HP siphon effect (recovering 2% max HP for the caster at skill level 5+, on PvP maps against players). Hercules currently applies this via `status_zap()`, which reduces the target's HP silently with no `clif->damage` packet, so the target sees no hit or damage number.

SP drain from the skill is unaffected and remains silent.

**Issues addressed:** #880


kyeme:
Tested and working~

Issue #1144, opened 2016-02-07
Misleading Status Window Info (DEF)


kyeme:
In Official, you see the right number of DEF on your Status Window according to the refinement of your Equips. But here in Hercules, it doesnt show the right info for DEF. (Theres nothing wrong with the Refine Formula, its just the Status Window that is bugged)

Official:
![official def](https://cloud.githubusercontent.com/assets/4656578/12871088/0520ddae-cda3-11e5-9575-2448c6e11928.jpg)

Hercules:
![def](https://cloud.githubusercontent.com/assets/4656578/12871090/096149ee-cda3-11e5-8037-cf5f5deb3224.jpg)

PS: It also show a wrong number ATK but I will make an different report about this since I still dont know the cause of this problem.

Photo source:
http://z4.invisionfree.com/pROsins/ar/t1471.htm

Playtester:
I want to note, that I like the way rAthena / Herc handle DEF more.
On Aegis, each refine increases DEF by 1 on the client, but only by 0.7 on the server.

rAthena/Herc show the real reduction: 60 DEF = 60% less damage

kyeme:
Its a good idea if we have a config of official/unofficial behavior :D

Jedzkie:
@malufett

MrKeiKun:
> Its a good idea if we have a config of official/unofficial behavior :D 👍

i gotchu 😺

PR #3433, opened 2026-08-25
Add option to display official-style (Aegis) refine DEF in the pre-renewal status window

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

On pre-renewal, the status window's DEF has always shown the real, effective DEF from armor refine (e.g. +7 at +10 refine, since `refine_db.conf`'s pre-re `StatsPerLevel` is 0.66/level). Official (Aegis) clients instead show a flat +1 DEF per armor refine level on screen, even though only the smaller amount is actually applied to damage reduction server-side, so at the same +10 refine, official displays +10 while the real reduction is only +7.

This adds an off-by-default `official_def_display` battle_config option. When enabled, the pre-renewal status window shows the official-style (+1/refine) number instead of the real one, for server admins who want visual parity with official servers. The real value used in damage calculations (`battle_status.def`) is never changed by this option.

**Issues addressed:** #1144

### DIFF
--- a/conf/map/battle/player.conf
+++ b/conf/map/battle/player.conf
@@ -209,6 +209,14 @@ costume_refine_def: true
 shadow_refine_def: true
 shadow_refine_atk: true
 
+// Show the official (Aegis) client status window DEF, which awards a flat +1 DEF
+// per armor refine level for display purposes, rather than the real value used in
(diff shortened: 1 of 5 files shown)


kyeme:
Tested, and it's working. Thanks!

Issue #576, opened 2015-06-28
Eff_Stone


AnisotropicDefixation:
This is from item scripts.

It's skipping OPT1_STONEWAIT entirely when it should still employ it but with a lower duration (a couple of seconds).

On a side note, where the hell are item scripts handled in src?

MishimaHaruna:
I don't see any code that handles it differently, SC_STONE always triggers OPT1_STONEWAIT (see `status_change_start`)... Which items scripts, specifically?

Scripts are executed in pc_equipitem (look for a call to `script->run()`).

AnisotropicDefixation:
1732 or 4315 for example, anything with this script.

AnisotropicDefixation:
This might be of some help:

WL_SIENNAEXECRATE petrifies you instantly (skipping OPT1_STONEWAIT ), which is the intended behavior, the function used is status->change_start while MG_STONECURSE uses sc_start4.

Now, in script.c, move to BUILDIN(sc_start) and notice that it's also using status->change_start. This could do the same thing and be the reason behind SC_BURNING expiring before ticking even once when triggered by item scripts as well.

AnisotropicDefixation:
Can anyone shed some light on this?

MishimaHaruna:
The problem isn't related to `sc_start4` versus `status->change_start` (if you check the macro in status.h, you can see that they're pretty much the same thing)

The actual problem is that `SC_STONE` expects the duration of the `OPT1_STONEWAIT` state in its `val4` field (`MG_STONECURSE` passes `skill->get_time(skill_id, skill_lv)`, while `WL_SIENNAEXECRATE` passes `500` -- half a second, not instant)

`bAddEffWhenHit`, though always passes `0` as `val4` argument to `status->change_start`, and offers no way to alter that.

@malufett what do you think? Do we need to add an extra field to the `struct s_addeffect` (and a bonus4 variant of `bAddEffWhenHit`), or there's a better way to implement those items' scripts?

malufett:
@MishimaHaruna
actually I found the code in aegis however I have lack of time to check it further..
according to it, it accepts a parameter for the delay before the effect take into action..like what Kyeme has discussed in his previous reports..however I need to confirm all instances where the delay is computed, all I found is 0 delay set on the param...regarding with SC_STONE I need to decode it first in aegis..
mostly in skills this is called the 0 delay
[...]

Issue #1141, opened 2016-02-05
Status Effect delay


kyeme:
Status Effects such as Sleep, Frost, Blind Stun etc.. have slight delays before the status fully take effect.

https://youtu.be/WJQSwdSANoI

kyeme:
@Playtester

Playtester:
I'm taken a look at it now, we apply the status change when the damage is applied. There's nothing that trigger later, so it would require to implement a new kind of timer. Not so easy. /hm

Playtester:
Upon further testing on officials:
- When you do a normal attack, the effect kicks it when your animation ends, for e.g. when you use a Katar it's the time right after your offhand damage appears, so it is influenced by aMotion it seems.
- For Scream / Frost Joke's effect kicks in only 3500ms after I pressed the skill key!
- ASPD doesn't seem to influence it, it was 3500ms with 20 ASPD but also with 190 ASPD.
- Bragi also doesn't seem to influence it, still 3500ms.
- For Envenom, the effect seems to always come simultaneous with the damage.
- For Lex Divina the delay is 1500ms. If I cast Lex Divina again on the silenced target, the silence effect is removed immediately without delay.
- For monsters, I measured delays between 700ms and 1100ms. Seems to be using something close to aMotion but I couldn't find an exact pattern.

Not so easy, seems to be quite skill specific and also specific whether PC or MOB uses it.

Playtester:
@kyeme After further testing I concluded that this only affects normal attacks and skills that don't deal any damage.

Can you elaborate why this is important to implement? Then I know what to focus on.

kyeme:
In Official, you still have a chance to use recovery items such as Mastela Fruit before you got frozen. This is so important when they're using Asura Strike with Stormy Knight Card. 

Example: A Champion uses the Frost Lock Method: Palm Push Strike then Asura Strike on the chosen target with a Quad Stormy Knight Card weapon.

Video:
https://youtu.be/QALg4D9orZA

Playtester:
Hmm this seems to be a special property of Asura. Usually it seems to only apply for non-damage skills, but I tested it and the delay is there for Asura too. I tried with Sonic Blow which is a similar skill that also takes status effects from the weapon, but there it's at the same time of the damage.

Playtester:
I noticed something new. The delay from the attack to the status change is actually substracted from the status change duration.

[...]

PR #3453, opened 2026-09-09
Fix SC_STONE skipping OPT1_STONEWAIT for item-triggered petrify

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

- Fixes the three call sites in `skill.c` to pass a real wait duration for `SC_STONE`, defaulting to `1000`ms (matching the existing minimum-duration clamp already in `status.c`) when the scripter doesn't specify one.
- Adds a `wait_duration` field to `s_addeffect` / `s_addeffectonskill` (`pc.h`) and a new `bonus5` form for `bAddEff`, `bAddEffOnSkill`, and `bAddEffWhenHit` so item/card scripts can set this wait duration explicitly instead of relying on the hardcoded default.
- Along the way, wires up `bAddEffWhenHit` to also support a fixed `duration` argument (via its new `bonus5` form), which it never had. Closing @MishimaHaruna TODO 😄 

**Issues addressed:** #576


kyeme:
With these changes, would it also be possible to fix the status effect delay issue I reported in https://github.com/HerculesWS/Hercules/issues/1141 ?

MrKeiKun:
> With these changes, would it also be possible to fix the status effect delay issue I reported in #1141 ?

maybe in another PR. better, i need to dig deep re that one.

Issue #969, opened 2015-12-16
SP potion still works upon death


kyeme:
SP potion still works upon death if I spam it.

Example If I cast Asura Strike on a Paladin with Reflect Shield and spam SP Potion upon death I still have SP (depends on the amount of spammed sp pots)
![fenjxb0](https://cloud.githubusercontent.com/assets/4656578/11839325/d38f0fbc-a427-11e5-8bf1-39f6571090a0.jpg)

FreddieFatso:
Is this really a bug? I have been trying to reproduce it, but all I see is this order of events:

1. Asura Strike cast (goes off, target is damaged)
2. All SP drained, because of 1.
3. Blue Potion restores SP, amount depending on number used before
4. Player dies; Blue Potion can no longer be used and hence SP restoration stops

I haven't tried suiciding against a Reflect Shield, but instead running a naked char into MVPs. To me, it seems to be a case of "working as intended", unless someone wants to test this on official servers? But I doubt the Reflect Damage effect is instant even there, as there must always be a delay in an ability that is reactive by nature.

PS: I faintly... very faintly... remember this being similar on the old official EuRO servers. Don't quote me on that, though.

MrKeiKun:
@kyeme i can't reproduce it with current Herc. can you still reproduce this problem?

kyeme:
@MrKeiKun Yes, I’m still experiencing this issue.

MrKeiKun:
@kyeme Thanks for confirming it still reproduces. I've traced a likely mechanism in the source and want to check it against your server config
before going further.

Could you share three values?

1. `delay_battle_damage` in `conf/map/battle/battle.conf`
2. Whether you're running renewal or pre-renewal
3. `item_use_interval` in `conf/map/battle/items.conf`

kyeme:
@MrKeiKun 

1. `delay_battle_damage: true`
2. I'm using Pre-Renewal.
3. The only setting I changed is `item_use_interval: 0`.

PR #3460, opened 2026-09-10
Fix items usable after death when delay_battle_damage is enabled

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

`pc_useitem()` only checked death status via a stale snapshot taken earlier at packet-parse time in `clif_parse_UseItem()`. With `delay_battle_damage` enabled, a lethal hit is applied later by a timer, so an item queued in that window (e.g. spammed with `item_use_interval: 0`) could still run its effect and restore HP/SP after the player had already died.

Re-checks death (`pc_isdead()` / `status->isdead()`) immediately before the item's effect script runs, right before `script->run_use_script()`. The item is still consumed if it was already flagged for removal (matching the `useitemack` already sent to the client), but its effect is skipped.

**Issues addressed:** #969

### DIFF
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -321,7 +321,13 @@ static int battle_delay_damage(int64 tick, int amotion, struct block_list *src, 
 	if (((d_tbl && sc && check_distance_bl(target, d_tbl, sc->data[SC_DEVOTION]->val3)) || e_tbl) && damage > 0 && skill_id != PA_PRESSURE && skill_id != CR_REFLECTSHIELD)
 		damage = 0;
 
-	if ( !battle_config.delay_battle_damage || amotion <= 1 ) {
+	// Lethal damage bypasses the delay even when delay_battle_damage is on:
+	// otherwise a player can act (e.g. use an item) in the window between a
+	// lethal hit being calculated here and it actually being applied by the
+	// delayed timer, since nothing yet marks them as dying. Non-lethal hits
+	// keep the delay for its normal (cosmetic packet-pacing / animation
+	// sync) purpose. (#969)
+	if ( !battle_config.delay_battle_damage || amotion <= 1 || damage >= status_get_hp(target) ) {
 		map->freeblock_lock();
 		status_fix_damage(src, target, damage, ddelay); // We have to separate here between reflect damage and others [icescope]
 		if( attack_type && !status->isdead(target) && additional_effects )
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -5496,6 +5496,13 @@ static int pc_useitem(struct map_session_data *sd, int n)
 	if (nameid == ITEMID_MEGAPHONE)
 		sd->state.using_megaphone = 1;
 
+	// Re-check death: delay_battle_damage can kill the player after clif_parse_UseItem()'s check. (#969)
+	if (pc_isdead(sd) || status->isdead(&sd->bl)) {
+		if (removeItem)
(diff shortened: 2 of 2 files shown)


MrKeiKun:
@kyeme this is the only solution i have in mind. can you try? 😄

kyeme:
I tested the fix, but unfortunately, it's not working.

MrKeiKun:
Thanks for testing this so quickly! I traced through the fix again and the logic checks out on my end (verified via an automated repro: Asura Strike into a Reflect Shield-holding Paladin, spamming Blue Potion with `item_use_interval: 0` — SP stops changing the instant death lands, across repeated runs).

Could you help me narrow down where it diverges from your case?

- Did you rebuild the map-server from this branch (not just pull the diff without recompiling)? Stale binaries are the most common reason a fix like this appears to do nothing.
- Your exact repro steps — same setup (Asura Strike + Reflect Shield), or something else now?
- `delay_battle_damage` and `item_use_interval` values you're testing with
- What you're observing exactly: SP still visibly increases after the death animation/packet, or is it something else (character not actually dying, no visible change at all, etc.)?

Want to make sure I'm chasing the same case you're hitting rather than guessing.

kyeme:
1. I downloaded a clean Hercules source and set it up with pre-renewal.

I manually applied this change in `pc.c`:

```c
// Re-check death: delay_battle_damage can kill the player after clif_parse_UseItem()'s check. (#969)
if (pc_isdead(sd) || status->isdead(&sd->bl)) {
	if (removeItem)
		pc->delitem(sd, n, 1, 1, DELITEM_NORMAL, LOG_TYPE_CONSUME);
	return 0;
}
```

Then I rebuilt and installed it on Windows using:

```text
cmake --build build --parallel
cmake --install build
```

2. My repro is basically the same setup: I use Asura Strike on a Paladin with Reflect Shield while continuously spamming Blue Potions during the cast. If the reflected damage kills me, I check whether my SP still increases from the Blue Potion after death.

3. My settings are:

`delay_battle_damage: true` (default)

`item_use_interval: 0`

4. Here is a video showing the behavior:



https://github.com/user-attachments/assets/78b3550d-c745-4f24-b747-1d5e7a7e1aa5

MrKeiKun:
Got it — "after Asura skill, he was able to use blue potion" is the part I need to pin down more precisely, because that description also matches the *expected* sequence rdw-software described on the original issue (potion lands and restores SP in the brief window before the reflected damage actually kills you, then it correctly stops once you're dead). My fix only targets potions that still take effect *after* the death packet/state has already landed — not the pre-death window.

Could you confirm one specific thing: at the moment SP visibly increases in your test, has the character already fallen over / is the client already showing the death state (lying down, can't move, etc.), or is this happening while the character is still standing/alive, just very close to death?

If it's genuinely after the death state is already showing client-side, that means the packet is still slipping through server-side even with the fix applied, and I'll need to dig further into why the guard isn't catching it in your case. If it's in the window before death, that's the existing (intended) behavior and the actual fix target is somewhere else — possibly making the client itself stop sending item-use packets once locally aware of death, rather than only guarding server-side.

kyeme:
By the way, with `delay_battle_damage: false`, this issue does not occur even before applying the fix in `pc.c`.

The issue only occurs when `delay_battle_damage: true`.

MrKeiKun:
@kyeme try again. after thinking this might be a delay damage thing.

kyeme:
Tested it again, and it’s working now. Thank you!


pic1

PR #3465, opened 2026-09-11
Delay status effects from attacks and skills to match official timing

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

- Adds `status->change_start_delayed()` / `status->change_start_delayed_timer()`, a self-contained timer-backed mechanism to apply a status change after a delay instead of instantly, without changing `status_change_start()`'s existing signature or any of its other call sites.
- Wires this into `skill_additional_effect()` so normal-attack triggered status changes (`bAddEff`-style card procs) take effect after the attacker's attack motion ends (`AttackMotion+500ms`) instead of the instant the damage lands, matching official behavior.
- Extends the same delay to single-hit weapon skills (e.g. Asura Strike), while multi-hit weapon skills (e.g. Sonic Blow, Double Strafe) stay instant since each hit already has its own timing. The distinguishing rule used is the skill's `NumberOfHits` (`skill->get_num()`): `abs(num) <= 1` is treated as single-hit.
- Adds skill-specific delays confirmed by official testing in the issue thread: Frost Joke/Scream (~3000ms after the skill goes off) and Lex Divina/Hammerfall (~1000ms after the skill goes off). Removing an existing Silence by recasting Lex Divina stays instant, matching reported behavior.
- Under `RENEWAL`, the applied delay is subtracted from the status duration.
- Fixes a regression found during review: if a hit breaks an existing Freeze/Stone/Sleep via damage, that same hit's delayed card-proc could reapply the status moments later once the delay elapsed (e.g. hitting a frozen target with a Stormy Knight Card weapon would unfreeze it and then immediately refreeze it). Tracks when an OPT1 status was last broken by damage and skips the delayed reapplication if it was broken by the same attack that scheduled it.

**Issues addressed:** #1141

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -1599,11 +1599,20 @@ static int skill_additional_effect(struct block_list *src, struct block_list *bl
 					flag = SCFLAG_NONE;
 				}
 
+				// On officials, "on attack" status changes from normal attacks only take
+				// effect after the attack animation ends (AttackMotion+500ms), not the
+				// instant the damage is applied. (bugreport:1141)
+				int delay = (skill_id == 0) ? sstatus->amotion + 500 : 0;
+#ifdef RENEWAL
+				if (delay > 0)
+					temp = max(1, temp - delay);
+#endif
+
 				if (sd->addeff[i].flag&ATF_TARGET)
-					status->change_start(src, bl, type, rate, 7, 0, (type == SC_BURNING) ? src->id : 0, 0, temp, flag, skill_id);
+					status->change_start_delayed(src, bl, type, rate, 7, 0, (type == SC_BURNING) ? src->id : 0, 0, temp, flag, skill_id, delay);
 
 				if (sd->addeff[i].flag&ATF_SELF)
-					status->change_start(src, src, type, rate, 7, 0, (type == SC_BURNING) ? src->id : 0, 0, temp, flag, skill_id);
+					status->change_start_delayed(src, src, type, rate, 7, 0, (type == SC_BURNING) ? src->id : 0, 0, temp, flag, skill_id, delay);
 			}
 		}
 
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -1599,10 +1599,14 @@ static int skill_additional_effect(struct block_list *src, struct block_list *bl
 					flag = SCFLAG_NONE;
 				}
 
-				// On officials, "on attack" status changes from normal attacks only take
-				// effect after the attack animation ends (AttackMotion+500ms), not the
-				// instant the damage is applied. (bugreport:1141)
-				int delay = (skill_id == 0) ? sstatus->amotion + 500 : 0;
+				// On officials, "on attack" status changes from normal attacks and single-hit
+				// weapon skills (e.g. Asura Strike) only take effect after the attack
+				// animation ends (AttackMotion+500ms), not the instant the damage is applied.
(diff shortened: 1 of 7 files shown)


MrKeiKun:
@kyeme this one

kyeme:
Thank you for this fix. I tested it and noticed one problem.

If the target is already frozen and I attack it using a knife with four Stormy Knight Cards, the target will unfreeze from the attack, which is normal. However, it immediately gets frozen again.

Video:

https://github.com/user-attachments/assets/ecddd271-ef38-4412-a6bd-b7ae7a321d29

kyeme:
> Thank you for this fix. I tested it and noticed one problem.
> 
> If the target is already frozen and I attack it using a knife with four Stormy Knight Cards, the target will unfreeze from the attack, which is normal. However, it immediately gets frozen again.
> 
> Video:
> 
>  9-11-2026.2-33-32.PM.mp4

I no longer have a way to test this scenario because I don't have access to an Aegis test server. However, as far as I remember, once your Freeze status is broken by an attack, you should become unfrozen and should not be frozen again immediately.

MrKeiKun:
@kyeme try now

kyeme:
It's working now. Thank you!

kyeme:
I saw that you also fixed the Frost Joke and Dancer status delay. Could you also mimic the official behavior for how the status delay is applied while the character/monster is walking? Thank you.

Video: 
Using Frost Joke (rathena https://github.com/rathena/rathena/issues/948 )
https://github.com/user-attachments/assets/da4b2fdb-3b53-4e8e-8248-af18857b09be

Official: 
Using Weapon with Stormy Knight Card
https://youtu.be/WJQSwdSANoI?t=30

MrKeiKun:
> I saw that you also fixed the Frost Joke and Dancer status delay. Could you also mimic the official behavior for how the status delay is applied while the character/monster is walking? Thank you.
> 
> Video: Using Frost Joke (rathena [rathena/rathena#948](https://github.com/rathena/rathena/issues/948) ) https://github.com/user-attachments/assets/da4b2fdb-3b53-4e8e-8248-af18857b09be
> 
> Official: Using Weapon with Stormy Knight Card https://youtu.be/WJQSwdSANoI?t=30

@kyeme Thanks for testing! Looking into the walk-interruption behavior now.

One thing I want to confirm before changing it: rAthena's current code makes Freeze/Stun/Sleep/Stone skip the walk-stop for *any* source (damage or non-damage) — it only stops walking if you're currently dancing. Your two linked videos compare Frost Joke (a non-damage skill) against Stormy Knight Card (a weapon/damage proc) — do you mean both should let the target keep walking to their queued destination, or should weapon-card-triggered Freeze still snap/stop movement the way damage-dealing hits normally do, and only skill-based Freeze (Frost Joke/Scream) should let you keep walking?

kyeme:
Based on my test video, both cases should allow the target to continue walking to the already queued destination.

For example, when my Champion attacks the Marina using a weapon with Stormy Knight Card, the Marina can still continue moving to its destination even after Freeze is applied.

The same behavior happens with Frost Joke: if the target is already walking when Freeze is applied, it should continue to the queued destination instead of stopping immediately.

MrKeiKun:
@kyeme Pushed a fix where Freeze/Stun/Sleep/Stone no longer interrupt movement (unless you're dancing, in which case it still stops), matching what you described for both the weapon-card and Frost Joke cases. Let me know if it looks right on your end.

kyeme:
It's working now thank you!

Issue #1273, opened 2016-04-29
Asura Strike vs Dead Target


kyeme:
When you cast Asura Strike and your target dies during the cast time, you still lose all sphere, fury mode and SP.

On rAthena the removal of Fury mode and drain of SP only happens if the target is still alive.

Credits: Playtester

memoryss1:
seems not only asura but all skills perform in the same way

MrKeiKun:
```
		//When Asura fails... (except when it fails from Fog of Wall)
		//Consume SP/spheres
		skill->consume_requirement(sd,ud->skill_id, ud->skill_lv,1);
		status->set_sp(src, 0, STATUS_HEAL_DEFAULT); 
		sc = &sd->sc;
		if (sc->count) {
			//End states
			status_change_end(src, SC_EXPLOSIONSPIRITS, INVALID_TIMER); 
```
Is this only on renewal?

kyeme:
This applies to both Renewal and Pre-Renewal.

Also, as @memoryss1 mentioned, this behavior applies to other skills as well. I'm referring specifically to cases where the target dies before your skill finishes casting.

MrKeiKun:
@kyeme can you try...

```c
	if (ud->skill_id == MO_EXTREMITYFIST && sd && !(sc && sc->data[SC_FOGWALL])) {
		//When Asura fails... (except when it fails from Fog of Wall)
```

to

```c
	if (ud->skill_id == MO_EXTREMITYFIST && sd != NULL && !(sc != NULL && sc->data[SC_FOGWALL] != NULL)
	 && (target == NULL || status->isdead(target) == 0)) {
		//When Asura fails... (except when it fails from Fog of Wall or the target died mid-cast)

```

this should work based on the target died mid-cast 😆 

> This applies to both Renewal and Pre-Renewal.
> 
> Also, as [@memoryss1](https://github.com/memoryss1) mentioned, this behavior applies to other skills as well. I'm referring specifically to cases where the target dies before your skill finishes casting.

other skills?

kyeme:
I tried this code, but unfortunately, it's not working.

PR #3482, opened 2026-09-16
Fix Asura Strike not draining SP and Fury Mode when target dies mid-cast

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

When a **player** target dies while Asura Strike is casting, the caster's spirit spheres are consumed, but their SP is not drained to 0 and Fury Mode (`SC_EXPLOSIONSPIRITS`) is not removed. Official behaviour is to consume everything as if the skill had been cast successfully.

The cause is that a dead player target slips between both of Asura's cost-handling paths:

- `pc_dead()` returns only `1` or `1|8`, never `&2`, so a dead player is **not** removed from the map (`status_damage()`). `target->prev` stays non-`NULL`.
- `battle_check_target()` has no death check, and `status_check_skilluse()` deliberately skips the dead check for skills ("Dead state is not checked for skills as some skills can be used on dead characters").
- So `skill_castend_id()` never breaks out to the "Skill failed" block, which is where the fail-path drain lives.
- Execution reaches `skill_consume_requirement()` (taking `SpiritSphereCost: 5` / `SPCost: 1`), then `skill_castend_damage_id()` returns early on `status->isdead(bl)` — before the switch case that performs `status->set_sp(src, 0, ...)` and ends `SC_EXPLOSIONSPIRITS`.

Mob targets were unaffected, since `mob_dead()` returns `3` (`&2`), removing the mob from the map so the existing fail-path drain is reached — which is likely why this went unnoticed.

**Issues addressed:** #1273


kyeme:
@MrKeiKun Thanks for looking into this issue. I think there may be a misunderstanding about the expected behavior. 

On the official server, if the target dies before Asura Strike finishes casting, the cast still completes and all requirements are consumed, including SP, Spirit Spheres, and Fury Mode, as if Asura Strike was successfully cast.

So the resource consumption itself is actually the expected behavior, even though the target died before the cast finished.

My original issue was about the behavior when the target dies before the skill finishes casting, but I did not mean that Asura Strike's requirements should not be consumed.

MrKeiKun:
> @MrKeiKun Thanks for looking into this issue. I think there may be a misunderstanding about the expected behavior.
> 
> On the official server, if the target dies before Asura Strike finishes casting, the cast still completes and all requirements are consumed, including SP, Spirit Spheres, and Fury Mode, as if Asura Strike was successfully cast.
> 
> So the resource consumption itself is actually the expected behavior, even though the target died before the cast finished.
> 
> My original issue was about the behavior when the target dies before the skill finishes casting, but I did not mean that Asura Strike's requirements should not be consumed.

@kyeme Thanks for the clarification, but this leaves me a bit confused — the original issue text (credited to Playtester) says:

> On rAthena/hercules the removal of Fury mode and drain of SP only happens if the target is still alive.

That reads as: consumption should be *skipped* when the target is dead, which is what this PR does. Your comment here says the opposite — that consumption is expected/correct regardless of target death, and that's not what you were originally reporting.

Could you clarify what the actual observed-vs-expected difference is? Specifically:
- What do you see happening on Hercules right now that's wrong?
- What does official do differently in that same scenario (if not resource consumption)?

Without a concrete before/after description I don't have a reliable target to fix against, and I'd rather not guess further.

kyeme:
**Current Hercules behavior:**
If the target dies before Asura Strike finishes casting → SP, Spirit Spheres, and Fury Mode are not consumed. [Bug]

**Official behavior:**
If the target dies before Asura Strike finishes casting → All SP, Spirit Spheres, and Fury Mode are still consumed, as if Asura Strike was successfully cast.

So, consuming the requirements when the target dies before the cast finishes is actually the expected official behavior. The issue I reported was that Hercules does not consume them in this situation.

MrKeiKun:
> **Current Hercules behavior:** If the target dies before Asura Strike finishes casting → SP, Spirit Spheres, and Fury Mode are not consumed. [Bug]
> 
> **Official behavior:** If the target dies before Asura Strike finishes casting → All SP, Spirit Spheres, and Fury Mode are still consumed, as if Asura Strike was successfully cast.
> 
> So, consuming the requirements when the target dies before the cast finishes is actually the expected official behavior. The issue I reported was that Hercules does not consume them in this situation.

@kyeme Found it. The bug only shows up when the target is a **player**  a dead player is never removed from the map (`pc_dead()` doesn't set the removal flag), so the code that normally drains SP and removes Fury Mode on a failed/interrupted Asura never runs, even though the spheres do get consumed. That's why it looked like "nothing was consumed"  the spheres go, but SP and Fury Mode were left behind.

Pushed a fix that drains SP and removes Fury Mode in that case too, matching official behavior. Could you test again on this branch when you get a chance?

kyeme:
Yes, it's working now. All SP, Vigor and Sphere are consumed correctly.

However, there is still one behavior that hasn't been fixed: the movement after casting Asura Strike. The character does not slide toward the target as it does on the official server.

MrKeiKun:
> Yes, it's working now. All SP, Vigor and Sphere are consumed correctly.
> 
> However, there is still one behavior that hasn't been fixed: the movement after casting Asura Strike. The character does not slide toward the target as it does on the official server.

@kyeme maybe in another PR since #1273 aint up for it. also can you create issue regarding this if possible give more details, like is this pre-re or re too?

kyeme:
It's the same for both Pre-Renewal and Renewal. I'll create a separate report for this. Thanks!

Issue #3479, opened 2026-09-16
Dispel can affect guildmates outside party on non-PvP maps


kyeme:
**Describe the bug**

Dispel can currently be used on guildmates in a non-PvP environment, even when the guildmate is not in the same party.

On official behavior, being in the same guild alone should not make a player a valid target for Dispel on a non-PvP map.

**Current behavior**
On a non-PvP map:
A guildmate can be affected by Dispel even if they are not in the same party.

**Expected behavior:**
Dispel should follow these targeting rules:

1. PvP/GvG: Dispel can be used against enemy players.
2. Party members: Dispel can be used on party members on any map, including non-PvP maps.
3. Guildmates: Being in the same guild alone should not allow Dispel. You cannot Dispel a guildmate who is not in the same party, even on WoE maps. If the guildmate is also in the same party, Dispel should work because they are a party member.
4. Duel: Dispel should work against the other player while participating in a Duel.

kyeme:
Dispel guildmates official video: https://www.youtube.com/watch?v=8leyc7ypog8

PR #3483, opened 2026-09-16
Fix SA_DISPELL affecting non-party members outside PvP and GvG/WoE

MrKeiKun:
### Pull Request Prelude

### Changes Proposed

Commit 63cc34c meant to let `SA_DISPELL` work between duelists but used `&&` instead of `||` on the `duel_group` check, so the "no dispel on non-party members outside PvP" guard almost never triggered. This let `SA_DISPELL` hit any non-party player, including guildmates, on non-PvP maps.

Additionally, on GvG/WoE maps the guard was skipped entirely (since those maps count as PvP-enabled), so players could Dispel guildmates who weren't in their party.

Fixed both cases: outside PvP, dispel again requires being in the same party (unless both players are in the same duel); on GvG/WoE maps, dispel is also blocked against guildmates unless they're in the same  party.

**Issues addressed:** #3479


kyeme:
Thanks for the fix. Could you also fix the behavior on GvG maps so that Dispel cannot be used on guildmates who are not in the same party?

Currently, players can Dispel their guildmates on GvG/WoE maps even when they are not in the same party.

**Official behavior for guildmates:**

```
Not in the same party: Cannot Dispel
In the same party: Can Dispel
```

kyeme:
Thanks! It's working now. <3

Issue #3485, opened 2026-09-17
Asura Strike does not slide toward the target if the target dies during casting


kyeme:
### Server Mode

Pre-Renewal, Renewal

### Result

When casting **Asura Strike (`MO_EXTREMITYFIST`)**, if the target dies before the cast finishes, the caster does not slide/move toward the target's position.

### Expected Result

On the official server, even if the target dies before Asura Strike finishes casting, the caster still slides toward the target's position as if the skill had been executed.

The Asura Strike requirements are also consumed (SP, Spirit Spheres, and Vigor), and the caster still performs the movement associated with the skill.

This should not behave the same way as when Asura Strike fails because the target is inside Wall of Fog. The target dying during the cast has different behavior: the caster should still perform the Asura Strike slide/movement toward the target's position.

### How to Reproduce

1. Use Asura Strike on a target.
2. While Asura Strike is still being cast, have the target die from another source.
3. Allow the Asura Strike cast to finish.
4. Notice that the caster remains in the same position instead of sliding toward the target.

### Official Behavior

If the target dies while Asura Strike is being cast, the skill **still behaves** as an executed Asura Strike after the cast completes:

* The skill requirements are consumed.
* The caster slides toward the target's position even though the target is already dead.
This behavior is different from an Asura Strike failure caused by Wall of Fog.

Related: https://github.com/HerculesWS/Hercules/issues/1273

PR #3486, opened 2026-09-17
Fix Asura Strike not sliding toward target when it dies mid-cast 

MrKeiKun:
### Pull Request Prelude

### Changes Proposed
When casting Asura Strike (`MO_EXTREMITYFIST`), if the target dies before the cast finishes, the caster does not slide toward the target's position, even though official servers still perform this slide.

`skill_castend_id`'s success path only reaches the caster's "skill failed" slide logic when the target is fully removed from the map (`target->prev == NULL`). If the target is dead but still present (e.g. corpse not yet cleaned up), `skill_castend_damage_id` returns early on `status->isdead(bl)` and the normal success path is taken instead, which never performs the slide.

This adds the same slide/movement handling (`unit->move_pos` + `clif->slide` + spiritball refresh) to that early-return branch for `MO_EXTREMITYFIST`, so the caster still slides toward the target's last position when the target died mid-cast.

**Issues addressed:** #3485


kyeme:
Tested and working! Thanks!

PR #3487, opened 2026-09-17
Implement code changes required to compile Hercules as C++

hemagx:
### Pull Request Prelude

### Changes Proposed

Implementing bare-minimum changes to compile hercules code as C++, the changes proposed are mostly fixes to warnings produced by C++ rules, type fixes and type casts and replacement of C features/compile features with their C++ counterpart.

**Issues addressed:** 


src/map/script.c
```diff
@@ -19873,7 +19866,7 @@ static BUILDIN(strtol)
 	int base = script_getnum(st, 3);
 	long value = strtol(string, NULL, base);
 #if LONG_MAX > INT_MAX || LONG_MIN < INT_MIN
-	value = cap_value(value, INT_MIN, INT_MAX);
+	value = std::clamp(value, (long)INT_MIN, (long)INT_MAX);
```
MrKeiKun, line 19869: `std::clamp` is undefined behavior when min > max.  The old `cap_value()` handled it. These values come straight from NPC scripts, and `cap_value` is a documented script command, so any third-party script passing a backwards range is affected. Can we guard the args here?

src/map/status.c
```diff
@@ -1298,14 +1299,14 @@ static int status_calc_pet_(struct pet_data *pd, enum e_status_calc_opt opt)
...
-			pstatus->dex = cap_value(pstatus->dex,1,battle_config.pet_max_stats);
-			pstatus->luk = cap_value(pstatus->luk,1,battle_config.pet_max_stats);
+			pstatus->rhw.atk = std::clamp(pstatus->rhw.atk, 1u, (unsigned int)battle_config.pet_max_atk1);
+			pstatus->rhw.atk2 = std::clamp(pstatus->rhw.atk2, 2u, (unsigned int)battle_config.pet_max_atk2);
+			pstatus->str = std::clamp((int)pstatus->str,1,battle_config.pet_max_stats);
+			pstatus->agi = std::clamp((int)pstatus->agi,1,battle_config.pet_max_stats);
+			pstatus->vit = std::clamp((int)pstatus->vit,1,battle_config.pet_max_stats);
+			pstatus->int_= std::clamp((int)pstatus->int_,1,battle_config.pet_max_stats);
+			pstatus->dex = std::clamp((int)pstatus->dex,1,battle_config.pet_max_stats);
+			pstatus->luk = std::clamp((int)pstatus->luk,1,battle_config.pet_max_stats);
```
MrKeiKun, line 1309: `pet_max_stats` is allowed to be 0 based on:

`{ "pet_max_stats", &battle_config.pet_max_stats, 99, 0, INT_MAX },`

src/map/status.c
```diff
@@ -3836,7 +3834,7 @@ static int status_base_atk(const struct block_list *bl, const struct status_data
 	if (bl->type == BL_PC)
 		str += dex / 5 + st->luk / 5;
 #endif // RENEWAL
-	return cap_value(str, battle_config.batk_min, battle_config.batk_max);
+	return std::clamp(str, battle_config.batk_min, battle_config.batk_max);
```
MrKeiKun, line 3837: Both bounds are admin-editable and only validated against their own ranges, though nothing checks min < max. `min_hitrate` has such a guard in battle.c, but none of the `*_limit` pairs do. Setting `batk_min_limit` above `batk_max_limit` is undefined behavior on every attack. Repeats for matk/watk/critical/hit/flee/flee2, ~40 spots in this file.

src/map/skill.c
```diff
@@ -10709,14 +10726,14 @@ static int skill_castend_nodamage_id(struct block_list *src, struct block_list *
 				// We then reduce the success chance based on the target's build.
 				rate -= rnd->value( tstatus->agi / 6, tstatus->agi / 3 ) + tstatus->luk / 10 + ( dstsd ? (dstsd->max_weight / 10 - dstsd->weight / 10 ) / 100 : 0 ) + status->get_lv(bl) / 10;
 				//Finally we set the minimum success chance cap based on the caster's skill level and DEX.
-				rate = cap_value( rate, skill_lv + sstatus->dex / 20, 100);
+				rate = std::clamp(rate, skill_lv + sstatus->dex / 20, 100);
```
MrKeiKun, line 10728: Minimum is computed from caster stats, maximum is a fixed 100... High DEX puts the minimum above the maximum. Used to just return 100. Capping the lower bound would fix it.

src/map/skill.c
```diff
@@ -3052,7 +3056,7 @@ static int skill_strip_equip(struct block_list *bl, unsigned short where, int ra
 	struct status_change *sc;
 	const int pos[5]             = {EQP_WEAPON, EQP_SHIELD, EQP_ARMOR, EQP_HELM, EQP_ACC};
 	const enum sc_type sc_atk[5] = {SC_NOEQUIPWEAPON, SC_NOEQUIPSHIELD, SC_NOEQUIPARMOR, SC_NOEQUIPHELM, SC__STRIPACCESSARY};
-	const enum sc_type sc_def[5] = {SC_PROTECTWEAPON, SC_PROTECTSHIELD, SC_PROTECTARMOR, SC_PROTECTHELM, 0};
+	const enum sc_type sc_def[5] = {SC_PROTECTWEAPON, SC_PROTECTSHIELD, SC_PROTECTARMOR, SC_PROTECTHELM, SC_NONE};
```
MrKeiKun, line 3058: `SC_NONE` is -1, and 0 is `SC_STONE`. The old code meant petrified targets couldn't have accessories stripped. Fix looks right, but it changes PvP behavior, so it probably needs a changelog entry.

MrKeiKun (commented):
Nicesu @MishimaHaruna @hemagx 

 `cap_value()` vs `std::clamp()` isn't equivalent. The old macro coped with a backwards min/max; `std::clamp` treats it as undefined behavior.

src/map/script.c
hemagx: That is actually untrue, the old macro also has the same UB as std::clamp, regardless it's a good point to check those while we're at it anyway.

MrKeiKun: as discussed, concern will be tackled in another MR but not this one ❤️

src/map/skill.c
MrKeiKun: old code was wrong, so `SC_NONE` was really the intention 👍🏻

MishimaHaruna (approved):

src/map/script.cpp
```diff
@@ -19873,7 +19866,7 @@ static BUILDIN(strtol)
 	int base = script_getnum(st, 3);
 	long value = strtol(string, NULL, base);
 #if LONG_MAX > INT_MAX || LONG_MIN < INT_MIN
-	value = cap_value(value, INT_MIN, INT_MAX);
+	value = std::clamp(value, (long)INT_MIN, (long)INT_MAX);
```
MishimaHaruna: We decided to address these in this PR even though the behavior has always been undefined, just in case the new UB doesn't match the old UB.

Most of these are extreme edge cases that will likely never happen in practice (i.e. configuration parameters that would make no sense to set), but there were a few (especially in scripts) that could happen with poorly written scripts.

src/map/skill.cpp
```diff
@@ -3052,7 +3056,7 @@ static int skill_strip_equip(struct block_list *bl, unsigned short where, int ra
 	struct status_change *sc;
 	const int pos[5]             = {EQP_WEAPON, EQP_SHIELD, EQP_ARMOR, EQP_HELM, EQP_ACC};
 	const enum sc_type sc_atk[5] = {SC_NOEQUIPWEAPON, SC_NOEQUIPSHIELD, SC_NOEQUIPARMOR, SC_NOEQUIPHELM, SC__STRIPACCESSARY};
-	const enum sc_type sc_def[5] = {SC_PROTECTWEAPON, SC_PROTECTSHIELD, SC_PROTECTARMOR, SC_PROTECTHELM, 0};
+	const enum sc_type sc_def[5] = {SC_PROTECTWEAPON, SC_PROTECTSHIELD, SC_PROTECTARMOR, SC_PROTECTHELM, SC_NONE};
```
MishimaHaruna: Yep, this was changed because once seen we couldn't unsee and ignore

src/map/clif.c
```diff
@@ -23262,7 +23260,7 @@ static void clif_rodex_add_item_result(struct map_session_data *sd, int16 idx, i
 	packet->index = idx + 2;
 	packet->count = amount;
 	packet->itemId = sd->status.inventory[idx].nameid;
-	packet->type = itemtype(sd->inventory_data[idx]->type);
+	packet->type = itemtype((enum item_types)sd->inventory_data[idx]->type);
```
guilherme-gm, line 23263: (probably I am not seeing something)
Wouldn't it be better for `sd->inventory_data[idx]->type` to become `enum item_types` instead of casting here? or is it too hard to switch this now?

hemagx: it would incur more code changes that we'd like for this PR

guilherme-gm (approved):

src/map/clif.c
guilherme-gm: ah, I see. Thanks!
Initially I thought "it probably will save casts", but I believe this will shift the need to cast to everywhere int is used. Makes sense to leave this for another moment.

Issue #3490, opened 2026-09-17
Asura Strike cannot be used immediately when the target is within 3 cells


kyeme:
### Server Mode

Pre-Renewal, Renewal

### Result

With `MO_EXPLOSIONSPIRITS` (Fury) already active, cast `CH_SOULCOLLECT` (Zen) and attempt to use Asura Strike on a target while Zen is still being cast.

The issue occurs when the caster is 3 cells or less from the target.

When Asura Strike is selected/queued while CH_SOULCOLLECT (Zen) is still being cast, Asura Strike fails to execute after Zen finishes if the distance between the caster and the target is 3 cells or less.

Asura Strike does not execute immediately after Zen finishes. Instead, "Skill has failed." appears in the chat box.

This seems to be specific to Asura Strike. If `MO_FINGEROFFENSIVE` (Finger Offensive), or another attack skill, is selected during the Zen cast, the skill can execute immediately after Zen finishes.

### Expected Result
On Official Asura Strike should be able to be selected/queued while Zen or  MO_CALLSPIRITS is being cast and execute immediately after Zen finishes, similar to Finger Offensive and other skills.

### How to Reproduce

1. Use `CH_SOULCOLLECT` (Zen).
2. While Zen is still being cast, target an enemy with `MO_EXTREMITYFIST` (Asura Strike).
3. After Zen finishes, Asura Strike does not execute and **"Skill has failed."** appears in the chat box.
4. Repeat the test, but use `MO_FINGEROFFENSIVE` (Finger Offensive) during the Zen cast.
5. Notice that Finger Offensive can execute immediately after Zen finishes.

Herc Video:

https://github.com/user-attachments/assets/75a7eece-bfd2-47f5-9c67-e4325a0d5278

Streusel:
Untested, but try this
```patch
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -15913,17 +15913,27 @@
 			if (sc && sc->data[SC_COMBOATTACK]) {
 				switch(sc->data[SC_COMBOATTACK]->val1) {
 					case MO_COMBOFINISH:
 					case CH_TIGERFIST:
 					case CH_CHAINCRUSH:
 						break;
 					default:
 						return 0;
 				}
-			} else if (!unit->can_move(&sd->bl)) {
-				//Placed here as ST_MOVE_ENABLE should not apply if rooted or on a combo. [Skotlex]
[...]

PR #3491, opened 2026-09-18
Fix Asura Strike being blocked by leftover walk delay

Streusel:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 

### Changes Proposed

Asura Strike (`MO_EXTREMITYFIST`) could not be cast right after another skill when the target was already within range.

`skill_check_condition_castbegin()` gates `MO_EXTREMITYFIST` behind `unit->can_move()`. That helper returns false for two very different reasons:

1. the character is genuinely immobilized, for example by Ankle Snare, Spider Web, Basilica, White Imprison, `opt1` states, sitting, vending or casting;
2. `ud.canmove_tick` is simply still in the future, because a previously used skill applied a walk delay.

Only the first case should prevent Asura Strike from being used. The second is a transient post-skill delay and is unrelated to the `ST_MOVE_ENABLE` requirement this check exists to enforce.

This is also why the problem only appeared at close range. Asura Strike uses `Range: -2` (weapon range), so when the target is within ~3 cells the cast begins immediately, while the walk delay left over by the preceding skill (typically `CH_SOULCOLLECT`) is still ticking, and the skill fails with `USESKILL_FAIL_LEVEL`. When the target is farther away the character first walks towards it, the walk delay expires on the way, and the skill succeeds. That made the bug look range-dependent when it is really a timing issue.

The condition is now evaluated with `ud.canmove_tick` temporarily set to the current tick and restored immediately afterwards. Every real immobilization check inside `unit_can_move()` still applies; only the leftover walk delay is ignored.

**Issues addressed:** #3490


hemagx (changes requested):
This workaround feels too hacky for me tbh, consider `unit->move` is only called in a few places, and may benefit in the future from such an argument anyway, please instead create a `skill_id` argument instead and handle the exception within `ud->can_move` itself.

Streusel:
Alternatively, how about adding a `unit_can_move_ignoring_walkdelay` on top of `unit_can_move`. Less of a hack, and allows for other skills to make use of it.
```diff
diff --git a/src/map/skill.c b/src/map/skill.c
index db3e3285e..fae1256ed 100644
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -15919,8 +15919,10 @@ static int skill_check_condition_castbegin(struct map_session_data *sd, uint16 s
 					default:
 						return 0;
 				}
-			} else if (!unit->can_move(&sd->bl)) {
+			} else if (unit->can_move_ignoring_walkdelay(&sd->bl) == 0) {
 				//Placed here as ST_MOVE_ENABLE should not apply if rooted or on a combo. [Skotlex]
+				//Only actual immobilization may block Asura Strike, so the walk delay left over
+				//by a previously used skill (i.e. CH_SOULCOLLECT) is ignored on purpose.
 				clif->skill_fail(sd, skill_id, USESKILL_FAIL_LEVEL, 0, 0);
 				return 0;
 			}
diff --git a/src/map/unit.c b/src/map/unit.c
index 04c818792..6f6d0d9fd 100644
--- a/src/map/unit.c
+++ b/src/map/unit.c
@@ -1211,9 +1211,12 @@ static int unit_is_walking(struct block_list *bl)
 }
 
 /*==========================================
- * Determines if the bl can move based on status changes. [Skotlex]
+ * Determines if the bl can move based on status changes, disregarding the
+ * walk delay left over by a previously used skill. [Skotlex]
+ * Skills which may be used while such a delay is still running, but not while
+ * the unit is actually immobilized, check this instead of unit->can_move().
  *------------------------------------------*/
-static int unit_can_move(struct block_list *bl)
+static int unit_can_move_ignoring_walkdelay(struct block_list *bl)
 {
 	struct map_session_data *sd;
 	struct unit_data *ud;
@@ -1240,9 +1243,6 @@ static int unit_can_move(struct block_list *bl)
 		}
 	}
 
-	if (DIFF_TICK(ud->canmove_tick, timer->gettick()) > 0)
-		return 0;
-
 	if (sd && (
 		pc_issit(sd) ||
 		sd->state.vending ||
@@ -1322,6 +1322,25 @@ static int unit_can_move(struct block_list *bl)
 	return 1;
 }
 
+/*==========================================
+ * Determines if the bl can move based on status changes. [Skotlex]
+ *------------------------------------------*/
+static int unit_can_move(struct block_list *bl)
+{
+	struct unit_data *ud;
+
+	nullpo_ret(bl);
+	ud = unit->bl2ud(bl);
+
+	if (ud == NULL)
+		return 0;
+
+	if (DIFF_TICK(ud->canmove_tick, timer->gettick()) > 0)
+		return 0;
+
+	return unit->can_move_ignoring_walkdelay(bl);
+}
+
 /*==========================================
  * Resume running after a walk delay
  *------------------------------------------*/
@@ -3260,6 +3279,7 @@ void unit_defaults(void)
 	unit->steptimer = unit_steptimer;
 	unit->stop_stepaction = unit_stop_stepaction;
 	unit->is_walking = unit_is_walking;
+	unit->can_move_ignoring_walkdelay = unit_can_move_ignoring_walkdelay;
 	unit->can_move = unit_can_move;
 	unit->resume_running = unit_resume_running;
 	unit->set_walkdelay_timer = unit_set_walkdelay_timer;
diff --git a/src/map/unit.h b/src/map/unit.h
index b04cac238..d3c984b08 100644
--- a/src/map/unit.h
+++ b/src/map/unit.h
@@ -127,6 +127,7 @@ struct unit_interface {
 	int (*steptimer) (int tid, int64 tick, int id, intptr_t data);
 	void (*stop_stepaction) (struct block_list *bl);
 	int (*is_walking) (struct block_list *bl);
+	int (*can_move_ignoring_walkdelay) (struct block_list *bl);
 	int (*can_move) (struct block_list *bl);
 	int (*resume_running) (int tid, int64 tick, int id, intptr_t data);
 	int (*set_walkdelay_timer) (int tid, int64 tick, int id, intptr_t data);
```
But if you still want the `skill_id` approach instead, then I'll do that instead and update the PR.

hemagx:
@Streusel The problem with this approach is are we going to make a new function for every new exception we find? and when this happen now we will have to maintain different versions with high chance of problems to arise in the future, Ideally we want our code to be as flexible as possible, with least amount of repetition and specific code when possible, as an emulator we're always finding new behaviors, mistakes etc so writing code that is simple flexible and with single responsibility makes everything much simpler on the long run.

Streusel:
This would be another option, passing the skill_id like you said but expanding skill db itself as well. Allowing for the walk-ability to be determined from skill db, rather than in source.
This removes the exception for each individual skill via ifs or switches, but does add 1 extra skill line, though it defaults to false, I prefer this option myself.
```diff
diff --git a/db/pre-re/skill_db.conf b/db/pre-re/skill_db.conf
index c28595586..827a751d8 100644
--- a/db/pre-re/skill_db.conf
+++ b/db/pre-re/skill_db.conf
@@ -87,6 +87,8 @@
 		RangeModByRadius: true/false            (boolean, defaults to false)
 		RangeModByResearchTrap: true/false      (boolean, defaults to false)
 		AllowPlagiarism: true/false             (boolean, defaults to false)
+		IgnoreWalkDelay: true/false             (boolean, defaults to false)
+							Usable while the walk delay of a previously used skill is still running.
 	}
 	AttackType: "Attack Type"                   (string, defaults to "None") (can be grouped by Levels)
 	                                            Types: "None", "Weapon", "Magic" or "Misc"
@@ -8043,6 +8045,7 @@ skill_db: (
 	SkillInfo: {
 		AllowReproduce: true
 		IsCombo: true
+		IgnoreWalkDelay: true
 	}
 	AttackType: "Weapon"
 	DamageType: {
diff --git a/db/re/skill_db.conf b/db/re/skill_db.conf
index 9c834044c..b769c45cc 100644
--- a/db/re/skill_db.conf
+++ b/db/re/skill_db.conf
@@ -87,6 +87,8 @@
 		RangeModByRadius: true/false            (boolean, defaults to false)
 		RangeModByResearchTrap: true/false      (boolean, defaults to false)
 		AllowPlagiarism: true/false             (boolean, defaults to false)
+		IgnoreWalkDelay: true/false             (boolean, defaults to false)
+							Usable while the walk delay of a previously used skill is still running.
 	}
 	AttackType: "Attack Type"                   (string, defaults to "None") (can be grouped by Levels)
 	                                            Types: "None", "Weapon", "Magic" or "Misc"
@@ -8247,6 +8249,7 @@ skill_db: (
 	SkillInfo: {
 		AllowReproduce: true
 		IsCombo: true
+		IgnoreWalkDelay: true
 	}
 	AttackType: "Weapon"
 	DamageType: {
diff --git a/src/map/mob.c b/src/map/mob.c
index bc1b00cb7..4a8e503c2 100644
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -1530,7 +1530,7 @@ static int mob_ai_sub_hard_slavemob(struct mob_data *md, int64 tick)
 
 		// Approach master if within view range, chase back to Master's area also if standing on top of the master.
 		if( (md->master_dist>MOB_SLAVEDISTANCE || md->master_dist == 0)
-		 && unit->can_move(&md->bl)
+		 && unit->can_move(&md->bl, 0)
 		) {
 			short x = bl->x, y = bl->y;
 			mob_stop_attack(md);
@@ -1645,7 +1645,7 @@ static int mob_randomwalk(struct mob_data *md, int64 tick)
 	nullpo_ret(md);
 
 	if(DIFF_TICK(md->next_walktime,tick)>0 ||
-	   !unit->can_move(&md->bl) ||
+	   !unit->can_move(&md->bl, 0) ||
 	   !(status_get_mode(&md->bl)&MD_CANMOVE))
 		return 0;
 
@@ -1746,7 +1746,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, int64 tick)
 		view_range = md->db->range2;
 	mode = status_get_mode(&md->bl);
 
-	can_move = (mode&MD_CANMOVE)&&unit->can_move(&md->bl);
+	can_move = (mode&MD_CANMOVE)&&unit->can_move(&md->bl, 0);
 
 	if (md->target_id) {
 		//Check validity of current target. [Skotlex]
@@ -2074,7 +2074,7 @@ static int mob_ai_sub_lazy(struct mob_data *md, va_list args)
 		return 0;
 	}
 
-	if( DIFF_TICK(md->next_walktime,tick) < 0 && (status_get_mode(&md->bl)&MD_CANMOVE) && unit->can_move(&md->bl) ) {
+	if( DIFF_TICK(md->next_walktime,tick) < 0 && (status_get_mode(&md->bl)&MD_CANMOVE) && unit->can_move(&md->bl, 0) ) {
 		if( rnd()%1000 < MOB_LAZYMOVEPERC(md) )
 			mob->randomwalk(md, tick);
 	}
diff --git a/src/map/pet.c b/src/map/pet.c
index c10272514..32c5f1850 100644
--- a/src/map/pet.c
+++ b/src/map/pet.c
@@ -1039,7 +1039,7 @@ static int pet_randomwalk(struct pet_data *pd, int64 tick)
 	nullpo_ret(pd);
 	Assert_ret(pd->msd == 0 || pd->msd->pd == pd);
 
-	if (DIFF_TICK(pd->next_walktime,tick) < 0 && unit->can_move(&pd->bl)) {
+	if (DIFF_TICK(pd->next_walktime,tick) < 0 && unit->can_move(&pd->bl, 0)) {
 		const int retrycount=20;
 		int i,c,d=12-pd->move_fail_count;
 		if (d < 5)
diff --git a/src/map/skill.c b/src/map/skill.c
index db3e3285e..aefa566ae 100644
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -4324,7 +4324,7 @@ static int skill_check_condition_mercenary(struct block_list *bl, int skill_id,
 	if( !type )
 		switch( state ) {
 			case ST_MOVE_ENABLE:
-				if( !unit->can_move(bl) ) {
+				if (!unit->can_move(bl, skill_id)) {
 					clif->skill_fail(sd, skill_id, USESKILL_FAIL_LEVEL, 0, 0);
 					return 0;
 				}
@@ -15376,7 +15376,7 @@ static int skill_check_condition_char_sub(struct block_list *bl, va_list ap)
 			return 1;
 		default: //Warning: Assuming Ensemble Dance/Songs for code speed. [Skotlex]
 			{
-				if(pc_issit(tsd) || !unit->can_move(&tsd->bl))
+				if(pc_issit(tsd) || !unit->can_move(&tsd->bl, 0))
 					return 0;
 
 				uint16 skill_lv = pc->checkskill(tsd, skill_id);
@@ -15919,7 +15919,7 @@ static int skill_check_condition_castbegin(struct map_session_data *sd, uint16 s
 					default:
 						return 0;
 				}
-			} else if (!unit->can_move(&sd->bl)) {
+			} else if (!unit->can_move(&sd->bl, skill_id)) {
 				//Placed here as ST_MOVE_ENABLE should not apply if rooted or on a combo. [Skotlex]
 				clif->skill_fail(sd, skill_id, USESKILL_FAIL_LEVEL, 0, 0);
 				return 0;
@@ -16504,7 +16504,7 @@ static int skill_check_condition_castbegin(struct map_session_data *sd, uint16 s
 			if (sc && sc->data[SC_COMBOATTACK] && sc->data[SC_COMBOATTACK]->val1 == skill_id)
 				sd->ud.canmove_tick = timer->gettick(); //When using a combo, cancel the can't move delay to enable the skill. [Skotlex]
 
-			if (!unit->can_move(&sd->bl)) {
+			if (!unit->can_move(&sd->bl, skill_id)) {
 				clif->skill_fail(sd, skill_id, USESKILL_FAIL_LEVEL, 0, 0);
 				return 0;
 			}
@@ -22469,6 +22469,7 @@ static void skill_validate_skillinfo(struct config_setting_t *conf, struct s_ski
 			{ "RangeModByRadius", INF2_RANGE_RADIUS },
 			{ "RangeModByResearchTrap", INF2_RANGE_RESEARCHTRAP },
 			{ "AllowPlagiarism", INF2_ALLOW_PLAGIARIZE },
+			{ "IgnoreWalkDelay", INF2_IGNORE_WALK_DELAY },
 		};
 
 		while ((tt = libconfig->setting_get_elem(t, i++)) != NULL) {
diff --git a/src/map/skill.h b/src/map/skill.h
index 5758b30cd..77c72dca6 100644
--- a/src/map/skill.h
+++ b/src/map/skill.h
@@ -156,6 +156,7 @@ enum e_skill_inf2 {
 	INF2_RANGE_RADIUS       = 0x04000000, ///< Range is modified by WL_RADIUS
 	INF2_RANGE_RESEARCHTRAP = 0x08000000, ///< Range is modified by RA_RESEARCHTRAP
 	INF2_ALLOW_PLAGIARIZE   = 0x10000000, ///< Allow skill to be copied via RG_PLAGIARISM[KeiKun]
+	INF2_IGNORE_WALK_DELAY  = 0x20000000, ///< Usable while the walk delay of a previously used skill is still running
 };
 
 
diff --git a/src/map/unit.c b/src/map/unit.c
index 04c818792..923ab172a 100644
--- a/src/map/unit.c
+++ b/src/map/unit.c
@@ -621,7 +621,7 @@ static int unit_walk_toxy(struct block_list *bl, short x, short y, int flag)
 		return 0;
 	}
 
-	if ((flag & 2) == 0 && ((status_get_mode(bl) & MD_CANMOVE) == 0 || unit->can_move(bl) == 0))
+	if ((flag & 2) == 0 && ((status_get_mode(bl) & MD_CANMOVE) == 0 || unit->can_move(bl, 0) == 0))
 		return 1;
 
 	ud->state.walk_easy = flag & 1;
@@ -691,7 +691,7 @@ static int unit_walktobl_timer(int tid, int64 tick, int id, intptr_t data)
 	if (ud->walktimer == INVALID_TIMER && ud->target == data) {
 		if (DIFF_TICK(ud->canmove_tick, tick) > 0) // Keep waiting?
 			timer->add(ud->canmove_tick + 1, unit->walktobl_timer, id, data);
-		else if (unit->can_move(bl) != 0 && unit->walk_toxy_sub(bl) == 0 && ud->state.attack_continue != 0)
+		else if (unit->can_move(bl, 0) != 0 && unit->walk_toxy_sub(bl) == 0 && ud->state.attack_continue != 0)
 			set_mobstate(bl);
 	}
 	return 0;
@@ -762,7 +762,7 @@ static int unit_walk_tobl(struct block_list *bl, struct block_list *tbl, int ran
 		return 0;
 	}
 
-	if (unit->can_move(bl) == 0)
+	if (unit->can_move(bl, 0) == 0)
 		return 1;
 
 	if (unit->walk_toxy_sub(bl) == 0) {
@@ -830,7 +830,7 @@ static bool unit_run(struct block_list *bl, struct map_session_data *sd, enum sc
 	if (sc == NULL || sc->data[type] == NULL)
 		return false;
 
-	if (unit->can_move(bl) == 0) {
+	if (unit->can_move(bl, 0) == 0) {
 		status_change_end(bl, type, INVALID_TIMER);
 		return false;
 	}
@@ -1212,8 +1212,10 @@ static int unit_is_walking(struct block_list *bl)
 
 /*==========================================
  * Determines if the bl can move based on status changes. [Skotlex]
+ * skill_id is the skill the check is performed for, or 0 when the check is not
+ * related to a skill. It lets a skill bypass the checks its db flags allow.
  *------------------------------------------*/
-static int unit_can_move(struct block_list *bl)
+static int unit_can_move(struct block_list *bl, uint16 skill_id)
 {
 	struct map_session_data *sd;
 	struct unit_data *ud;
@@ -1240,7 +1242,10 @@ static int unit_can_move(struct block_list *bl)
 		}
 	}
 
-	if (DIFF_TICK(ud->canmove_tick, timer->gettick()) > 0)
+	// Skills flagged as IgnoreWalkDelay may be used while the walk delay left over by a
+	// previously used skill is still running, but not while the unit is actually immobilized.
+	if (DIFF_TICK(ud->canmove_tick, timer->gettick()) > 0
+	    && (skill->get_inf2(skill_id) & INF2_IGNORE_WALK_DELAY) == 0)
 		return 0;
 
 	if (sd && (
@@ -1378,7 +1383,7 @@ static int unit_set_walkdelay(struct block_list *bl, int64 tick, int delay, int
 			return 0;
 	} else {
 		//Don't set walk delays when already trapped.
-		if (!unit->can_move(bl))
+		if (!unit->can_move(bl, 0))
 			return 0;
 		//Immune to being stopped for double the flinch time
 		if (DIFF_TICK(ud->canmove_tick, tick-delay) > 0)
diff --git a/src/map/unit.h b/src/map/unit.h
index b04cac238..0a9227496 100644
--- a/src/map/unit.h
+++ b/src/map/unit.h
@@ -127,7 +127,7 @@ struct unit_interface {
 	int (*steptimer) (int tid, int64 tick, int id, intptr_t data);
 	void (*stop_stepaction) (struct block_list *bl);
 	int (*is_walking) (struct block_list *bl);
-	int (*can_move) (struct block_list *bl);
+	int (*can_move) (struct block_list *bl, uint16 skill_id);
 	int (*resume_running) (int tid, int64 tick, int id, intptr_t data);
 	int (*set_walkdelay_timer) (int tid, int64 tick, int id, intptr_t data);
 	int (*set_walkdelay) (struct block_list *bl, int64 tick, int delay, int type);
```

Streusel:
@kyeme please let me know if this requires any additional changes and if everything works.

kyeme:
@Streusel  Tested the latest changes and it's working properly now. Thank you!

Issue #3474, opened 2026-09-13
Using non-offensive itemskill cancels player_invincible_time


kyeme:
**Describe the bug**
Using a non-offensive itemskill, such as a Blessing Scroll or Heal Scroll, immediately removes `player_invincible_time`.

Because the invincibility is removed, nearby monsters can immediately target and attack the player, even though the configured `player_invincible_time` has not yet expired.

**To Reproduce**
1. Enter a state where `player_invincible_time` is applied. For example, warp to an area with aggressive monsters nearby. Do not move, then use an item such as a Blessing Scroll (ID 12215).
2. Notice that `player_invincible_time` is immediately removed, allowing the monster to attack the player.

**Current Behavior**
Using a non-offensive itemskill cancels player_invincible_time, allowing monsters to attack the player immediately.

**Expected behavior**
On Aegis, non-offensive item skills such as Blessing or Heal can be used while the player is protected by player_invincible_time without canceling the protection. The player should remain protected until the configured invincibility duration expires.

Offensive actions/skills may still cancel the protection as intended.

**System specs (please complete the following information):**
 - Mode: pre-renewal
 - Packet version: 20160106

**Additional context**
I’ll remember that this itemskill / player_invincible_time issue is an old, pre-existing Hercules bug, not a recent regression.

kyeme:
Herc Video:

https://github.com/user-attachments/assets/8ea53771-d7cd-457c-b0ff-ca478e3a908f

By the way, I noticed in the video that after the player is resurrected, nearby monsters start moving toward the player but do not attack. I believe this behavior is incorrect. While `player_invincible_time` is active, monsters should not target or move toward the player until the invincibility period expires.

kyeme:
Official Video: Using Heal Scroll

https://github.com/user-attachments/assets/6fc28634-c194-48f3-a3cf-b73c46190187

kyeme:
**Official Video:** Using a Fire Wall Scroll (ground skill)

**Result:** Casting a Fire Wall Scroll does not remove the player's invincibility.


https://github.com/user-attachments/assets/2e2d8e79-7c61-4e77-88ab-ad6849fbc7ea

kyeme:
**Official Video:** Using a Cold Bolt Scroll (targeted skill)

**Result:** Casting a Cold Bolt Scroll does not remove the player's invincibility.

https://github.com/user-attachments/assets/39e2ca26-027e-41cd-ab8d-db00e05d1a90

[...]

PR #3493, opened 2026-09-18
Fix item skills canceling player_invincible_time

Streusel:
### Pull Request Prelude

### Changes Proposed

Using a non-offensive item skill, such as a Blessing Scroll or a Heal Scroll, immediately cancelled `player_invincible_time`, so nearby aggressive monsters could attack the player before the configured duration had expired.

Hercules cancelled the invincibility at the moment the skill was *requested*, in three places in `clif.c`:

* `clif_useSkillToIdReal()`, for any targeted skill that is not `INF_SELF_SKILL`, including skills cast through items;
* `clif_parse_UseSkillToPosSub()`, for every ground skill;
* `clif_parse_UseSkillMap()`, for the warp destination dialog.

The first one carried a comment stating that target skills used through items are meant to cancel invincibility. Testing on the official server shows the opposite: scroll-cast skills do not cancel it at cast time, whatever their target type.

Verified against official behavior:

| Scroll | Skill type | Invincibility after casting |
| --- | --- | --- |
| Blessing Scroll | self (via `ISF_CASTONSELF`) | kept |
| Cold Bolt Scroll / Fire Bolt Scroll | targeted | kept |
| Fire Wall Scroll | ground | kept |
| Heal Scroll on a normal monster | targeted | kept |

The protection is not unconditional. It ends as soon as the skill actually connects and deals damage, which is why a Heal Scroll used on an Undead monster does remove it: Heal is redirected to `skill_castend_damage_id()` for Undead targets and deals damage. A Fire Bolt Scroll behaves the same way, keeping the invincibility while the bolt is being cast and losing it once the bolt hits.

This patch moves the cancellation to match that distinction:

1. the three `clif.c` call sites no longer cancel invincibility when `sd->auto_cast_current.type` is `AUTOCAST_ITEM`;
2. `skill_attack()` cancels it once the final damage total for a player-sourced attack is greater than zero.

Skills the player casts normally are unaffected, since they still cancel invincibility at request time as before, which makes the new check in `skill_attack()` a no-op for them.

**Issues addressed:** #3474


kyeme:
@Streusel Tested everything, and it works fine now. Thank you!

hemagx (approved):

PR #3430, opened 2026-08-25
Let non-damage skills bypass player_invincible_time

MrKeiKun:
### Pull Request Prelude

- [x] I have followed [proper Hercules code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

`battle_check_target()` blocked *all* skill targeting against a player under `player_invincible_time` (post-spawn/teleport grace period), except for a hardcoded two-skill allowlist (`PR_SANCTUARY`, `PR_MAGNIFICAT`) added in 2015 as an admittedly incomplete fix for bugreport:8397.

On official servers, invincibility only protects against damage and other hostile/offensive effects — non-damage skills such as `SA_DISPELL`, songs (`BA_POEMBRAGI`, `DC_SERVICEFORYOU`), and other buffs/debuffs should apply immediately regardless of the target's invincible state.

This replaces the two-skill allowlist with a general check against `skill->get_nk(skill_id) & NK_NO_DAMAGE`, so any non-damage skill bypasses the invincibility gate by default instead of needing to be special-cased individually. Equip-strip skills (`RG_STRIPWEAPON`/`STRIPSHIELD`/`STRIPARMOR`/`STRIPHELM`, `ST_FULLSTRIP`, `SC_STRIPACCESSARY`, `GC_WEAPONCRUSH`) are kept blocked despite being `NK_NO_DAMAGE`, since they're still an offensive debuff. Plain melee attacks (`skill_id == 0`) are unaffected and remain blocked.

**Issues addressed:** #966

Thanks @kyeme and @Playtester

[cont]: https://github.com/HerculesWS/Hercules/blob/master/CONTRIBUTING.md
[code]: https://github.com/HerculesWS/Hercules/wiki/Coding-Style

PR #3494, opened 2026-09-18
Fix monsters targeting players protected by player_invincible_time

Streusel:
### Pull Request Prelude

### Changes Proposed

After being resurrected, a player protected by `player_invincible_time` would see nearby monsters walk towards them without ever attacking. On official servers monsters neither target nor approach a player while that protection is running.

The attack itself was already blocked. `battle_check_target()` returns `-1` for a player whose `invincible_timer` is running, so `unit_attack()` never goes through. The movement was not blocked, because three separate monster targeting paths never consult that timer.

**`mob_linksearch()`**

Links monsters of the same class onto a target when one of them is attacked. It performed no target validation at all, so an invincible player could be handed to every nearby monster of that class.

**`mob_target()`**

Validated the target with `status_check_skilluse()` only, which does not look at `invincible_timer`. It is reached from five call sites in `skill.c`, so a monster could lock onto a player it was not allowed to attack.

**`mob_ai_sub_hard()`**

This one did already drop an invincible player from `md->target_id`, but only after two early returns: the warp chase check, and the `mob_chase_refresh` grace period that deliberately keeps a monster walking for a few cells before it gives up a target. A monster that was mid-walk therefore kept moving towards the player and only released the target later.

All three now reject a player whose `invincible_timer` is running. In `mob_ai_sub_hard()` the grace period is skipped for that specific case, so the target goes straight to `mob_unlocktarget()`, which calls `mob_stop_walking()` for chasing states and actually stops the approach rather than letting it run to completion.

The `mob_chase_refresh` behavior is otherwise untouched. It still applies to every other reason a target becomes invalid, which is what it exists for.

**Issues addressed:** #3474


kyeme:
When respawn or logging in:


https://github.com/user-attachments/assets/c5e9c52f-6d45-4396-82cb-e6826f8a5c3a

kyeme:
@Streusel  Ohh, sorry. I just realized that I had already applied PR [#3430](https://github.com/HerculesWS/Hercules/pull/3430), which is why I was seeing this issue with the monster behavior.

Streusel:
> @Streusel Ohh, sorry. I just realized that I had already applied PR [#3430](https://github.com/HerculesWS/Hercules/pull/3430), which is why I was seeing this issue with the monster behavior.

Oh, dear.
Uhhh, but the behaviour is still wrong, so I hope this new update fixes that.

Streusel:
> @Streusel Ohh, sorry. I just realized that I had already applied PR [#3430](https://github.com/HerculesWS/Hercules/pull/3430), which is why I was seeing this issue with the monster behavior.

Wait so, is the video behaviour due to 3430 or my change? And were the monsters behaving (chasing/targeting) incorrectly due to 3430 or without it? Also I think I DMd you on Discord, since it's quicker to resolve this, assuming it's still necessary.

kyeme:
Sorry for the confusion. The behavior shown in my video was caused by PR #3430, which I forgot I still had applied during my testing.

Without PR #3430 and PR # https://github.com/HerculesWS/Hercules/pull/3494 the monsters behave correctly: while `player_invincible_time` is active, they do not target or chase the player.

So this issue is not caused by your change. Sorry about that!

Issue #3476, opened 2026-09-14
Missing / Incorrect Basilica Behaviors


kyeme:
Missing / Incorrect Behaviors:

**1. On the official server, Deactivating Basilica should still have a cast time and should not consume/recheck the requirements.**

**Official behavior:**
thRO (02:22)
https://youtu.be/dEZMX6kJA04?t=137

**Current Hercules behavior:**
When deactivating Basilica, there is no cast time, and the player is still required to have Holy Water and Gemstones.

**2. On official, Basilica requires a 7x7 clear area, including ground skills**

The wiki description states that Basilica requires a 5x5 clear space, but on the official server it appears to require a **7x7 clear area around the caster**.

The restriction is also not limited to walls, obstacles, players, or monsters. Existing ground skills within the required area can also prevent Basilica from being cast.

For example, the video shows that Basilica cannot be cast beside Sanctuary because the required 7x7 area is not clear.

**Official behavior:**
thRO (02:43)
https://youtu.be/dEZMX6kJA04?t=163

**3. On official,  casting Basilica should not remove the caster's existing ground skills**

On the official server, creating a Basilica does not remove the caster's other active ground skills.

The video demonstrates that Sanctuary and Basilica can exist at the same time.

**Official behavior:**
thRO (03:02)
https://youtu.be/dEZMX6kJA04?t=183

**Current Hercules behavior:**
The caster's existing ground skills, such as Sanctuary or Safety Wall, are immediately removed when Basilica is cast.

I'm also wondering about the interaction between Magnus Exorcismus and Basilica. Since official behavior allows Basilica to coexist with the caster's other ground skills, I would like to know whether Magnus Exorcismus still deals damage while its caster is protected by their own Basilica. Maybe add config for this? XD

I currently don't have access to an Aegis server, so I have no way to test and confirm this scenario.

> Please note that my current source for these behaviors is a **Renewal official server**. However, I don't remember these Basilica mechanics being changed between Pre-Renewal and Renewal, so it is possible that the same behavior also applies to Pre-Renewal.
> 
> Unfortunately, I'm currently unable to verify all of these behaviors on a Pre-Renewal Aegis server.
> 
> If anyone has a reliable Pre-Renewal source showing that any of the behaviors described above are different or that this report does not apply to Pre-Renewal, please let me know.
> 
> **Source:** Ragnarok Online Thailand official server (thRO)

Streusel:
Let me know if there are any issues.

PR #3495, opened 2026-09-18
Fix pre-renewal Basilica to match official behavior (issue #3476)

Streusel:
### Pull Request Prelude

### Changes Proposed

Fixes the pre-renewal Basilica behaviors reported in #3476. Renewal is unaffected, the touched code is all behind `#ifndef RENEWAL` already.

- Cancelling Basilica has a cast time again. `unit_skilluse_id2()` forced `casttime = -1` when the caster had `SC_BASILICA`.
- The requirements and the free area are now checked in `skill_check_condition_castbegin()`, so a Basilica that cannot be created fails immediately instead of after the cast. Hercules only checked items in `skill_check_condition_castend()`, which runs when the cast completes.
- The requirements are consumed on activation only. They have to be consumed before the status change starts, or the turn off check in `skill_get_requirement()` skips them, and cancelling returns early so the consumption at the end of `skill_castend_pos2()` is not reached.
- The clear area check is 7x7 instead of 5x5, and now covers `BL_SKILL` and non walkable cells rather than just characters, so ground skills, walls and Ice Wall block the cast. Per the idRO Classic wiki the free area is only required to begin casting, so characters entering the area during the cast do not prevent the Basilica from being created.
- Casting no longer calls `skill->clear_unitgroup()`, so Basilica coexists with Sanctuary and the like. The group id goes in the status `val3` so ending the status deletes only the Basilica group, and `skill_delunitgroup()` ends the caster's status if the group goes away by other means, same as Gospel does.

I'm unable to address the Magnus Exorcismus mention due to lack of information.

**Issues addressed:** #3476


kyeme:
@Streusel 
I tested the changes and found two remaining issues:

1. The Gemstones and Holy Water should only be consumed when **activating** Basilica. Currently, the requirements are not consumed on activation, but are instead consumed when **deactivating** Basilica.

2. If the required items are missing, or if there is an obstruction within the required area, Basilica should be prevented from casting immediately and `Skill has failed` should be shown.

   Currently, the cast starts first, and only after the cast finishes does it check for missing requirements or obstructions and then fail the skill.

Streusel:
Please try now~

kyeme:
It's working now. Thank you!

Streusel:
Fixed the `skill_interface` issue that caused 2 builds to fail.

kyeme:
@Streusel Additional information regarding Basilica's obstruction and monster behavior:

According to the idRO Classic Wiki, the 7x7 clear-area requirement should only be checked when beginning the cast:

"The 7 X 7 area of free space is only required to begin casting Basilica. Once the player starts casting it, other characters (and monsters) may freely enter the area where Basilica will be placed."

This means that if the 7x7 area is clear when Basilica begins casting, but a monster/player enters the area during the cast, Basilica should still successfully activate instead of failing due to the new obstruction.

Currently, with this PR, Basilica can fail after the casting is completed if an obstruction enters the area while it is being cast. Based on the documented behavior above, it seems the obstruction check should only happen when the cast begins.

Source:
https://www.idrowiki.org/klasik/Basilica

I also attached a screenshot of the idRO Wiki page as a reference because the website is sometimes inaccessible.

idrowiki_basilica

Streusel:
Adjusted to match, let me know if that resolves everything.

kyeme:
Tested, and it's working now. Thank you!


https://github.com/user-attachments/assets/19fcd4e2-ceaa-43df-a42f-e737d9c26a9f

Issue #3471, opened 2026-09-12
Using an item with a dialog during NPC interaction causes "Double continuation" warning


kyeme:
Describe the bug
When using a Kafra Card while the player is currently interacting with another NPC that has an active dialog, the map-server displays a Double continuation warning in the console.

This happens when item usage during NPC interaction is enabled through:

Mode: Pre-Renewal
// map/battle/items.conf
item_enabled_npc: 0x3

To Reproduce

Set the following in map/battle/items.conf:
item_enabled_npc: 0x3

Talk to an NPC that opens a dialog, such as a Bulletin Board.
While the NPC dialog is still open, use a Kafra Card.

Check the map-server console.
The following warning appears:
```
[Warning]: Unable to restore stack! Double continuation!
[Debug]: Previous script (lost):
[Debug]: Source (NPC): Bulletin Board#09 at prontera (148,49)
[Debug]: Current script:
[Debug]: Source (NPC): FAKE_NPC (invisible/not on a map)
```

rAthena had the same issue with item_enabled_npc and already implemented a fix:

https://github.com/rathena/rathena/commit/3c93e30798618805791e0d598761c90f49becc12

Streusel:
Let me know if this requires any additional attention.

PR #3496, opened 2026-09-18
Fix "Double continuation" warning when using an item during NPC interaction

Streusel:
### Pull Request Prelude

### Changes Proposed

With `item_enabled_npc` set to allow item usage during NPC interaction, using an item that opens a dialog of its own while an NPC conversation is already open printed a `Double continuation` warning and silently discarded the conversation the player was in:

```
[Warning]: Unable to restore stack! Double continuation!
[Debug]: Previous script (lost):
[Debug]: Source (NPC): Bulletin Board#09 at prontera (148,49)
[Debug]: Current script:
[Debug]: Source (NPC): FAKE_NPC (invisible/not on a map)
```

`pc_useitem()` ran the item script while the player was still attached to the NPC's script. `script_attach_state()` therefore backed the NPC's state up into the item script's `bk_st`. When the item script then paused to show its own dialog, `run_script_main()` reached the `st->state != END && st->rid != 0` branch, found `st->bk_st` set, emitted the warning and called `script_free_state()` on the backup. That freed state was the NPC conversation.

The player is now detached from the running script before the item script executes, and reattached to it once the item script is done, so no backup chain is ever created. Whatever state the item script leaves pending is discarded, because it may not take over a conversation that is already in progress. When no NPC script is attached, which is the ordinary case, the added code is inert.

The fix is a port of the equivalent rAthena change for the same bug. The upstream commit is reproduced in full in the commit message, as done previously for other ported changes in this repository.

### Testing

Reproduction from the issue report, in pre-renewal with `item_enabled_npc: 0x3`:

1. talk to an NPC that opens a dialog, such as a Bulletin Board;
2. while the dialog is open, use an item that opens a dialog of its own, such as a Kafra Card;
3. the warning no longer appears and the NPC dialog remains usable.

**Issues addressed:** #3471


kyeme:
I tested this PR and found another issue.

While talking to an NPC with an active dialog, I tried to use a `Kafra_Card`. After doing so, the following debug message appears in the console:

`[Debug]: npc_scriptcont: Shion#nv1 (sd->npc_id=110010076) is not FAKE_NPC (id=110010601).`

After this happens, my character becomes stuck.

I also tested the same scenario on rAthena, and the same issue occurs there.

Streusel:
Welp, thanks for the report.
Could you retest both cases when you get a chance? An item with no window of its own, such as a Blessing Scroll, should leave the NPC dialog usable, and a Kafra Card should open its window normally without locking you up.

kyeme:
> Welp, thanks for the report. Could you retest both cases when you get a chance? An item with no window of its own, such as a Blessing Scroll, should leave the NPC dialog usable, and a Kafra Card should open its window normally without locking you up.

It's working now. The `Kafra_Card` can be used while an NPC dialog is active. The previous dialog is canceled, and the Kafra Storage opens correctly.

However, the `Blessing Scroll` cannot be used while an NPC dialog is active. I'm not sure if this is the intended behavior, since I'm unable to test it on an official Pre-Renewal server because I don't have access to one.

Streusel:
Great. As for the blessing scroll we'll have to wait for more info and address it when it comes to that.

hemagx (approved):

hemagx:
@skyleo may you be able to confirm the blessing scroll?

kyeme:
@Streusel 
I found another bug scenario while testing this PR again.

### Scenario

1. Talk to an NPC that opens a `menu`.
2. While the NPC menu is still active, use an item that also opens a menu, such as `DUN_TELE_SCROLL1`.
3. The `DUN_TELE_SCROLL1` menu does not appear.
4. Instead, the character is automatically warped immediately.

It seems that using an item with its own menu while an NPC menu is already active causes the item script to continue with a menu selection automatically instead of waiting for the player to select an option.


https://github.com/user-attachments/assets/588c21d1-0d53-4e9b-87af-10b69baf75b6

Streusel:
@kyeme Do you have an example of this in official? Is the NPC menu supposed to be dismissed? What's the expected behaviour? Other than the menu obviously being accessible.

kyeme:
https://github.com/user-attachments/assets/2dc9eb78-9de1-4372-a32e-25b34cb9b503



If the NPC dialog **does not have a menu shown**, I can use the Dungeon Teleport Scroll normally.

The issue only happens when the NPC dialog currently has a **menu/selection shown**. If I use `DUN_TELE_SCROLL1` while that NPC menu is open, the scroll automatically warps the character instead of showing its own menu/selection.

So it seems this case is specifically related to using an item with its own menu while an **NPC menu is already active**.

I'm not sure what the correct official pre-renewal behavior should be, since I don't currently have access to a pre-renewal Aegis server to verify it.

For comparison, on the official Renewal server I tested, usable items cannot be used at all while interacting with an NPC, so I can't use Renewal behavior to confirm how this should work in pre-renewal.

kyeme:
Tested the new PR, and it's working now. Thank you!

Streusel:
I'm not entirely happy with this. But the proper solution would be touching more files with #3508 would be good.

Issue #3472, opened 2026-09-12
Pre-Renewal Gravitational Field


kyeme:
# Pre-Renewal Gravitational Field

## Describe the bug

There are several incorrect behaviors with **Gravitational Field** in Pre-Renewal compared to the official Aegis behavior.

### Current Hercules behavior / Issues

1. **Gravitational Field cannot damage the Emperium.**
   On Aegis Pre-Renewal, Gravitational Field always deals **400 damage to the Emperium regardless of skill level**.

2. **The damage can be increased by Lex Aeterna.**
   On Aegis, Gravitational Field's damage cannot be increased or decreased, similar to the behavior of **Pressure**.

3. **The caster can perform normal attacks while Gravitational Field is active.**
   According to iRO Wiki Classic, the caster should **not be able to perform normal attacks while Gravitational Field is active**.

4. **Casting another Gravitational Field does not remove the previous Gravitational Field group.**
   The existing Gravitational Field should be removed when the caster casts another Gravitational Field.
Playtester note on the fix:
"Added the possibility to link different skill unit groups together; if a skill unit group gets deleted, the linked skill unit groups are deleted as well; this was needed because when being hit, all Gravitational Fields of the person being hit need to be removed."

5. **The AfterCastActDelay is incorrect.**

## Aegis AfterCastActDelay

* Level 1: 5000 ms
* Level 2: 6000 ms
* Level 3: 7000 ms
* Level 4: 8000 ms
* Level 5: 9000 ms

## References

iRO Wiki Classic — Gravitational Field:
https://irowiki.org/classic/Gravitational_Field

rAthena fixes:

1. https://github.com/rathena/rathena/commit/861112b89b58eeeb875d36b0c59f3e999b5f528d
2. https://github.com/rathena/rathena/commit/3089bbb6c1cc364b2dfe9801a26167b92401667b

## Mode

Pre-Renewal

Streusel:
Let me know if this requires any additional attention.

PR #3497, opened 2026-09-18
Fix/pre re gravitation 3472

Streusel:
### Pull Request Prelude

### Changes Proposed

Corrects the pre-renewal behavior of Gravitational Field, which differed from Aegis in several independent ways. Renewal treats the skill as plain magic damage since the 2018.10 rebalance and is left untouched throughout: every code change is either inside an existing `#ifndef RENEWAL` block, or relies on a status change the caster only receives in pre-renewal.

Each defect is a separate commit, so they can be reviewed and accepted independently.

**Gravitational Field could not damage the Emperium.** It now deals a flat 400 damage to it on every skill level, matching Aegis.

**The damage could be modified.** Like Pressure, Gravitational Field's damage cannot be increased or decreased by any means. It was reaching the damage increasing effects in `battle_calc_damage()` and was in particular doubled by Lex Aeterna. It now returns early alongside Pressure.

**The caster could perform normal attacks.** Per iRO Wiki Classic the caster cannot attack for as long as their own field is up. Added to `pc_can_attack()`.

**Recasting did not remove the previous field.** Skill unit groups can now be linked: deleting a group deletes the group it links to. Each new Gravitational Field links to the caster's previous one, taken from the group id already stored in `SC_GRAVITATION`'s `val4`. This is also what removes every field of a caster at once when they are hit. The skill is additionally added to the list of skills that cannot be placed on top of themselves, so overlapping can no longer refresh its duration. The link is cleared before the group is freed, so a linked group cannot delete its origin back.

**The AfterCastActDelay was a flat 2000ms.** It now scales from 5000ms at level 1 to 9000ms at level 5.

One further change is included because it completes an earlier partial port rather than adding new scope: the check that stops other skills from dealing damage while the caster's field is active already exists in `skill_attack()`, but it was missing the exception for auto-casted skills, which do deal damage on official servers.

### Notes for review

The four behavioral fixes come from Playtester's rAthena implementation, and the Emperium change from his later follow up. Both upstream commits are reproduced in full in the relevant commit messages.

Two notes:

The rAthena implementation commit also carries an unrelated renewal fix, changing `status_weapon_atk()` from `wa.range > 1` to `wa.range > 3` so that weapons with a range of 2 or 3 scale with STR instead of DEX. This is not included here.

rAthena also changed the skill range from 9 to 14. Hercules currently has 18, the issue does not mention range, and the correct value is not obvious from the issue, as such the range is left as is.

**Issues addressed:** #3472


kyeme:
I tested it, and it's working now. Thank you!

I just have two questions regarding the rAthena implementation:

1. I noticed that latest rAthena added `PathCheck: true` for Gravitational Field. Do we need this as well? If so, could you please include it in this PR?

~~2. I also noticed that they removed the flinching animation/effect when a target is hit by Gravitational Field. I couldn't find any report or reference explaining this change.~~

Streusel:
Great! Thanks for testing!

`PathCheck` isn't part of either commit this PR ports. rAthena gave Gravitational Field that flag in 2016 as part of a broad line of sight overhaul (7906f0a3f), which also changed what the flag checks, and a 2025 PR (#9341) gave it a second meaning on top. Hercules never ported either, so the flag doesn't mean the same thing here and wouldn't reproduce rAthena's behaviour. Renewal is unaffected regardless, since our renewal entry has no Unit block.

I would leave it for a proper line of sight port, if we need it, than fold it in here, same as the range change I mentioned above.

Issue #3492, opened 2026-09-18
Monster Sightrasher and Stormgust Level 10


kyeme:
**Server Mode**
Pre-Renewal, Renewal

**Expected Result**
When a monster uses Sightrasher or Stormgust Level 10, it will do the same damage as when used by a player. There should not be increased levels for these two skills.

Thanks to @Playtester
https://github.com/rathena/rathena/commit/0e4d6a4f24de7813d73907660420badc791e9d10

Streusel:
Should be fixed with this, once merged.

PR #3498, opened 2026-09-18
Fix monster Sightrasher dealing more damage than a player's

Streusel:
### Pull Request Prelude

### Changes Proposed

Six pre-renewal monsters were set to cast Sightrasher at level 20. The skill's maximum level is 10, so they dealt considerably more damage with it than a player casting the same skill.

| Monster | Skill | Before | After |
| --- | --- | --- | --- |
| KATRINN | `WZ_SIGHTRASHER` | 20 | 10 |
| THA_DESPERO | `WZ_SIGHTRASHER` | 20 | 10 |
| G_THA_DESPERO | `WZ_SIGHTRASHER` | 20 | 10 |
| GLOOMUNDERNIGHT | `WZ_SIGHTRASHER` | 20 | 10 |
| KASA | `WZ_SIGHTRASHER` | 20 | 10 |
| IMP | `WZ_SIGHTRASHER` | 20 | 10 |

Only `db/pre-re/mob_skill_db.conf` is touched. Two things the upstream fix covered turned out to need no change here:

* the renewal database already uses level 10 for every one of these monsters;
* Stormgust is already capped at level 10 for Ktullanux in both databases, whereas rAthena still had it at 11.

No other `WZ_SIGHTRASHER` entry is affected. The remaining ones in the pre-renewal database are at level 5, 7 or 10 and were left alone, as were all unrelated skills that legitimately use level 20.

**Issues addressed:** #3492


hemagx:
Can we confirm with Aegis? the original issue and commit from rA doesn't recite a source.

Streusel:
30d75f5e0 originally implemented it as level 10. (2006)
cce28be79 is when Playtester raised the levels from 10 to 11 & 20. (2007)
954363476 Playtester added Kasa level 20 Fireball (2007)
9321a0241 L0ne_W0lf changed some level from 20 to 10. (2007)
792dd1097 changed it back from 10 to 20 (4 days later). (2007)
c20671c6c changes some level 20 to 10, but leaves other, mistake? And the mysterious "match official files better" statement (2009) slave vs non-slave kasa? (i.e. mvp summons)
0e4d6a4 Playtester changed it back to level 10 (2016)

The question is, is level 10 really max? Maya, Phreeoni, Moonlight Flower & others used to have Heal level 11. The only one remaining, in the current workspace, with a >10 skill is `TIMEHOLDER` with level 11 `WZ_METEOR`.

kyeme:
Could this be related to the episode of the source used at that time? Maybe the source back then had it at level 11 or 20, while the Episode 11.3 source currently used by rAthena has it at level 10?
https://github.com/rathena/rathena/blob/e985006171d2eb320ee512a653f4c83aea3d81b6/db/pre-re/mob_skill_db.txt#L2

Issue #3480, opened 2026-09-16
Soul Siphon (Soul Burn) should only target players


kyeme:
**Server Mode**
Pre-Renewal

**Result**
You can cast Soul Siphon on monsters.

**Expected Result**
You cannot cast Soul Siphon on monsters. If you try, it does not consume SP, it does not set any delay and it gives you have "Skill has failed" message.

**How to Reproduce**
Cast Soul Siphon on Poring.

**Official Information**
Tested on official servers.

Thanks to @Playtester
https://github.com/rathena/rathena/issues/9706

Streusel:
Should be fixed with this, once merged.

PR #3499, opened 2026-09-18
Restrict Soul Siphon to player targets

Streusel:
### Pull Request Prelude

### Changes Proposed

Soul Siphon (`PF_SOULBURN`) could be cast on monsters. On official servers it only works on players, and an attempt on anything else fails outright without costing the caster anything.

The target check is added to the skill specific section of `skill_check_condition_castend()`, next to the existing player-target checks for `MO_KITRANSLATION` and `MO_ABSORBSPIRITS`, which use the same `BL_CAST(BL_PC, target)` pattern.

That placement gives all three behaviors the report asks for:

* the player is shown the usual "skill has failed" message, through `USESKILL_FAIL_LEVEL`;
* no SP is consumed, because `skill_castend_id()` leaves the block on a failed condition check before it reaches `skill_consume_requirement()`;
* no skill delay is set, because the skill failed path that follows assigns `ud->canact_tick = tick`.

The upstream change is not restricted by mode and the rAthena pull request lists its server mode as both, so no renewal guard is used here either. The reporter tested in pre-renewal.

**Issues addressed:** #3480


kyeme:
Tested and working. Thank you.

src/map/skill.c
```diff
@@ -16945,6 +16945,15 @@ static int skill_check_condition_castend(struct map_session_data *sd, uint16 ski
 			break;
 		}
 
+		case PF_SOULBURN: {
```
hemagx, line 16948: very minor there's no need to open a new scope/curly brackets here.

hemagx (approved):

hemagx (approved):

Issue #3489, opened 2026-09-17
NJ_ISSEN (Final Strike) facing direction


kyeme:
### Server Mode

Pre-Renewal, Renewal

### Result

**Incorrect facing direction**

After using Final Strike, the caster changes direction to face the target.

### Expected Result

Based on official behavior:

After using Final Strike, the caster should retain their original facing direction instead of turning to face the target.

### How to Reproduce

1. Use `NJ_ISSEN` on a target.
2. Observe the caster's facing direction after using the skill.

### Official Information

Tested on the official server.

Streusel:
And that hopefully fixes that issue.

PR #3500, opened 2026-09-18
Keep the caster's facing direction after Final Strike

Streusel:
### Pull Request Prelude

### Changes Proposed

After using Final Strike (`NJ_ISSEN`) the caster turned to face their target. On official servers the caster keeps the direction they were already facing.

At the end of a successful `skill_castend_id()` the caster is turned towards the target:

```c
if (src->id != target->id && !is_asura)
    unit->set_dir(src, map->calc_dir(src, target->x, target->y));
```

Asura Strike was already excluded from this. Final Strike needs the same treatment: the two skills share a single case in `skill_castend_damage_id()`, and both reposition the caster relative to the target before this point, so turning the caster afterwards undoes the facing the skill left them with.

Final Strike is added to that exception. Since the flag no longer refers to Asura Strike alone it is renamed from `is_asura` to `keeps_facing`.

**Issues addressed:** #3489


kyeme:
Tested and working. Thank you.

hemagx (approved):

Issue #2702, opened 2020-04-26
Waterball / Jupitel Thunder double cast


kyeme:
https://github.com/rathena/rathena/issues/907

I can confirm this issue here in hercules.

Fixes: https://github.com/rathena/rathena/commit/6ebcb67601117f0e6a39aad67a1a4de11c9b8d86

guilherme-gm:
working on that

kyeme:
Hi @Streusel, could you please take a look at this when you have a chance? Thanks!

Streusel:
Please test this fix, wasn't able to make it a 1:1 copy, so will need confirmation.

PR #3502, opened 2026-09-19
Fix/waterball jupitel magicpower 2702

Streusel:
### Pull Request Prelude

### Changes Proposed

Ports Playtester's Waterball, Jupitel Thunder and Mystical Amplification work from rAthena. The upstream commit closes two rAthena issues at once and is reproduced in the relevant commit messages. It is split here into four commits, one per behavior, so they can be reviewed and accepted separately.

**Jupitel Thunder no longer deals its damage at cast end.** It schedules a timer 150ms out and applies the after cast delay when that timer fires. This is what allows the caster to begin another spell before the target is knocked back, so the second spell can still reach a target that would otherwise be pushed out of range.

**Waterball now uses skill units.** Previously it counted the water cells around the caster at cast end, consumed Deluge and Suiton cells as part of that count, and fired the resulting number of balls on a timer. Water, Deluge and Suiton cells within the layout are now turned into waterball units, and the timer consumes them one at a time, each consumed unit producing one ball. Several behaviors follow from this:

* multiple Waterball timers compete for the same units instead of each holding a private count;
* a ball that deals no damage, for instance against a target that fully resists water, cancels the rest and returns control to the caster;
* the special casing moves out of the code and into the skill database, so ordinary unit rules apply;
* monsters deploy a 9x9 area on level 10.

**Mystical Amplification toggles when a cast begins** rather than when it ends, so a spell is amplified according to the state at the moment it was started.

**MATK is recalculated immediately when Mystical Amplification ends.** The normal path defers recalculation for players while damage is being processed, which could leave a spell already under way still amplified after the status had expired.

### Database changes

`WZ_WATERBALL` gains a `Unit` block in both the pre-renewal and renewal skill databases, with unit id `0x8c`, a per level layout and `UF_PATHCHECK`. `Target` is deliberately omitted, which leaves the unit target at `BCT_NOONE`, matching the upstream entry. `SkillData1` is set to 10000 as the unit duration.

### Notes

The two second limit on the double cast window is not official behavior. It is an anti-abuse measure Playtester added upstream, in his words "to prevent exploits this behavior can only be used every 2 seconds". The 150ms delay is the official part.

Upstream removed a `skill_toggle_magicpower()` call from the preserved spell path inside `WL_RELEASE`. Hercules has that call in a slightly different position, at the top of `WL_RELEASE` and applying to Release itself. It is removed here too, because Release is now toggled when its own cast starts and toggling again would end the status before the released spell is cast. As such this is an interpretation rather than a direct translation.

`skill_get_unit_layout()` gains a special case giving monsters a 9x9 area for Waterball on level 10. This preserves the previous behavior, which applied the same exception through the old cell counting code.

This is a larger change than a typical fix and it touches shared skill timing, specifically `canact_tick`, the timer skill path and walk delay. It has not been tested on a live server. Testing Waterball over natural water, over Deluge and over Suiton, at several skill levels and against a water immune target, would be the most valuable confirmation.

**Issues addressed:** #2702


kyeme:
Tested, and it works fine now. Thank you!

I also want to mention rAthena's first commit related to this change:
https://github.com/rathena/rathena/commit/525e8178c27327f000e15ba557c6ed37a371c0b2

In that commit, they mention that all skills that are "chain-able" on official servers should also be 100% chain-able.

I also noticed that their first implementation changed `TIMER_MIN_INTERVAL` to `20`. Does Hercules also need this change to properly reproduce the official skill-chaining behavior, or is it unnecessary with the current implementation?

kyeme:
@Streusel 
I also noticed that rAthena made some follow-up fixes for Waterball/Jupitel after the original implementation:

https://github.com/rathena/rathena/commit/4fd4c1d5bf9c91efe65fec9a03178695ed0451dd

The follow-up mentions that Waterball/Jupitel should be able to hit through shootable walls, and that Waterball cells should be able to overlap.

I checked the current PR and it looks like these follow-up changes may not be included yet. Could you please check if we also need them here?
Thanks!

Streusel:
@kyeme Please test for me again, I hope this resolve all the issues.

As for the conflicts, I'll have a look later, likely related to the compile_cpp branch being merged.

kyeme:
Thanks for the fixes! I tested Water Ball and noticed one issue when the target takes no damage.

I also removed this check:

```c
		case WZ_WATERBALL:
			//A waterball that deals no damage is not shown at all. (issue #2702)
			if (damage > 0)
				dmg.dmotion = clif->skill_damage(dsrc, bl, tick, dmg.amotion, dmg.dmotion, damage, dmg.div_, skill_id, (flag & SD_LEVEL) != 0 ? -1 : skill_lv, type);
			break;
```

This allows the client to show `MISS` for every Water Ball hit. This behavior also exists in pre-renewal, although it seems it was changed later to hide the misses (I'm not sure when).

Because of that, I think this:

```c
						if (skill->attack(BF_MAGIC, src, src, target, skl->skill_id, skl->skill_lv, tick, skl->flag) == 0 && skl->type > 1) {
							// A waterball that deals no damage cancels the remaining ones
							su = NULL;
						}
```

should also just be:

```c
skill->attack(BF_MAGIC, src, src, target, skl->skill_id, skl->skill_lv, tick, skl->flag);
```

Otherwise, the remaining Water Balls are cancelled after a `MISS`.


Aside from this, I tested the other changes and everything seems to be working correctly. Thank you!

Streusel:
Adjustments made and conflicts resolved. Should show the MISS now.

